PR #21, opened 2013-04-21
Add translation support

pokecheck:
Extracted translations from the Japanese, French, Italian, German and Spanish ROMS.
Messages have been reformatted to use @ for formatting (similar to how it is handled in-game). A demo translation function _ is also provided, but not made use of in battle.js. To actually enable translations, in addition to loading the big translation file and setting a language in Config, every game message in battle.js must be refactored to use translation, e.g. "Pointed stones dug into " + poke.getLowerName() + "!" should be rewritten to _("Pointed stones dug into @0!", poke.getLowerName());

### DIFF
--- a/data/translations.js
+++ b/data/translations.js
@@ -1,0 +1,13334 @@
+exports.Translations = {
+	"The @0 your Pokémon saw in the dream appeared!": [
+		"ゆめで　であった @0が　あらわれた！",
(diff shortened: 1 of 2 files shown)


Zarel:
@pokecheck, do you have an updated gen 6 version of this file?

Zarel:
(If so, could you post it in #247 ?)

PR #92, opened 2013-12-10
Attempt to use set.baseSpecies as name before set.species on the teambui...

Joimer:
...lder

When adding a new Pokémon to the team, the default name should be 
the same as in-game, thus it should try to use set.baseSpecies if 
existant and then use set.species as a last resort, should there 
be no name and no baseSpecies.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -318,7 +318,7 @@
 			}
 			buf += '<div class="setmenu"><button name="moveSet"><i class="icon-move"></i>Move</button> <button name="deleteSet"><i class="icon-trash"></i>Delete</button></div>';
 			buf += '<div class="setchart-nickname">';
-			buf += '<label>Nickname</label><input type="text" value="'+Tools.escapeHTML(set.name||set.species)+'" name="nickname" />';
+			buf += '<label>Nickname</label><input type="text" value="'+Tools.escapeHTML(set.name||set.baseSpecies||set.species)+'" name="nickname" />';
 			buf += '</div>';
 			buf += '<div class="setchart">';
 


Joimer:
It's been a long while. What do we do about this?

Zarel:
I mean, i asked you a yes-or-no question a year and a half ago, and you haven't answered it yet...

Again:

It's currently done like so:

> Bob sent in Fluffy (Rotom-Wash)!

Right now, your implementation would also do, for un-renamed Pokémon:

> Bob sent in Rotom (Rotom-Wash)!

There's nothing majorly wrong with doing it this way (although you need to update a lot of things, like Nickname Clause, server-side). But I need a yes or no on whether or not you think this is the right way to do things.

Zarel:
I personally think "Rotom (Rotom-Wash)!" is kind of ugly (although not horrible).

I don't see anything hugely wrong with the current system, even if it doesn't match how names work on-cartridge.

Zarel:
I guess I'm mostly neutral here? I might take a side after more discussion.

Joimer:
It was my intention, but since it's been a long time, I'd rather ask people around for their opinions. I think I'll ask Inside Scoop as well.

Joimer:
http://www.smogon.com/forums/threads/forme-names-in-teambuilder-and-battles-in-ps.3554704/

js/client-teambuilder.js
```diff
@@ -318,7 +318,7 @@
 			}
 			buf += '<div class="setmenu"><button name="moveSet"><i class="icon-move"></i>Move</button> <button name="deleteSet"><i class="icon-trash"></i>Delete</button></div>';
 			buf += '<div class="setchart-nickname">';
-			buf += '<label>Nickname</label><input type="text" value="'+Tools.escapeHTML(set.name||set.species)+'" name="nickname" />';
+			buf += '<label>Nickname</label><input type="text" value="'+Tools.escapeHTML(set.name||set.baseSpecies||set.species)+'" name="nickname" />';
```
Zarel: This needs to be `template.baseSpecies`, by the way - sets don't have base species.

PR #115, opened 2014-01-12
Statistics-based suggestions for OU team builder

tolmar:
When building a team in OU, if your team already has at least one other
pokemon, every time it brings up a pokemon list, a “Suggestions”
category is listed based on your team.


ReklessZombie:
So is this going to happen?

Zarel:
Probably not.

monsanto:
Thanks for the PR but it doesn't seem like there was enough interest at the time, and it's been a really long time since.

PR #126, opened 2014-02-03
Friend list

ghost:


ReklessZombie:
Also would like to see this happen

Zarel:
Login server no longer dying, but I wanted to make a better UI for a friends list anyway.

ReklessZombie:
Thanks for the response @Zarel 
Yeah the UI needs work

PR #221, opened 2014-07-13
Small optimizations to parseMessage()

Morfent:
Replaced capturing groups with non-capturing groups when their match isn't used, and removed redundant capturing groups.


Zarel:
I honestly still don't understand what this does.

Zarel:
nvm, I got it.

PR #228, opened 2014-07-28
Order rooms alphabetically preference

SolarisFox:
Defaults to order by user count
can change preference with /order or by using the drop down in the options menu
Changing ordering preference forces the Rooms room to update.

options


Zarel:
Closing, a number of things, and especially manual room rearranging support (which is planned) make this pretty unnecessary.

PR #251, opened 2014-09-29
/rank format targeting

SolarisFox:
Allows users to specify which formats they want to see rank for.
Syntax: /rank [username], [format 1], [format 2]...
(preexisting syntax still works the same way)
Returns an empty table if a meta is specified which the user has not
laddered any matches in or if only non-existent formats are specified.

I considered adding some error messages for these cases but it seemed like it'd make the code look way more messy for what it's worth the only ways I could think to do so. If anyone has some suggestions for ways to improve the empty table though I'd appreciate it.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -596,17 +596,24 @@
 			case 'rating':
 			case 'ladder':
 				if (!target) target = app.user.get('userid');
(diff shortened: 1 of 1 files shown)


Zarel:
Superseded by #299

PR #254, opened 2014-10-03
Folders feature for teambuilder

sirDonovan:
Here are some screenshots: http://imgur.com/a/FLliy#0

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -609,6 +609,7 @@
 			}
 			return '<button class="select formatselect'+(noChoice?' preselected':'')+'" name="format" value="'+formatid+'"'+(noChoice?' disabled':'')+'>'+Tools.escapeFormat(formatid)+'</button>';
 		},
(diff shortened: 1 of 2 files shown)


Joimer:
Hello there, we've gotten (again) about a hundred requests for this. How can this be improved so we can get this live? May I help @sirDonovan in any way?

parivard:
thats really a good idea sirDonovan moreover also its not complicated, sometimes its too difficult for the pps to locate there teams on time and either get dqed or forced to change there teams.Imo it should be implemented.Its up to you!

Zarel:
3a41c51c37e4 mostly obviates the need for this.

In addition, drag-and-drop team rearranging add an extra layer of team organization on top of format folders.

I understand some people want to organize/categorize their teams independently from formats, but I get the impression such a system would need to be built in a way that coexists with format folders, and in particular in a way that understands exactly what problems users are still having and how to best solve them, so I'm closing this.

PR #258, opened 2014-10-19
Fix getPokemon for inactive mons

Slayer95:
If the search ID didn't include a slot and no details were provided, it would always return an active Pokémon.
This should fix most graphical glitches related to Red Card.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -363,7 +363,6 @@ function Pokemon(species) {
 		return false;
 	};
 	this.getIdent = function() {
-		if (selfP.side.active.length === 1) return selfP.ident;
 		var slots = ['a','b','c','d','e','f'];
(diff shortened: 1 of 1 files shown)


ascriptmaster:
Obligatory bump because this might be useful for Zarel/Pokemon-Showdown#1452

And even if it turns out it isn't, it's still good for the purposes of Aromatherapy and Gen 4 Beat Up.

Zarel:
Needs to be rebased...

Slayer95:
This isn't ready as I said above. I'd like a different approach for this, but my ideas are still in crisalide stage. I've only kept this open due to the discussion..

Zarel:
If there's anything in here that 848e90afc05 doesn't address, feel free to make a new issue for it.

PR #275, opened 2014-11-22
Support XY/ORAS icons and displaying gender/item on Team Preview

Slayer95:
Previews:
http://imgur.com/cpSe3RP,MVfkbbW,Vj17ZmV,DUt9ASN

Files for sprites/
sprites/xyicons-pokeball-sheet.png http://i.imgur.com/DTykNE0.png
sprites/xyicons-sheet.png http://i.imgur.com/uaiWiUJ.png

Protocol: |poke|[PLAYER]|[DETAILS]|item


ascriptmaster:
It's been a while since this has been touched. I would suggest this: have gender and whether there's an item in the mouse-over text, like HP and status. Then, during the actual team preview at the beginning, have the icons maybe appear next to the sprites at the center screen?

Layell:
I've been getting a few comments here and there about the ORAS icons, can we get those up at least while we figure out the other stuff.

Zarel:
>It's been a while since this has been touched. I would suggest this: have gender and whether there's an item in the mouse-over text, like HP and status. Then, during the actual team preview at the beginning, have the icons maybe appear next to the sprites at the center screen?

That's exactly what I recommend (well, my idea is fancier, but that one works).

>I've been getting a few comments here and there about the ORAS icons, can we get those up at least while we figure out the other stuff.

How do you propose fitting them?

Layell:
I would ideally like to see no overlap in the icons, because we simply don't know how future icons will border on the regular size limits (see kyurem formes bw2).

However I see the logic in not reworking the layout, so I am fine with no layout changes as well. This appears to be the easiest option so I'd say go for it.

Joimer:
I just want to note that showing whether a Pokémon has an item or not is competitively important. One could discuss gender, but that's minor.

Zarel:
Yes, Joim, we all know. :|

Layell, I sort of want to resize the trainer bars, but right now the layout is sort of precisely calculated to be 4:3 (480x360) without trainerbars and 16:9 (640x360) with trainerbars.

Zarel:
Also, have you tested how they look in teambuilder?

Also note that this will probably break PSdex, which uses these icons, too.

Zarel:
Oh, you have looked in teambuilder, and it looks weird. :|

Zarel:
Point is, this is a pretty big change to the design. :/

Zarel:
Seriously, though, every single one of those screenshots looks uglier than what we currently have.

ascriptmaster:
ok now what

Zarel:
This design won't be accepted; further discussion in #590

Issue #278, opened 2014-12-05
Weather remaining turns' tooltip is inaccurate


Slayer95:
If the weather is set by an ability due to a switch-in at the end of a turn or beginning of the battle, the counter will be offset, and the tooltip will display data as if an additional turn had passed.

E.g.
http://replay.pokemonshowdown.com/customgame-189512953
Turn 4: Predicts 1 or 4 turns left.
Turn 5: Predicts 3 turns left. However, that turn rain stops.

PR #289, opened 2014-12-17
Fixes Zarel/Pokemon-Showdown-Client#278 (weather prediction bug)

woodrush:
Fixes weather prediction inaccuracy bug when
- When weather is changed by ability on the first turn
- When weather is changed by ability on switch after a Pokemon faints

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2572,7 +2572,7 @@ var Battle = (function () {
 			this.endLastTurnPending = true;
 		}
 		this.turn = turnnum;
-		this.updateWeatherLeft();
+		this.updatePseudoWeatherLeft();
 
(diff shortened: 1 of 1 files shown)


woodrush:
The weather upkeep message idea worked the best. The first commit uses the upkeep for the fix. Tested by battling myself, and also by watching that battle as a third person from another client. The battles I used for testing are these 2 below.
http://replay.pokemonshowdown.com/ubers-194822105
http://replay.pokemonshowdown.com/ubers-199144453
The second commit fixes the additional issues I mentioned in issue #289 (also tested by the above).

Zarel:
I like this. Does it require a corresponding server-side update?

woodrush:
Thanks. No, it works with the current server.

PR #293, opened 2014-12-28
Implement old gen sprites for past gens

Joimer:
Implement old gens to use the old sprites as well as an option to disable it that may interact with enable BW sprites.
If old gens are not disallowed, those will be used for past gens (1-5).
If BW sprites are allowed, those will be used for XY and BW.
If old gens are disallowed and BW allowed, all gens will use BW sprites.

Tested all options and gens interactions with testclient.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -2293,6 +2293,7 @@
 		events: {
 			'change input[name=noanim]': 'setNoanim',
 			'change input[name=bwgfx]': 'setBwgfx',
+			'change input[name=nopastgens]': 'setNopastgens',
 			'change input[name=notournaments]': 'setNotournaments',
 			'change input[name=nolobbypm]': 'setNolobbypm',
 			'change input[name=temporarynotifications]': 'setTemporaryNotifications',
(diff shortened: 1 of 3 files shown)


js/client.js
```diff
@@ -2315,7 +2316,8 @@
 			buf += '<hr />';
 			buf += '<p><label class="optlabel">Background: <select name="bg"><option value="">Charizards</option><option value="#344b6c url(/fx/client-bg-horizon.jpg) no-repeat left center fixed">Horizon</option><option value="#546bac url(/fx/client-bg-3.jpg) no-repeat left center fixed">Waterfall</option><option value="#546bac url(/fx/client-bg-ocean.jpg) no-repeat left center fixed">Ocean</option><option value="#344b6c">Solid blue</option>'+(Tools.prefs('bg')?'<option value="" selected></option>':'')+'</select></label></p>';
 			buf += '<p><label class="optlabel"><input type="checkbox" name="noanim"'+(Tools.prefs('noanim')?' checked':'')+' /> Disable animations</label></p>';
-			buf += '<p><label class="optlabel"><input type="checkbox" name="bwgfx"'+(Tools.prefs('bwgfx')?' checked':'')+' /> Enable BW sprites</label></p>';
+			buf += '<p><label class="optlabel"><input type="checkbox" name="bwgfx"'+(Tools.prefs('bwgfx')?' checked':'')+' /> Enable BW sprites for XY</label></p>';
+			buf += '<p><label class="optlabel"><input type="checkbox" name="nopastgens"'+(Tools.prefs('nopastgens')?' checked':'')+' /> Use modern sprites for past generations</label></p>';
```
Joimer: I was thinking about it, what about people whose computer is too bad to use XY sprites on XY but wants to play XY?

Slayer95: The natural option would be to turn off animations, which should switch to BW sprites.

Zarel: What @Slayer95 said.

We might want to add XY stills, but then CAP would have to make sprites that matched the Xy style...

PR #296, opened 2015-01-01
Add support for multi-gen battle backgrounds

Joimer:


js/battle.js
```diff
@@ -2107,8 +2107,8 @@ var Battle = (function () {
 		// 5 = seeking
 		this.playbackState = 0;
 
-		this.backdropImage = BattleBackdrops[0];
-		this.backdropImage = BattleBackdrops[Math.floor(Math.random() * BattleBackdrops.length)];
+		this.backdropImage = BattleBackdrops.gen6[0];
+		this.backdropImage = BattleBackdrops['gen' + this.gen][Math.floor(Math.random() * BattleBackdrops['gen' + this.gen].length)];
```
Zarel: maybe cache `BattleBackdrops['gen' + this.gen]`? It's being used twice here.

Layell:
This should also probably skip the old gen backgrounds if the user is using contemporary sprites. The backgrounds will be really designed only for the old gen ones, might look awkward with the new sprites (if you did this and I didn't notice apologies).

Layell:
see: http://i.imgur.com/uCk7ELC.png

Joimer:
I've updated this.

js/battle.js
```diff
@@ -2107,8 +2107,9 @@ var Battle = (function () {
 		// 5 = seeking
 		this.playbackState = 0;
 
-		this.backdropImage = BattleBackdrops[0];
-		this.backdropImage = BattleBackdrops[Math.floor(Math.random() * BattleBackdrops.length)];
+		this.backdropImage = BattleBackdrops.gen6[0];
+		var gentext = (Tools.prefs('nopastgens'))? 'gen6' : 'gen' + this.gen;
+		this.backdropImage = BattleBackdrops[gentext][Math.floor(Math.random() * BattleBackdrops[gentext].length)];
```
Zarel: I said to save `BattleBackdrops[gentext]`, not `gentext`. :| It's a pointer, you know...

js/battle.js
```diff
@@ -2107,8 +2107,9 @@ var Battle = (function () {
 		// 5 = seeking
 		this.playbackState = 0;
 
-		this.backdropImage = BattleBackdrops[0];
-		this.backdropImage = BattleBackdrops[Math.floor(Math.random() * BattleBackdrops.length)];
+		this.backdropImage = BattleBackdrops.gen6[0];
```
Zarel: Does this actually do anything?

Joimer: Well, just changing the default of this.backdropImage from BattleBackdrops[0] to BattleBackdrops.gen6[0], which would be the appropiate default. Both old and new line is pretty much useless, I guess, but I just kept it.

js/battle.js
Joimer: Oh, I didn't get you. I thought you meant the classic "if you use it twice, make a variable for it".

js/battle.js
Zarel: The default isn't used for anything. I'd say remove it so it's clear it's not used for anything.

js/battle.js
Zarel: `BattleBackdrops[gentext]` _is_ used twice.

Joimer:
I've seen that the old gen backgrounds are loaded lazily anyways since the gen for the battle is also loaded lazily, so I've just gone ahead and added the new gen 1 background for gens 1 and 2, left bg.png for gens 3, 4 and 5 meanwhile.

I'll figure something better out once Layell has done the rest of the bgs, meanwhile this request has the new update that will make Old Gen players happy (as they are already asking for this):
https://github.com/Zarel/Pokemon-Showdown-Client/pull/304

PR #297, opened 2015-01-02
Sideroom tab reordering

YaYana:


Zarel:
Hmmmmmm... This is nice and low-dependency. Remind me to take a look at it.

Morfent:
I tested this out since this is something I've wanted implemented for a long time now. Closing room tabs with this is next to impossible on Firefox >=36.0, not including battle room reordering seems inconsistent, and being able to drag room tabs outside the boundaries of the tab list looks unappealing, though it's only a minor issue.

Zarel:
I guess if we ever do this, we should probably do this differently, then. :/

PR #298, opened 2015-01-03
Move client command documentation to client

asgdf:
As per https://github.com/Zarel/Pokemon-Showdown-Client/pull/280#issuecomment-65842063

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -436,7 +436,10 @@
 				return false;
 
 			case 'ignore':
-				if (!target) return '/help ignore';
+				if (!target) {
+					this.parseCommand('/help ignore');
+					return false;
+				}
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -684,6 +692,45 @@
 					Tools.prefs('avatar', avatar);
 				}
 				return text; // Send the /avatar command through to the server.
+				
+			// documentation of client commands
+			case 'help':
+				switch (target) {
```
Zarel: `(target||'').toLowerCase()`?`

Zarel: idr if our commands are case-sensitive

Zarel: Apparently they're not

asgdf:
Ah right, forgot case-sensitivity. Fixed that.

PR #299, opened 2015-01-03
/rank format targeting

YaYana:
Seeing how nobody worked on #251 anymore, I've implemented what was asked.
It was a bit tricky because of the HTML sanitizing though, making this not the most elegant solution, I guess...

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -607,18 +607,35 @@
 			case 'rating':
 			case 'ladder':
 				if (!target) target = app.user.get('userid');
+
+				var targets = target.split(',');
+				var formatTargeting = false;
+				var formats = {};
+				for (var i = 1, len = targets.length; i < len; i++) {
+					formats[toId(targets[i])] = 1;
+					formatTargeting = true;
+				}
+
 				var self = this;
-				$.get(app.user.getActionPHP() + '?act=ladderget&user='+encodeURIComponent(target), Tools.safeJSON(function(data) {
+				$.get(app.user.getActionPHP() + '?act=ladderget&user='+encodeURIComponent(targets[0]), Tools.safeJSON(function(data) {
 					try {
 						var buffer = '<div class="ladder"><table>';
-						buffer += '<tr><td colspan="8">User: <strong>'+target+'</strong></td></tr>';
+						buffer += '<tr><td colspan="8">User: <strong>'+targets[0]+'</strong></td></tr>';
 						if (!data.length) {
 							buffer += '<tr><td colspan="8"><em>This user has not played any ladder games yet.</em></td></tr>';
 						} else {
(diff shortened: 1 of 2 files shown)


js/client-chat.js
```diff
@@ -641,8 +658,31 @@
...
 						buffer += '</table></div>';
 						self.add('|raw|'+buffer);
+
+						if (hiddenFormats.length) {
+							self.$chatFrame.find('button.show-other-formats:last').click(function (e) {
+								var autoscroll = (self.$chatFrame.scrollTop() + 60 >= self.$chat.height() - self.$chatFrame.height());
+
+								var $target = $(e.currentTarget);
+								var $table = $target.closest('table');
+								$table.find('tr.hidden').css('display', 'table-row');
```
Zarel: lol, I was about to say "not IE6/IE7-compatible" before remembering we don't have to worry about supporting them anymore.

Zarel: Still, though, wouldn't `.show()` be easier?

js/client-chat.js
```diff
@@ -641,8 +658,31 @@
...
+								buffer += '<tr class="no-matches"><td colspan="8"><em>This user has not played any ladder games that match the format targeting.</em></td></tr>';
+							}
+
+							buffer += '<tr><td colspan="8"><button class="show-other-formats">' + hiddenFormats.slice(0, 3).join(', ') + (hiddenFormats.length > 3 ? ' and ' + (hiddenFormats.length - 3) + ' other formats' : '') + ' not shown</button></td></tr>';
+						}
 						buffer += '</table></div>';
 						self.add('|raw|'+buffer);
+
+						if (hiddenFormats.length) {
+							self.$chatFrame.find('button.show-other-formats:last').click(function (e) {
```
Zarel: This appears to be the wrong place to attach an event listener.

Zarel: You don't even need another event listener; presumably our `button name` system should be enough.

YaYana:
I was unaware of the `button name` system. It's really nifty.
Anyhow, I've made the changes you requested.

js/client-chat.js
```diff
@@ -607,18 +607,35 @@
...
-							for (var i=0; i<data.length; i++) {
+							var hiddenFormats = [];
+							var formatLength = data.length;
+							for (var i=0; i<formatLength; i++) {
 								var row = data[i];
-								buffer += '<tr><td>'+row.formatid+'</td><td><strong>'+Math.round(row.acre)+'</strong></td><td>'+Math.round(row.gxe,1)+'</td><td>';
+								if (!formatTargeting || formats[row.formatid]) {
+									buffer += '<tr>';
+								} else {
+									buffer += '<tr class="hidden">';
```
Zarel: You probably shouldn't use the HTML sanitizer, then... There are a few workarounds.

If you can't figure out any other workaround, I'd accept replacing `.ladder .hidden {` with `.hidden {` in the CSS files.

YaYana: I couldn't figure out a way to make it not use the HTML sanitizer without changing loads of functions or practically copy-pasting whole functions, so I opted for your alternative.

PR #302, opened 2015-01-06
Add support for removal of messages from chats

sirDonovan:
Client-side part of <a>http://www.smogon.com/forums/threads/3524852/</a>

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -5260,9 +5260,9 @@ var Battle = (function () {
 			}
 			var message = args.join('|');
 			if (message.substr(0,2) === '//') {
-				this.log('<div class="chat"><strong style="' + hashColor(toUserid(name)) + '">' + clickableName + ':</strong> <em>' + Tools.parseMessage(message.substr(1), name) + '</em></div>', preempt);
+				this.log('<div class="chat chatmessage-' + toId(name) + '"><strong style="' + hashColor(toUserid(name)) + '">' + clickableName + ':</strong> <em>' + Tools.parseMessage(message.substr(1)) + '</em></div>', preempt);
 			} else if (message.substr(0,4).toLowerCase() === '/me ') {
-				this.log('<div class="chat"><strong style="' + hashColor(toUserid(name)) + '">&bull;</strong> <em>' + clickableName + ' <i>' + Tools.parseMessage(message.substr(4), name) + '</i></em></div>', preempt);
+				this.log('<div class="chat chatmessage-' + toId(name) + '"><strong style="' + hashColor(toUserid(name)) + '">&bull;</strong> <em>' + clickableName + ' <i>' + Tools.parseMessage(message.substr(4)) + '</i></em></div>', preempt);
 			} else if (message.substr(0,14).toLowerCase() === '/data-pokemon ') {
 				if (window.Chart) this.log('<div class="chat"><ul class=\"utilichart\">'+Chart.pokemonRow(Tools.getTemplate(message.substr(14)),'',{})+'<li style=\"clear:both\"></li></ul></div>', preempt);
 			} else if (message.substr(0,11).toLowerCase() === '/data-item ') {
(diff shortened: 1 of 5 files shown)


js/client-chat.js
```diff
@@ -958,6 +958,13 @@
 					});
 					break;
 
+				case 'remove':
+					if (app.user.get('isStaff')) break;
+					$('.chatmessage-' + toId(row[1])).each(function() {
+						$(this).remove();
+					});
```
Zarel: Wow, you really don't know how to jQuery, do you?

That is literally just a slow and unreadable way to write:

    $('.chatmessage-' + toId(row[1])).remove()

Zarel: Anyway, using a raw `$` is pretty unacceptable for something as complex as PS; we'd generally use `this.$` in place of that.

js/battle.js
```diff
@@ -5458,6 +5458,7 @@ var Battle = (function () {
 			this.log('<div class="debug"><div class="chat"><small style="color:#999">[DEBUG] ' + Tools.escapeHTML(name) + '.</small></div></div>', preempt);
 			break;
 		case 'unlink':
+		case 'remove':
```
Zarel: You're not going to implement it here, too?

js/client-chat.js
```diff
@@ -958,6 +958,13 @@
 					});
 					break;
 
+				case 'remove':
+					if (app.user.get('isStaff')) break;
```
Zarel: Does `isStaff` exist client-side? Probably not.

Anyway, this should be a preference. We probably shouldn't force censorship on users.

js/client-chat.js
```diff
@@ -958,6 +958,13 @@
 					});
 					break;
 
+				case 'remove':
```
Zarel: I'm not sure we want to use as general a word as `remove`. We could repurpose `unlink` or something.

js/client.js
```diff
@@ -845,7 +846,8 @@
 					name: name,
 					userid: toUserid(name),
 					named: named,
-					avatar: parts[3]
+					avatar: parts[3],
+					isStaff: !!+parts[4]
```
Zarel: Are you actually going to add `isStaff` to userdetails? Ugh. I'd say just make it a preference.

js/client-chat.js
Zarel: Default should probably hide the messages but add a small-font notice "Some spam messages were hidden, click here to restore."

Zarel:
This is actually pretty good other than the issues I mentioned. It was exactly how I was thinking about doing it myself, so good job.

sirDonovan:
The comment of repurposing <code>unlink</code> gave me an idea that led to this, what do you think? https://github.com/sirDonovan/Pokemon-Showdown-Client/commit/a5fa226c85a921c6c9abb097b88dbee2d511153b

Zarel:
For the most part, yeah, that seems fine to me.

sirDonovan:
Added changes from https://github.com/sirDonovan/Pokemon-Showdown-Client/commit/a5fa226c85a921c6c9abb097b88dbee2d511153b.

js/battle.js
```diff
@@ -5458,6 +5458,13 @@ var Battle = (function () {
 			this.log('<div class="debug"><div class="chat"><small style="color:#999">[DEBUG] ' + Tools.escapeHTML(name) + '.</small></div></div>', preempt);
 			break;
 		case 'unlink':
+			var user = toId(args[2]) || toId(args[1]);
+			var messages = $('.chatmessage-' + user);
```
Zarel: `var $messages` makes it clearer it's a jQuery collection.

js/battle.js
```diff
@@ -5458,6 +5458,13 @@ var Battle = (function () {
...
+			messages.find('a').contents().unwrap();
+			if (args[2]) {
+				messages.css("display", "none");
```
Zarel: you mean `messages.hide()`.

sirDonovan:
Updated.

PR #303, opened 2015-01-06
Improve old gen sprites functionality

Joimer:
Always load mini sprites from bw when no animations or bw.
Delete the now redundant option to enable BW sprites.
You just need no animations option load BW sprites for XY.
Make the option text more fitting to the overall style.
Make sure sprites have the proper size.

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -836,14 +836,25 @@ var Tools = {
 		// Decide what gen sprites to use.
 		var gen = {1:'rby', 2:'gsc', 3:'rse', 4:'dpp', 5:'bw', 6:'xy'}[options.gen];
 		if (Tools.prefs('nopastgens')) gen = 'xy';
-		if (Tools.prefs('bwgfx') && gen === 'xy') gen = 'bw';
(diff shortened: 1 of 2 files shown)


Joimer:
btw, we should probably think of adding all sprite data to a single pokedex-mini, it's just a few bytes more of text we'd be loading and would make it easier and working better.

js/battledata.js
```diff
@@ -856,8 +867,9 @@ var Tools = {
 		}
 		// if there is no entry or enough data in pokedex-mini.js or the animations are disabled or past gen, use the proper sprites
 		gen = (gen === 'xy')? 'bw' : gen;
+		// Do this here in case gen was XY but animations were disabled.
+		Tools.loadSpriteData('bw');
```
Zarel: "In case" scares me. Can't we wrap this in an `if`-statement detecting that case?

Layell:
BW animated isn't redundant, if anything for animated BW sprites.

Can we keep it.

Zarel:
As far as I can tell, there's no need for this change. As @Layell points out, there are people who actually prefer BW sprites to XY models.

PR #307, opened 2015-01-14
Make move types and base power match battle status

Joimer:
When a Plate is active, it will change the type of Judgment.
Display that on the move selector to help players know and remember what attack type they're using.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -398,14 +398,15 @@
 						if (move.id === 'Struggle' || move.id === 'Recharge') pp = '&ndash;';
 						if (move.id === 'Recharge') move.type = '&ndash;';
 						if (name.substr(0, 12) === 'Hidden Power') name = 'Hidden Power';
+						var moveType = this.getMoveType(move, switchables[pos]);
 						if (moveData.disabled) {
 							movebuttons += '<button disabled="disabled"' + this.tooltipAttrs(moveData.move, 'move') + '>';
 							hasDisabled = true;
 						} else {
-							movebuttons += '<button class="type-' + move.type + '" name="chooseMove" value="' + (i+1) + '" data-move="' + Tools.escapeHTML(moveData.move) + '"' + this.tooltipAttrs(moveData.move, 'move') + '>';
+							movebuttons += '<button class="type-' + moveType + '" name="chooseMove" value="' + Tools.escapeHTML(moveData.move) + '"' + this.tooltipAttrs(moveData.move, 'move') + '>';
 							hasMoves = true;
 						}
-						movebuttons += name + '<br /><small class="type">' + (move.type || "Unknown") + '</small> <small class="pp">' + pp + '</small>&nbsp;</button> ';
+						movebuttons += name + '<br /><small class="type">' + (moveType || "Unknown") + '</small> <small class="pp">' + pp + '</small>&nbsp;</button> ';
 					}
 					if (!hasMoves) {
 						controls += '<button class="movebutton" name="chooseMove" value="0" data-move="Struggle">Struggle<br /><small class="type">Normal</small> <small class="pp">&ndash;</small>&nbsp;</button> ';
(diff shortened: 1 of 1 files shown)


js/client-battle.js
```diff
@@ -396,6 +396,11 @@
 						if (move.id === 'Struggle' || move.id === 'Recharge') pp = '&ndash;';
 						if (move.id === 'Recharge') move.type = '&ndash;';
 						if (name.substr(0, 12) === 'Hidden Power') name = 'Hidden Power';
+						// Update Judgment's type to match that of the Plate type, if any.
+						var item = Tools.getItem(switchables[pos].item);
+						if (move.name === 'Judgment' && item.onPlate) {
+							move.type = item.onPlate;
```
Zarel: `move` is a pointer to global storage. Make a copy before messing with its type.

js/client-battle.js
```diff
@@ -396,6 +396,11 @@
 						if (move.id === 'Struggle' || move.id === 'Recharge') pp = '&ndash;';
 						if (move.id === 'Recharge') move.type = '&ndash;';
 						if (name.substr(0, 12) === 'Hidden Power') name = 'Hidden Power';
+						// Update Judgment's type to match that of the Plate type, if any.
+						var item = Tools.getItem(switchables[pos].item);
```
Zarel: You probably want to move this inside the if-statement. `Tools.getBlah` aren't super heavy functions, but they aren't super light functions, either.

Joimer: The point is to check for item's onPlate for the if to avoid string manipulation.

Zarel: You can check for it after checking `move.name`. I guess either way is fine, but that's how I'd do it.

Joimer:
Opinion on loading the item instead of checking for its id / name:
If an item that is not a plate with plate properties and an onPlate is added, we don't need to update this functionality to reflect that. It will work. However, if we check for plate on its name or id, it won't.

Personal preference here.

js/client-battle.js
```diff
@@ -396,14 +396,20 @@
...
 						if (moveData.disabled) {
 							movebuttons += '<button disabled="disabled"' + this.tooltipAttrs(moveData.move, 'move') + '>';
 							hasDisabled = true;
 						} else {
-							movebuttons += '<button class="type-' + move.type + '" name="chooseMove" value="' + Tools.escapeHTML(moveData.move) + '"' + this.tooltipAttrs(moveData.move, 'move') + '>';
+							movebuttons += '<button class="type-' + moveType + '" name="chooseMove" value="' + Tools.escapeHTML(moveData.move) + '"' + this.tooltipAttrs(moveData.move, 'move') + '>';
 							hasMoves = true;
 						}
-						movebuttons += name + '<br /><small class="type">' + move.type + '</small> <small class="pp">' + pp + '</small>&nbsp;</button> ';
+						movebuttons += name + '<br /><small class="type">' + moveType + '</small> <small class="pp">' + pp + '</small>&nbsp;</button> ';
```
Zarel: The type will still be wrong on mouseover, I think.

Zarel: Do we want to handle that, too?

Joimer: Oh, yeah, this change makes it necessary to re-do this for mouseover as moveType is local. Let's handle it as well.

Joimer:
Ok, I think this is ready.
Move type is shown properly on singles and doubles.
It is shown properly on tooltip on both formats as well.
Item is only gotten if the move is Judgment.
Move type is stored on a var to add it to the HTML, not touching move.type.

Marty-D:
This will probably give the wrong information in Magic Room or while the Plate holder is under the effect of Embargo. Judgment is Normal type in those situations because the Plate effect that changes its type is ignored (Arceus itself remains whatever its type was).

Joimer:
Yeah, Marty, thanks for the note, I've improved my Judgment-type implementation.

js/client-battle.js
```diff
@@ -926,8 +936,19 @@
...
+				var moveType = move.type;
+				if (move.name === 'Judgment') {
+					var pokemon = this.battle.mySide.pokemon[this.choice.choices.length];
+					if (this.battle.hasPseudoWeather('Magic Room') || pokemon.volatiles['embargo']) {
+						moveType = 'Normal';
+					} else {
+						var item = Tools.getItem(pokemon.item);
+						if (item.onPlate) moveType = item.onPlate;
+					}
+				}
```
Zarel: Code duplication... can we put this in a separate function? getModifiedMove(move, battle) or something?

Joimer: Hm. I can add that so we can re-use it for other moves as well, yeah.

Slayer95: Probably as a <code>Pokemon</code> method?

Joimer: It can be a var move = Pokemon.modifyMove(move);
And it returns a copy of the move with the changed variables (in this case, merely movetype).

Zarel:
It could, but it's not really a battle method, just a sim UI method. I'd probably want it in `client-battle.js` rather than `battle.js`.

js/client-battle.js
Zarel: It could, but it's not really a battle method, just a sim UI method. I'd probably want it in `client-battle.js` rather than `battle.js`.

Joimer:
I thought of cloning the move object, but since this is javascript I simply added a client-battle UI function to get the move type, since it's varying. Also added the functionality for techno drive as I was at it.

js/client-battle.js
```diff
@@ -396,14 +396,15 @@
 						if (move.id === 'Struggle' || move.id === 'Recharge') pp = '&ndash;';
 						if (move.id === 'Recharge') move.type = '&ndash;';
 						if (name.substr(0, 12) === 'Hidden Power') name = 'Hidden Power';
+						var moveType = this.getMoveType(move, switchables[pos]);
```
Zarel: I still think it'd be better if for moves that vary, you got everything you knew about the move. Like, for moves with variable base power, you could display what base power it could be.

You could do that as two functions, though, in which case you wouldn't need to clone the object. Thoughts?

Joimer: We can make a new object that only has move information in a single function (and it will be usable in the future if there's more changes on moves... like pp or whatever) or use two functions, I think both options are good.

Zarel: I like two functions for now.

PP is already sent from the server so it's always accurate.

Zarel: So... are you going to change it to use two functions?

Joimer: Yes, I'm going to.

Joimer:
See #356

PR #309, opened 2015-01-15
Update learnsets-g6 in `update` githook

Slayer95:
- Since it's mandatory that file hashes are only updated after the file contents themselves have been updated, the index update is now wrapped in a function to be called once no files are pending update.
- The original version logged every new move acquired, but it's probably too spammy for a not-dedicated command.
- Adds `sugar` as a dependency.

### DIFF
--- a/githooks/update
+++ b/githooks/update
@@ -3,6 +3,7 @@
 /**
  * This script parses index.html and sets the version query string of each
  * resource to be the MD5 hash of that resource.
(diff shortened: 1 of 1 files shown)


Zarel:
Hm, "Updating file `data/learnsets-g6`..." takes a long time. Can you make it compare modified times so it doesn't run each time?

PR #312, opened 2015-01-17
Show tooltip for fainted pokemon

sirDonovan:
Moving the tooltip to a <code>span</code> allows it to be displayed for disabled buttons


Zarel:
Alternatively, make "disabled" a class instead of actually disabling the button.

I think I'd rather do that.

sirDonovan:
Yeah that's simpler, updated.

Zarel:
Do you correctly handle what happens when someone clicks a button that looks disabled? :|

sirDonovan:
Is something like this better? The messages are from ORAS.

PR #313, opened 2015-01-17
Update stat spreads on battle for gen 1

Joimer:
On gen 1, all stats are always maxed out so you may know them exactly.
Also, on gen 2 max DVs are 15, so max IVs in conversion are 30.
Also remove useless operations.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1043,16 +1043,26 @@
 					}
 					text += pokemon.stats['spe'] + '&nbsp;Spe</p>';
 				} else if (template.baseStats) {
-					var minSpe;
-					var maxSpe;
-					if (this.battle.tier === 'Random Battle') {
-						minSpe = Math.floor(Math.floor(2*template.baseStats['spe']+(0)+Math.floor((0)/4))*pokemon.level / 100 + 5);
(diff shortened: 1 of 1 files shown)


Slayer95:
Why not display the full available range? When the server gets fixed to allow all DVs combinations the client will be immediately compatible.
Also, it would be awesome if the Gen 2 tooltip accounted for shinyness.

Zarel:
Isn't gen 2 like that, too?

And shouldn't we display the range just in case? There might be cases in which someone intentionally wants to be slow or something.

Joimer:
Gen 2 has hidden powers and rollback stuff that makes people have less DV.
On Gen 1 there's literally zero reason to run lower speed. Unless you play to lose, that's it.

Slayer95:
Suicide Cup ftw

Zarel:
I'd say Suicide Cup is enough reason. As long as it's possible for users to set, I don't want to get bug reports "it said their Speed was faster but mine was faster, please fix".

Zarel:
I'm not hearing a counterargument to my reasoning, so I'm closing.

Joimer:
I had to update this to show the proper ranges on past gens anyways (as they are still different), but ok. I'll do it on a new pull request eventually.

PR #321, opened 2015-01-30
Protocol: include aliases in |queryresponse|rooms|

Morfent:
The server now includes a map of room aliases in the response to /cmd
rooms so the client can properly resolve them to their IDs.


Zarel:
See Zarel/Pokemon-Showdown#1483.

PR #323, opened 2015-01-30
Update |unlink|hide| message

sirDonovan:
Display the reveal button based on each user's screen instead of the current room's chat for cases such as a user being locked/banned for a PM offense without having any messages in the room.


Zarel:
I'm confused. Can you explain more clearly what this does?

Zarel:
Oh, I understand now.

PR #324, opened 2015-01-30
Toggle hiding unlinked users' chat messages

Morfent:
Users should be able to judge whether or not they want locked/banned users' chat messages to be hidden by default or not. Allowing staff to leave PMs and recent chat history revealed by default would help them moderate more easily as well.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -5464,6 +5464,7 @@ var Battle = (function () {
 			this.log('<div class="debug"><div class="chat"><small style="color:#999">[DEBUG] ' + Tools.escapeHTML(name) + '.</small></div></div>', preempt);
 			break;
 		case 'unlink':
+			if (Tools.prefs('nounlink')) return;
 			var user = toId(args[2]) || toId(args[1]);
 			var $messages = $('.chatmessage-' + user);
 			$messages.find('a').contents().unwrap();
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -518,6 +518,15 @@
 				Tools.prefs('showbattles', false);
 				return false;
 
+			case 'unpackhidden':
+				this.add('Locked/banned users\' chat messages: ON');
+				Tools.prefs('nounlink', true);
+				return false;
+			case 'packhidden':
+				this.add('Locked/banned users\' chat messages: HIDDEN');
+				Tools.prefs('nounlink', false);
+				return false;
+
 			case 'timestamps':
 				var targets = target.split(',');
 				if ((['all', 'lobby', 'pms'].indexOf(targets[0]) === -1) || targets.length < 2 ||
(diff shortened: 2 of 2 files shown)


js/client.js
```diff
@@ -2320,6 +2321,7 @@
 			buf += '<p><label class="optlabel"><input type="checkbox" name="nopastgens"'+(Tools.prefs('nopastgens')?' checked':'')+' /> Use modern sprites for past generations</label></p>';
 			buf += '<p><label class="optlabel"><input type="checkbox" name="notournaments"'+(Tools.prefs('notournaments')?' checked':'')+' /> Ignore tournaments</label></p>';
 			buf += '<p><label class="optlabel"><input type="checkbox" name="nolobbypm"'+(Tools.prefs('nolobbypm')?' checked':'')+' /> Don\'t show PMs in lobby chat</label></p>';
+			buf += '<p><label class="optlabel"><input type="checkbox" name="nounlinkhide"'+(Tools.prefs('nounlinkhide')?' checked':'')+' /> Don\'t hide locked/banned users\' chat messages</label></p>';
```
Zarel: Seeing as you can already click "show" to re-show the messages, I don't see a need to make it a checkbox. Console command is enough, I think.

Morfent: I don't think users should have to click "show" to reveal messages if they don't want any hidden in the first place,

Zarel: Which is why I said console command is enough.

Morfent: Oh, I completely misunderstood what you meant at first.

Morfent:
While I'm at it, should there be help documentation for the commands for battle messages and debug battle ones along with this?

Zarel:
If you'd like, yes.

js/client-chat.js
```diff
@@ -518,6 +518,15 @@
 				Tools.prefs('showbattles', false);
 				return false;
 
+			case 'showunlink':
```
Slayer95: Maybe /showremoved? (And, analogically, /hideremoved)

js/client-chat.js
```diff
@@ -518,6 +518,15 @@
...
+				this.add('Unlink chat messages: ON');
+				Tools.prefs('nohideunlink', true);
+				return false;
+			case 'hideunlink':
```
Zarel: I'm not sure if "show" and "hide" are very clear here.

js/client-chat.js
Morfent: What about /packhidden and /unpackhidden? Removed isn't really accurate for what ``|unlink|hide|`` is doing, but using /show and /hide and ending up /hidehidden as a command doesn't sound like a good idea

Morfent:
Changed the command names to /unpackhidden and /packhidden, added missing help documentation, and fixed inconsistencies in their grammar

js/battle.js
```diff
@@ -5464,6 +5464,7 @@ var Battle = (function () {
 			this.log('<div class="debug"><div class="chat"><small style="color:#999">[DEBUG] ' + Tools.escapeHTML(name) + '.</small></div></div>', preempt);
 			break;
 		case 'unlink':
+			if (Tools.prefs('nounlinkhide')) return;
```
Zarel: You use `nohideunlink` in the commands and `nounlinkhide` here.

If you're having trouble remembering which it is, this usually means that the pref is too long. How about `nounlink`?

Morfent: ``nounlink`` seems reasonable since both commands parsing ``|unlink|`` just deal with hiding messages. I shouldn't have forgotten to check that the commands work while I was focusing on writing /help instead

Morfent:
Changed the toggle from ``nounlinkhide``/``nohideunlink`` to ``nounlink``

Zarel:
I'd put the help changes in a different commit, but that's not _too_ important. Just as a note for next time.

Morfent:
Ok

PR #325, opened 2015-02-01
Fix crash in Precipice Blades animation

sirDonovan:


ascriptmaster:
I could have sworn this was tested, but I guess it wasn't

Maybe I got too excited and only tested Origin Pulse, but really I thought this was working

PR #326, opened 2015-02-01
Show the number of messages hidden in |unlink|hide|

sirDonovan:


Zarel:
As an aside, we have a rule against the text "Click here" which everyone seems to forget about :|

TheImmortal:
How about "Restore hidden messages."

Zarel:
"View N hidden messages"

PR #327, opened 2015-02-02
Only check if an address is already registered by hostname, not by IP

kotarou3:
When attempting to go on an unregistered server when there is already another registered server sharing the same IP/Port pair (such as in virtual hosting), you get redirected to the registered server, which
is undesirable.

This is important for me particularily because due to the recent change in server registration policy (only one server to be registered to a single user, and no registering on the behalf of others), I can't get my clients to register servers themselves due to this issue. (I host multiple servers on a single IP, but all through port 80)

(Code untested for obvious reasons)

### DIFF
--- a/crossdomain.php
+++ b/crossdomain.php
@@ -56,19 +56,14 @@ if (isset($PokemonServers[$config['host']])) {
 
 	// see if this is actually a registered server
 	if ($config['host'] !== 'localhost') {
-		$ip = gethostbyname($config['host']);
 		foreach ($PokemonServers as &$server) {
-			if (!isset($server['ipcache'])) {
-				$server['ipcache'] = gethostbyname($server['server']);
-			}
-			if ($ip === $server['ipcache']) {
-				if (($config['port'] === $server['port']) ||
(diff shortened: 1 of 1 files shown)


Zarel:
Wait, how is that even possible? I guess WebSocket has namespaces but does SockJS? Is that even an officially documented PS feature?

Well, that obviously isn't an officially documented PS feature, but how does it work?

kotarou3:
I have nginx acting as a reverse proxy to each server

Zarel:
but how do you decide where to send which thing? referer?

kotarou3:
The same way a single IP can service multiple websites - the Host HTTP header

Zarel:
Oh, yeah, that. WebSocket supports the Host HTTP header? Do all other SockJS transports support that?

kotarou3:
Even if they don't, nginx seamlessly proxies the connection. Both websocket and xhr work.

CreaturePhil:
I also had this problem when hosting on heroku. It redirected me to a already registered server.

PR #328, opened 2015-02-02
Reintroduce auto-linking without closing parenthesis

xfix:
This commit reintroduces change in 6b8b68d without regression on links like http://i.imgur.com/rcOnWA2.png.

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -330,7 +330,7 @@ var Tools = {
 				options.hidestrikethrough ? '$1' : '<s>$1</s>');
 		// linking of URIs
 		if (!options.hidelinks) {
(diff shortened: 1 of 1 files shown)


kotarou3:
Maybe it's time to replace that hell of a regex with a proper function for finding links?

Zarel:
It could at least be multiple simpler regexes.

Zarel:
On second thought, I'm not sure it could. I don't want it to detect a link inside some text that's already been linkified, which is why it's done this way.

xfix:
Well, that's a general issue with JavaScript - there are no multiline regexes (`/x`) that make regexes way more readable in other programming languages. No spaces, no linebreaks, it's line noise.

I think it possible to make this more readable by some trickery with `new RegExp` and string concatenation, but it still will be ugly. I'll look into that later.

xfix:
Closing because of a bug.

xfix:
See #330 now. This pull request is superior to this one.

PR #329, opened 2015-02-03
Prevent arbitrary code in Pokemon names

Morfent:
Pokemon's names can use /eval and /evalbattle without admin permission.
This probably shouldn't be the case.
Edit: this needs to be fixed to happen when the sides are generated


Zarel:
...wut?

PR #330, opened 2015-02-03
Rewrite URL parsing.

xfix:
Regular expression is now written using new RegExp constructor to make a more readable regular expression (even if I admit it's not that great because of `+` everywhere, but this is JavaScript, so no neat `/x` flag for regular expressions).

This heavily changes behaviour of auto-linking, in particular.

-   URLs can be in parenthesis, see 6b8b68d for more details and justification of a change.

-   More top level domains are allowed in URLs without http:// prefix. Anything that is two letter is allowed, as long followed by slash or colon.

-   Allow auto-linking URLs starting with `www.`, even with unrecognized TLD. Chances are if a domain has this particular subdomain, it's a website address.

-   Removed qmark.tk special case, this site is dead anyway.

-   Added support for `.co` TLD. This domain gets a lot of popularity, and it doesn't look like it could harm anything.

-   Added support for `.info` TLD. This is because the special case of handling 2-3 letter TLD doesn't handle `.info`.

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -278,6 +278,42 @@ var baseSpeciesChart = {
 	// others are hardcoded by ending with 'mega'
 };
 
+// Precompile often used regular expression for links.
+var domainRegex = '[a-z0-9\\-]+(?:[.][a-z0-9\\-]+)*';
+var parenthesisRegex = '[(][^\\s()<>]*[)]';
+var linkRegex = new RegExp(
+	'\\b' +
+	'(?:' +
+		'(?:' +
+			// When using www. or http://, allow any-length TLD (like .museum)
+			'(?:https?://|www[.])' + domainRegex +
+			'|' + domainRegex + '[.]' +
+				// Allow a common TLD, or any 2-3 letter TLD followed by : or /
+				'(?:com?|org|net|edu|info|us|jp|[a-z]{2,3}(?=[:/]))' +
+		')' +
+		'(?:[:][0-9]+)?' +
+		'\\b' +
+		'(?:' +
+			'/' +
+			'(?:' +
+				'(?:' +
+					'[^\\s()<>]' +
+					'|' + parenthesisRegex +
+				')*' +
+				// URLs usually don't end with punctuation, so don't allow
+				// punctuation symbols that probably aren't related to URL.
+				'(?:' +
+					'[^\\s`()<>\\[\\]{}\'".,!?;:]' +
+					'|' + parenthesisRegex +
+				')' +
+			')?' +
+		')?' +
+		'|[a-z0-9.]+\\b@' + domainRegex + '[.][a-z]{2,3}' +
+	')'
+,
+	'ig'
+);
+
 var Tools = {
 
 	resourcePrefix: (function() {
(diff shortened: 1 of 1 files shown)


js/battledata.js
```diff
@@ -330,8 +330,41 @@ var Tools = {
...
+					'(?:[:][0-9]+)?' +
+					'\\b' +
+					'(?:' +
+						'/' +
+						'(?:' +
+							'(?:' +
+								'[^\\s()<>]' +
+								'|' + parenthesisRegex +
+							')*' +
+							// URL cannot end with punctation characters
```
Zarel: Actually, they totally can

xfix: Well, sure they can (technically), but generally speaking it's odd to find an URL ending with a period, for example. While it's fine to allow punctuation at end if it's let's say URL field on a website (no reason to block it here), in this case I parse plain text, which is far more tricky than a simple URL field. If you see a period at end of URL, chances are it's not a part of URL, but rather a period at end of sentence.

Zarel: Yeah, but that should probably be clarified in the comment

Zarel: Also do we want to move this regex definition out of this function? I don't know how good JIT compilers are, but this is a function that's called multiple times per second.

ascriptmaster:
Great job! The commits should probably be squashed together, though.

Zarel:
What @ascriptmaster said.

xfix:
In my opinion, it's better to use `git merge --squash` on `master` branch side than to destroy commit history on feature branch, but I digress. Done.

ascriptmaster:
You can't `git merge --squash` when dealing with pull requests. That's why we need people who submit to squash, because if we do it post then it might mess up someone who happened to pull before we finish the process of pulling, squashing, and force-pushing again.

Slayer95:
We totally can do <code>git merge --squash</code> with no need to force-push from either side, but the green "Merge pull request" button is lovely :P

xfix:
@ascriptmaster At bottom of a page, there is a text saying "You can also merge branches on the command line." (assuming you can merge to the given repository). You can follow instructions from it, just using `--squash` on `git merge` step.

(that aside, I'm not advocating doing that, in my own opinion merging by squash [breaks `git bisect`](http://paul.stadig.name/2010/12/thou-shalt-not-lie-git-rebase-ammend.html), which is a quite useful tool, but I understand this project prefers clean history to that)

Slayer95:
Well, having WIP commits in a branch can also break <code>git bisect</code>, as it assumes that the history is divided between <code> Recent bad states | Old good states </code>. WIP commits can break and fix tests, so... [I'll just let one of the Node genius minds talk for me](http://blog.izs.me/post/37650663670/git-rebase).

Zarel:
@Slayer95 is correct on both counts.

We _could_ do `git merge --squash`, but the green Merge Pull Request button is seriously much easier.

And I honestly don't see any point to keeping your branch history messy. _Especially_ if the change in total is as simple as this one. How much does it really help anyone to know that you once had a buggier version of this that never went anywhere?

Zarel:
By the way, @xfix, do you go on PS very often? What's your PS username? You've contributed plenty of things to PS so far and I'd like to talk to you more.

xfix:
@Zarel: It's xfix, but I doubt you will see me often, because from what I see, we live in different timezones (I live in Europe). I'm quite often, although often it's just leaving the tab open (and the only rooms I visit is Skillmons and RarelyUsed, I never visit Lobby).

PR #331, opened 2015-02-06
Update jQuery to version 2.1.3

Morfent:
``$`` is undefined once the script loads on newer versions of Firefox.


Slayer95:
I don't have this issue in Firefox 35.0.1 (Win 7 x86) ... Are you using Aurora?

Morfent:
Yes

Slayer95:
May I say "don't"? Or is it a "feature" expected in the stable branch?

Morfent:
It looks more like an issue with the old version of jQuery than one with Aurora. Apart from e10s, I don't think there are any new features that could cause this to happen. I couldn't find anything in Bugzilla related to jQuery or browser scripts that matches the problem here, and the latest version fixed it when I tested it.

PR #332, opened 2015-02-07
Graphics: Process the [animalias] tag for move animations

ascriptmaster:
The [animalias] tag allows moves (usually created in format mods) to use
already-established move animations.

Reliant on Zarel/Pokemon-Showdown#1511


Zarel:
You don't need this. Just send a [still] move and a -moveanim separately.

PR #334, opened 2015-02-09
Several improvements to |unlink|hide|

sirDonovan:
- Prevent messages from being hidden in replays
- Improve wording on the reveal button
- Highlight revealed messages


Zarel:
>Prevent messages from being hidden in replays

Why would this be desirable? I don't want spam in replays, either.

PR #335, opened 2015-02-15
Improve sprite functionality

sirDonovan:
- Move all sprite data to one file (<code>pokedex-mini.js</code>)
- Ensure all sprites are shown with the correct size
- Remove <code>loadSpriteData</code> calls that have been rendered unnecessary
- Update instances of <code>BattlePokemonSprites</code>
- Update sprite option text


Slayer95:
Do we really want this? I'd say that ideally we would only load past gen data on-demand. Probably by doing it when issuing/receiving challenges in past gens or going to a past gen ladder / teambuilder.

Zarel:
PS data would probably benefit from being loaded on-demand rather than having the huge up-front download.

On the other time, `pokedex-mini-bw.js` by itself is small enough that I'm not sure if it really deserves to be split up and loaded on-demand.

Discuss.

Zarel:
In other words: the current system is probably the worst of both worlds. The only hard question is, where do we go from here?

Slayer95:
And what about the future <code>pokedex-mini-dpp.js</code>, <code>pokedex-mini-rse.js</code>, etc? Are we going to pack them all together in a single file? There is probably no point in doing that if we are going to load old move/abilities data separately.

The most important thing to remember here is that a huge fraction of PS userbase only plays current gen. I guess we can ensure that old data is loaded timely by hooking at battle preparation events (as I said before: challenges, teambuilding, etc).

Additionally, we could add an expirable automatically set preference so that if a player plays X or more games of a certain gen in a single browsing session, the data for that gen will be loaded as early as possible for the next month or something ?_?

sirDonovan:
Animated battle sprites weren't used until BW, why would we need to add <code>pokedex-mini-dpp.js</code>, <code>pokedex-mini-rse.js</code>, etc?

One option I thought of is getting the numbers for Pokemon cries from <code>BattlePokedex</code>, since they're already loaded in that file. Removing the <code>num</code> fields from <code>pokedex-mini.js</code> and <a href="https://github.com/sirDonovan/Pokemon-Showdown-Client/blob/test/data/pokedex-mini.js">restructuring the file a bit</a> gets it down to 70.755kb, compared to the existing file's 61.46kb.

Slayer95:
Oh, I was under the impression that the fixed sprites for old gens did not have a fixed size... (maybe that was the case last year?) It would still apply for future gens though...

<code>pokedex-mini.js</code> needs the numbers to figure out the cries in replays, as <code>pokedex.js</code> isn't loaded for them.

Zarel:
How many sprite gens do we really need? I can't imagine anyone caring enough about the difference between Gen 7 and Gen 6 models for people to complain that we're using Gen 7 model animations for Gen 6.

Zarel:
The `num` fields in `pokedex-mini.js` are necessary because the replay player doesn't load `pokedex.js`.

sirDonovan:
Yeah, forgot about that. Would renaming the files to the species' names as Slayer mentioned be an option?

Zarel:
They're still needed for icons.

sirDonovan:
Hm, that's true. Adding the numbers <a href="https://github.com/sirDonovan/Pokemon-Showdown-Client/blob/test/data/pokedex-mini.js">back to the file</a> brings it to 77.511kb. If that's not a favorable increase from the existing file, I can try to think of something else.

Zarel:
We should rename the cries to use the same naming scheme as the sprites, though.

Zarel:
@sirDonovan, is anything from here still needed?

I don't think the bw and xy spritedata should be in the same file.

sirDonovan:
No, the other issue has been fixed.

PR #337, opened 2015-02-18
SyntaxError handling when adding regex highlights

Morfent:
If the user attempts to add any regular expressions that throw a syntax
error on RegExp object construction, they can see the first pattern that
throws an error along with its error message without breaking their
entire highlight list.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -570,8 +570,16 @@
 					switch (targets[0]) {
 					case 'add':
 						for (var i=1, len=targets.length; i<len; i++) {
-							highlights.push(targets[i].trim());
+							if (/[\\^$*+?()|{}[\]]/.test(targets[i])) {
+								// Catch any errors thrown by newly added regular expressions so they don't break the entire highlight list
+								try {
+									new RegExp(targets[i]);
+								} catch (e) {
+									return this.add(e.message.substr(0, 28) === 'Invalid regular expression: ' ? e.message : 'Invalid regular expression: /' + targets[i] + '/: ' + e.message);
+								}
+							}
 						}
+						highlights = highlights.concat(targets.slice(1));
 						this.add("Now highlighting on: " + highlights.join(', '));
 						// We update the regex
 						app.highlightRegExp = new RegExp('\\b('+highlights.join('|')+')\\b', 'i');


js/client-chat.js
```diff
@@ -567,11 +567,20 @@
...
-							highlights.push(targets[i].trim());
-						}
+						var test;
+						for (var i=0, len=targets.length; i<len; i++) {
+							if (/[\\^$*+?()|{}[\]]/.test(targets[i])) {
+								// Catch any errors thrown by newly added regular expressions so they don't break the entire highlight list
+								try {
+									test = new RegExp(targets[i]);
+								} catch (e) {
+									return this.add(e.message.substr(0, 28) === 'Invalid regular expression: ' ? e.message : 'Invalid regular expression: /' + targets[i] + '/: ' + e.message);
```
Zarel: You don't even tell people that the valid regexes earlier are added but the valid regexes later aren't?

Morfent: What?

Morfent: The error message on browsers other than Chrome/Chromium don't include "Invalid regular expression: " or the regex in which the error occurs in the error message, so this line adds those to keep the error message consistent across all browsers

Morfent: Oh, I see what you mean now

Morfent:
It now adds the highlights only if they all check out

js/client-chat.js
```diff
@@ -567,11 +567,20 @@
 					for (var i=0, len=targets.length; i<len; i++) {
 						targets[i] = targets[i].trim();
 					}
-					switch (targets[0]) {
+					switch (targets.shift()) {
 					case 'add':
-						for (var i=1, len=targets.length; i<len; i++) {
-							highlights.push(targets[i].trim());
-						}
```
Zarel: Just move this code down to replace `highlights = highlights.concat(targets);`

Zarel: It would also eliminate the need to `.shift()`

Morfent: So you want it to loop through ``targets`` twice? What would be the difference between that and ``Array.prototype.push.apply(highlights, targets)`` or ``highlights = highlights.concat(targets)``?

Zarel: 1. internally, `concat` loops through `targets`, anyway; it's only slightly faster than actually writing out the loop.

2. `targets[i].trim()` is applied, which your code doesn't do.

Morfent: Ah. ``targets[i].trim()`` is already applied before the switch statement, though

Morfent:
Removed the need for ``.shift()``

js/client-chat.js
```diff
@@ -569,9 +569,18 @@
...
+							if (/[\\^$*+?()|{}[\]]/.test(targets[i])) {
+								// Catch any errors thrown by newly added regular expressions so they don't break the entire highlight list
+								try {
+									test = new RegExp(targets[i]);
+								} catch (e) {
+									return this.add(e.message.substr(0, 28) === 'Invalid regular expression: ' ? e.message : 'Invalid regular expression: /' + targets[i] + '/: ' + e.message);
+								}
+							}
 						}
+						highlights = highlights.concat(targets);
```
Zarel: Don't you mean `highlights.concat(targets.slice(1));`?

js/client-chat.js
Zarel: Oh, you're right.

js/client-chat.js
Morfent: Argh, yes

Morfent:
Fixed the ``concat``

js/client-chat.js
```diff
@@ -569,9 +569,18 @@
 					}
 					switch (targets[0]) {
 					case 'add':
+						var test;
 						for (var i=1, len=targets.length; i<len; i++) {
-							highlights.push(targets[i].trim());
+							if (/[\\^$*+?()|{}[\]]/.test(targets[i])) {
+								// Catch any errors thrown by newly added regular expressions so they don't break the entire highlight list
+								try {
+									test = new RegExp(targets[i]);
```
Zarel: We don't need a var test; we can just `new RegExp(targets[i]);`

Even if we needed the var test, it should be declared here, not at the top of the case.

Morfent:
Removed the var test

PR #339, opened 2015-02-26
Tournaments: highlight users in the bracket

sirDonovan:
Screenshot: http://i.imgur.com/7KWoNwh.png

### DIFF
--- a/js/client-chat-tournament.js
+++ b/js/client-chat-tournament.js
@@ -504,6 +504,7 @@
 		};
 
 		TournamentBox.prototype.generateBracket = function (data) {
+			var user = app.user.get('name');
 			if (data.type === 'tree') {
 				if (!data.rootNode)
 					return;
@@ -594,6 +595,7 @@
 					var elem = d3.select(this);
 					if (node.children.length === 0) {
 						elem.classed('tournament-bracket-tree-node-team', true);
+						if (node.team && node.team === user) elem.classed('tournament-bracket-tree-node-user', true);
 						elem.append('svg:text').text(node.team || "Unavailable");
 					} else {
 						elem.classed('tournament-bracket-tree-node-match', true);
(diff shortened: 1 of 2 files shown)


Zarel:
Hm. We could use username colors for this.

sirDonovan:
It's not as effective for users with darker colors: http://i.imgur.com/nwlqM2F.png. If you meant using their colors instead of the blue, that wouldn't help users who have colors similar to the existing red/green.

Zarel:
We could lighten the existing red/green until they look good.

sirDonovan:
I tried multiple opacity levels, from the current down to 1%. Highlighted usernames don't stand out enough to be easily seen while scrolling through the bracket, compared to highlighted nodes: http://i.imgur.com/VgGwm3e.png http://i.imgur.com/lW8ccmf.png http://i.imgur.com/K9V6Nq1.png

TheImmortal:
@Zarel is there a need to use username colors? I like the blue, personally.

kotarou3:
Agreeing with @TheImmortal

Zarel:
I don't see the need for any highlighting at all, in that case...

TheImmortal:
@sirDonovan didn't mention why this change but I assume because it's useful for huge tournaments.

Zarel:
Well, at least pick nicer colors. The example screenshot was really ugly, which might be part of why it seems useless to me.

Slayer95:
What about adding some way to center the tournament bracket on the current user battle? Maybe listen to the Spacebar key pressing?

Zarel:
PS has text boxes that should be usually-focused, which makes Spacebar a bit of a suboptimal choice for keyboard shortcut.

kotarou3:
@Slayer95 
I was originally thinking of adding a button of sorts to centre the user, but I never got around to that.

sirDonovan:
How's this color? http://i.imgur.com/vo71KBH.png

sirDonovan:
Superseded by https://github.com/Zarel/Pokemon-Showdown-Client/pull/515.

PR #340, opened 2015-03-06
/accept with no userid accepts sole challenges

Morfent:
Merge this before the server pullreq.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -420,6 +420,58 @@
 				challenge(targets);
 				return false;
 
+			case 'accept':
+				var userid = toId(target);
+				if (userid) {
+					var $challenge = $('.pm-window').filter('div[data-userid="' + userid + '"]').find('button[name="acceptChallenge"]');
+					if (!$challenge.length) {
+						this.add('You do not have any pending challenge from ' + target + ' to accept.');
+						return false;
+					}
+					$challenge[0].click();
+					return false;
+				}
+
+				var $challenges = $('.challenge').find('button[name=acceptChallenge]');
+				if (!$challenges.length) {
+					this.add('You do not have any pending challenges to accept.');
+					return false;
+				}
+				if ($challenges.length > 1) {
+					this.add('You need to specify a user if you have more than one pending challenge to accept.');
+					this.parseCommand('/help accept');
+					return false;
+				}
+
+				$challenges[0].click();
+				return false;
+			case 'reject':
+				var userid = toId(target);
+				if (userid) {
+					var $challenge = $('.pm-window').filter('div[data-userid="' + userid + '"]').find('button[name="rejectChallenge"]');
+					if (!$challenge.length) {
+						this.add('You do not have any pending challenge from ' + target + ' to reject.');
+						return false;
+					}
+					$challenge[0].click();
+					return false;
+				}
+
+				var $challenges = $('.challenge').find('button[name="rejectChallenge"]');
+				if (!$challenges.length) {
+					this.add('You do not have any pending challenges to reject.');
+					this.parseCommand('/help reject');
+					return false;
+				}
+				if ($challenges.length > 1) {
+					this.add('You need to specify a user if you have more than one pending challenge to reject.');
+					this.parseCommand('/help reject');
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -420,6 +420,43 @@
 				challenge(targets);
 				return false;
 
+			case 'accept':
+				if (toId(target)) return text;
```
Zarel: You should still send the team even if you specify the user...

js/client-chat.js
```diff
@@ -420,6 +420,43 @@
...
+					this.parseCommand('/help reject');
+					return false;
+				}
+				if ($challenges.length > 1) {
+					app.addPopupMessage('You need to specify a user if you have more than one pending challenge to reject.');
+					this.parseCommand('/help reject');
+					return false;
+				}
+
+				return text;
```
Zarel: You might as well specify the username you're rejecting from here, then the server doesn't need any changes.

js/client-chat.js
```diff
@@ -420,6 +420,43 @@
...
+					return false;
+				}
+				if ($challenges.length > 1) {
+					app.addPopupMessage('You need to specify a user if you have more than one pending challenge to accept.');
+					this.parseCommand('/help accept');
+					return false;
+				}
+
+				var format = $('button[name="format"]', $challenges[0]).val();
+				if (window.BattleFormats[format].team === 'preset' || Object.keys(Storage.teams).length) return text;
```
Zarel: You still need to actually send the team. :|

You can just simulate a click on the accept button if you don't want to go through all that work. In fact, that's in general probably better for  you.

Morfent: Simulating a click with both commands would also handle the HTML of the challenge box properly, which I hadn't caught earlier.

Morfent:
The commands now simulate clicking the accept/reject button.

js/client-chat.js
```diff
@@ -747,6 +778,14 @@
 					this.add('/challenge - Open a prompt to challenge a user to a battle.');
 					this.add('/challenge [user] - Challenge the user [user] to a battle.');
 					return false;
+				case 'accept':
+					this.add('/accept - Accept a challenge if only one is pending.');
+					this.add('/accept [user] - Accept a challenge from the specified user.');
```
Zarel: You still haven't implemented this part.

This still doesn't send the team etc.

Morfent: Is it supposed to send the team even if the command fails?

Zarel: You made it so `/accept` sends the team.

You haven't made it so `/accept [user]` sends the team.

Morfent: Oh. It might help if ``/accept [user]`` actually worked. I really need to get better with testing

Morfent:
``/accept [user]`` works properly now.

js/client-chat.js
```diff
@@ -420,6 +420,58 @@
 				challenge(targets);
 				return false;
 
+			case 'accept':
+				var userid = toId(target);
+				if (userid) {
+					var $challenge = $('.pm-window').filter('div[data-userid="' + userid + '"]').find('button[name="acceptChallenge"]');
+					if (!$challenge.length) {
+						app.addPopupMessage('You do not have any pending challenge from ' + target + ' to accept.');
```
Zarel: Should it really be a popup message? I think inline would work better.

js/client-chat.js
```diff
@@ -420,6 +420,58 @@
...
+						return false;
+					}
+					$challenge[0].click();
+					return false;
+				}
+
+				var $challenges = $('.challenge').find('button[name=acceptChallenge]');
+				if (!$challenges.length) {
+					app.addPopupMessage('You do not have any pending challenges to accept.');
```
Zarel: Ditto your other popup messages.

Popup messages in general are really bad UI; they should in general be replaced with other things.

Morfent: If that's the case, should the popup message in ``acceptChallenge`` also be made inline? I only made these as popups to follow its style.

Morfent:
Changed the popups to regular responses.

PR #341, opened 2015-03-07
Add "Ignore Spectators" button

sirDonovan:


js/client-battle.js
```diff
@@ -683,6 +684,9 @@
 		setTimer: function(setting) {
 			this.send('/timer '+setting);
 		},
+		toggleIgnoreSpects: function() {
+			this.battle.ignoreSpects = !this.battle.ignoreSpects;
```
Zarel: You're not even going to update the controls after doing this?

sirDonovan: Oh right, updated.

js/client-battle.js
```diff
@@ -571,13 +571,14 @@
 				if (this.choice && this.choice.waiting && !this.finalDecision) {
 					buf += '<button name="undoChoice">Cancel</button>';
 				}
-				buf += '</p>';
+				buf += '</p><p class="timer">';
```
Zarel: Do you have a screenshot?

I think I forgot we even had a `class="timer"`

sirDonovan: Here's a screenshot: http://i.imgur.com/oTo2Hog.png

Zarel: My main issue is that this is a place where the player won't see it normally...

Zarel: Like, if they haven't decided on a move yet.

sirDonovan: Maybe we could add options to the main control screen? http://i.imgur.com/d8jHTXv.png

Ridaz: May I suggest using using a tone of gray instead of green, for the options color?
Might blend better with the UI colors imo

Zarel:
sirDono, ideally those would be in a fixed place, and they wouldn't be level with the other ones.

Joimer:
@sirDonovan

sirDonovan:
Sorry for the delay. How is this? http://i.imgur.com/v6uwV4U.png

Zarel:
Still isn't a fixed/consistent place for it...

Ridaz:
How about next to the battle chat box, or under "What will x do?"

sirDonovan:
For reference, here's what that looks like: http://i.imgur.com/QECUBua.png

Joimer:
I'd sugges top right, just next to the chat.

Zarel:
I really don't have a good solution to this. We don't have a good spot to put this at all.

Maybe in the chat box? Between the input area and the log itself?

sirDonovan:
Here's what it looks like in the chat box: http://i.imgur.com/qggvgj0.png

Zarel:
You could put it like actually in the chat box. Like, under the preempt-box.

SolarisFox:
If it's under the preempt-box that would create a space for potentially adding another button in the future if it seems to be wanted, like may be a "Ignore Opponent" button or something.

sirDonovan:
Alright, moved it under. Yeah @SolarisFox , I have something like that in mind which I'll submit once this is done.

Zarel:
Hm, does this affect how autoscroll works?

sirDonovan:
I tested a few battles and it doesn't appear to affect it.

Zarel:
Well, specifically I think a naïve implementation would interfere with the autoscroll threshold (the amount you have to scroll up before autoscroll disables itself).

Zarel:
https://github.com/sirDonovan/Pokemon-Showdown-Client/blob/ebe554a90d60a91f64072a271f46a47ff497d8c6/js/battle.js#L2308

This line, specifically, needs to account for the battle options panel.

js/client-battle.js
```diff
@@ -22,6 +22,9 @@
 
 			this.$chat = this.$chatFrame.find('.inner');
 
+			this.$options = $('<div class="battle-options"></div>').insertAfter(this.$chatFrame.find('.inner-preempt')).html('<button name="toggleIgnoreSpects">' + (this.battle.ignoreSpects ? 'Unignore' : 'Ignore') + ' Spectators</button>');
+			this.battle.optionsElem = this.$options;
```
Zarel: According to the new Google thing we saw, we should initialize this in the constructor/prototype in `battle.js`.

And then we'd just do `this.$options = this.battle.optionsElem;` here.

Also prevents `Battle.prototype.log` from crashing in replays where `optionsElem` wouldn't otherwise exist.

js/client-battle.js
sirDonovan: Updated.

Zarel: To clarify, by "new Google thing we saw", I meant the V8 optimization thing: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers

js/battle.js
```diff
@@ -2215,6 +2215,8 @@ var Battle = (function () {
 		this.logElem = this.logFrameElem.children();
 		this.logFrameElem.append('<div class="inner-preempt"></div>');
 		this.logPreemptElem = this.logFrameElem.children().last();
+		this.logFrameElem.append('<div class="battle-options"></div>');
+		this.optionsElem = this.logFrameElem.children().last().html('<button name="toggleIgnoreSpects">' + (this.ignoreSpects ? 'Unignore' : 'Ignore') + ' Spectators</button>');
```
Zarel: This second line should be moved to `client-battle.js`. It's not like it does anything in replays.

Zarel: Oh, wait, just the `.html(...)` part should be moved. The rest can stay.

sirDonovan: Updated

Slayer95:
@Zarel, you still want these to be fixed at the top, right?

js/battle.js
```diff
@@ -2305,7 +2307,7 @@ var Battle = (function () {
 		if (window.console && console.log) console.log(text);
 	};
 	Battle.prototype.log = function (html, preempt) {
-		var willScroll = (this.logFrameElem.scrollTop() + 60 >= this.logElem.height() + this.logPreemptElem.height() - this.logFrameElem.height());
+		var willScroll = (this.logFrameElem.scrollTop() + 60 >= this.logElem.height() + this.logPreemptElem.height() - this.optionsElem.height() - this.logFrameElem.height());
```
Slayer95: ``logFrameElem`` already includes ``optionsElem``. Do I miss something?

Slayer95: It seems that this change made it so that a single mouse-wheel scroll (2.5 lines) no longer prevents autoscroll.

Zarel:
Define "stickied at the top". Do you have alternate ideas?

js/battle.js
Zarel: Perhaps this should be hardcoded, then; replace the 60 with a larger number instead of adding an element that shouldn't be added...

Zarel:
I don't have any problem with the current implementation but if you have ideas for improvement, I'm always open to that.

js/battle.js
Slayer95: In principle we can just remove the ``optionsElem`` term from the RHS, can't we?
I am mildly bothered by the magical number 60, but I am not sure if that's what you mean. Why should it be a larger number anyway?

Slayer95:
The Ignore Spectators / Players checkboxes are not very discoverable, and often people have no idea that they exist, because they are currently hidden when the log is scrolled. Instead, the scroll could be restricted to the ``.inner`` element, as shown in [the screenshot](http://i.imgur.com/ubAI9nn.png).

Zarel:
We really should implement a battle clock (to do sometime during the timer refactor) but I think it belongs in the battle window itself. All my battle redesigns so far have included a clock.

The checkboxes aren't supposed to be very in-your-face like they would be if they were always-there. I definitely don't want them always-there; they'd get so annoying...

Slayer95:
Fair enough. Then I'll leave this as is. And regarding the clock, I'll probably stub the markup until you get to the redesign.

Zarel:
You can just put it under the corresponding trainer, for now.

PR #342, opened 2015-03-07
Team/format button style fixes

Morfent:
The button's text is no longer offset vertically by the arrow added by
::after, and the background gradients now display properly on Firefox.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -484,7 +484,7 @@ p.or:after {
 	background: #EEEEEE;
 	box-shadow: inset 0 1px 0 #FFFFFF;
 	background: -webkit-gradient(linear, left top, left bottom, from(#edf2f8), to(#d7e3ec));
-	background: linear-gradient(top,  #edf2f8,  #d7e3ec);
+	background: linear-gradient(to bottom,  #edf2f8,  #d7e3ec);
 	color: black;
 }
 .select.teamselect {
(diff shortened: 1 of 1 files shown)


style/client.css
```diff
@@ -494,6 +495,7 @@ p.or:after {
 	border-color: #555555;
 	background: #FFFFFF;
 	background: -webkit-gradient(linear, left top, left bottom, from(#dce4ec), to(#c6d3e0));
+	background: -moz-linear-gradient(top,  #dce4ec, #c6d3e0);
```
Zarel: Are you telling me Firefox still doesn't support unprefixed `linear-gradient`? It's literally 2015. I find this hard to believe.

Zarel:
Can you clarify what this does? Preferably with before/after screenshots?

style/client.css
kotarou3: [It is supported](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient#Browser_compatibility)

Morfent:
These were the changes from just ``-moz-linear-gradient`` http://i.imgur.com/2kiHpUB.png
The arrow's top margin is screwing up its alignment with the text. I'll remove it after fixing ``linear-gradient``.

Morfent:
Added.

style/client.css
Zarel: Okay, can you change it to `to bottom` and see if it still works in Chrome?

Zarel:
I'm still confused by:

> The button's text is no longer offset vertically by the arrow added by `::after`

Morfent:
That was wrong. On Firefox, the text and the arrow properly like Chrome does. Since the text's height is lesser than the arrow's, it's offset from the centre of the button. The arrow's top margin doesn't make much of a difference in how it's displayed, so removing it makes the offset less noticable. The team select button's arrow is still off centre, so I'll need to write a proper fix for this.

PR #347, opened 2015-04-09
Add hover text showing battlers in battles

ScottehMax:
Screenshot:
Screenshot


Zarel:
This doesn't look like your usual name. Am I going crazy?

Zarel:
Git blame brought me back four years!

Anyway you can put `title` directly on the `<a>` tag, you don't need to add a `<span>` tag just to add a title.

PR #348, opened 2015-04-12
Add /clearpms command

sirDonovan:
Suggested on the forums with decent support: http://www.smogon.com/forums/threads/suggestions.3534365/page-2#post-6158958

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -521,6 +521,19 @@
 				}
 				return false;
 
+			case 'clearpms':
+				var $pms = $('.pm-window');
+				if (!$pms.length) {
+					this.add('You do not have any PM windows open.');
+					return false;
+				}
+				$pms.each(function () {
+					app.rooms[''].closePM($(this).data('userid'));
+					$(this).find('.inner').empty();
+				});
+				this.add("All PM windows cleared and closed.");
+				return false;
+
 			case 'nick':
 				if (target) {
 					app.user.rename(target);


sirDonovan:
@Zarel is this okay?

Zarel:
For that use case, selecting a PM box and mashing Esc works well enough...

Zarel:
This command, on the other hand, doesn't interact well with challenges and some other situations. Why not just use `closePM`?

sirDonovan:
The suggester wanted to clear all of the chat histories along with closing the PM windows, which <code>closePM</code> doesn't achieve. I could add on to this so that it cancels challenges and any other PM situations?

Zarel:
You could still call `closePM` and then clear chat histories afterwards?

You probably want some sort of feedback, too, so it's clear exactly what happened.

sirDonovan:
Updated to use <code>closePM</code> and added a message.

js/client-chat.js
```diff
@@ -521,6 +521,14 @@
 				}
 				return false;
 
+			case 'clearpms':
+				$('.pm-window').each(function () {
+					app.rooms[''].closePM($(this).data('userid'));
+					$(this).find('.inner').empty();
+				});
+				this.add("All PM windows cleared.");
```
Zarel: Maybe a different message if no PM windows are open?

Zarel: Also, it might be clearer to write: `All PM windows cleared and closed.`

So you know that "cleared" doesn't just mean "closed".

sirDonovan: Updated that line and added the other message.

PR #350, opened 2015-04-12
Fix gen 5 sprite sizes in battles

QuiteQuiet:
Reloading the BattePokemonSprites to bw for gen 5 battles.

This is by no means perfect (and I'm unsure what the correct location for this is) and require a bit of time for pokedex-mini / pokedex-mini-bw to load properly.

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -811,17 +811,17 @@ var Tools = {
 		return type;
 	},
 
-	loadedSpriteData: 'xy',
+	loadedSpriteData: {'xy':1, 'bw':0},
 	loadSpriteData: function(gen) {
-		if (this.loadedSpriteData === gen) return;
-		this.loadedSpriteData = gen;
+		if (this.loadedSpriteData[gen]) return;
+		this.loadedSpriteData[gen] = 1;
 
 		var path = $('script[src*="pokedex-mini.js"]').attr('src');
-		var qs = path.split('?')[1] || '';
+		var qs = '?' + (path.split('?')[1] || '');
 		path = (path.match(/.+?(?=data\/pokedex-mini\.js)/) || [])[0] || '';
 
 		var el = document.createElement('script');
-		el.src = path + 'data/pokedex-mini' + (gen !== 'xy' ? '-' + gen : '') + '.js' + (qs ? '?' + qs : '');
+		el.src = path + 'data/pokedex-mini-bw.js' + qs;
 		document.getElementsByTagName('body')[0].appendChild(el);
 	},
 	getSpriteData: function(pokemon, siden, options) {
@@ -847,7 +847,12 @@ var Tools = {
 			facing = 'back';
 		}
 
-		var animationData = window.BattlePokemonSprites && BattlePokemonSprites[pokemon.speciesid];
+		var animationData = {};
+		if (options.gen === 5 && window.BattlePokemonSpritesBW) {
+			animationData = BattlePokemonSpritesBW && window.BattlePokemonSpritesBW[pokemon.speciesid];
+		} else {
+			animationData = BattlePokemonSprites && window.BattlePokemonSprites[pokemon.speciesid];
+		}
 		if (animationData) {
 			var num = '' + animationData.num;
 			if (num.length < 3) num = '0' + num;
--- a/js/client-chat-tournament.js
+++ b/js/client-chat-tournament.js
@@ -274,8 +274,11 @@
 						break;
 
 					case 'start':
-						if (!this.info.isJoined)
+						if (!this.info.isJoined) {
 							this.toggleBoxVisibility(false);
+						} else if (this.info.format.substr(0, 4) === 'gen5' && !Tools.loadedSpriteData['bw']){
+							Tools.loadSpriteData('bw');
+						}
 						this.room.$chat.append("<div class=\"notice tournament-message-start\">The tournament has started!</div>");
 						break;
 
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -383,6 +383,7 @@
 				$formatButton.addClass('preselected')[0].disabled = true;
 				$teamButton.addClass('preselected')[0].disabled = true;
 				$searchForm.find('button.big').html('<strong><i class="icon-refresh icon-spin"></i> Searching...</strong>').addClass('disabled');
+				if ($formatButton.val().substr(0, 4) === 'gen5' && !Tools.loadedSpriteData['bw']) Tools.loadSpriteData('bw');
 			} else {
 				var format = $formatButton.val();
 				var teamIndex = $teamButton.val();
@@ -405,6 +406,7 @@
 				this.openPM(' '+i, true);
 			}
 			var self = this;
+			var atLeastOneGen5 = false;
 			this.$('.pm-window').each(function(i, el) {
 				var $pmWindow = $(el);
 				var userid = $pmWindow.data('userid');
@@ -420,6 +422,7 @@
 					buf += '<p><label class="label">Team:</label>'+self.renderTeams(format)+'</p>';
 					buf += '<p class="buttonbar"><button name="acceptChallenge"><strong>Accept</strong></button> <button name="rejectChallenge">Reject</button></p></form>';
 					$challenge.html(buf);
+					if (format.substr(0, 4) === 'gen5') atLeastOneGen5 = true;
 				} else {
 					var $challenge = $pmWindow.find('.challenge');
 					if ($challenge.length) {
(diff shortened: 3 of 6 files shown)


Zarel:
This seems like the wrong fix.

Who made this loader in the first place? @Slayer95? @Joimer?

Joimer:
The loader was already there when I added old gen sprites. Btw this fix does not solve it properly due to cache, I've already some changes in my client, the proper fix is to change the loader itself.

Slayer95:
Quinella

QuiteQuiet:
@Zarel, does this look better? It loads the BW sprite data the first time you make/receive a challenge, when you attempt to search for a battle, or select a gen 5 meta in the teambuilder. Then it keeps the sprite information as long as the session is ongoing so it's reusable.

js/client-mainmenu.js
```diff
@@ -405,6 +406,7 @@
 				this.openPM(' '+i, true);
 			}
 			var self = this;
+			var gen5battle = false;
```
Zarel: Might want to name this something clearer, like `atLeastOneGen5`?

js/battledata.js
```diff
@@ -811,17 +811,17 @@ var Tools = {
...
+		if (this.loadedSpriteData[gen]) return;
+		this.loadedSpriteData[gen] = 1;
 
 		var path = $('script[src*="pokedex-mini.js"]').attr('src');
 		var qs = path.split('?')[1] || '';
 		path = (path.match(/.+?(?=data\/pokedex-mini\.js)/) || [])[0] || '';
 
 		var el = document.createElement('script');
-		el.src = path + 'data/pokedex-mini' + (gen !== 'xy' ? '-' + gen : '') + '.js' + (qs ? '?' + qs : '');
+		el.src = path + 'data/pokedex-mini-bw.js' + (qs ? '?' + qs : '');
```
Zarel: uh, what are we querystringing here?

js/battledata.js
```diff
@@ -811,17 +811,17 @@ var Tools = {
...
-	loadedSpriteData: 'xy',
+	loadedSpriteData: {'xy':1, 'bw':0},
 	loadSpriteData: function(gen) {
-		if (this.loadedSpriteData === gen) return;
-		this.loadedSpriteData = gen;
+		if (this.loadedSpriteData[gen]) return;
+		this.loadedSpriteData[gen] = 1;
 
 		var path = $('script[src*="pokedex-mini.js"]').attr('src');
 		var qs = path.split('?')[1] || '';
```
Zarel: We might want to add `'?'` here so we don't have to do the convoluted `(qs ? '?' + qs : '')` thing later.

js/battledata.js
Zarel: nvm, figured it out

js/battledata.js
```diff
@@ -847,7 +847,8 @@ var Tools = {
 			facing = 'back';
 		}
 
-		var animationData = window.BattlePokemonSprites && BattlePokemonSprites[pokemon.speciesid];
+
+		var animationData = ((options.gen === 5 && typeof(BattlePokemonSpritesBW) !== 'undefined') ? BattlePokemonSpritesBW && window.BattlePokemonSpritesBW[pokemon.speciesid] : BattlePokemonSprites && window.BattlePokemonSprites[pokemon.speciesid]);
```
Zarel: We may want to split this into multiple lines.

js/battledata.js
QuiteQuiet: I assume you mean <code>'?' + path.split('?')[1] || ''</code>, so it looks like <code>(qs ? qs : '')</code>?

Zarel: `('?' + path.split('?')[1] || '')` is equivalent to `'?' + path.split('?')[1]` because `'?' + path.split('?')[1]` is never falsy.

On the other hand, `'?' + (path.split('?')[1] || '')` is fine.

And since `qs` is now guaranteed to be a string, we can just use `qs` instead of `(qs ? qs : '')`.

js/battledata.js
```diff
@@ -847,7 +847,12 @@ var Tools = {
 			facing = 'back';
 		}
 
-		var animationData = window.BattlePokemonSprites && BattlePokemonSprites[pokemon.speciesid];
+		var animationData = {};
+		if (options.gen === 5 && typeof(BattlePokemonSpritesBW) !== 'undefined') {
```
Zarel: Convention is to use `window.BattlePokemonSpritesBW` instead of `typeof`. Either way, you're using `typeof` against convention - note that `typeof` is an operator, not a function.

QuiteQuiet:
I did notice that this approach misses out on people who simply open up an ongoing battle without doing any of the steps mentioned before. I'll try to think of something to get past that, if I can.

js/client-teambuilder.js
```diff
@@ -450,6 +450,7 @@
 		formatChange: function(e) {
 			this.curTeam.format = e.currentTarget.value;
 			this.save();
+			if (this.currTeam.format.substr(0, 4) === 'gen5' && !Tools.loadedSpriteData['bw']) Tools.loadSpriteData('bw');
```
Zarel: You spelled `curTeam` wrong? Was this tested?

js/battledata.js
```diff
@@ -811,17 +811,17 @@ var Tools = {
...
+	loadedSpriteData: {'xy':1, 'bw':0},
 	loadSpriteData: function(gen) {
-		if (this.loadedSpriteData === gen) return;
-		this.loadedSpriteData = gen;
+		if (this.loadedSpriteData[gen]) return;
+		this.loadedSpriteData[gen] = 1;
 
 		var path = $('script[src*="pokedex-mini.js"]').attr('src');
-		var qs = path.split('?')[1] || '';
+		var qs = '?' + path.split('?')[1] || '';
```
Zarel: This will be the string `"?undefined"` if `path` does not contain a `?` character.

Like I said, this should be

    var qs = '?' + (path.split('?')[1] || '');

js/battledata.js
```diff
@@ -811,17 +811,17 @@ var Tools = {
...
 		path = (path.match(/.+?(?=data\/pokedex-mini\.js)/) || [])[0] || '';
 
 		var el = document.createElement('script');
-		el.src = path + 'data/pokedex-mini' + (gen !== 'xy' ? '-' + gen : '') + '.js' + (qs ? '?' + qs : '');
+		el.src = path + 'data/pokedex-mini-bw.js' + (qs ? qs : '');
```
Zarel: `qs` is guaranteed to be a string, so this should just be

    el.src = path + 'data/pokedex-mini-bw.js' + (qs ? qs : '');

js/client-teambuilder.js
QuiteQuiet: I apparently missed to test the actual condition here when I made sure I matched the correct substring.

js/battledata.js
```diff
@@ -811,17 +811,17 @@ var Tools = {
...
-		var qs = path.split('?')[1] || '';
+		var qs = '?' + (path.split('?')[1] || '');
 		path = (path.match(/.+?(?=data\/pokedex-mini\.js)/) || [])[0] || '';
 
 		var el = document.createElement('script');
-		el.src = path + 'data/pokedex-mini' + (gen !== 'xy' ? '-' + gen : '') + '.js' + (qs ? '?' + qs : '');
+		el.src = path + 'data/pokedex-mini-bw.js' + qs;
 		document.getElementsByTagName('body')[0].appendChild(el);
 	},
 	getSpriteData: function(pokemon, siden, options) {
```
Zarel: Wait, have we always passed `options`?

Seems like it'd be easier just to read `pokemon.side.battle.gen`...

Slayer95: <code>getSpriteData</code> also accepts an string as the first parameter, for forme change implementation.

js/client-chat-tournament.js
```diff
@@ -277,6 +277,7 @@
 						if (!this.info.isJoined)
 							this.toggleBoxVisibility(false);
 						this.room.$chat.append("<div class=\"notice tournament-message-start\">The tournament has started!</div>");
+						if (this.info.format.substr(0, 4) === 'gen5' && !Tools.loadedSpriteData['bw']) Tools.loadSpriteData('bw');
 						break;
```
QuiteQuiet: This implementation loads the sprite data for everyone in the room as the tournament begins, which might not be the most elegant way to do it.

Zarel: I think we should have a more general solution for watching games.

We should at least load the sprite data when the game is joined.

QuiteQuiet: The problem with loading the data on joined room is that it takes longer for that data to load than it takes for the room to be made, so unless the room is forced to wait, that would only display properly at the second game.

Zarel: Or once people start switching in the first game. It's not ideal, but it's better than nothing.

QuiteQuiet: I guess there's two ways to go from here:
1. Wait on any of the earlier specified events and load the sprite data before they begin a battle (like the pullreq now, as well as loading it for the joined players in a tour), and the load the data on join for spectators.
2. Only load sprites when you join a gen 5 battle, even if you're a player, and accept that the team preview is pretty ugly. 

Neither one is perfect, but for the players either method would work.

Zarel: We could do 1.

Zarel:
Not under normal conditions, no.

Zarel:
If a third-party server doesn't begin a gen 5 format with `gen5`, we can't stop them, though. Does that affect you?

QuiteQuiet:
That's okay I suppose.

PR #351, opened 2015-04-17
Add a special command for room intro.

xfix:
This command prevents `</div>` hack allowing ignoring height limit of an infobox. Requires server-side support, although if it isn't here, old infoboxes will still display, except without `</div>` safety feature.


Zarel:
This is unnecessary; the only people who can use the `</div>` hack are ROs (who are few enough to be easily demoted), and they can't exactly do significant damage.

If we really need validation, we can check for mismatched divs completely server-side in O(n).

PR #352, opened 2015-04-18
Add sorting options for Pokemon and moves in teambuilder

QuiteQuiet:
This can sort Pokemon by name or any of their stats. Moves can be sorted by name, base power, category, or type.

Examples: http://puu.sh/hhUjW.jpg http://puu.sh/hijMP.jpg

I tested some things, but I couldn't find if anything required <code>BattleChart.pokemon</code> to be in alphabetical order.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -42,6 +42,8 @@
 			'keyup .chartinput': 'chartKeyup',
 			'focus .chartinput': 'chartFocus',
 			'change .chartinput': 'chartChange',
+			'change select[name=pokemonsort]': 'sortChart',
+			'change select[name=movesort]': 'sortChart',
 
 			// clipboard
 			'click .teambuilder-clipboard-data .result': 'clipboardResultSelect',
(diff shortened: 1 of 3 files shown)


TheImmortal:
Why is Meteor Mash first? Body Slam is also Physical and should be first alphabetically.

QuiteQuiet:
Only because I forgot that would be a good idea. I'll get that sorted out.

Zarel:
This conflicts with a pending optimization.

QuiteQuiet:
This is most likely less important, so if you want to wait it's probably easy enough to merge this afterwards.

Zarel:
Wow, the "pending optimization" wasn't actually pushed until December 2015... oh well, that's not too slow as far as PS development goes.

Teambuilder sorting is still planned, so I guess we can keep this pullreq open for that. I get the impression we should really make issues for pullreqs we want to implement but not in the way the pullreq does.

Zarel:
Teambuilder sorting is done for Pokémon! Now to do it for moves...

Zarel:
Pokémon sorting was implemented in 385c284c9691b15bb

Move sorting was implemented in 256885354e54def443a

PR #353, opened 2015-04-19
Allow picking a custom background.

xfix:


js/client.js
```diff
@@ -2686,4 +2690,34 @@
...
+			buf += '<input type="file" accept="image/*" name="bgfile">';
+
+			buf += '<p><button name="close">Cancel</button></p>';
+			this.$el.html(buf);
+		},
+		setBg: function(e) {
+			var file = e.currentTarget.files[0];
+			var reader = new FileReader();
+			reader.onload = function (e) {
+				e.target.result;
```
kotarou3: ?

kotarou3:
Also, you might want allow for URI entry as well

js/client.js
```diff
@@ -2686,4 +2690,34 @@
...
+				var bg = '#344b6c url(' + e.target.result + ') no-repeat left center fixed';
+				Tools.prefs('bg', bg);
```
kotarou3: Be careful of running out of `localStorage` space

js/client.js
```diff
@@ -2686,4 +2690,33 @@
...
+
+			buf += '<p><button name="close">Cancel</button></p>';
+			this.$el.html(buf);
+		},
+		setBg: function(e) {
+			var file = e.currentTarget.files[0];
+			var reader = new FileReader();
+			reader.onload = function (e) {
+				var bg = '#344b6c url(' + e.target.result + ') no-repeat left center fixed';
+				Tools.prefs('bg', bg);
```
kotarou3: Be careful of running out of `localStorage` space

Slayer95:
Cool!

xfix:
@kotarou3: URI entry can be added separately if someone is interested in doing that. Currently I'm implementing just the basic feature.

(also, web browser file pickers should allow URLs, but currently the only browser I know that allows that is Konqueror... that's outside of scope of this pull request however)

js/client.js
```diff
@@ -2686,4 +2690,42 @@
...
+			buf += '<p><button name="close">Cancel</button></p>';
+			this.$el.html(buf);
+		},
+		setBg: function(e) {
+			$('.bgstatus').text('Changing background image.');
+			var file = e.currentTarget.files[0];
+			var reader = new FileReader();
+			var self = this;
+			reader.onload = function(e) {
+				var bg = '#344b6c url(' + e.target.result + ') no-repeat left center fixed';
```
Zarel: Data URL is actually a pretty interesting idea.

Zarel:
My main fear is that this is storing too much data in localStorage and/or RAM

xfix:
Hm, that's understandable. RAM usage shouldn't be too much of an issue, an image has to be stored in RAM anyway to display it. I guess the image is stored essentially twice, but with how much RAM is in today's machines, it shouldn't be an issue (the image is 3.5MB max).

As for localStorage limit, it's 5MB, except for Internet Explorer 8 which has 128 KB limit. I guess this could be using IndexedDB instead of localStorage, because IndexedDB has higher size limits (50MB), however.

Zarel:
`localStorage` limit may be 5MB, but larger data amounts tend to get pruned more often afaik.

This only affects people who actually use custom backgrounds, though, so I don't think it's too much of a problem.

Zarel:
You might want to warn about large images, though? `base64` is like 75% efficient, and we use `localStorage` for other things, so we might want to warn at 1MB or 1MB or something.

Zarel:
(I'd recommend making a new pull request for that)

PR #354, opened 2015-04-19
Fix curse targeting in doubles/triples

QuiteQuiet:


Slayer95:
The request data is supposed to have a <code>target</code> property for each move, which is what should be checked here.

QuiteQuiet:
Unless I'm misunderstanding something, <code>Tools.getMove(move).target</code> fetches the target for curse, which is <code>normal</code> and then it's server side decided if curse hit yourself or the opponent based on typing. In singles this doesn't matter, but in doubles/triples you always get prompted to select a target for curse, even if the pokemon in question isn't ghost type.

ascriptmaster:
The request data does have a `target` property, but Curse doesn't work properly with it yet. Give me a second.

Zarel:
Is this still relevant at all? @Marty-D

Marty-D:
If you mean is it fixed already, then no. It's not hugely relevant besides being a strange UI quirk; the only cases I can think of mechanically that are adversely affected include:

1. Getting hit by Trick-or-Treat the same turn you use Curse. It will target the Pokemon you chose while you were not a Ghost type, when it should be a random opponent because you're not supposed to have selected a target (usually; there's some buggy behaviour in-game if you're the Pokemon on the right side in Doubles and use Curse this way, it will target your ally instead of a random opponent).

2. Same as 1, except having Color Change and getting hit by a Ghost move after having selected Curse.

Should still be fixed though, despite its general irrelevancy.

Zarel:
Well, this should be fixed server-side as well as client-side, then...

Zarel:
@Marty-D, do @Slayer95's Nov 1 changes fix this?

Marty-D:
The skipped tests in Curse still fail, so no, this remains unresolved.

Zarel:
Okay, fixed in Zarel/Pokemon-Showdown@99a6e73df0

PR #356, opened 2015-04-20
Make move types and base power match battle status

Joimer:
Move Type and Base Power are volatile and the information is often not accurate.
When a Plate is active, it will change the type of Judgment.
When a Drive is active, it will change the type of Techno Blast.
Other items that change move type and base power will also reflect their changes.
Moves that have a deterministic varying base power will show the current base power.
In doubles, moves that have different base powers depending on target will show all possibilities.
Display all of that on move selector and tooltip to help players know and remember what attack type and what move power they have available.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -425,14 +425,15 @@
 					if (move.id === 'Struggle' || move.id === 'Recharge') pp = '&ndash;';
 					if (move.id === 'Recharge') move.type = '&ndash;';
 					if (name.substr(0, 12) === 'Hidden Power') name = 'Hidden Power';
+					var moveType = this.getMoveType(move, switchables[pos]);
 					if (moveData.disabled) {
 						movebuttons += '<button disabled="disabled"' + this.tooltipAttrs(moveData.move, 'move') + '>';
 						hasDisabled = true;
 					} else {
-						movebuttons += '<button class="type-' + move.type + '" name="chooseMove" value="' + (i + 1) + '" data-move="' + Tools.escapeHTML(moveData.move) + '"' + this.tooltipAttrs(moveData.move, 'move') + '>';
+						movebuttons += '<button class="type-' + moveType + '" name="chooseMove" value="' + (i + 1) + '" data-move="' + Tools.escapeHTML(moveData.move) + '"' + this.tooltipAttrs(moveData.move, 'move') + '>';
 						hasMoves = true;
 					}
-					movebuttons += name + '<br /><small class="type">' + (move.type || "Unknown") + '</small> <small class="pp">' + pp + '</small>&nbsp;</button> ';
+					movebuttons += name + '<br /><small class="type">' + (moveType || "Unknown") + '</small> <small class="pp">' + pp + '</small>&nbsp;</button> ';
 				}
 				if (!hasMoves) {
 					controls += '<button class="movebutton" name="chooseMove" value="0" data-move="Struggle">Struggle<br /><small class="type">Normal</small> <small class="pp">&ndash;</small>&nbsp;</button> ';
@@ -952,13 +953,70 @@
 				var move = Tools.getMove(thing);
 				if (!move) return;
 				var basePower = move.basePower;
-				if (!basePower) basePower = '&mdash;';
+				var basePowerText = '';
+				var additionalInfo = '';
+				var yourActive = this.battle.yourSide.active;
+				var myPokemon = this.battle.mySide.pokemon;
+
+				// Check if there are more than one active Pokémon to check for multiple possible BPs.
+				if (yourActive.length > 1) {
+					// We check if there is a difference in base powers to note it.
+					// Otherwise, it is just shown as in singles.
+					// The trick is that we need to calculate it first for each Pokémon to see if it changes.
+					var previousBasepower = false;
+					var difference = false;
+					var basePowers = [];
+					for (var i = 0; i < yourActive.length; i++) {
+						basePower = this.getMoveBasePower(move, this.battle.mySide.active[this.choice.choices.length], yourActive[i]);
+						if (previousBasepower === false) previousBasepower = basePower;
+						if (previousBasepower !== basePower) difference = true;
+						if (!basePower) basePower = '&mdash;';
+						basePowers.push('Base power for ' + yourActive[i].name + ': ' + basePower);
+					}
+					if (difference) {
+						basePowerText = '<p>' + basePowers.join('<br />') + '</p>';
+					}
+					// Falls through to not to repeat code on showing the base power.
+				}
+				if (!basePowerText) {
+					basePower = basePower || this.getMoveBasePower(move, myPokemon[this.choice.choices.length], yourActive[0]);
+					if (!basePower) basePower = '&mdash;';
+					basePowerText = '<p>Base power: ' + basePower + '</p>'
+				}
 				var accuracy = move.accuracy;
 				if (!accuracy || accuracy === true) accuracy = '&mdash;';
 				else accuracy = '' + accuracy + '%';
+
+				// Handle move type for moves that vary their type.
+				var moveType = this.getMoveType(move, myPokemon[this.choice.choices.length]);
+
+				// Deal with Nature Power special case, indicating which move it calls.
+				if (move.id === 'naturepower') {
+					if (this.battle.gen === 6) {
+						additionalInfo = 'Calls ';
+						if (this.battle.hasPseudoWeather('Electric Terrain')) {
+							additionalInfo += Tools.getTypeIcon('Electric') + 'Thunderbolt';
+						} else if (this.battle.hasPseudoWeather('Grassy Terrain')) {
+							additionalInfo += Tools.getTypeIcon('Grass') + 'Energy Ball';
+						} else if (this.battle.hasPseudoWeather('Misty Terrain')) {
+							additionalInfo += Tools.getTypeIcon('Fairy') + 'Moonblast';
+						} else {
+							additionalInfo += Tools.getTypeIcon('Normal') + 'Tri Attack';
+						}
+					} else if (this.battle.gen > 3) {
+						// In gens 4 and 5 it calls Earthquake.
+						additionalInfo = 'Calls ' + Tools.getTypeIcon('Ground') + 'Earthquake';
+					} else {
+						// In gen 3 it calls Swift, so it retains its normal typing.
+						additionalInfo = 'Calls ' + Tools.getTypeIcon('Normal') + 'Swift';
+					}
+				}
+
 				text = '<div class="tooltipinner"><div class="tooltip">';
-				text += '<h2>' + move.name + '<br />' + Tools.getTypeIcon(move.type) + ' <img src="' + Tools.resourcePrefix + 'sprites/categories/' + move.category + '.png" alt="' + move.category + '" /></h2>';
-				text += '<p>Base power: ' + basePower + '</p>';
+				text += '<h2>' + move.name + '<br />' + Tools.getTypeIcon(moveType) + ' <img src="' + Tools.resourcePrefix;
+				text += 'sprites/categories/' + move.category + '.png" alt="' + move.category + '" /></h2>';
+				text += basePowerText;
+				text += additionalInfo;
 				text += '<p>Accuracy: ' + accuracy + '</p>';
 				var flags = {
 					"authentic": "Ignores a target's substitute.",
@@ -1096,16 +1154,7 @@
 					}
 					text += pokemon.stats['spe'] + '&nbsp;Spe</p>';
 				} else if (template.baseStats) {
-					var minSpe;
-					var maxSpe;
-					if (this.battle.tier === 'Random Battle') {
-						minSpe = Math.floor(Math.floor(2 * template.baseStats['spe'] + (0) + Math.floor((0) / 4)) * pokemon.level / 100 + 5);
-						maxSpe = Math.floor(Math.floor(2 * template.baseStats['spe'] + (31) + Math.floor((85) / 4)) * pokemon.level / 100 + 5);
-					} else {
-						minSpe = Math.floor(Math.floor(Math.floor(2 * template.baseStats['spe'] + (0) + Math.floor((0) / 4)) * pokemon.level / 100 + 5) * 0.9);
-						maxSpe = Math.floor(Math.floor(Math.floor(2 * template.baseStats['spe'] + (31) + Math.floor((252) / 4)) * pokemon.level / 100 + 5) * 1.1);
-					}
-					text += '<p>' + minSpe + ' to ' + maxSpe + ' Spe (before items/abilities/modifiers)</p>';
+					text += '<p>' + this.getTemplateMinSpeed(template, pokemon.level) + ' to ' + this.getTemplateMaxSpeed(template, pokemon.level) + ' Spe (before items/abilities/modifiers)</p>';
 				}
 				if (pokemon.moves && pokemon.moves.length && (!isActive || isActive === 'foe')) {
 					text += '<p class="section">';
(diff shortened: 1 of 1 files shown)


Joimer:
@Zarel #307 was move type specific so I went ahead and I just made a new PR with all the updated stuff.

There's some information the client doesn't keep track off, so some moves like Return can't show the proper base power for now. I've edited this so you can keep track of most types and base power though, the most intersting change is seeing the multiple base powers in doubles triples. 

See:
Natural Gift http://i.imgur.com/FkQQIIq.png
Multiple target Grass Knot http://i.imgur.com/B5j1UZt.png
Single Target / same BP Grass Knot http://i.imgur.com/STZGABD.png

Zarel:
Looks good. Would you say this is ready for merge?

Zarel:
I think Return should be sent from the server the same way Hidden Power is sent.

Joimer:
Yeah, it's ready for merge. And I agree, it's probably easier for Return and Frustration to get the BP from server just like Hidden Power. I think I didn't miss any move?

QuiteQuiet:
If you want to go that far, you could change moves that are effected by Refrigirate, Aerilate, Pixilate and Normalize.

Weather Ball and Nature Power do change based on battle condition though, so you probably should to add those.

Marty-D:
How is Grass Knot's power 80 on Genesect in the first GK sample image and 120 in the next?

You forgot Reversal with Flail; they also have a different implementation in Gen 4. Not sure if/how you want to handle that.

I disagree with having Crush Grip and Wring Out here, because it's just additional information about how much HP the opponent has. Using the listed power, you can easily calculate a more precise percentage than what is given normally. You didn't put Gyro Ball here for good reason, but this is not that different.

I also disagree with Technician since it's only one of 14 or so Abilities that modify a move's power, and just makes things needlessly complicated. This should ideally only be for moves that change power based on their own mechanics rather than some other effect's mechanics.

Joimer:
Agreed, will not add abilities nor Crush Grip and Wring Out then.
Indeed, I can handle gen 4 for those moves, thanks for pointing it out!
Checking now base power, maybe it was the one whose weight I edited manually to see if it updated properly. Adding more screenshoots later on.

Zarel:
The showing base power for Crush Grip etc is fine because it's calculated from the percentage the client knows, so it's not any more information than is already given. Instead, it should just report the estimated power as approximate.

I'd support abilities; heck, I'd even want to include STAB if it's not already included. And maybe have another number underneath that reports it after weaknesses/resistances.

Joimer:
I'm re-checking it right now.

Joimer:
I've taken into account all of you guys, so I've updated the commit accordingly. I've also improved the code slightly and tested this with a ubers and doubles battle.

Since STAB does not modify BP but damage after other calculations, I just show it is present. As for weaknesses and resistances, more of the same, but maybe this should be for a future PR.

I've added notes to moves that can be confusing as to why the base power is changed, indicated all ability boosted moves that they are being boosted by its ability (only on the most basic ones that change type as well, so far).

Since I had to make functions for min-max speed calculation to not to repeat code, I've also implemented #313 properly. If you want, I can make it a separate commit, though I don't think it's necessary since this commit and PR is basically an overhaul of tooltips and move descriptions in general.

How do you like this, @Marty-D @Zarel ?

js/client-battle.js
```diff
@@ -1083,6 +1112,202 @@
...
+					basePower = 40;
+					if (pokemonWeight > targetWeight * 5) basePower = 120;
+					else if (pokemonWeight > targetWeight * 4) basePower = 100;
+					else if (pokemonWeight > targetWeight * 3) basePower = 80;
+					else if (pokemonWeight > targetWeight * 2) basePower = 60;
+					if (target.volatiles.autotomize) basePower += ' (Approximation)';
+				}
+			}
+			// Other ability boosts.
+			if (pokemon.ability === 'technician' && basePower <= 60) {
```
Zarel: It seems like `basePower` can be a string at this point, like `"45 to 47"` or `"80 (Approximation)""`. What happens if you compare such a string with `<= 60` or multiply it like `* 1.5`?

Joimer: You are right. Improving this and updating the PR.

js/client-battle.js
```diff
@@ -1083,6 +1112,202 @@
...
+				if (move.pp === 1) basePower = 200;
+				else if (move.pp === 2) basePower = 80;
+				else if (move.pp === 3) basePower = 60;
+				else if (move.pp === 4) basePower = 50;
+			}
+			if (move.id === 'venoshock') {
+				if (target.status === 'psn' || target.status === 'tox') basePower = (basePower * 2) + ' (Boosted by status)' ;
+			}
+			if (move.id === 'wakeupslap') {
+				if (target.status === 'slp') basePower = (basePower * 2) + ' (Boosted by status)' ;
```
Zarel: It seems inconsistent to show BP based on opponent status if we don't show things like weakness/resistance.

Joimer: I thought that because those are moves that are boosted specifically by a status only if you use those moves, as opposed to a general boost to damage due to status (Guts), while weakness / resistance being something to account for separatedly. Want me to add an extra next to STAB about resistances?

Zarel:
How I'd personally do it is to show two numbers: "Effective Base Power (EBP)" and "EBP against the opposing <name>". First one includes modifiers on your own pokemon (including STAB), second includes your opponent's pokemon (including weaknesses/resistances).

I'd rather have "effective" base power because to a user, it doesn't matter whether STAB affects the base power directly or not. At most it will make like a 1hp difference because of rounding, and it's not like anyone manually tracks rounding direction anyway (well, at least outside of LC).

Joimer:
I updated this so basePower is a number until the end, where the possible comment gets added. If it's an empty string, means it needs to be an &mdash;.

Can we talk a bit more about the EBP showing? I think it's redundant information often. It could be shown only when different. Anyway this means adding a client-side typechart check (and IMO it should include abilities), so can we leave this for a next iteration and maybe discuss on dev forums on what's the best way to show this information?

js/client-battle.js
```diff
@@ -1083,6 +1112,238 @@
...
+				if (target.volatiles.formechange) template = Tools.getTemplate(target.volatiles.formechange[2]);
+				var min = (Math.floor(25 * this.getTemplateMinSpeed(template, target.level) / pokemon.stats.spe) || 1);
+				var max = (Math.floor(25 * this.getTemplateMaxSpeed(template, target.level) / pokemon.stats.spe) || 1);
+				if (min > 150) min = 150;
+				if (max > 150) max = 150;
+				// Special case due to range as well.
+				basePower = 0;
+				if (pokemon.ability === 'technician') {
+					if (min <= 60) min *= 1.5;
+					if (max <= 60) max *= 1.5;
```
Zarel: if `max > 60`, max could very well be `90`.

Consider gen 5 Hidden Power. Ranged from 30 to 70, but its Technician range was 45 to 90, not 45 to 70.

Zarel: Here's one way to do it:

    if (max <= 60) max = Math.max(max * 1.5, 90);

Zarel: See also Electro Ball

Zarel:
Sure, we can leave the detailed EBP stuff for the next iteration.

I'd leave off the "has STAB" message entirely in this iteration, then. I think it's confusing if you say (technician) and it changes the displayed BP but you say (STAB) and it doesn't.

js/client-battle.js
Joimer: I'm not sure I understand, if the `max` calculated equals `60` it will be multiplied by 1.5 since I'm using `<=`, which leads to `90`, but the actual max for Gyro Ball, technician or not, is `150`. Am I missing something?

js/client-battle.js
```diff
@@ -1083,6 +1112,238 @@
...
+					var item = Tools.getItem(pokemon.item);
+					if (item.naturalGift) moveType = item.naturalGift.type;
+				}
+			}
+			// Weather and pseudo-weather type changes.
+			if (move.id === 'naturepower') {
+				if (this.battle.hasPseudoWeather('Electric Terrain')) moveType = 'Electric';
+				if (this.battle.hasPseudoWeather('Grassy Terrain')) moveType = 'Grass';
+				if (this.battle.hasPseudoWeather('Misty Terrain')) moveType = 'Fairy';
+			}
```
ascriptmaster: What about the case for Gen 5 Nature Power, which becomes Earthquake and therefore is Ground-type?

js/client-battle.js
ascriptmaster: Based on the opponent's speed, Gyro Ball could max out at a different BP. Does this code deal with stat boosts (Dragon Dance, Curse, etc)?

Also, the code should be 

    if (max <= 60) max *= 1.5;
    else max = Math.max(max, 90);

js/client-battle.js
```diff
@@ -1083,6 +1112,238 @@
...
+					if (targetWeight >= 200) basePower = 120;
+					else if (targetWeight >= 100) basePower = 100;
+					else if (targetWeight >= 50) basePower = 80;
+					else if (targetWeight >= 25) basePower = 60;
+					else if (targetWeight >= 10) basePower = 40;
+					if (target.volatiles.autotomize) basePowerComment = ' (Approximation)';
+				}
+				if (move.id === 'heavyslam' || move.id === 'heatcrash') {
+					basePower = 40;
+					if (pokemonWeight > targetWeight * 5) basePower = 120;
```
Marty-D: `pokemonWeight` isn't defined anywhere!

Marty-D:
:+1: Nice changes, although I'm not sure about Nature Power. It's not a type change like the others; it literally uses a move other than itself based on terrain. It also can't take into account the -ate Abilities anyway since Nature Power has no set power. Thoughts?

js/client-battle.js
Joimer: Damn, must've been lost in one of the rebases I've done. Good catch.

js/client-battle.js
Joimer: Ok, you guys mean in the case that the min is in Technician range whereas max is not, I needed more sugar in my brain yesterday, lol. Also good note on boosts.

Zarel: If we're not going to do STAB/weaknesses/resistances in this iteration, I think boosts should also be left for next iteration.

Zarel: Oh, you mean boosts for the Gyro Ball calculation explicitly.

Zarel:
Well, we can still show Nature Power stuff on the popup. Maybe show Normal but have its called move type somewhere else?

Joimer:
I can add an additional text that says "Calls: [Ground icon] Earthquake, 100 bp"
Also updating this PR later, have had a lot of work today.

Zarel:
This is still a really good change... can we get it finished up and merged?

DanUgore:
>I'd leave off the "has STAB" message entirely in this iteration, then. I think it's confusing if you say (technician) and it changes the displayed BP but you say (STAB) and it doesn't.

Maybe but STAB doesnt change BP... Technician does >.> <.<

Joimer:
@Zarel @DanUgore @Marty-D @ascriptmaster I've updated this Pull Request now that I'm checking code status again. I've added the special case for Nature Power and left STAB off, to be present in a future iteration.

Let me know if I can improve this or if it's good to go.

Marty-D:
Currently on mobile so I just glanced through quickly. Hex got a power boost in Gen 6, so you'll probably want to do `*= 2` there too. I'll take a better look later tonight, but it's come along nicely! :+1:

js/client-battle.js
```diff
@@ -1150,6 +1200,234 @@
...
+					var item = Tools.getItem(pokemon.item);
+					if (item.onPlate) moveType = item.onPlate;
+				}
+				if (move.id === 'technoblast') {
+					var item = Tools.getItem(pokemon.item);
+					if (item.onDrive) moveType = item.onDrive;
+				}
+				if (move.id === 'naturalgift') {
+					var item = Tools.getItem(pokemon.item);
+					f (item.naturalGift) moveType = item.naturalGift.type;
```
Marty-D: Typo.

Joimer:
Fixed Marty's concerns!

js/client-battle.js
```diff
@@ -1150,6 +1200,231 @@
...
+					var item = Tools.getItem(pokemon.item);
+					if (item.naturalGift) moveType = item.naturalGift.type;
+				}
+			}
+			// Weather and pseudo-weather type changes.
+			if (move.id === 'weatherball') {
+				if (this.battle.weather === 'sunnyday' || this.battle.weather === 'desolateland') moveType = 'Fire';
+				if (this.battle.weather === 'raindance' || this.battle.weather === 'primordialsea') moveType = 'Water';
+				if (this.battle.weather === 'sandstorm') moveType = 'Rock';
+				if (this.battle.weather === 'hail') moveType = 'Ice';
```
ascriptmaster: Should we have this accurately account for Cloud Nine and Air Lock?

js/client-battle.js
```diff
@@ -951,14 +952,72 @@
 			case 'move':
 				var move = Tools.getMove(thing);
 				if (!move) return;
+				// Declare variables we are going to need.
```
Zarel: ...I don't think this needs a comment

Zarel: My personal rule for comments is: Never document _what_ the code does (programmers know how to read code and don't need to be told that), document _why_ the code does it.

Joimer: True, sometimes I tend to write too much comments.

js/client-battle.js
Joimer: It's not possible to know whether Cloud Nine or Air Lock are actually active, since it could be a hackmons format or something.

Joimer:
Updated the PR, eliminated unnecessary comment, also added a couple of checks to avoid crashes with pokemon.volatiles being undefined.

js/client-battle.js
Zarel: What does hackmons change?

Zarel: Don't Cloud Nine and Air Lock announce on switch-in?

Zarel:
Other than the Cloud Nine issue, everything else looks good to me.

Joimer:
Right, I forgot cloud nine and air lock announce. Added it.

Marty-D:
I guess the only other case that might be useful is Bulldoze, Earthquake, and Magnitude having halved power in Grassy Terrain. The nice part about it is there's no need to check that anything is grounded, since those moves can't hit airborne Pokemon anyway.

(The power boost on Grass moves in Grassy Terrain / power boost on Electric moves in Electric Terrain / halved power on Dragon moves in Misty Terrain would need some kind of check though.)

Joimer:
Those modifiers are in damage and not in bp though, right?
We can add that on next iteration when adding EBP.

Marty-D:
No, they're base power modifiers, but yeah they can wait until later.

Zarel:
I kind of want to get rid of BP-against-specific-opponents and EBP entirely.

I sort of want to use Damage Rating and Effective Damage Rating. Damage Rating would be percent damage against a typeless opponent with 100 base Def/SpD. Effective Damage Rating would be percent damage against the opponent in question (so a regular damage calc).

Joimer:
That is a good idea!

Zarel:
Speaking of good ideas, no one's commented on my Offline Messaging implementation sketch yet

https://github.com/Zarel/Pokemon-Showdown/issues/2151

js/client-battle.js
```diff
@@ -952,13 +953,70 @@
...
+				// Check if there are more than one active Pokémon to check for multiple possible BPs.
+				if (yourActive.length > 1) {
+					// We check if there is a difference in base powers to note it.
+					// Otherwise, it is just shown as in singles.
+					// The trick is that we need to calculate it first for each Pokémon to see if it changes.
+					var previousBasepower = false;
+					var difference = false;
+					var basePowers = [];
+					for (var i = 0; i < yourActive.length; i++) {
+						basePower = this.getMoveBasePower(move, this.battle.mySide.active[this.choice.choices.length], yourActive[i]);
```
urkerab: Why does this use `this.battle.mySide.active[this.choice.choices.length]`...

js/client-battle.js
```diff
@@ -952,13 +953,70 @@
...
+						if (!basePower) basePower = '&mdash;';
+						basePowers.push('Base power for ' + yourActive[i].name + ': ' + basePower);
+					}
+					if (difference) {
+						basePowerText = '<p>' + basePowers.join('<br />') + '</p>';
+					}
+					// Falls through to not to repeat code on showing the base power.
+				}
+				if (!basePowerText) {
+					basePower = basePower || this.getMoveBasePower(move, myPokemon[this.choice.choices.length], yourActive[0]);
```
urkerab: ...but this uses `myPokemon[this.choice.choices.length]`? Note that if this version is actually the correct version, and you need to fix the above case, then you probably don't need dd760f1480504cf7f14673adcd83b0bbd98abdbc any more.

Zarel: @urkerab, it looks like this is the correct formulation. Can you make the change and revert dd760f1?

Marty-D: I noticed this too, but I'm choosing not to do anything about it until I have time to audit the entire thing. dd760f1 is probably still necessary either way though, since Gyro Ball was crashing in both Singles and Doubles, and this line is the default for Singles.

PR #359, opened 2015-04-21
Implement team sorting in teambuilder

Joimer:
Looks like this http://i.imgur.com/iMKEmSe.png
Can't believe it was this easy.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -169,12 +169,15 @@
 						formatText = '['+team.format+'] ';
 					}
 
-					buf += '<li><button name="edit" value="'+i+'" style="width:400px;vertical-align:middle">'+formatText+'<strong>'+Tools.escapeHTML(team.name)+'</strong><br /><small>';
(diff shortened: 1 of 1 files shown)


Zarel:
Looks sort of ugly. :/

Zarel:
It's also sort of unclear what the arrows mean, which is probably a bigger problem (rest of that page looks ugly anyway).

Joimer:
Well, the whole page is ugly, heh. I think the arrows are pretty straightforward, you have them in both directions unless it's the topmost one or bottommost one, or no arrows if only one team.

Any suggestions?

Zarel:
The thing is that arrows mean way too much. The way the in-team move buttons work is with a Move button with a Move icon, which I think is clearer.

Joimer:
But 1 click > 2 clicks imo. The move buttons and menu would look worse if you have, say, more than 20 teams, don't you think so?

Zarel:
How so? We might be thinking of different things if you think it'd look worse.

Zarel:
Well, we now have drag-and-drop sorting, which is better, so I assume this can be closed.

PR #360, opened 2015-04-22
Teambuilder team sorting

Ecuacion:
Implement a move button and remove the useless edit button (users can just click the team button). Also, teams can be sorted drag and dropping them with the mouse (requires jQuery UI).

Looks like this: http://imgur.com/a/tzjjC#2

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -102,6 +102,7 @@
 		 * Team list view
 		 *********************************************************/
 
+		curMoving: false,
 		deletedTeam: null,
 		deletedTeamLoc: -1,
 		updateTeamList: function() {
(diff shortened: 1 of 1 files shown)


js/client-teambuilder.js
```diff
@@ -200,11 +235,33 @@
...
 		},
 		edit: function(i) {
+			if (this.curmoving !== false) return;
 			var i = +i;
 			this.curTeam = teams[i];
 			this.curTeamIndex = i;
 			this.update();
 		},
+		"move": function(i) {
+			if (this.curmoving !== false) return;
```
SolarisFox: Any particular reason why these aren't written `if (this.curmoving === true) return;` ?

Ecuacion: Because that var has the A position and false if not moving.

SolarisFox: In that case you probably want to set it to `null` as it's default state so it doesn't appear to be a Boolean. You might even want to compare to `if (this.curmoving >= 0) return;` if the position can be 0 or just `if (this.curmoving) return;` otherwise.

On another style note, to match the casing used with other variables you probably want the variable to be named `curMoving`.

Slayer95:
jQuery UI is too much of a dependency for a drag-and-drop feature. We would ideally have our own implementation of drag-and-drop.

Zarel:
In addition to being a rather huge dependency, sorting tends to be hard to make "feel right", and it doesn't look like jQuery UI does so. :(

Zarel:
Anything involving drag-and-drop tends to be very hard to get the little things right, and you sort of have to do it yourself.

Ecuacion:
Remaking this now with buttons (removing drag and drop). Added a move option for teams which allow users move a team to any position with 2 clicks.
Some screenshots: http://imgur.com/a/cU2TK#0

PR #362, opened 2015-04-24
Allow negative attack nature for U-turn and Volt Switch

xfix:
U-turn and Volt Switch are unusual in an aspect that they are dual purpose moves. Sometimes they are used as damaging moves (main attacking move even, like with Mega Beedrill or Genesect). In other cases, they are used to preserve momentum, and the damage from them doesn't exactly matter (see Cobalion, Xatu, Mega Pidgeot or Hydreigon).

This change allows giving a negative attacking nature to an user of U-turn or Volt Switch if it appears to be used in a momentum-sort of way, as opposed to an usage as a damaging move.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1843,9 +1843,9 @@
 				if (evTotal < 508) {
 					var remaining = 508 - evTotal;
 					if (remaining > 252) remaining = 252;
-					if (!evs['atk'] && moveCount['PhysicalAttack']) {
+					if (!evs['atk'] && moveCount['PhysicalAttack'] >= 1) {
 						evs['atk'] = remaining;
-					} else if (!evs['spa'] && moveCount['SpecialAttack']) {
+					} else if (!evs['spa'] && moveCount['SpecialAttack'] >= 1) {
 						evs['spa'] = remaining;
 					} else if (stats.hp == 1 && !evs['def']) {
 						evs['def'] = remaining;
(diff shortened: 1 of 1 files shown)


Zarel:
This would give XY OU Mixed Greninja a -Atk nature, wouldn't it?

I'm not opposed to giving -Atk natures, but they should probably be given more conservatively.

xfix:
XY Greninja is an interesting case. Indeed, this EV suggester would suggest negative Attack, while the current EV proposer would propose negative Defense. I would rather go with negative Special Defense, because negative Defense increases fraility of Greninja to priority moves (the biggest worry, really). Sets I saw for XY Greninja with U-turn varied (some lowered Special Defense, some Speed, some Defense), in fact, I believe the main set on Smogon mentioned Hasty / Timid as the nature.

Not that XY Greninja matters anymore, because first of all, XY Greninja is not relevant anymore (it's ORAS now, Gunk Shot is a thing), and secondly U-turn was considered to be a poor choice on Greninja (I don't know exactly why, I don't play OU).

For Xatu, Mega Pidgeot, Hydreigon, Yanmega, and few other mons I couldn't bother to look for, the proposed change is an improvement. These sets pretty much always use Timid or Modest, even if they do use U-turn.

I tried to look for sets where this change would propose a worse nature, and I found one current set - Doubles Scarf Genesect (U-Turn, Ice Beam, Thunderbolt, Flamethrower). It's Doubles set however, and the current proposed spread is quite bad anyway.

Here is where a limitation of EV suggester lies - it doesn't really handle mixed sets well, putting even U-turn aside. I don't have any particular suggestions to this problem, making mixed sets is quite tricky, usually EVs are given in one stat to defeat a particular threat, and other EVs are given to other stat, while Speed also gets some EVs to outspeed something in particular. Jirachi mixed set uses 132 Atk / 160 SpA / 216 Spe (Mild) for example - 132 Atk 2HKOes Clefable, 216 Spe outspeeds Kyurem-Black, and 160 are left for special attack. Those benchmarks are usually Pokemon sensititve however, which makes automation of these quite tricky

Zarel:
We don't need to automate them; we already hardcode the EVs to defeat certain threats; see Jirachi's default spread. We can hardcode those cases, too.

PR #363, opened 2015-04-24
Ignore spectators forever and ignore opponent

Ecuacion:
Add Ignore spectators forever and ignore opponent options for battles.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2194,6 +2194,7 @@ var Battle = (function () {
 		this.p1 = this.mySide;
 		this.p2 = this.yourSide;
 		this.gen = 6;
+		this.ignoreSpects = Tools.prefs('ignorespectsforever');
 	};
 	Battle.prototype.updateGen = function () {
 		if (this.gen < 3) this.backdropImage = 'bg-gen1.png';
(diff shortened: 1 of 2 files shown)


js/battle.js
```diff
@@ -5317,7 +5317,8 @@ var Battle = (function () {
 		case 'chat':
 		case 'c':
 			name = args[1];
-			if (this.ignoreSpects && !this.getSide(name).battle) break;
+			if ((Tools.prefs('ignorespectsforever') || this.ignoreSpects) && !this.getSide(toId(name)).battle) break;
```
sirDonovan: Instead of calling <code>Tools.prefs('ignorespectsforever')</code> for every chat message, you could store it in <code>this.ignoreSpects</code> when initializing the battle.

Ecuacion: Done, this makes the specific menu of the battle has priority.

Zarel:
So far, none of our ignore features are permanent, because people forget that they've changed settings all the time (that, or their friends change settings while using their computer), and then I get bug reports "my friend is trying to talk to me and I'm not getting their messages, please fix"

One workable solution would be to have a big red "You are ignoring spectators" at the beginning of every battle log or something. Anyway, @sirDonovan is currently working on a different-but-related feature that might make it easier to do that sort of thing.

PR #364, opened 2015-04-26
Room autojoin

Morfent:
Remaking this pullreq, since it's on the dev to-do list.

Allows users to autojoin additional rooms on login. They must be in the
rooms they add.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -373,6 +373,13 @@
 				self.addPopup(ReconnectPopup, {cantconnect: true});
 			});
 
+			this.once('init:choosename', function() {
+				var autojoins = Tools.prefs('autojoins');
+				if (autojoins && autojoins.length) {
+					self.send('/autojoin ' + autojoins.join(','));
+				}
+			});
+
 			this.user.on('login:invalidname', function(name, reason) {
 				self.addPopup(LoginPopup, {name: name, reason: reason});
 			});
@@ -824,6 +831,14 @@
 				} else {
 					this.joinRoom(roomid, roomType, true);
 				}
+
+				if (roomType === 'chat' && roomid !== 'staff') {
+					var autojoins = Tools.prefs('autojoins') || [];
+					if (autojoins.indexOf(roomid) < 0 && autojoins.length < 6) {
+						autojoins = this.sliceAutojoins();
+						Tools.prefs('autojoins', autojoins);
+					}
+				}
 			} else if ((data+'|').substr(0,8) === '|expire|') {
 				var room = this.rooms[roomid];
 				if (room) {
@@ -858,6 +873,13 @@
 				} else {
 					if (isdeinit) { // deinit
 						this.removeRoom(roomid, true);
+						var autojoins = Tools.prefs('autojoins');
+						if (!autojoins) {
+							Tools.prefs('autojoins', []);
+						} else if (autojoins.indexOf(roomid) >= 0) {
+							autojoins = this.sliceAutojoins();
+							Tools.prefs('autojoins', autojoins);
+						}
 					} else { // noinit
 						this.unjoinRoom(roomid);
 						if (roomid === 'lobby') this.joinRoom('rooms');
@@ -1385,6 +1407,19 @@
 				gui.Shell.openExternal(e.target.href);
 				return false;
 			}
+		},
+
+		// returns the first 6 valid sideRooms from app.rooms
+		sliceAutojoins: function() {
+			var autojoins = [];
+			for (var i in this.rooms) {
+				if (this.rooms[i].type !== 'chat' || i === 'staff') {
+					continue;
+				}
+				autojoins.push(i);
+				if (autojoins.length >= 6) break;
+			}
+			return autojoins;
 		},
 
 		/*********************************************************


Slayer95:
Ref https://github.com/Zarel/Pokemon-Showdown-Client/pull/268

js/client-chat.js
```diff
@@ -671,6 +671,62 @@
...
+							if (!tarRoom || tarRoom === 'rooms' || tarRoom === 'staff') continue;
+							if (autojoins.indexOf(tarRoom) < 0) autojoins.push(tarRoom);
+						}
+						this.add('Now autojoining rooms: ' + autojoins.join(', '));
+					} else {
+						var illegalRooms = [];
+						for (var i = 0, len = args.length; i < len; i++) {
+							var tarRoom = toId(args[i]);
+							if (!tarRoom) return this.add('The global room cannot be added to the autojoins list.');
+							if (tarRoom === 'staff') return this.add('The staff room cannot be added to the autojoins list.');
```
Slayer95: Why? It should be useful for servers without staff autojoin.

Morfent: Hm, didn't think of that

Zarel:
I finally ended up speccing out an autojoin feature I liked:

- Whenever a room is joined or left, save the first 6 rooms you're in (skipping `staff`) as your autojoin list.

- On connect, send `/autojoin <roomid1>, <roomid2>, <roomid3>, <...>` instead of `/autojoin` (server will need to be modified accordingly

- Remember that different rooms have different autojoin lists

- No /commands, no Options menu entry

Joimer:
I still would prefer to choose manually which rooms to autojoin and which rooms not to.

Slayer95:
I'm not sure if the feature I mention is something we want, but is it feasible for automatic multi-targetting to be accomodated with password-protected rooms?

Morfent:
I find that awfully unintuitive. Would the average user be able to figure out how autojoin works if the first 6 rooms in the room list were automatically added? Do all users want autojoin in the first place? I think making rooms autojoin similar to how Staff/Help/Lobby do now isn't the right way of going about this if it's supposed to be user configurable. Having a command would at least give them the choice of which rooms to autojoin and in what order to join them without having to keep said rooms open at all times so the list doesn't get screwed up, if how rooms are added isn't a mystery to them.

I do think ``/autojoin room1, room2, ... roomn`` is a good idea, but why would the list be limited to 6 rooms when they'd all be joined through one message?

Zarel:
The limit is 6 rooms to discourage users from spending time in too many rooms (because more users in more rooms means more server resource usage i.e. lag, and they're not going to be particularly active in all those rooms anyway).

And having rooms saved automatically is much more intuitive than having a /command. A command is literally the opposite of intuitive.

The system basically works like: As long as you're in 6 or fewer rooms, the next time you open PS, you'll be in the same rooms you were in when you left PS. Very straightforward/intuitive.

Zarel:
> Do all users want autojoin in the first place?

Sure, we can make a checkbox in Options, "Rejoin previously open chat rooms when reopening PS (limit 6 rooms)". I'd probably default it to on, but either works.

> I think making rooms autojoin similar to how Staff/Help/Lobby do now isn't the right way of going about this if it's supposed to be user configurable.

Configuration is often a bad thing.

- most users don't look at options menus or memorize commands, and so lose out on the feature.
- users will sometimes choose an option, and then forget they chose it, and get frustrated about the program not working as expected
- users will sometimes accidentally choose an option (misclick etc), and then forget about it, and get frustrated about the program not working as expected
- users will sometimes intentionally choose a setting that's worse than the default/best option (users are, surprise-surprise, often not very good UI designers), and get frustrated at the program not working as well as it could
- friends will sometimes change the option as a prank, and then the user won't even know how to change it back and get frustrated about the program not working as expected

And this doesn't even get into the problem that every option increases the amount of possible code paths (for each on/off option, you double the number of possible states the program could be in), which drastically increases potential for bugs.

And of course:

- instead of hiding an option in an obscure configuration menu or command, the program should just do the right thing all the time

This isn't always an "always right" answer, but a lot of the time there is. A lot of the time, options are just political additions, because different programmers think different ways of doing things are best.

But I digress. There are advantages to configuration, I know. But all advantages have to be weighed against that list of disadvantages.

> Having a command would at least give them the choice of which rooms to autojoin and in what order to join them without having to keep them open at all times to maintain the list, if this way of adding rooms isn't a mystery to them.

You're sacrificing a lot of intuitiveness/ease-of-use for the ability to choose the order of their rooms on next startup???

Seriously, though, think of it this way: How many people have Chrome set to "reopen the tabs I had the last time I closed Chrome" and how many people have Chrome set to "open this specific list of tabs on each startup"?

Your problem here is that you're conflating two features:

1. not losing all your rooms when you close/reopen PS or when PS restarts
2. remembering a list of rooms even after you leave them or add more rooms

By trying to do both in one feature, you're making both worse.

Joimer:
I have a solution that would fit both advanced users and ten year old users:
Make your suggestion the default option, but add a configuration option in which you can edit the autojoin list yourself, should you want to.

That is, keep ajoin list updated automatically upon joining / leaving rooms up to the six limit, but if you choose this option, you may use commands to add and remove rooms from ajoin as well as see the ajoin state.

Zarel:
I find it mildly ironic that your solution to "too much configuration is sometimes a bad thing" is to add even more configuration.

Honestly, I do understand why you'd want to maintain an autojoin list separately from the list of rooms you're currently in. You want to have a two-tiered system: a long-term list and a short-term list.

But there's seriously no strong reason why your long-term list should be tied to an autojoin feature. I mean, seriously, you're overcomplicating matters:

> without having to keep them open at all times to maintain the list

Or maybe you shouldn't assume someone wants to stay in a room when they leave it, or that they want to leave a room when they join it.

On the off chance a user wants to leave a room when they join it, we have a close button for that.

On the off chance a user wants to rejoin a room later when they leave it, we have a Join room button for that.

And on the off chance you really do want to remember a list of rooms you care about long term, separate from the rooms you're currently in, make it a separate feature, don't shove it into the autojoin feature.

Joimer:
But this is a case of "alternative configuration" more than "more configuration". It's transparent for the user, the user does not need to know about this, it's just an advanced configuration option for advanced PS users that will in no way affect other regular users that will be happy with your suggested feature.

I think this is part of an auto-join feature, at least that is how autojoin works on IRC and other chat channel systems, several MMOs implement the feature like this.

It can be another PR, that's for sure, but I'd say it is an extension to the autojoin feature, just an extra option that doesn't collide with the base feature. It's a way to have both functionalities, so the user can decide how they want to autojoin rooms.

Morfent:
>Seriously, though, think of it this way: How many people have Chrome set to "reopen the tabs I had the last time I closed Chrome" and how many people have Chrome set to "open this specific list of tabs on each startup"?
I think we have two different ideas on what autojoin's supposed to do. The list is simply meant to be rooms that would be joined anyways on startup, not a list of what rooms were open when PS was last closed. Keeping a list of bookmarks to visit when Chrome starts vs. going to each site from memory describes this better, but even then, browser history and PS chatrooms aren't very analogous. There's no reason to keep a history with the amount of rooms there are, and they aren't changing from day to day like the sites one'd visit.

>And this doesn't even get into the problem that every option increases the amount of possible code paths (for each on/off option, you double the number of possible states the program could be in), which drastically increases potential for bugs.
...but adding/removing rooms from the autojoin list every time a user joins or leaves a room would make this more of a problem than it would be using commands on a configured list.

> Make your suggestion the default option, but add a configuration option in which you can edit the autojoin list yourself, should you want to.
This only adds the configuration Zarel doesn't want for the default behaviour that I don't want.

Zarel:
The problem is, your "advanced" feature boils down to: "assume a user doesn't _really_ want to leave a room when they leave a room, and assume a user doesn't _really_ want to join a room when they join a room"

It's true, that's how IRC servers and some clients used to implement autojoin. But that was for technical reasons, because it was tacked onto a system that didn't have it in the past. Other clients, like Textual, do implement the straightforward-style "reopen the channels you last had open".

I really think the only reason you or Morfent want "persist rooms through reopening" and "manually keep track of a long-term list of rooms" in the same feature is because you're used to it for historical reasons, not because it's actually a good idea.

Zarel:
> The list is simply meant to be rooms that would be joined anyways on startup, not a list of what rooms were open when PS was last closed.

So, you don't really want to stay in a room when you join a room but you also think it's too much work to click the close button. Sure, I guess some people are like that.

But do you really think that desire is so common that you'd want to design a feature around it _and_ make that the default behavior?

> ...but adding/removing rooms from the autojoin list every time a user joins or leaves a room would make this more of a problem than it would be using commands on a configured list.

It would be done on every user's computer, so if there was a bug in it, it would be obvious. It's a single path that everyone goes down.

On the other hand, a configured list would probably only be used by a few users, so if it breaks some other rarely-used feature, that code-path wouldn't see much use, and we'd probably never realize that.

_That's_ the problem with having too many code-paths.

TalkTakesTime:
It more boils down to "let the user decide if they want to keep rejoining a room without requiring them to do maintenance on their autojoins every time they leave the server". Forcing a user to trim the rooms they're in before they close PS every time is an unnecessary burden, and one that makes assumptions that aren't really true. I know that I personally often join rooms when I'm on PS and don't want to rejoin them but don't close all of them before I leave because why should I have to? PS shouldn't be making the decision about which rooms someone wants to join, the user should. Sure, the user can make the decision by closing and opening rooms, but they shouldn't be having to do that every time they close PS...

Not only that, but when users don't want to be in any rooms, or they decide to leave a room they normally spend a lot of time in for some reason, they shouldn't be forced to rejoin all the other rooms they regularly use before they close PS just so they can autojoin them. When someone closes a room they're regularly in, it doesn't necessarily mean they no longer want to join it every session; it just means they're opting out for the rest of this session. Why should PS be telling users how they want their autojoin list to work, or forcing them to do *more* to get it to work how they wish?

> And having rooms saved automatically is much more intuitive than having a /command. A command is literally the opposite of intuitive.

Considering how heavily PS uses commands, I think the vast majority of users are familiar enough with them that a command is more intuitive than forcing them to adjust the rooms they're in just so that they can leave. In addition, a command is a mostly one-off cost (presumably with infrequent updates thereafter) whereas making it automated means the user has to either opt out to get rid of the cost, or they have to trim the rooms they're in before they leave regularly. Which of these makes more sense?

> The limit is 6 rooms to discourage users from spending time in too many rooms (because more users in more rooms means more server resource usage i.e. lag, and they're not going to be particularly active in all those rooms anyway).

We shouldn't be policing the rooms a user is in, unless it's for something like a roomban. Why should we punish the end users for a performance issue which isn't even vaguely their fault? Let them join and stay in whatever rooms they like, and instead seek to reduce the cause of the issue. It's not their fault and we shouldn't treat them as second class citizens because *we've* written code that isn't suitable for the number of users we get.

> So, you don't really want to stay in a room when you join a room but you also think it's too much work to click the close button. Sure, I guess some people are like that.

> But do you really think that desire is so common that you'd want to design a feature around it and make that the default behavior?

Uh, short answer: yes. People join rooms for single sessions regularly. Because someone suggested they should, because they were curious, because something was happening in that room, or many other reasons. A lot of the time they'll close the rooms before they close PS, but a lot of the time they won't. Autojoin should be the rooms a user *chooses* to join every session, not the ones PS tells them they should. Forcing a user to do *more* work for a feature that is supposed to be about convenience is extremely counter-intuitive.

> But there's seriously no strong reason why your long-term list should be tied to an autojoin feature.
> [...]
> And on the off chance you really do want to remember a list of rooms you care about long term, separate from the rooms you're currently in, make it a separate feature, don't shove it into the autojoin feature.

"Here's a list of rooms that I would like to automatically join every time I go on PS, regardless of what I might have done last time." What exactly is this long-term list other than *exactly* what autojoin is? I'm not really sure what you're meaning there. The long-term list of regularly joined rooms is precisely what autojoin is desired for, according to every user I've encountered who either uses an autojoin script or wants a PS native one.

Morfent:
>So, you don't really want to stay in a room when you join a room but you also think it's too much work to click the close button. Sure, I guess some people are like that.

PS chatrooms can/t be idled in the same way IRC channels can. Of course I have to leave some rooms if I'm busy with coding or something, but others are fine regardless. The latter would be fine for autojoin, while the former would get caught up in automatic autojoin. I'm not exactly sure what your point is here.

>On the other hand, a configured list would probably only be used by a few users, so if it breaks some other rarely-used feature, that code-path wouldn't see much use, and we'd probably never realize that.
>That's the problem with having too many code-paths.

By this logic /highlight would also be buggy and rarely used, yet bugs were still fixed and the command is used fairly often. I honestly don't see /autojoin being an unpopular command considering how often it's suggested,

TalkTakesTime:
You're right, configuration *is* often a bad thing. So why would you add a feature that requires reconfiguring every time the user comes on PS and does something that isn't part of their normal routine? 

I understand the desire to not have the command and for it to be automated. The issue is that you're removing the up front cost in exchange for requiring the user to perform constant maintenance on their configuration. That would be ok if autojoin were a short term thing, but talking to the users in my time as both staff and a voice, I've only ever heard requests for a *long-term* list of rooms that get rejoined every session. I think the main issue here is that you're assuming users want it to be a short term feature when to the best of my knowledge, and to those I've seen mention or request this feature, that assumption is untrue. 

Not only that, but I've only ever known users to request that there be a feature which allows *them* to specify a list of rooms to automatically join. Sure, I haven't talked to everyone so that might not be the view of the majority but when it's the view of *everyone* I've talked to about it, it sure seems like it's the desired feature

kotarou3:
How about we add these features:
1. Allow room reordering, and saving that order even if they're closed
2. Keep track of frequently visited/used rooms in the room list to allow easy rejoining of closed but often used rooms

Then just auto-joining the first 6 rooms isn't that big of a problem anymore

Zarel:
> We shouldn't be policing the rooms a user is in, unless it's for something like a roomban. Why should we punish the end users for a performance issue which isn't even vaguely their fault? Let them join and stay in whatever rooms they like, and instead seek to reduce the cause of the issue. It's not their fault and we shouldn't treat them as second class citizens because we've written code that isn't suitable for the number of users we get.

What does it matter whose fault it is? Problems don't just go away just because they're not your fault.

You are welcome to try to fix our performance problems, but in my over-three years on PS, I've never seen anyone other than me and a few times Cathy/bmelts have any effect on lag besides make it worse.

It seems like you are laser-focused on whatever pet feature you care about at the moment and are unwilling to work with others with other priorities.

Zarel:
> It more boils down to "let the user decide if they want to keep rejoining a room without requiring them to do maintenance on their autojoins every time they leave the server".

So forcing a user to leave a room when they close PS makes more sense? The only reason you like your approach more is because you're used to it.

> Not only that, but I've only ever known users to request that there be a feature which allows them to specify a list of rooms to automatically join.

“If I had asked people what they wanted, they would have said faster horses.”
– attributed to Ford but he probably didn't actually say it

This isn't because people are dumb, but more because people don't actually realize that there's an option better than a faster horse.

> Considering how heavily PS uses commands, I think the vast majority of users are familiar enough with them that a command is more intuitive than forcing them to adjust the rooms they're in just so that they can leave.

Yes, "clicking a close button just so they can leave". Oh, the humanity! How will users ever survive leaving a room the normal way? What next, users will have to click the room or type `/join roomname` to join rooms? How can people survive that way? They'd definitely be better off if TTT decided for them when they want to join rooms and when they want to leave rooms!

I'm kidding, but I think it's an important point. A user can decide for themself when to leave a room by pressing the Close button; they don't _really_ need you to decide for them.

Instead of privileging your own entrenched preferences, let's look from scratch: A user joins a room, then closes PS. What does this mean?

- Well, maybe PS just restarted. Then it'd be nice to resume where you left off.

- Or maybe you accidentally closed PS. Then it'd be nice to resume where you left off.

Seriously, even if you do want to temporarily join rooms: what if you're interested in a room but forget to add it to autojoin?

I thought about it a lot, and in the end, there are too many reasons why you'd want to close PS, and in only a few of them would you actually want to leave all rooms, even if you do otherwise enjoy the long-term room-tracking IRC-style autojoin provides. And on the off chance you do want to leave a room, the close button is still right there.

Since the close button works consistently, there's no need to introduce another feature that closes rooms, sometimes when the user might not want to.

Joimer:
To be honest we should tone down the sarcasm and condescence a bit (especially towards users, who may be reading this), especially since the discussion is whether to add both systems, Zarel's by default, and not either.

My two cents.

Slayer95:
Appearances, appearances... That's why I dislike so much the emphasis you give to "What's going on PS dev" and similar stuff...

My stance on this:

Any implementation of autojoin will be better than what we currently have, so I don't really care any way.

However, I am convinced that the great majority of users -i.e. regulars- will be better served by the "automatically remember" implementation. What would bring the most benefit to staff, voices, and long-standing lurkers, I don't know, but it doesn't necessarily coincide with what they <i>think</i> will benefit them the most.

Settings and the power of choice are often overrated. However, I am really having a hard time imagining how much more difficult and bug-prone supporting both options would be.

Joimer:
You missed the point on that list, then, because it is double purpose: to let people know what's being done in PS at the moment and to let developers know who is doing what to avoid double work and so new contributors can pick an issue that is not being worked and do it (like it's been happening since I first posted the list).

That is not relevant to this PR, though, and I also have a hard time figuring out what's the problem in having the two features. Power of choice might be overrated, but it is a requested feature. That shouldn't be disregarded.

Morfent:
After thinking over this for a while, I've decided that the automatic autojoin would be better off for the avarage user. I wouldn't expect them to figure out on their own about the command or how to use it, considering many of them go mostly unknown because they're not documented outside their own /help command. Hell, I even forgot what the name for ``/packhidden`` was earlier when I was the one who wrote it in the first place. Automatic autojoin would allow a greater number of users to use the feature without having to lurk in the PS community until they find out about it.

However,  I don't necessarily think that picking just the automatic or the command autojoin alone is a good idea, since they appeal to separate groups  of users. With the command, those who don't keep up to date or hang around in chat for long periods of time won't know that the command even exists, while with the automatic autojoin, those who would gain from the control that the command gives them would instead be inconveneinced by having to manage their room tabs and deal with closing them every time the page loads to accomplish what the command could've done on its own. Despite your condescencion towards my arguments about havimg to close the rooms on page load, I think of it as being along the same kind of logic as not having autojoin at all because ``/join`` exists. Forcing users to autojoin is not a perfect solution for this, and would benefit users who'd use that at the expense of those who's rather not, or choose where they join manually. To me, not being able to control behaviour like this shows that there's a flaw in how it's designed. Ideally, users should choose whether or not to use autojoin at all, or to be able to have a choice in how the feature works.

>I thought about it a lot, and in the end, there are too many reasons why you'd want to close PS, and in only a few of them would you actually want to leave all rooms, even if you do otherwise>enjoy the long-term room-tracking IRC-style autojoin provides. And on the off chance you do want to leave a room, the close button is still right there.

Honestly, I would not use autojoin at all, so this would happen in all situations, not in just a few. Clicking the close button, then joining the rooms when I want to later is even more time-consuming than not having autojoin at all. What about other users who don't want to use this either? I originally decided on a command so users can have the choice not to To be blunt, you also seem to be lazer-focused on your own priorities for this.

>Settings and the power of choice are often overrated. However, I am really having a hard time i>magining how much more difficult and bug-prone supporting both options would be in this case.

Honestly it wouldn't be any more difficult than writing either of the two on their own, considering I've already thought through how it would be laid out in case I do end up adding both.

Anyways, Zarel's autojoin should be finished.

js/client.js
```diff
@@ -1374,6 +1387,50 @@
...
 
+		// returns the first 6 valid sideRooms from app.rooms
+		sliceAutojoins: function() {
+			var autojoins = [];
+			for (var i in this.rooms) {
+				if (!this.rooms[i].isSideRoom || i === 'rooms' || i === 'staff') {
+					continue;
+				}
+				autojoins.push(i);
+				if (autojoins.length === 6) break;
```
Zarel: `>= 6`.

Less brittle.

js/client.js
```diff
@@ -1374,6 +1387,50 @@
 			}
 		},
 
+		// returns the first 6 valid sideRooms from app.rooms
+		sliceAutojoins: function() {
+			var autojoins = [];
+			for (var i in this.rooms) {
+				if (!this.rooms[i].isSideRoom || i === 'rooms' || i === 'staff') {
```
Zarel: `this.rooms[i].type !== 'chat' || i === 'staff'` might be a better filter.

js/client.js
```diff
@@ -1374,6 +1387,50 @@
...
+					continue;
+				}
+				autojoins.push(i);
+				if (autojoins.length === 6) break;
+			}
+			return autojoins;
+		},
+		addAutojoin: function(id, auto) {
+			var autojoins = Tools.prefs('autojoins') || [];
+			if (autojoins.indexOf(id) >= 0 || autojoins.length === 6) {
```
Zarel: `>= 6`

js/client.js
```diff
@@ -361,6 +361,13 @@
 				self.addPopup(ReconnectPopup, {cantconnect: true});
 			});
 
+			this.once('init:choosename', function() {
+				var autojoins = Tools.prefs('autojoins');
+				if (autojoins && autojoins.length) {
+					self.send('/autojoin ' + autojoins.join(', '));
```
Zarel: `','` not `', '`

js/client.js
```diff
@@ -806,12 +813,17 @@
...
+
 				if (this.rooms[roomid] || roomid === 'staff') {
 					// staff is always joined in background
 					this.addRoom(roomid, roomType, true);
 				} else {
 					this.joinRoom(roomid, roomType, true);
 				}
+
+				if (roomType === 'chat' && roomid !== 'staff') {
+					this.addAutojoin(roomid, true);
```
Zarel: Rerunning `sliceAutojoin` might be more reliable than manually doing `addAutojoin` and `removeAutojoin` each time. It's also more likely to be in the right order.

Morfent: ``addAutojoin`` and ``removeAutojoin`` already do run ``sliceAutojoin`` except when removing a room from the list when its length is less than 6. Would there be a way for the rooms to get out of order in this case, or is the difference in performance between splicing the room out of ``autojoins`` and using ``sliceAutojoins`` not worth the risk?

None of the three methods are actually necessary if no command is ever added.

Zarel: The difference in performance is really not worth it: remember that `sliceAutojoins` will loop a maximum of six times and is basically O(1). More important is the difference in complexity and brittleness (same reason why I said to use `>= 6` instead of `=== 6`).

Morfent: Ah, ok

Morfent:
I guess there's no point in stalling over this if the essential part's already finished.

TheImmortal:
>Seriously, though, think of it this way: How many people have Chrome set to "reopen the tabs I had the last time I closed Chrome" and how many people have Chrome set to "open this specific list of tabs on each startup"?

I'm sold. That is literally the best browser feature imo and I was skeptical about your idea at first but it is definitely the better implementation.

Morfent:
To clarify my last comment, I'm fine with just adding your autojoin system. If any form of customization can be added, it can be added at another point in time. I didn't intend for it to come across as having any other kind of meaning like it did there.

Zarel:
The problem right now is more because `/autojoin` isn't committed server-side because I'm still investigating issues with its implementation.

Zarel:
Done in b5bcbb13dd88 and 547629ba282

PR #365, opened 2015-04-28
Improve parsing of HTML entities for URLs.

xfix:
Previously, link like `"http://example.com/"` was changed into `"http://example.com/";` because of wrong parsing of entities.


Zarel:
I don't understand.

Zarel:
Oh, I do now.

xfix:
What is status of this pull request?

PR #366, opened 2015-04-30
Various Graphics Animation Changes

Pikachuun:
Basically, I tried to make them more accurate to the games, except for Spark + Nuzzle, which I believe have too "strong" animations at this time.
The list:

Mist Ball [Old animation moved to Moonblast]; Luster Purge; Hyperspace Hole; Spark + Nuzzle; Icy Wind + Powder Snow; Blue Flare [Update; fire is now blue]; Dragon Ascent [Update; collision effects are greener].


Zarel:
@ascriptmaster what do you think of these?

ascriptmaster:
I remember someone mentioning that it was a bad idea to have a graphic that's dedicated to a single move only. Although, Gen 4 and up have blue/purple flames for Sacred Fire (especially in gen 6, which is entirely blue), so you could use the blue flame for that move as well.

Layell:
We could do blue flame for Wil-O-Wisp too, ingame ORAS it's purplish but description is: "The user shoots a sinister, bluish-white flame at the target to inflict a burn."

ascriptmaster:
The flame for Will-o-Wisp is purple in Gen 6 from what I know. But we could use blue.

Zarel:
Blue is close enough, yes.

ascriptmaster:
fireball - copy

I thinkered around with the hue/saturation/brightness on the fireball image and got this, I wonder if it'll look better haha

Zarel:
Our original fireball image sucks; can you find a freely-licensed fireball that doesn't suck?

ascriptmaster:
Alas I'm not good at finding freely-licensed fireballs or flames. Maybe I should just Photoshop something, but then I can't guarantee that it's any better than what we have.

Zarel:
Worth a try.

ascriptmaster:
fire

If you say it's worth a try, then here is my try.

Zarel:
On first glance it looks better than what we had before, but it might be nice to see screenshots side-by-side.

Also, multiple fireball sprites might make some animations look better.

Slayer95:
@ascriptmaster @Pikachuun  what's the state of this?

ascriptmaster:
Oh, I forgot to finish crafting a better flame image.

@Pikachuun, can you shelve Blue Flare change for now? I'll work on it later and we can merge the rest for now.

PR #368, opened 2015-05-05
Update 'Ignore Spectators' option

sirDonovan:


Zarel:
Now that we show ranks in battle rooms, we might want to change Ignore Spectators to work by rank.

sirDonovan:
Good idea, updated to ranks.

Zarel:
I've implemented it myself. The main remaining issue was that I'd rather have it allow everything but unauthed and voice.

PR #369, opened 2015-05-05
Improve teambuilder imports

Marty-D:
PO exports a Pokemon with no item as "POKEMON @ (No Item)", which is not
a valid item in this teambuilder.


Zarel:
is that the usual way to make a set without an item? Not like `curSet.item = '';` or something?

Marty-D:
I don't think there is a usual way, but your way would probably be more in line with most of the codebase. Changed!

PR #370, opened 2015-05-05
Add option to ignore opponents

sirDonovan:
Places opponents on the ignore list and causes their Pokemon to display
species instead of nicknames

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -520,14 +520,14 @@ var Pokemon = (function () {
 		if (this.side.n === 0) {
 			return Tools.escapeHTML(this.name);
 		} else {
-			return "The opposing " + Tools.escapeHTML(this.name);
+			return "The opposing " + (this.side.battle.ignoreOpponent ? this.species : Tools.escapeHTML(this.name));
 		}
 	};
 	Pokemon.prototype.getLowerName = function () {
 		if (this.side.n === 0) {
 			return Tools.escapeHTML(this.name);
 		} else {
-			return "the opposing " + Tools.escapeHTML(this.name);
+			return "the opposing " + (this.side.battle.ignoreOpponent ? this.species : Tools.escapeHTML(this.name));
 		}
 	};
 	Pokemon.prototype.getTitle = function () {
@@ -540,8 +540,8 @@ var Pokemon = (function () {
 		return titlestring;
 	};
 	Pokemon.prototype.getFullName = function (plaintext) {
-		var name = Tools.escapeHTML(this.name);
-		if (this.name !== this.species) {
+		var name = this.side.n && this.side.battle.ignoreOpponent ? this.species : Tools.escapeHTML(this.name);
+		if (name !== this.species) {
 			if (plaintext) {
 				name += ' (' + this.species + ')';
 			} else {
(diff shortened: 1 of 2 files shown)


Zarel:
Apparently our previous "Ignore Spectators" feature is unpopular, and a strawpoll had 90% wanting it gone from the UI.

We might want to put this on hold for now.

sirDonovan:
I've updated the code to match the new location of 'Ignore Spectators'. The backlash has died down since moving it and changing it from a button to a check-box. Can we consider this now?

js/client-battle.js
```diff
@@ -22,7 +22,7 @@
 
 			this.$chat = this.$chatFrame.find('.inner');
 
-			this.$options = this.battle.optionsElem.html('<div style="padding-top: 3px; text-align: right"><label style="font-size: 8pt; padding: 3px 5px"><input type="checkbox" name="ignorespects" /> Ignore Spectators</label></div>');
+			this.$options = this.battle.optionsElem.html('<div style="padding-top: 3px; text-align: right"><label style="font-size: 8pt; padding: 3px 5px"><input type="checkbox" name="ignorespects" /> Ignore Spectators</label><label style="font-size: 8pt; padding: 3px 5px"><input type="checkbox" name="ignoreopp" /> Ignore Opponent</label></div>');
```
Zarel: Space between the two labels.

js/client-battle.js
```diff
@@ -691,6 +692,17 @@
 		toggleIgnoreSpects: function(e) {
 			this.battle.ignoreSpects = !!e.currentTarget.checked;
 		},
+		toggleIgnoreOpponent: function(e) {
+			if (!!e.currentTarget.checked) {
+				app.ignore[toUserid(this.battle.yourSide.name)] = 1;
+				this.battle.ignoreOpponent = true;
+				this.battle.add('User ' + this.battle.yourSide.name + ' ignored.');
```
Zarel: We might want to do this message for `toggleIgnoreSpects` too.

Zarel:
Anyway, yes, I figured changing the location would be enough, which is why I kept this pullreq open instead of closing it.

Zarel:
Also do you also ignore opponent's chat messages? I couldn't find the line where you did.

sirDonovan:
Yeah, the line that adds them to <code>app.ignore</code> ignores their chat messages https://github.com/Zarel/Pokemon-Showdown-Client/blob/master/js/battle.js#L5334

js/client-battle.js
```diff
@@ -690,6 +691,18 @@
 		},
 		toggleIgnoreSpects: function(e) {
 			this.battle.ignoreSpects = !!e.currentTarget.checked;
+			this.battle.add('Spectators ' + (this.battle.ignoreSpects ? '' : 'no longer ') + 'ignored.');
+		},
+		toggleIgnoreOpponent: function(e) {
+			if (!!e.currentTarget.checked) {
+				app.ignore[toUserid(this.battle.yourSide.name)] = 1;
```
Zarel: I'm not sure if this is the right implementation. It's not clearly too clearly communicated that the user is ignored globally and not just in the battle. We might want to ignore the user just in the battle.

sirDonovan: That's a good point, updated to ignore only in battle.

js/battle.js
```diff
@@ -5331,6 +5331,7 @@ var Battle = (function () {
 		case 'c':
 			name = args[1];
 			if (this.ignoreSpects && (name.charAt(0) === ' ' || name.charAt(0) === '+')) break;
+			if (this.ignoreOpponent && this.getSide(name) === this.yourSide) break;
```
Zarel: We could just test for Player symbol.

sirDonovan: I thought about doing that, but it would result in users being unable to see their own messages, like using /ignore on yourself. It would probably be confusing if that happened instead of only ignoring the opponent's messages.

Zarel: Good thinking. What if we tested for Player symbol other than your own name?

Zarel: I think the way it's currently written, it'd behave weirdly if the opponent changed their username (or username capitalization), and how do we want /addplayer to behave?

sirDonovan: Updated. Not sure what to do with <code>/addplayer</code>, but I don't think we'll see very many situations where someone has both their opponent ignored and the need to replace them.

js/battle.js
```diff
@@ -5331,6 +5331,7 @@ var Battle = (function () {
 		case 'c':
 			name = args[1];
 			if (this.ignoreSpects && (name.charAt(0) === ' ' || name.charAt(0) === '+')) break;
+			if (this.ignoreOpponent && name.charAt(0) === '★' && toUserid(name) !== app.user.get('userid')) break;
```
Zarel: You're not going to use the \unumbers form? I tend not to like Unicode in code files.

sirDonovan: Oh my mistake, fixed that.

PR #371, opened 2015-05-07
Implement Drag and Drop Team reordering

QuiteQuiet:
I'm aware there are two pullreqs for this already, but this is a very "grounded" implementation based on mouse events rather than external dependencies. Visually it's a bit sub-par, but that's fixable and I don't know what the proper "PS look" of this would be.

In addition, I don't know how Drag-and-Drop should "feel" so this might be way off the mark...

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -28,6 +28,13 @@
 			'change select[name=format]': 'formatChange',
 			'change input[name=nickname]': 'nicknameChange',
 
+			// team reordering
+			'mousedown button.movable-btn': 'mouseDown',
+			'mousemove button.movable-btn': 'mouseMoved',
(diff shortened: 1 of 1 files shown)


SolarisFox:
Attempting to drag a team outside of the team list causes the team to be stuck selected even if the mouse click is released. You should probably force the team to drop if it moves outside the expected range.

Also to improve visual clarity, it might be nice to put a spacing buffer above and below the selected team so it's a bit easier to see it moving.

QuiteQuiet:
Updated to release a move if you stray too far from the list (exactly how far "too far" is can be argued) and changed the visual image of reordering so it looks like this: http://puu.sh/hGL17

QuiteQuiet:
Mostly worse than 64c22a9

PR #373, opened 2015-05-08
Sprite gen preferences should override battle gen

### LINKED ISSUE #372: BW animations strech
http://i.imgur.com/MEDXPBD.png

https://github.com/Zarel/Pokemon-Showdown-Client/commit/299777f297254e39399bcd42ab5738e67384c6c6

Try switching to bw animated and go to a battle, still loading xy stuff.

I'm not sure what broke here but I know part of the original addition was that there would be a sprite reload every time we selected bw animations over xy stuff. If it is this commit that broke it can we get it reverted until it is fixed?

QuiteQuiet:
Checking for preferences if bw or xy sprites should be loaded, even if the gen you're playing want something else. Fixes #372.

(This really should've been in the original pull req.)

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -847,8 +847,13 @@ var Tools = {
 			facing = 'back';
 		}
 
+		// Decide what gen sprites to use.
+		var gen = {1:'rby', 2:'gsc', 3:'rse', 4:'dpp', 5:'bw', 6:'xy'}[options.gen];
+		if (Tools.prefs('nopastgens')) gen = 'xy';
+		if (Tools.prefs('bwgfx') && gen === 'xy') gen = 'bw';
+
 		var animationData = {};
-		if (options.gen === 5 && window.BattlePokemonSpritesBW) {
+		if (gen === 'bw' && window.BattlePokemonSpritesBW) {
 			animationData = BattlePokemonSpritesBW && window.BattlePokemonSpritesBW[pokemon.speciesid];
 		} else {
 			animationData = BattlePokemonSprites && window.BattlePokemonSprites[pokemon.speciesid];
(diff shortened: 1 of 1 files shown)


js/battledata.js
```diff
@@ -848,7 +848,7 @@ var Tools = {
 		}
 
 		var animationData = {};
-		if (options.gen === 5 && window.BattlePokemonSpritesBW) {
+		if ( (options.gen === 5 || Tools.prefs('bwgfx') ) && window.BattlePokemonSpritesBW && !Tools.prefs('nopastgens')) {
```
Rhythms88: Why do you have spaces between each parentheses

QuiteQuiet: I have them because I find spaces make it easier to pair up each pair of parentheses in multiconditional statements like `(options.gen === 5 || Tools.prefs('bwgfx')`. If PS do things without those, I'll remove them.

Rhythms88: PS! does things without spaces between parentheses.

Slayer95:
Protip: we determine the gen in https://github.com/QuiteQuiet/Pokemon-Showdown-Client/blob/fix372/js/battledata.js#L875

QuiteQuiet:
Moving that section above animation selection doesn't change anything else, so it's a better way of deciding the generation...

js/battledata.js
```diff
@@ -847,8 +847,13 @@ var Tools = {
 			facing = 'back';
 		}
 
+		// Decide what gen sprites to use.
+		var gen = {1:'rby', 2:'gsc', 3:'rse', 4:'dpp', 5:'bw', 6:'xy'}[options.gen];
+		if (Tools.prefs('nopastgens')) gen = 'xy';
+		if (Tools.prefs('bwgfx') && gen === 'xy') gen = 'bw';
```
Zarel: Why only for gen 6? o_O

Why do we only have an override for gen 6 -> 5 anyway?

Zarel:
Also, how does this fix that? Shouldn't the displayed sprite be the same as the file the sprite dimensions were loaded from?

Zarel:
Oh, I understand now.

js/battledata.js
Slayer95: Because you rejected having more options here qwq
https://github.com/Zarel/Pokemon-Showdown-Client/pull/293#discussion_r22302066

Zarel: I rejected having a gen-5-only option too. Why did we add one anyway? :|

PR #375, opened 2015-05-09
handle long team name overflow

SolarisFox:
prevents this from happening: http://prntscr.com/736fcq


Zarel:
I'm pretty sure you don't need to add an extra span for this.

SolarisFox:
Updated

PR #377, opened 2015-05-12
Adding tiled custom background option

mashirochan:
Added a drop down box during custom background selection to have the option to make it tile, rather than stretch the image across the whole screen. Would be a useful feature to have if some users find the need, but not necessary if they don't want it.

screenshot 2015-05-11 at 11 02 13 pm

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -319,7 +319,7 @@
 				if (bg) {
 					$(document.body).css({
 						background: bg,
-						'background-size': 'cover'
+						'background-size': Tools.prefs('coverType') ? Tools.prefs('coverType') : 'cover'
 					});
 				} else if (Config.server.id === 'smogtours') {
 					$(document.body).css({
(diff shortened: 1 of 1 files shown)


SolarisFox:
@xfix thoughts?

Zarel:
This isn't great UI... now that the window has more than one option, there should probably be an "OK" button instead of automatic-take-effect-when-file-is-selected.

But I'm not sure how useful this is. Do you personally want to use this feature?

mashirochan:
There are a lot of cool tiled backgrounds out there that I would probably use myself, given the chance. And I could move the drop-down menu and the file input on the same line and then add a Submit button underneath next the the Cancel button (still working on getting both inputs on the same line).

screenshot 2015-05-12 at 3 32 46 pm

Zarel:
That looks better, yes.

Zarel:
We've overhauled the background menu. If you feel tiled backgrounds are important enough to implement in the new menu, feel free to pullreq an updated version.

PR #379, opened 2015-05-15
Move index.template.html character set declaration to HTTP

xfix:
According to [WHATWG specification][1], there are a few of requirements that a character set declaration must meet. One of these is a following one.

> The element containing the character encoding declaration must be serialised completely within the first 1024 bytes of the document.

Beginning of a document contains a **huge** comment (pretty much 2 kilobytes). Either it would have to be moved to be after character set declaration, character set declaration would have to be moved to HTTP layer, or a comment would have to be remove entirely. I decided on a second option, because I couldn't think of a way to move this comment without making it uglier.

[1]: https://html.spec.whatwg.org/multipage/semantics.html#charset

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -5,6 +5,7 @@ AddType application/x-web-app-manifest+json .webapp
 
 <filesMatch "^index\.html$">
   FileETag None
+  AddCharset UTF-8 .html
   <ifModule mod_headers.c>
      Header unset ETag
      Header set Cache-Control "max-age=0, no-cache, no-store, must-revalidate"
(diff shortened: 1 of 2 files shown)


Zarel:
I'd rather move the comment. HTTP headers are horrible places to put anything, honestly.

Zarel:
Specifically, this is because the charset is a property of the page (most HTTP headers like it are), so it should stay with the page. If you save the page as `.html`, things like charset should stick with it, but they don't unless they're in meta tags rather than as a header.

PR #380, opened 2015-05-15
Disable Chrome Frame

xfix:
Chrome Frame is not supported for two years, which means removing support for it on Pokémon Showdown should be safe.


Zarel:
I think Chrome Frame is mentioned in other places; can you remove those, too?

xfix:
bump?

PR #382, opened 2015-05-17
linking to pullreqs

SolarisFox:
Thought this might be useful.

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -396,19 +396,10 @@ var Tools = {
 					'" target="_blank" onclick="' + onclick + '">' + uri + '</a>';
 			});
 			// google [blah]
-			// google[blah]
 			//   Google search for 'blah'
 			str = str.replace(/\bgoogle ?\[([^\]<]+)\]/ig, function(p0, p1) {
 				p1 = Tools.escapeHTML(encodeURIComponent(Tools.unescapeHTML(p1)));
 				return '<a href="http://www.google.com/search?ie=UTF-8&q=' + p1 +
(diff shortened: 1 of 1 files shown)


SolarisFox:
On a somewhat related note, we might want to remove `gl [blah]` because I don't think anyone uses that or is even aware that it exists. `[[blah]]` notation seems to be the common one.

Morfent:
I use the Google and Wiki linking notations occasionally, but I think they're rarely used because they're undocumented. Adding them alongside the ``[[foo]]`` notation in the setting to enable/disable that type of linking might help bring more exposure to them.

SolarisFox:
I wasn't saying to remove the google and wiki linking, just the I'm Feeling Lucky one because it's redundant with `[[blah]]` and `gl [blah]` isn't very self-explanatory with what it does.

Zarel:
Yeah, I agree with removing `gl`.

For this, I'd want it to be less likely to get accidentally trigger. Something like "client issue #"

SolarisFox:
Updated to use "issue" in linking to pullreqs
removed `gl [blah]` notation

js/battledata.js
```diff
@@ -396,21 +396,12 @@ var Tools = {
 					'" target="_blank" onclick="' + onclick + '">' + uri + '</a>';
 			});
 			// google [blah]
-			// google[blah]
```
SolarisFox: I removed this line note because it is now the only entry that indicates the space is optional with a line note, so I felt it was not only unnecessary but also inconsistent with the other entries.

PR #384, opened 2015-05-19
Fix modjoin problem

ascriptmaster:
Solution as described in https://github.com/Zarel/Pokemon-Showdown/issues/1787

I don't know if this actually works because this exact problem isn't exactly the easiest thing to test given the `testclient.html` stuff we have.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -849,7 +849,7 @@
 				}
 				// handle error codes here
 				// data is the error code
-				if (data === 'namerequired') {
+				if (data === 'namerequired' || (!isdeinit && !this.user.named)) {
 					this.removeRoom(roomid);
 					var self = this;
 					this.once('init:choosename', function() {


ascriptmaster:
@Zarel, do you think this is a good solution?

ascriptmaster:
errrrrgh. bump @Zarel again

Zarel:
Not a horrible solution

ascriptmaster:
Beautiful. Is there any way to even test this? I've found it virtually impossible to check with `testclient.html`

Slayer95:
It works. Tested at http local custom client.
But I don't think there is a need to break support for `'|deinit|namerequired`. It could be useful for roomkick-like commands, or to hook it to /modjoin.

ascriptmaster:
Did I accidentally break `namerequired`? Whoops.

ascriptmaster:
Hmmm, I don't understand. Did I really break `namerequired` with this code?

Slayer95:
What we currently use is `|noinit|namerequired|`. You broke support for `|deinit|namerequired|`, which is potentially useful; that's what I mean.

ascriptmaster:
So when I get the chance I should change it to `if (data === 'namerequired' || (!isdeinit && !this.user.named))`? All right.

ascriptmaster:
something like this, except idk again because idk how to set it up to test

ascriptmaster:
poke poke

Zarel:
This actually is a pretty good fix.

Zarel:
I ended up fixing it server-side, though.

ascriptmaster:
Since Zarel is evidently fixing the whole thing serverside, we don't need this.

PR #385, opened 2015-05-19
Saturate nick colors a bit more

xfix:
Now the saturation range is 51-100, and not 50-99.


Zarel:
...I was joking

TheImmortal:
So, should this be closed then?

Zarel:
Yes.

PR #386, opened 2015-05-20
Listen for -center messages

ascriptmaster:
Relavent server-side PR: https://github.com/Zarel/Pokemon-Showdown/pull/1896

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -3128,6 +3128,9 @@ var Battle = (function () {
 			if (!kwargs.simult) nextAnimDelay++;
 
 			switch (args[0]) {
+			case '-center':
+				actions += "Automatic center!";
+				break;
 			case '-damage':
 				var poke = this.getPokemon(args[1]);
 				var damage = poke.healthParse(args[2], true);


Zarel:
Does the message "Automatic center!" actually appear in-game?

Zarel:
Also, nothing appears to actually be centered here.

ascriptmaster:
Hmmm, good point. @Marty-D?

Also the server-side implementation sends the correct `swap` messages with `[silent]` to deal with the actual centering.

Zarel:
Oh, in that case `-center` is correct, you don't need `center` if the actual swapping is somewhere else.

Zarel:
But yeah, that's only if it even gives a message in the first place.

Marty-D:
Yeah, it gives a message. I updated the old move messages doc that's linked to on Trello months ago for Gen 6 stuff (with Gen 5 for comparison).

Zarel:
Might want to send `|swap|`...`[center]` then?

ascriptmaster:
Wouldn't `|swap|...[center]` end up being called twice during the automatic centering?

Zarel:
Would it? How/why?

ascriptmaster:
From server-side PR:

		if (this.gameType === 'triples' && this.sides.map('pokemonLeft').count(1) === this.sides.length) {
			// If both sides have one Pokemon left in triples and they are not adjacent, they are both moved to the center.
			var center = false;
			for (var i = 0; i < this.sides.length; i++) {
				for (var j = 0; j < this.sides[i].active.length; j++) {
					if (!this.sides[i].active[j] || this.sides[i].active[j].fainted) continue;
					if (this.sides[i].active[j].position === 1) break;
					this.swapPosition(this.sides[i].active[j], 1, '[silent]');
					center = true;
					break;
				}
			}
			if (center) this.add('-center');
		}

It forces both remaining Pokemon into the center position and send the `swap|...[silent]` message if they were moved. If neither Pokemon was in the center, and we change it to `swap|...[center]`, it could display twice.

So I guess you want me to change the server-side PR to send the `[center]` version once and the other one be `[silent]` if needed?

Zarel:
Oh, in that case, `-center` is fine.

Zarel:
btw, @ascriptmaster, this is the Trello board: https://trello.com/b/GvO2XGVg/pokemon-showdown

Is your message accurate?

ascriptmaster:
The Google Doc indicates that the message is unchanged from Gen 5, and is accurate.

Zarel:
That's... a really weird message.

PR #390, opened 2015-05-28
Add pokemon Icons to teambuilder teams list

Ecuacion:
I think adding Pokemon icons to each team in teambuilder teams list improves the appearance of the list. It also makes easier to recognize individual teams when you are looking for one and you have a lot.

An example of how it looks: http://imgur.com/0TQrOrd


Zarel:
Yeah, this was actually planned for the eventually.

Zarel:
The reason I didn't merge this is because I was in the middle of a major refactor that changed this as well as some other things: bbc2188a402b5

That refactor is now done.

PR #391, opened 2015-05-29
Add option to ignore in PM windows

sirDonovan:
Screenshot (updated): http://imgur.com/SQy6fML


Slayer95:
Nice feature.
Though I think the UI would benefit the most from the button being next to the chatbox. The button could be `<i class="icon-exclamation-sign"></i>`

sirDonovan:
I agree, it does look better that way. Updated screenshot: http://imgur.com/SQy6fML

ascriptmaster:
It looks nice, but does it actually ignore?

sirDonovan:
Yes, it works.

ascriptmaster:
No, just making sure since the actual disabling code isn't in the diff.

Slayer95:
(Missclicked <_<)

Assigning @Zarel because UI

Zarel:
I was planning on doing it in a different way...

This way is sort of ugly, and the buttons are a bit too close to each other.

Slayer95:
Ugly? D: Well, your call...
I noted the closeness issue as well, hence why I alternatively suggested a "next to chatbox" position.

We could add the ignore button in the user popup instead maybe?

Zarel:
That might be better. I had a different idea though.

sirDonovan:
Here's the button in the user popup, doesn't look as good as the PM window in my opinion: http://imgur.com/LdxbESq

What was your idea Zarel?

Zarel:
I would rather have a "More" menu, anyway. Ignore shouldn't be given equal weight as Challenge and Chat.

Irraquated:
I think it'd look nice inside a small settings icon top right of the user pop-up

Joimer:
I like the first version, it's done like that in many UIs as well. It's fast and easy to use.

sirDonovan:
Something happened while trying to update and fix merge conflicts, so I'll open a new pull request.

PR #392, opened 2015-05-30
Fix issues caused by accidental kwargs in battles

xfix:
This makes in particular messages starting with [ and ending with ] work in battles (in particular, messages with timestamps were affected by this bug). This also removes hack needed to make tier command with [Gen X] in format name work.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2162,7 +2162,6 @@ var Battle = (function () {
 	Battle.prototype.mute = false;
 	Battle.prototype.messageDelay = 8;
 
-
 	Battle.prototype.removePseudoWeather = function (weather) {
 		for (var i = 0; i < this.pseudoWeather.length; i++) {
 			if (this.pseudoWeather[i][0] === weather) {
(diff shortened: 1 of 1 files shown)


js/battle.js
```diff
@@ -5566,6 +5563,7 @@ var Battle = (function () {
 		}
 	};
 	Battle.prototype.run = function (str, preempt) {
+		var takesKwargs = {swap:1, move:1, cant:1, callback:1};
```
Zarel: I would rather have a whitelist of `nokwargs`.

Honestly, I'd rather go to a different approach altogether. We might want `|` as well as `[` as valid characters (e.g. in `|c|`), so it makes more sense to slice off the end of the string for a final argument of certain commands somehow.

xfix: Sounds reasonable, this would help implementing extra kwargs while preserving the compatibility. I would fix that... but I lately don't have much time, although that should be dealt with in this week.

Zarel: >this week

>nearly half a year ago

ascriptmaster:
One day... one day this will be resolved

Zarel:
Looks like this was fixed in f6c03e0370a81ac0dc1 (in Nov 2018)

(I added a `Battle.lineParse`, which is now `BattleTextParser.parseLine`, which uses a whitelist like I suggested in this PR.)

PR #393, opened 2015-05-31
Improve support for searching battles in multiple formats

Slayer95:
- `|updatesearch|{searching}` now allows an array of formats as a `searching` value.
- Old gen sprite data will be loaded based on the data sent by the server rather than the current DOM state

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -379,12 +379,7 @@
 			var $searchForm = $('.mainmenu button.big').closest('form');
 			var $formatButton = $searchForm.find('button[name=format]');
 			var $teamButton = $searchForm.find('button[name=team]');
-			if (this.searching) {
-				$formatButton.addClass('preselected')[0].disabled = true;
-				$teamButton.addClass('preselected')[0].disabled = true;
-				$searchForm.find('button.big').html('<strong><i class="icon-refresh icon-spin"></i> Searching...</strong>').addClass('disabled');
-				if ($formatButton.val().substr(0, 4) === 'gen5' && !Tools.loadedSpriteData['bw']) Tools.loadSpriteData('bw');
-			} else {
+			if (!this.searching || $.isArray(this.searching) && !this.searching.length) {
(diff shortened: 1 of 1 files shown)


js/client-mainmenu.js
```diff
@@ -379,11 +379,18 @@
 			var $searchForm = $('.mainmenu button.big').closest('form');
 			var $formatButton = $searchForm.find('button[name=format]');
 			var $teamButton = $searchForm.find('button[name=team]');
-			if (this.searching) {
+			if (this.searching && (!$.isArray(this.searching) || this.searching.length)) {
```
Zarel: What is `(!$.isArray(this.searching) || this.searching.length)` testing?

Slayer95: That `this.searching` is a plain object (like JSON data sent in current protocol), or an array with one or more elements (like JSON data sent in protocol proposed in https://github.com/Zarel/Pokemon-Showdown/pull/1910).

Zarel: It seems like it would be `true` for scalars, too (e.g. `235`, `true`, or `'foobar'`).

In other words, the conditional is true unless `this.searching` is `[]`. Which... is there a clearer way to write that?

Zarel: `!($.isArray(this.searching) && this.searching.length === 0)`

Slayer95: That would break backwards compatibility with the current protocol:
- Currently searching randbats: `{searching: {format: 'randombattle'}}`
- Currently not searching: `{searching: false}`

The clearer way to write the condition is invert the if/else blocks:

```
if (!this.searching || $.isArray(this.searching) && !this.searching.length) {
// do stuff
} else {
// do different stuff
}
```

But I didn't want to add unneeded deltas.

Zarel: I'm just saying this isn't particularly readable...

Slayer95:
Ok, then. Switched the conditionals for clarity.

PR #396, opened 2015-06-06
Sidebar update for dragIn

kubetz:
Simple fix for an open issue https://github.com/Zarel/Pokemon-Showdown/issues/1872 

`updateStatbar` is now called in `dragIn` same way as it is called in `switchIn`.


Zarel:
Thanks!

PR #399, opened 2015-06-07
Remove crossdomain issues on load for testclient over https

kubetz:
When testclient is run over https browser loves to complain and block mixed content which includes:
* pokemonshowdown's `//play.pokemonshowdown.com/js/config.js`
* calls to `action.php` like `upkeep`

Fix consists of using protocol-relative URL for `config.js` and not hardcoding protocol to `http://` in test mode for `getActionPHP`.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -143,7 +143,7 @@
 		getActionPHP: function() {
 			var ret = '/~~' + Config.server.id + '/action.php';
 			if (Config.testclient) {
-				ret = 'http://' + Config.origindomain + ret;
+				ret = 'https://' + Config.origindomain + ret;
 			}
 			return (this.getActionPHP = function() {
 				return ret;
--- a/testclient.html
+++ b/testclient.html
@@ -74,7 +74,7 @@
 			document.getElementById('loading-message').innerHTML += ' DONE<br />Loading data...';
 		</script>
 
-		<script src="http://play.pokemonshowdown.com/js/config.js"></script>
+		<script src="https://play.pokemonshowdown.com/js/config.js"></script>
 		<script src="js/battledata.js"></script>
 		<script src="data/pokedex-mini.js"></script>
 		<script src="data/typechart.js"></script>


Zarel:
Does this still work correctly if `testclient` is loaded over a `file://` URL?

Zarel:
Commit summary should be `Remove crossdomain...`

kubetz:
Good point with the `file://` protocol. It was not working ...

I modified the code to use `https://` if testclient's protocol is `https://` and otherwise `http://` will be used. 

`config.js` script tag is now writen into the document using javascript same way google analytics is doing it (see [Standard Setup here](https://developers.google.com/analytics/devguides/collection/gajs/gaTrackingOverview)) except triple equals operatr is used.

Zarel:
It may be more straightforward to load over HTTPS always.

Zarel:
Are there any disadvantages to doing so?

kubetz:
There should not be any problem with that according to what I read about the topic. It will be just a very little bit slower than accessing it over http whenever possible.

My goal was to minimize the effect of changes so that only use-case where it was not working would be really affected, but I can switch to https completely, test the changes and add them here.

kubetz:
Changed the code to use only `https:` protocol for `action.php` call and `config.js` script as suggested.

Tested the testclient over `file:`, `http:` and `https:` and it worked everywhere for me.

Zarel:
I'd rather dynamically decide between `http:` and `https:` (defaulting to `https:` for `file:`) for the `action.php` call. The `config.js` include in `testclient` can be `https`-only.

My main concern is ISPs that block HTTPS. `testclient` may default to HTTPS but the main client should prefer HTTP in case it's necessary.

kubetz:
Not sure I understand. The code in `client.js` I modified is executed only in the testclient mode. Main client is calling `action.php` on the same domain as loaded client page using the same protocol. Therefore I am not getting your:

> .. main client should prefer HTTP ...

as main client functionality is intact.

I can modify `client.js` as you suggest, I just want to be sure we are talking about the same thing :).

Zarel:
Oh, nevermind then; I didn't read the code closely enough. This looks good.

kotarou3:
ISPs block HTTPS?

AFAIK, ISPs only block HTTPS if the site is on a blacklist, in which case HTTP traffic will also be blocked.
Do you know of any cases where HTTPS is blocked via whitelist instead?

Zarel:
I know some ISPs try to MitM HTTPS poorly, leading Chrome to block HTTPS, leaving me no way of accessing PS except by HTTP. This isn't in theory, this is something that's actually happened to me. :|

PR #406, opened 2015-06-25
Change -endability message structure

ascriptmaster:
With Simple Beam and Worry Seed now using -endability as well to reveal
the old ability, it is more appropriate to use a new message. Parentheses
are traditionally used by the client to indicate messages that do not
actually appear as in-game text.


Marty-D:
Hmm, I was thinking more like "[POKEMON's OLDABILITY] -> [POKEMON's Insomnia]", but I guess the existing implementation of `-endability` kind of impeded that. Oh well.

ascriptmaster:
`-ability|POKEMON|ABILITY|[oldability] OLDABILITY`? We'd have to account for backwards compatibility with replays, of course.

PR #408, opened 2015-07-03
Don't use toId for room names while joining a room using popup

xfix:
This fixes the issue that the Join room popup couldn't be used to enter battle rooms.


Slayer95:
It should use ``toRoomid`` instead, which correctly handles battle rooms.

Zarel:
What @Slayer95 said.

xfix:
Fixed.

xfix:
@Zarel Sorry, I was quite busy lately. My fault.

PR #409, opened 2015-07-03
Fix <<roomid>> linking

Ecuacion:
Fix an issue with << roomid >> formatting. That kind of links force to restart the website on click if your web browser has "self" as default target for links, and they should work like internal links.

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -374,7 +374,7 @@ var Tools = {
 				options.hidestrikethrough ? '$1' : '<s>$1</s>');
 		// <<roomid>>
 		str = str.replace(/&lt;&lt;([a-z0-9-]+)&gt;&gt;/g,
-				options.hidelinks ? '&laquo;$1&raquo;' : '&laquo;<a href="/$1">$1</a>&raquo;');
+				options.hidelinks ? '&laquo;$1&raquo;' : '&laquo;<a href="/$1" target="_blank" onclick="return selectTab(\'$1\');">$1</a>&raquo;');
 		// linking of URIs
 		if (!options.hidelinks) {
 			str = str.replace(linkRegex, function(uri) {


Zarel:
What browser in this? We currently intercept link clicks so this shouldn't be an issue.

Ecuacion:
I'm using Chrome. When i click these links they don't work like other links (they force to restart the page).

Zarel:
What version of Chrome? Do you have any extensions that might be interfering with click events?

Ecuacion:
My version is `43.0.2357.130 m` and the only extension i have is "Avast Online Security"

Zarel:
I think Avast Online Security might be interfering with the click capture. Can you try again with that extension disabled?

Ecuacion:
I disabled the extension and restarted the browser and it still does not work properly.

Zarel:
Can we TeamViewer?

Zarel:
If so, find me in http://play.pokemonshowdown.com/development

kotarou3:
I can confirm the current implementation doesn't work as expected on Firefox 38

It can be fixed by just adding `class="ilink"` to the link, though, much like what `/whoare` does

Zarel:
https://github.com/Zarel/Pokemon-Showdown-Client/blob/64c22a9be0069cb3c70e396e3620061fcf1e8f0d/js/client.js#L8-21

The relevant code should in theory not deepend on `class="ilink"`...

Ecuacion:
After seeing the implementation now i understand why it did not work for me. I was testing it in other server (psim.us) and this seems that only checks "play.pokemonshowdown.com" links. I tested it in main server and it works.

Zarel:
This newest change should have fixed it, can you confirm?

5393b57455670

Ecuacion:
Yes, it works perfectly now.

PR #410, opened 2015-07-04
Add /regdate

Irraquated:
Add regdate command, as mentioned [here](https://github.com/Zarel/Pokemon-Showdown/pull/2024#issuecomment-118454069).

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -808,7 +808,25 @@
 					}
 				}), 'text');
 				return false;
-
+				
+			case 'regdate':
+				if (!target) target = app.user.get('userid');
+				target = toId(target);
+				var self = this;
+				$.get("http://pokemonshowdown.com/users/" + target + ".json",
+				{}, Tools.safeJSON(function(data) {
+					var buffer = "";
+					var regTime = Number(data.registertime);
+					if (regTime) {
+						var regDate = new Date(regTime * 1000);
+						buffer += "User: <strong>" + data.username + "</strong> registered since <i>" + regDate.toString() + "</i>";
+					} else {
+						buffer += "User: <strong>" + target + "</strong> is not registered";
+					}
+					self.add('|raw|'+buffer);
+				}), 'text');
+				return false;
+			
 			case 'buttonban':
 				var self = this;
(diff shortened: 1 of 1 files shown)


sirDonovan:
<code>Revamped regdate command</code> doesn't make sense for the commit title because the command doesn't exist yet. You can use something like <code>Add /regdate</code>.

Irraquated:
In regards to the closed pull request at the client repo, sorry @sirDonovan will change the name.

Zarel:
The pull request name doesn't matter; we could change it ourselves if we cared. The important one is the commit name.

Zarel:
Ironically, "Revamped regdate command" makes sense as a pull request name, since we've previously had a regdate pull request. It just doesn't make sense as a commit name, since we haven't previously had a regdate client commit in master.

js/client-chat.js
```diff
@@ -808,7 +808,25 @@
 					}
 				}), 'text');
 				return false;
-
+				
+			case 'regdate':
+				if (!target) target = app.user.get('userid');
+				target = toId(target);
+				var self = this;
+				$.get("http://pokemonshowdown.com/users/" + target + ".json",
```
Zarel: This will be a security error in HTTPS, but I don't know enough about XHR to know if this is considered cross-origin. It's also unnecessary database load, you're getting a ladder list for no reason. You probably want to do it more similarly to how `/rank` is currently implemented.

Irraquated:
Changed the commit message. @Zarel

js/client-chat.js
xfix: @Zarel It's a different protocol, therefore it's not same-origin.

Zarel: The two things I mentioned above will need to be changed before I can merge this.

Zarel: @xfix, I mean I don't know if `http://play.pokemonshowdown.com/` to `http://pokemonshowdown.com/` is cross-origin.

Another thought: `*.psim.us` to `pokemonshowdown.com` definitely is.

Zarel:
I really don't want this to be too commonly used, anyway, so since this command appears unmaintained, I'm just going to close it.

PR #411, opened 2015-07-05
Change /buttonkick to /buttonwarn

Irraquated:
(Also keeping /buttonkick as an alias) Makes the command much more understandable.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -828,9 +828,10 @@
 				return false;
 
 			case 'buttonkick':
+			case 'buttonwarn':
 				var self = this;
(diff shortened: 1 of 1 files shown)


QuiteQuiet:
Why is there a commit to remove tabs you yourself added, rather than not just add the tabs in the first place?

Rhythms88:
@quiteQuiet they probably were automatically added from the program he used, and he just forgot to remove them

also he's probably squashing the commits

Irraquated:
Yes I will be squashing the commits. @QuiteQuiet & @Rhythms88

Irraquated:
Ok messed that up

Irraquated:
Ok messed that up

PR #417, opened 2015-07-10
Don't map highlights list to toName function

ascriptmaster:
'|' is a valid regex character and highlight phrases can occasionally pass
18 characters. If they are mapped, it can become difficult to remove
specific entries on the list.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -673,7 +673,7 @@
 					var targets = target.match(/[^,]+(,\d*}[^,]*)?/g);
 					// trim the targets to be safe
 					for (var i=0, len=targets.length; i<len; i++) {
-						targets[i] = targets[i].trim();
+						targets[i] = targets[i].replace(/\n/g, '').trim();
 					}
 					switch (targets[0]) {
 					case 'add':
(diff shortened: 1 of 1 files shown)


ascriptmaster:
@Slayer95 since his commit was what caused this change

Slayer95:
Maybe we should use ``JSON.stringify`` here. The idea is that the message sent should not contain newlines.

ascriptmaster:
We don't want newlines, right? So it's probably more along the lines of `s/\n/ /g`

And it should probably be done in highlight validation as well

Slayer95:
Doing it in input validation should suffice. Chat messages received by the client are splitted in lines, anyway.

ascriptmaster:
Input validation?

Slayer95:
Yea. But, now JSON conversion is unneeded -and it would itself make edition of highlight list complicated, due to escaping.

Zarel:
We have Tools.escapeHTML, don't we?

Zarel:
Actually, what do we need to validate here?

ascriptmaster:
Just the newline messing up things, I think?

The only reason why I'm replacing the newline on add is because I don't even think our highlights checker can even read the newline. Although it's probably better to replace it with an empty string instead of a space.

Also I'm pretty sure `Tools.escapeHTML` is absolutely the wrong filter to use anyways

PR #418, opened 2015-07-11
Link to Showdown dex in /data

xfix:


style/utilichart.css
```diff
@@ -12,7 +12,7 @@
 	padding-top: 1px;
 }
 
-.utilichart li a
+.utilichart li > a, .utilitychart li > div
```
Slayer95: Hmm, we don't have a ``.utilitychart``

xfix: I fail... wait... how does it work then.

monsanto:
> Link to XY Smogon analysis in /data to Link to Showdown dex in /data

wow, thanks for the show of support everyone

Slayer95:
@monsanto, we are still planning to link to Smogdex from Teambuilder.

In other news, wasn't the plan to link moves and abilities to PSdex? :|

PR #421, opened 2015-07-14
Change codestyle to follow server codestyle

Freigeist:


Zarel:
I want to point out that past PS standardized on `!=` and `==` for numeric comparisons

Switching them to `!==` and `===` could introduce bugs

I can't guarantee that I haven't compared, say, `55 == '55'` in the past

Zarel:
The rest of this is good, although you'll need to rebase on my latest client changes.

PR #422, opened 2015-07-15
Remove article in tournament creation

MonoHearted:
Add "An" instead of "A" if the format name begins with a vowel.

### DIFF
--- a/js/client-chat-tournament.js
+++ b/js/client-chat-tournament.js
@@ -251,7 +251,7 @@
 				case 'create':
 					var format = Tools.getEffect(data[0]).name;
 					var type = data[1];
-					this.room.$chat.append("<div class=\"notice tournament-message-create\">A " + Tools.escapeHTML(format) + " " + Tools.escapeHTML(type) + " Tournament has been created.</div>");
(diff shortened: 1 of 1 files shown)


js/client-chat-tournament.js
```diff
@@ -251,7 +251,11 @@
 				case 'create':
 					var format = Tools.getEffect(data[0]).name;
 					var type = data[1];
-					this.room.$chat.append("<div class=\"notice tournament-message-create\">A " + Tools.escapeHTML(format) + " " + Tools.escapeHTML(type) + " Tournament has been created.</div>");
+					if ("aeiou".includes(format.charAt(0).toLowerCase())) {
```
Rhythms88: The thing I don't like about this is that `A UU Tournament...` sounds better than `An UU Tournament...` imo

it might just be my personal preference, but grammatically correct doesnt mean it always has to be used, which is what you're aiming for here.

MonoHearted:
Yes, but I'm not sure where to balance "grammatically correct" and "sounds good".

Zarel:
Just remove the article, I guess. `"UU tournament created."` Terseness is a virtue!

PR #423, opened 2015-07-15
Fix CSS issues related to spoiler display

xfix:
Now codes and links are properly hidden by spoilers. Codes didn't have proper styles to hide them in spoilers, while links had a wrong color (copied from .spoiler:hover, rather than .spoiler).


ascriptmaster:
:+1:

Zarel:
Done in 838d855c727.

PR #424, opened 2015-07-16
Add small animation when an Ability activates

ascriptmaster:
The idea is that `|-ability|POKEMON|ABILITY|[activate]` will trigger when abilities activate but don't need a further message; server-side will be updated once we figure out what's the best protocol.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -3980,12 +3980,16 @@ var Battle = (function () {
 					// do nothing
 				} else if (effect.id) switch (effect.id) {
 				case 'trace':
+					this.resultAnim(poke, "Traced " + ability.name, 'good', animDelay);
+					actions += "[" + poke.getName() "'s Trace!]<br />";
 					actions += '' + poke.getName() + ' traced ' + ofpoke.getLowerName() + '\'s ' + ability.name + '!';
 					break;
 				case 'roleplay':
+					this.resultAnim(poke, "Copied " + ability.name, 'good', animDelay);
 					actions += '' + poke.getName() + ' copied ' + ofpoke.getLowerName() + '\'s ' + ability.name + ' Ability!';
 					break;
 				case 'mummy':
+					this.resultAnim(poke, "Acquired Mummy", 'neutral', animDelay);
 					actions += "" + poke.getName() + "'s Ability became Mummy!";
 					break;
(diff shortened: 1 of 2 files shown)


js/battle.js
```diff
@@ -4004,43 +4008,50 @@ var Battle = (function () {
...
+						break;
+					case 'darkaura':
+						actions += "" + poke.getName() + " is radiating a dark aura!";
+						break;
+					case 'airlock':
+					case 'cloudnine':
+						actions += "The effects of the weather disappeared.";
+						break;
+					default:
+						if (!kwargs.activate) {
```
Zarel: Do we really need the `[activate]` part? How common is `"POKEMON has ABILITY!"`? Maybe that's the one that should be special-cased.

ascriptmaster: AFAIK, `POKEMON has ABILITY` is used in exactly one case: Gen 4 Mold Breaker (which means nothing in XY). @Marty-D should verify, though.

Zarel: In other words, we don't need the `[activate]`, no message should just be the default behavior.

Marty-D:
@ascriptmaster is correct, I searched the HGSS, B2W2, and ORAS battle text and Gen 4 Mold Breaker is the only one to use that message.

My only concern is, Abilities now exist that activate on a Pokemon that may or may not be the target of the move (Aroma Veil, Flower Veil, Sweet Veil). I was going to add their activation messages a while ago but I thought it was better to decide how all this Ability activation stuff would work beforehand. Now here we are and I don't remember exactly what it is I wanted to be done first.

ascriptmaster:
What's wrong with the owner of Aroma Veil not being the target of the Ability? We just do something like this

    this.add('-ability', this.effectData.target, 'Aroma Veil')

And the proper `POKEMON's Aroma Veil!` will show up.

Marty-D:
Oh, yeah I see what you mean. I was still thinking about it in terms of whatever implementation I was more or less working on. (This is much better!)

ascriptmaster:
All right then, here is the new version.

I'm not sure how Drought/Drizzle/Sand Stream/Snow Warning would interact with Desolate Land/Primordial Sea/Delta Stream protocol-wise, so for now those three messages stay the same.

PR #425, opened 2015-07-17
Support 'choice' battle messages

Slayer95:
Intended for parsing player battle choices and possibly doing something with that information in the future -doing nothing for now.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -5563,6 +5563,8 @@ var Battle = (function () {
 			name = args.join(' ');
 			this.log('<div class="debug"><div class="chat"><small style="color:#999">[DEBUG] ' + Tools.escapeHTML(name) + '.</small></div></div>', preempt);
 			break;
+		case 'choice':
+			break;
 		case 'unlink':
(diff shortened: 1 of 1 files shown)


Zarel:
Can you clarify?

Slayer95:
With the ref-ed server-side PR, when battle decisions are commited, players' decisions will be stored in the replay split log, and sent back to them as confirmation.

- Storage in replay, as I mentioned in the server PR, will be useful to deal with bug reports. (But I clicked move X! No, you cancelled. Yes, and were encored, etc.)
- Confirmation, as it was discussed in Smogon PM, is potentially useful to keep a better track of own Pokémon, specially in the context of Illusion, lack of Nickname Clause, etc.

ascriptmaster:
Wait, is the server continuously sending `choice` during a battle? Wouldn't that be a leak?

Or does it send in replays only

Slayer95:
It uses the ``split`` feature, @ascriptmaster. So no leaks.

PR #426, opened 2015-07-18
Do not show Ability Activation in the battle frame log

ascriptmaster:


TheImmortal:
Also, don't use `poke.getName()` for the activation. The game just says "Pokemon's Ability!" even for opposing Pokemon.

TheImmortal:
Oh and remove the Intimidate message. The game doesn't have it. It was just here because we didn't have activations.

Marty-D:
@TheImmortal, re: "Pokemon's Ability!" for opposing Pokemon: https://github.com/Zarel/Pokemon-Showdown-Client/commit/69bd9148065cc78b548727649595d0a758e78d0f#commitcomment-12234567

TheImmortal:
@Marty-D but it would say opposing in the Attack fell message

Marty-D:
That's true for Intimidate, but there are Abilities that activate which don't show a message at all (like Dry Skin), or that don't mention the user, like Forewarn or Trace which use "It was alerted" and "It traced", respectively.

Zarel:
We might do something like align left for your pokemon, and align right for opposing pokemon.

Zarel:
What's wrong with showing ability activation in the frame log?

Marty-D:
>We might do something like align left for your pokemon, and align right for opposing pokemon.

This sounds like a fine solution.

TheImmortal:
Yeah that's a good idea.

What's wrong is that there's no point. The frame has the activation animation. It also doesn't really look nice in there. [this] is meant for battle log.

Zarel:
Oh, I see.

ascriptmaster:
Do I just do like, an extra `<div style="align:right">` for that?

Zarel:
Well, that might be a bit ugly in the log.

Maybe

allySide: `[POKEMON's ABILITY]`

foeSide: `PLAYER's [POKEMON's ABILITY]`

ascriptmaster:
Would we need to do something if Ignore Opponent is set, so that their username won't be shown?

TheImmortal:
Is their name hidden in the rest of the battle log? Like when they send out Pokemon.

Also, Zarel forgot the `!`.

ascriptmaster:
A thing.

I made the executive decision to put the opponent's name at the end instead of at the front. I think it looks nicer that way.

Also, opponent's name never disappears, even if you Ignore Opponent.

Zarel:
Man, that still looks weird.

`[POKEMON's ABILITY]`, `[Foe POKEMON's ABILITY]`?

ascriptmaster:
I went with `[Opposing POKEMON's ABILITY]` (basically the original message w/o the "The")

js/battle.js
```diff
@@ -516,18 +516,18 @@ var Pokemon = (function () {
 		this.clearTurnstatuses();
 		this.clearMovestatuses();
 	};
-	Pokemon.prototype.getName = function () {
+	Pokemon.prototype.getName = function (nameOnly) {
```
Zarel: You might as well implement "Opposing" here, since it's the only use thereof.

ascriptmaster: What should I call the parameter name, then?

Zarel: `shortName`?

Zarel: Or `useShortName`?

ascriptmaster:
Should be fine now. Wi-Fi situation is still a problem for me so I haven't had the opportunity to client test, though.

PR #427, opened 2015-07-20
Support `seed` battle messages

Slayer95:
For battle starting seeds


Zarel:
We should probably comment exactly why these simply `break`.

Slayer95:
Ok, I'll add a comment at some point.

PR #429, opened 2015-07-24
Support hidden [non-searchable non-challengeable] formats

Slayer95:
In the formats list, each entry will now support a trailing mask -sent as hexadecimal- whose binary form encodes the data regarding its availability and team requirements

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -113,7 +113,7 @@
 		});
 	}
 
-	Config.version = '0.10.0';
+	Config.version = '0.10.1';
 	Config.origindomain = 'play.pokemonshowdown.com';
 
 	// `defaultserver` specifies the server to use when the domain name in the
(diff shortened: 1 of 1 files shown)


js/client.js
```diff
@@ -1114,20 +1114,31 @@
...
+					var lastCommaIndex = name.lastIndexOf(',');
+					var code = lastCommaIndex >= 0 ? parseInt(name.substr(lastCommaIndex + 1), 16) : NaN;
+					if (!isNaN(code)) {
+						name = name.substr(0, lastCommaIndex);
+						if (code & 1) team = 'preset';
+						if (!(code & 2)) searchShow = false;
+						if (!(code & 4)) challengeShow = false;
+						if (!(code & 8)) tournamentShow = false;
+					} else {
+						/* Backwards compatibility: late 0.9.0 -> early 0.10.0 */
```
Zarel: > early 0.10.0

Just increment the version number to 0.10.1 if you're going to change the protocol.

Zarel: Also: prefer `//` for comments.

Slayer95: That's the spirit!

Zarel:
Remember to increment server version.

PR #431, opened 2015-07-26
Fix issue with HP colours

Irraquated:


Zarel:
?

This doesn't seem right...

Irraquated:
How so @Zarel ?

Zarel:
It seems that until now, the guarantee was that a Pokémon in red HP would have both the `hp-yellow` and `hp-red` classes.

This changes it so that it only has one of them, which may have unpredictable consequences.

Zarel:
It also doesn't do it very cleanly, so it has the same problems the previous implementation had.

Irraquated:
Fair call, @Zarel what if instad of changing what is currently changed in the pull to instead have: On line 2933 add `$hp.addClass('hp-yellow');` ?

Zarel:
It depends. Most likely it's getting out of sync at some other place.

Zarel:
We could replace all the `$hp.addClass('hp-red');` with `$hp.addClass('hp-yellow hp-red');`

Irraquated:
That was the other way I was thinking of going bout it @Zarel, should I also do this with `else $prevhp.addClass('prevhp-red');` and replace with `else $prevhp.addClass('prevhp-yellow prevhp-red');` ?

Zarel:
Yes, probably.

Irraquated:
There's one more thing I have to do before this can be committed, will have to do it later as i'm on mobile.

Slayer95:
Superceeded by https://github.com/Zarel/Pokemon-Showdown-Client/pull/444

Zarel:
Apparently the correct spelling is 'supersede'. 'Supercede' is an alternate spelling, but it's less common (at least in US English).

PR #436, opened 2015-08-02
Allow linking to old gen analyses in teambuilder.

xfix:


Zarel:
idk how this went unnoticed for so long...

PR #438, opened 2015-08-06
Unify chat message parsing

Slayer95:


Zarel:
Can you clarify what "Don't hide suppressed spoilers in notifications" means?

Slayer95:
Currently, if a user has spoiler formatting disabled, and is sent a highlight containing spoilers, the text of the notification will be ``(spoiler)``, rather than the content of the message. Said commit fixes this issue.

Zarel:
I see.

PR #439, opened 2015-08-10
Ability activations not linked to -ability reveal Ability to opponent

ascriptmaster:


Zarel:
You should probably replace the existing ability/baseAbility overwrites with markAbility as well

ascriptmaster:
Ok, I did what I could with that.

Testing things manually is such a pain... I've been spoiled by Mocha's test suite

js/battle.js
```diff
@@ -4010,9 +4019,10 @@ var Battle = (function () {
 				var ability = Tools.getAbility(args[2]);
 				var effect = Tools.getEffect(kwargs.from);
 				var ofpoke = this.getPokemon(kwargs.of);
-				poke.ability = ability.name;
 				if (!effect.id || kwargs.fail) {
```
Zarel: Perhaps this part should also be implemented in `markAbility`? Perhaps as a second argument.

js/battle.js
```diff
@@ -4129,16 +4139,24 @@ var Battle = (function () {
 					actions += "" + poke.getName() + " transformed!";
 					if (toId(template.species) === 'shaymin') break;
 				} else if (toId(template.species) === 'darmanitanzen') {
```
Zarel: Perhaps `[from] Zen Mode` would be appropriate here.

ascriptmaster: That's a server-side change, yes? I'll work on that in a bit, then.

Speaking of Zen Mode, @Marty-D, what happens if a Zen-Mode Darmanitan Skill Swaps another Zen-Mode Darmanitan in-game? (assuming both are under half health)

js/battle.js
ascriptmaster: ``markAbility(ability, noBase)``?

js/battle.js
Marty-D: @ascriptmaster, I finally managed to test it. Zen Mode ended after the Skill Swap, so yeah, it's like Flash Fire / Unburden.

ascriptmaster:
Most things are good. The only thing that remains is the part about having Ability activation appear below the other `resultAnim` messages

Issue #314, opened 2015-01-17
Illusion, search IDs, and UI assistance


Slayer95:
Illusion currently has a bug in the fact that it displays the level of the fake mon in the simulator, while in-game it displays the real level of the Pokémon with Illusion. This is -as an immediate cause- due to Pokémon getting identified client-side by their <code>searchid</code>, which is built from server-side <code>details</code>. As far as I can see, the overall server-side approach to Illusion in <code>BattlePokemon.getDetails</code> is sound (though I can see some room to improvement by taking advantage of the battle log <code>|split|</code> feature).

Currently, <code>searchid</code> includes.
- Species
- Nickname
- Gender
- Shinyness
- Level

Additionally to its not copying level, Illusion should -when it becomes possible- display the copied Poké Ball (related: #306). Similarly, <code>searchid</code> should include Poké Ball information, and it would be straightforward to remove the level data from <code>searchid</code>.

However, it stands that a different displayed level can be used in a metagame without Species Clause to discover that the opponent has 2 or more of a species (if there is no Team Preview), or associate items/moves/abilities to one or more of them. Similarly, it's theoretically possible to identify Pokémon based on whether they have a certain item or activate an ability on switch-in (or even afterwards, but its implementation is unrealistic imo). Automating the process of identifying different Pokémon on the basis of items/abilities would be potentially a very significant assistance in formats such as Hackmons.

Which concept would we like to implement?
Restrict <code>details</code> to the information conveyed by Illusion? If we don't do that, we would need to replace the level when Illusion is active. In that case, would we like to keep the amount of information in <code>details</code> the same (plus Poké Ball)? Or maybe expand it to obvious items and abilities? Something else?

Zarel:
Illusion no longer displays the fake level. I don't know what the rest of this is about, but a number of other Illusion issues have been fixed since then, and we don't get complaints about Illusion UI anymore.

PR #441, opened 2015-08-12
Support parsing versions and config from servers

Slayer95:
Version support client-side and login-server-side.
Config support stubbed only client-side. Possible usage for it includes custom ranks / rank sorting (# vs ~ comes to mind)

(Be harsh in PHP review)

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -5447,6 +5447,9 @@ var Battle = (function () {
 	};
 	Battle.prototype.runMajor = function (args, kwargs, preempt) {
 		switch (args[0]) {
+		case 'version':
+			this.version = args[1];
+			break;
 		case 'start':
 			this.teamPreview(false);
 			this.mySide.active[0] = null;
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -931,6 +931,8 @@ var Tools = {
 		return type;
 	},
 
+	loadConfig: function() {},
+
 	loadedSpriteData: {'xy':1, 'bw':0},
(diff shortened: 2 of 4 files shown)


Zarel:
Is there a need for this so far?

Slayer95:
Not really, but can we get this covered before thinking of removing the "Beta" label?

Zarel:
I don't think we'll ever have a breaking change that requires something like this.

If we did, I'd honestly prefer to post-process old replays than support multiple protocol versions.

Slayer95:
I think that's overly optimistic in the light of all these issues with pending protocol updates:

https://github.com/Zarel/Pokemon-Showdown-Client/issues/314
https://github.com/Zarel/Pokemon-Showdown/pull/1269
https://github.com/Zarel/Pokemon-Showdown/issues/1452

Replay postprocessing sounds like a good idea, but it would still need to accurately determine the version of the protocol, and checking dates is not enough.

Zarel:
Replay postprocessing would presumably mean that that there's only one version in the database at any given time.

Zarel:
Also, none of those protocol updates seem backwards-incompatible.

Zarel:
And I still don't like them.

Zarel:
I'm closing this. Of the issues you've named so far, I don't think any need to be backwards incompatible.

Zarel:
If we ever need to make a backwards incompatible change, we'll probably need to fit the solution to the problem; this generic solution probably won't work for it anyway.

PR #442, opened 2015-08-13
Fix slow -boost animations

sobolews:
A delay was added to -boost in c99de62 to prevent overlap of boost and ability
animations. This caused normal -boost animations to be slow, and was
inconsistent with -unboost not being delayed. Changed to delay the ability
instead.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -3363,11 +3363,11 @@ var Battle = (function () {
 				if (this.gen === 1 && stat === 'spd') break;
 				if (this.gen === 1 && stat === 'spa') stat = 'spc';
 				var amount = parseInt(args[3]);
+				var fromAbility = false;
 				if (!poke.boosts[stat]) {
 					poke.boosts[stat] = 0;
 				}
 				poke.boosts[stat] += amount;
-				this.resultAnim(poke, poke.getBoost(stat), 'good', 2);
 
 				var amountString = '';
 				if (amount === 2) amountString = ' sharply';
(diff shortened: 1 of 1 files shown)


ascriptmaster:
Ability activation needs to be first, though.

What you'll need to do is check if the `[from]` effect is an ability, and if it is, arrange the animations accordingly.

sobolews:
Thanks, changed to only delay the boost when it is [from] an ability.

Zarel:
Or we could just change how ability activations work

Zarel:
For instance:

We could put ability activations lower and make them stay still instead of moving upwards.

What do people think of that idea? I think that'd be better than ugly hacks with delaying boosts.

TheImmortal:
I like that. Still makes it's easier to see too, imo. @ascriptmaster can you do that?

ascriptmaster:
I'll see what I can do since I already have a PR related to ability activation, but don't close this just yet until I have a fix to make this redundant.

js/battle.js
```diff
@@ -3380,6 +3380,7 @@ var Battle = (function () {
 					switch (effect.id) {
 					default:
 						if (effect.effectType === 'Ability') {
+							fromAbility = true;
 							this.resultAnim(poke, effect.name, 'ability', animDelay);
```
urkerab: An alternative approach is to reset animDelay to 2 here so that it's always correct later on when you need it.

sobolews:
Closing; this appears to have been [fixed](http://git.io/vW8cl)

PR #451, opened 2015-08-21
Make teambuilder more resistant to zoom changes

urkerab:
Changing the zoom level can disrupt the teambuilder display as shown in this screenshot that I [found on the Smogon forums](http://www.smogon.com/forums/threads/bug-reports-v2-0-read-op-before-posting.3469932/page-217#post-6296053):
Screenshot
Notice how the Item is slightly larger than the Nickname, Pokemon (sic) or Ability fields. Removing this item-specific style allows the teambuilder to display naturally at other zoom levels.

I didn't remove the style block because there was a similar placeholder style block a few lines above for `.setchart-nickname` but if this is an error then I can simply remove both blocks.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -2306,7 +2306,10 @@ a.ilink:hover {
 }
 .setchart .setcell-item input {
 	width: 114px;
-	margin-right: 1px;
+	margin-right: 0px;
+}
+.setchart .setcell-ability input {
+	margin-left: 0px;
 }
 .setchart .setcol-moves input {
 	width: 129px;
--- a/style/client.css
+++ b/style/client.css
@@ -2096,6 +2096,9 @@ a.ilink:hover {
 .setcell-details {
 	float: none;
 }
+.setcell-ability {
+	float: right;
+}
 .setcol-icon .itemicon {
 	display: block;
 	margin-left: 88px;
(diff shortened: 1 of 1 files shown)


Zarel:
This presumably misaligns the normal-zoom version (a screenshot could help here). It would make more sense to do it in a way that doesn't have this problem.

urkerab:
I see what you mean, so I've now taken a different approach; by floating the Ability input to the right (which has no effect at normal zoom) I can afford to reduce the margin on the Item input, thus allowing them to fit at various zoom levels.

Zarel:
Would it be easier to reduce the ability box's right margin?

urkerab:
That would work, although you do then lose the alignment of the Ability box with the Details box when you zoom.

Zarel:
Good point, this way is fine.

style/client.css
```diff
@@ -2303,6 +2306,7 @@ a.ilink:hover {
 }
 .setchart .setcell-item input {
 	width: 114px;
+	margin-right: 1px;
```
Zarel: Why not just make this `0px`?

urkerab: It was 1px because I was experimenting with changing margins in various places to see what worked and found that 1px lets me zoom to 67%; to zoom to 50% I'd need to set both this and the left margin on the ability to zero.

Zarel: That may be preferable. Thoughts?

Zarel:
For future reference, please comment on the pull request after doing the changes I request. I don't get notified of pushes to a pullreq.

PR #452, opened 2015-08-24
Autojoin battles in event of disconnection

xfix:


xfix:
Hm, I just realized is not going to work well with server restarts. Might as well wait until I implement persistent battles.

js/client.js
```diff
@@ -368,6 +368,18 @@
 						}
 					}
 					this.send('/autojoin ' + autojoinIds.join(','));
+
+					var battleAutojoin = (Tools.prefs('battleautojoin') || '').split(',');
```
Zarel: We don't need a separate `battleAutojoin`; can't we just add it to the normal autojoin list?

Zarel: This also might interact weirdly with opening multiple tabs; you probably don't want battles to get focused instantly.

Zarel: Might be more straightforward for the server to send you a list of battles you're currently in, and just make the "rejoin battle" button very prominent.

PR #453, opened 2015-08-26
Update the window title as the fragment changes

urkerab:
Currently the browser's history (both from the back button but also global history) just shows "Showdown!" for any page under http://play.pokemonshowdown.com/. On the other hand, all replays get nice titles e.g. "Challenge Cup 1-vs-1 replay: conbeef vs. urkerab - Pokémon Showdown". I've gone for a slightly simpler "conbeef vs. urkerab - Showdown!" to show up during an actual battle.

The title is refreshed when the fragment changes as this is what the browser stores in its history, so this not only includes the teambuilder and ladder rooms but also chat rooms if your window is small enough, or if you joined by loading a direct link.

The case of switching to a room whose title has not been read from the server yet is also handled.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -98,6 +98,7 @@
 			if (log.length && log[0].substr(0, 7) === '|title|') {
 				this.title = log[0].substr(7);
 				log.shift();
+				app.roomTitleChanged(this);
 			}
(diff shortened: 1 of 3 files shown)


Zarel:
Very nice.

PR #454, opened 2015-08-28
Quality of Life addition for users of Markdown-based forums

StAlRuth:
This commit makes life easier for some growing Reddit communities based around competitive Pokémon by making the teams easier to paste there in a format that allows them to be pasted back into the teambuilder after being parsed by a Markdown parser, while having little to no effect on everyone else.

### DIFF
--- a/js/storage.js
+++ b/js/storage.js
@@ -661,18 +661,18 @@ Storage.exportTeam = function (team) {
 		if (curSet.item) {
 			text += ' @ ' + curSet.item;
 		}
-		text += "\n";
+		text += "  \n";
 		if (curSet.ability) {
-			text += 'Ability: ' + curSet.ability + "\n";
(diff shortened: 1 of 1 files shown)


Zarel:
Hm, what does this do to our PO compatibility?

Zarel:
Another option might be to prepend 4 spaces.

StAlRuth:
Just tested, works as expected in PO.

Prepending 4 spaces would require changing the teambuilder-client.js TeambuilderRoom.importSet function to add the extra spaces at the beginning of individual pokes, whereas this is uneccesary otherwise as the last line would be followed by a double line break so that people can see what they're writing.

Zarel:
Fair enough. I've thought for a bit and can't think of any disadvantages, so I guess this goes through.

PR #456, opened 2015-08-31
Display battle formats in groups

urkerab:
When teambuilding or searching for a battle you are currently presented with a plain list, unlike when challenging or laddering. This improves the display of those lists to include all the same sections.


Zarel:
Can you clarify what you mean by "searching for a battle" as opposed to "laddering"?

Zarel:
Oh, you mean looking at the battle list.

PR #457, opened 2015-09-01
Battle backgrounds gen 3 + 5

Layell:
Adding in the gen 3 bg, and fixing it so that gen 5 can use the rotating backgrounds that it could before.

Preview: http://i.imgur.com/qQB0mj5.png

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2222,7 +2222,8 @@ var Battle = (function () {
 	};
 	Battle.prototype.updateGen = function () {
 		if (this.gen < 3) this.backdropImage = 'bg-gen1.png';
-		else if (this.gen < 6) this.backdropImage = 'bg.jpg';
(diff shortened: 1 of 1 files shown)


js/battle.js
```diff
@@ -2222,7 +2222,8 @@ var Battle = (function () {
 	};
 	Battle.prototype.updateGen = function () {
 		if (this.gen < 3) this.backdropImage = 'bg-gen1.png';
-		else if (this.gen < 6) this.backdropImage = 'bg.jpg';
+		if (this.gen == 3) this.backdropImage = 'bg-gen3.png';
```
Zarel: This needs to be `else if`, or gen 1 will get the gen 4 background.

Layell:
@Zarel done.

Zarel:
In your next pullreq (I hear you're doing gen 4 soon), can you change the `<` to `<=`? It'll be more readable here, I think.

PR #461, opened 2015-09-03
Background loader update

Layell:
Adding in gen 4 background, in addition if the user has modern sprites for past gens checked old gen backgrounds will not load.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2226,8 +2226,7 @@ var Battle = (function () {
 			else if (this.gen <= 3) this.backdropImage = 'bg-gen3.png';
 			else if (this.gen <= 4) this.backdropImage = 'bg-gen4.png';
 			};
-		if (this.bgElem) this.bgElem.css('background-image', 'url(' + Tools.resourcePrefix + 'fx/' + this.backdropImage + ')');
-
+		if (this.bgElem) this.bgElem.css('background-image', 'url(' + Tools.resourcePrefix + 'fx/' + this.backdropImage + ')')
 	};
 	Battle.prototype.reset = function (dontResetSound) {
 		// battle state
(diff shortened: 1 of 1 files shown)


js/battle.js
```diff
@@ -2221,10 +2221,12 @@ var Battle = (function () {
...
 	Battle.prototype.updateGen = function () {
-		if (this.gen < 3) this.backdropImage = 'bg-gen1.png';
-		else if (this.gen == 3) this.backdropImage = 'bg-gen3.png';
-		else if (this.gen < 5) this.backdropImage = 'bg.jpg';
-		if (this.bgElem) this.bgElem.css('background-image', 'url(' + Tools.resourcePrefix + 'fx/' + this.backdropImage + ')');
+		if (!Tools.prefs('nopastgens')){
+			if (this.gen <= 2) this.backdropImage = 'bg-gen1.png';
+			else if (this.gen <= 3) this.backdropImage = 'bg-gen3.png';
+			else if (this.gen <= 4) this.backdropImage = 'bg-gen4.png';
+			};
```
Zarel: This should be `}` and it should be unindented one level.

js/battle.js
```diff
@@ -2221,10 +2221,12 @@ var Battle = (function () {
 		this.gen = 6;
 	};
 	Battle.prototype.updateGen = function () {
-		if (this.gen < 3) this.backdropImage = 'bg-gen1.png';
-		else if (this.gen == 3) this.backdropImage = 'bg-gen3.png';
-		else if (this.gen < 5) this.backdropImage = 'bg.jpg';
-		if (this.bgElem) this.bgElem.css('background-image', 'url(' + Tools.resourcePrefix + 'fx/' + this.backdropImage + ')');
+		if (!Tools.prefs('nopastgens')){
```
Zarel: This should be `if (!Tools.prefs('nopastgens')) {`

PR #469, opened 2015-09-13
Preview your username colour as you type

urkerab:
This is loosely based on the first suggestion from http://www.smogon.com/forums/posts/6424903 which asked for a link to the colour preview. I thought it would be neater simply to preview the username as you choose it.


Zarel:
If we do this, we might as well bold the username?

Zarel:
I don't think we colorize usernames anywhere where we don't also bold the username, for readability reasons.

ascriptmaster:
Where does the color change? In the input field as you type? If that's the case I'd rather have a box that changes color as you type. Otherwise, it'll depend where it's changing.

Also, I think users without room staff aren't bold in the user list.

urkerab:
Well, I thought the input field was the most obvious place to change the colour as it's already got the username in it.

Zarel:
Well, I still have worries that it would be distracting. Maybe the border color of the box would be less distracting?

ascriptmaster:
I'd absolutely loathe it if the color of the text changed as I was typing it. Border or box is preferred, or if you *really wanted* I guess you could have an extra line that displayed the color w/ the name like the color preview does, but I'd prefer you didn't because that's a lot of extra space.

Zarel:
You could just have a bullet somewhere, and colorize that, like the bullet in `/me`.

urkerab:
Colourised bullet it is.

Zarel:
Screenshot?

urkerab:
username

urkerab:
Oops, bug crept into 59c06fd.

Zarel:
Maybe right-align the bullet.

Maybe even have it right-aligned inside the textbox.

...my next fear is that people will see a red bullet and think it means "invalid" instead of "your username color will be red".

Maybe instead of a bullet, just the text "COLOR" in bold small font.

urkerab:
How about this version?
username

Zarel:
Looks really weird, especially having the name twice. I think just having it right-aligned and saying "COLOR" or "(color)" might be better.

urkerab:
username

style/client.css
```diff
@@ -2668,6 +2668,10 @@ a.ilink:hover {
 	color: #555555;
 }
 
+.bull {
+	float: right;
```
Zarel: This need a decreased font size, as well. Looks a bit ugly at its current size.

urkerab:
username
Also renamed the CSS class as I'm no longer using a `&bull;`.

DanUgore:
Damn I'm late but I think having the name twice was a good idea :(

PR #470, opened 2015-09-13
Make the username in PM headers clickable

urkerab:
One of the suggestions in http://www.smogon.com/forums/posts/6197110 is to make the username in PM headers clickable.

The code to create the clickableName was shamelessly copied and pasted from `client-chat.js`.


Zarel:
I'm unfortunately planning on making the PM header an expand/collapse button, so this would conflict with that.

A bar under the header with profile/challenge/other options might be useful, though.

Slayer95:
@urkerab, I see you recently updated this (i.e. after the last comment).

What did you change? If it's still relevant, could you rebase and post screenshots?

Zarel:
I usually leave PRs open when I disapprove of the way it's implemented but have a related change planned. Maybe I should just close the PR and open an issue when that happens, to be clearer...

urkerab:
Sorry, I don't remember what I changed; it might have been something as simple as a typo or a merge conflict.

Zarel:
PM headers now minimize/maximize the PM when clicked, superseding this pullreq.

PR #473, opened 2015-09-14
Instant replays should not withdraw

urkerab:
When instant replaying a doubles or triples battle any Pokémon that were still on the field (other than the first, which is currently cleared out by Battle.start()) are shown as coming back or being withdrawn. The game type setting block appears to be a convenient place to clear out the active Pokémon since we know how many there should be anyway, in much the same way as BattleSide works on the server.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -5482,13 +5482,20 @@ var Battle = (function () {
 			break;
 		case 'gametype':
 			this.gameType = args[1];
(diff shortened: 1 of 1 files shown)


Zarel:
Old singles battles don't have `gametype`, which needs to be accounted for.

urkerab:
In that case I'll revert the change to Battle.start() which will take care of that case.

PR #475, opened 2015-09-16
Add a validate button to the teambuilder

sobolews:
The validate button sends the current team and a validation request to the
server. The server will respond with a positive acknowledgement for valid teams,
or else a list of problems (the same list as when seeking a battle).

NOTE: This PR is dependent on https://github.com/Zarel/Pokemon-Showdown/pull/2150 being pulled in first.
Screenshots: http://imgur.com/a/dVj3F

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -509,7 +509,7 @@
 						}
 					}
 					if (curSection) buf += '</optgroup>';
-					buf += '</select></li>';
+					buf += '</select><button name="validate"><i class="icon-check"></i> Validate</button></li>';
 				}
 				if (!this.curSetList.length) {
 					buf += '<li><em>you have no pokemon lol</em></li>';
@@ -656,6 +656,14 @@
 		save: function () {
 			this.saveFlag = true;
 			Storage.saveTeams();
+		},
+		validate: function () {
+			if (!this.curTeam.format) {
+				app.addPopupMessage('You need to pick a format to validate your team.');
+				return;
+			}
+			app.sendTeam(this.curTeam);
+			app.send('/vtm ' + this.curTeam.format);
 		},
 		teamNameChange: function (e) {
 			this.curTeam.name = ($.trim(e.currentTarget.value) || 'Untitled ' + (this.curTeamLoc + 1));
--- a/style/client.css
+++ b/style/client.css
@@ -2018,7 +2018,7 @@ a.ilink:hover {
 }
 .teamchart li.format-select select,
 .teamchart li.format-select option {
-	font: 12pt Verdana, sans-serif;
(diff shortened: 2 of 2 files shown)


Zarel:
I'm not sure the topbar needs more things squeezed into it.

How about at the bottom, somewhere near "Add Pokemon"?

sobolews:
The bottom is kind of hard to see unless you know to go looking for it, at least once you have a full team. Would beside the Format dropdown be okay? Since its function is pretty closely related to the choice of format?
If not, I'm fine with at the bottom.

Slayer95:
We may not want it to be very visible as people would go and click it everytime they build a new team, effectively duplicating -or at least increasing noticeably- the workload of the validator processes.

Still, fwiw, next to the format would be a great place aesthetically and functionally imo.

sobolews:
If server load is an issue for validating teams, would it be worth it to let the client run a copy of the Validator for this? (of course, validating for battles would still have to be server side)

Slayer95:
Probably not, as formats and their rules could change anytime.

Zarel:
It's an interesting idea, but I think we'd need to work on on-demand resource/code loading before doing something like that. It might take a decent amount of refactoring, too.

Maybe we could use something like RequreJS.

Slayer95:
Note that such feature would need to be restricted to main server formats, to prevent XSS.
It could fallback to server-side validation.

Zarel:
Nah, we can safely take JSON files from servers.

Slayer95:
We would still need a server-side fallback for formats with custom team validation functions.

Zarel:
Good point.

sobolews:
Okay, this is ready to go in now that the server-side code is pushed.
I just need a final decision on the button location... here are [some options](http://imgur.com/a/Jse9Y). I personally prefer any of the first 3 since validation and format are closely related, but if you'd prefer it under the team then I'm happy with that too.

kotarou3:
@Slayer95 
If the stuff is sandboxed in a caja instance, XSS shouldn't be possible

Zarel:
"Validate" next to the format dropdown isn't horrible.

Zarel:
I don't want to deal with sandboxing and Caja, the server-side solution works fine

sobolews:
Moved the validate button next to the formats list.

Zarel:
Someone should merge this after the next server restart, and poke me to update client when that happens.

TheImmortal:
@Zarel

urkerab:
If client-side prevalidation ever happens this would also help the teambuilder out on other ways, for instance Doubles UU doesn't currently offer you UU by default, and Pikachu Cup defaulted to level 100 which was a pain when trying to allocate your EV spread.

PR #476, opened 2015-09-16
Make tooltips work in instant replays

urkerab:
During an instant replay the controls are not updated and therefore the tooltips don't work, unless you are lucky and the battle was forfeited with all the Pokémon still playing.

By setting up the tooltips prior to the start we can ensure that they will always work. I had to change the way the Pokémon is selected as `Battle.getPokemon` works on the logical rather than the visual sides and also if you give it a fainted Pokémon it will happily substitute another Pokémon for you. The new code will automatically ignore fainted Pokémon or Pokémon that are out of range (e.g. position 2 in singles).

The `thing` for the `'pokemon'` tooltip is now a concatenation of either `'my'` or `'your'` depending on the visual side and the integer offset in the active array, thus in doubles the possible values are `my0`, `my1`, `your0` and `your1`.

The tooltips are created with a loop to make it easier to see what's going on, rather than using a large unreadable block of HTML.

There is one other place that field tooltips are used and that is when selecting the target of a move in doubles or triples, so I have adapted that case accordingly.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -15,7 +15,26 @@
 			this.$chatFrame = this.$el.find('.battle-log');
 			this.$chatAdd = this.$el.find('.battle-log-add');
 			this.$join = null;
+
+			// tooltips
+			var buf = '';
+			var tooltips = {
+				your2: { top: 70, left: 250, width: 80, height: 100 },
+				your1: { top: 85, left: 320, width: 90, height: 100 },
+				your0: { top: 90, left: 390, width: 100, height: 100 },
+				my2: { top: 210, left: 410, width: 140, height: 160 },
+				my1: { top: 210, left: 270, width: 160, height: 160 },
+				my0: { top: 210, left: 130, width: 180, height: 160 }
+			};
+			for (var active in tooltips) {
+				buf += '<div style="position:absolute;';
+				for (var css in tooltips[active]) {
+					buf += css + ':' + tooltips[active][css] + 'px;';
+				}
+				buf += '"' + this.tooltipAttrs(active, 'pokemon', true, true) + '></div>';
+			}
 			this.$foeHint = this.$el.find('.foehint');
+			this.$foeHint.html(buf);
 
 			BattleSound.setMute(Tools.prefs('mute'));
 			this.battle = new Battle(this.$battle, this.$chatFrame);
@@ -194,30 +213,6 @@
 
 			}
 
-			// tooltips
-			var myActive = this.battle.mySide.active;
-			var yourActive = this.battle.yourSide.active;
-			var buf = '';
-			if (yourActive[2]) {
-				buf += '<div style="position:absolute;top:70px;left:250px;width:80px;height:100px;"' + this.tooltipAttrs(yourActive[2].getIdent(), 'pokemon', true, 'foe') + '></div>';
-			}
-			if (yourActive[1]) {
-				buf += '<div style="position:absolute;top:85px;left:320px;width:90px;height:100px;"' + this.tooltipAttrs(yourActive[1].getIdent(), 'pokemon', true, 'foe') + '></div>';
-			}
-			if (yourActive[0]) {
-				buf += '<div style="position:absolute;top:90px;left:390px;width:100px;height:100px;"' + this.tooltipAttrs(yourActive[0].getIdent(), 'pokemon', true, 'foe') + '></div>';
-			}
-			if (myActive[0]) {
-				buf += '<div style="position:absolute;top:210px;left:130px;width:180px;height:160px;"' + this.tooltipAttrs(myActive[0].getIdent(), 'pokemon', true, true) + '></div>';
-			}
-			if (myActive[1]) {
-				buf += '<div style="position:absolute;top:210px;left:270px;width:160px;height:160px;"' + this.tooltipAttrs(myActive[1].getIdent(), 'pokemon', true, true) + '></div>';
-			}
-			if (myActive[2]) {
-				buf += '<div style="position:absolute;top:210px;left:410px;width:140px;height:160px;"' + this.tooltipAttrs(myActive[2].getIdent(), 'pokemon', true, true) + '></div>';
-			}
-			this.$foeHint.html(buf);
-
 			if (this.battle.done) {
 
 				// battle has ended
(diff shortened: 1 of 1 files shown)


Zarel:
So, if controls are up, pokemon data should be read from `.myPokemon`, since those contain full move/item/ability data, but if the battle is playing/replaying, pokemon data should be read from `.active`. Does your change account for this?

urkerab:
Tooltips for switching and targeting your own Pokémon use the `'sidepokemon'` type, which is not affected by this change in any way.
Tooltips for the battle area and for targeting opposing Pokémon use the `'pokemon'` type, which initially uses the `.active` Pokémon; only the way the `.active` Pokémon is chosen has changed. However the `'pokemon'` type falls through to the existing `'sidepokemon'` type which also checks to see whether you are one of the players in which case it substitutes the `.myPokemon` for the `.active` Pokémon.

Zarel:
Tooltips for the battle area should only substitute `.myPokemon` if we're up to date and the move/switch chooser is open. Like, `.myPokemon` contains data for the end of the turn, so it's inaccurate if you mouseover at the beginning of the turn animation.

urkerab:
That existing `'sidepokemon'` code activates during normal play so if it was wrong I would have thought someone would have noticed by now.

Zarel:
It _is_ wrong, it's just that it's not particularly important so none of us have bothered to fix it.

But if we're going to fix instant replays, all the more reason to fix it, because it means that it could be out-of-date for more than a turn.

urkerab:
As I haven't actually noticed any errors in my instant replay tooltips, can you give me an example where reading from the wrong place will give me the wrong tooltip, so I can verify the fix?

urkerab:
Not using `.myPokemon` isn't really going to work because it won't show you your Pokémon from your point of view.

urkerab:
There are three cases to consider here:
1. The `'pokemon'` case for an opposing Pokémon. The ability and item are shown if they are known.
2. The `'sidepokemon'` case where the Pokémon is one that the server would have just sent as part of its request. While uses the code path for an opposing Pokémon, the ability and item are always known so they are always shown. The stats are also shown in this case.
3. The `'pokemon'` case when your Pokémon is active in the battle and therefore whose last known item and ability might not be up-to-date. In this case the existing code already checks for a volatile ability and uses that instead of the last known ability from the server, however the code always displays the last known item.
What this additional change does is to change the value corresponding to "Pokémon lost its item" from `''` to `null` thus making it possible to distinguish this case in the volatile item case. For good measure I avoid showing the ability or item if the opponent already knows it (it will then be shown in that section anyway).

Zarel:
>What this additional change does is to change the value corresponding to "Pokémon lost its item" from '' to null thus making it possible to distinguish this case in the volatile item case.

This is a brittle approach. Just test whether or not the battle is up-to-date.

>For good measure I avoid showing the ability or item if the opponent already knows it (it will then be shown in that section anyway).

We intentionally show this so it's clear whether or not the opponent already knows it.

urkerab:
> Just test whether or not the battle is up-to-date.
Sorry, I don't know what you mean by that, and it doesn't solve the case of your opponent knocking off an item they didn't know you had anyway.

Zarel:
> Sorry, I don't know what you mean by that,

Testing whether the battle is up-to-date generally involves comparing `battle.activityStep` with `battle.activityQueue.length`. Probably give one or two steps worth of leeway.

> and it doesn't solve the case of your opponent knocking off an item they didn't know you had anyway.

Can you clarify what about that case needs to be solved?

urkerab:
> We intentionally show this so it's clear whether or not the opponent already knows it.

I just thought the tooltip was getting a bit big and the information was redundant anyway, but I'll revert that bit.

For the knock off case, if your Pokémon had an item at the start of the turn then that is stored in `myPokemon.item`, while if your Pokémon was given an item during the turn then that is stored in `pokemon.item`. However there is no flag to indicate that your Pokémon's item was removed (or consumed) during the turn, so the (lack of) item won't update until the next request.

urkerab:
> Testing whether the battle is up-to-date generally involves comparing `battle.activityStep` with `battle.activityQueue.length`. Probably give one or two steps worth of leeway.

I don't suppose checking either `this.callbackWaiting` or `this.controlsShown` makes any sense?

urkerab:
(Latest version uses `this.callbackWaiting` as an example, although I wasn't able to distinguish the difference with or without that additional check.)

Zarel:
> I don't suppose checking either `this.callbackWaiting` or `this.controlsShown` makes any sense?

`controlsShown` makes more sense, yes. `callbackWaiting` should probably be removed entirely.

urkerab:
Now using `controlsShown` and added some comments for my sanity.

js/client-battle.js
```diff
@@ -1002,18 +997,23 @@
...
-			case 'sidepokemon':
+			case 'sidepokemon': // Tooltip for switching pokemon; myPokemon contains the latest server state.
 				var battlePokemon;
 				var myPokemon;
 				if (this.myPokemon) {
 					if (!pokemon) {
+						// We didn't fall through, so we're switching. Just show the server state.
 						pokemon = this.myPokemon[parseInt(thing)];
 						battlePokemon = this.battle.getPokemon('other: old: ' + pokemon.ident, pokemon.details);
 					} else if (pokemon.side === this.battle.mySide) {
```
Zarel: In theory, this is the part that shouldn't happen if `controlsShown` is false, right?

urkerab: Sorry, I'm not sure what you mean there. There are several cases:
- Choosing one of your own Pokémon for a switch (or targeting an ally in doubles/triples). This uses the `'sidepokemon'` case, and controlsShown is always true. This case uses the noted line and none of the other cases do.
- Targeting one of your opponent's Pokémon in doubles/triples. This uses the `'pokemoin'` case, and controlsShown is always true, but it's your opponent, so it makes no difference.
- Hovering over your opponent's Pokémon in the battle. This uses the `'pokemon'` case, and controlsShown may or may not be true, but it's your opponent, so it makes no difference.
- Hovering over your own Pokémon in the battle. This uses the `'pokemon'` case, and controlsShown may or may not be true. The `else if` part of the noted condition takes effect allowing the stats to be shown (and the ability even if the opponent doesn't know it). The current code assumes that if the opponent knows your Ability (e.g. due to in-turn activation, Skill Swap, Entrainment, Worry Seed etc.) then that's always at least as up-to-date as the server, so it doesn't check `controlsShown`.

Zarel: I'm saying that this `else if` should be

    } else if (pokemon.side === this.battle.mySide && this.controlsShown) {

In other words, hovering over your own pokemon in the battle-scene should be treated the same way as hovering over an opponent pokemon when controls are not shown.

urkerab: If you did that you wouldn't be able to see your own item, ability or stats in the battle-scene...

Zarel: Well, they would sometimes be incorrect, anyway, so it's better just not to show them at all.

urkerab: Out of interest, when would they be incorrect?

Zarel: I guess your following code is carefully designed to prevent this situation.

Still, though, I don't like the complication that results from mixing information sources like this, I prefer for information sources to be clearer.

urkerab:
This version shows the full display only when the controls are shown while when the game is animating or has finished you just get the spectator view.

Zarel:
Didn't we display possible abilities in the past? What happened?

urkerab:
The spectator view shows possible abilities, or the real ability if it is known via activation, trace, skill swap, entrainment, worry seed etc. This is done later on in showTooltip.
The controls view always knows the real ability of course.

PR #479, opened 2015-09-21
Support chat history in PM chatboxes

urkerab:
kota mentioned this when I was writing PR #477; kota also asked for a deduplication feature, although not exactly the same as the one I've actually implemented.

I made the chat history object global and split some of the key handling code out to it to reduce duplication.

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -170,6 +170,8 @@
 					animate: false,
 					extraSpace: 0
 				});
+				// create up/down history for this PM
+				this.chatHistories[userid] = new ChatHistory();
 			} else {
 				$pmWindow.show();
 				if (!dontFocus) {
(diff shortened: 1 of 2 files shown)


kotarou3:
Personally I prefer just checking if the previous message was a duplicate, and I think that's what most chat history features in other programs do.

Other than that, lgtm

Zarel:
I'm not sure global chat history is desirable. Can you poll people?

Zarel:
I think the chat history object can be left in `client-chat.js`

urkerab:
Sorry for the delay, but I misread your comment: this isn't global chat history; each room and PM creates their own chat history.

js/client-mainmenu.js
```diff
@@ -295,6 +296,7 @@
 							$chat.append('<div class="chat">User ' + userid + ' no longer ignored.</div>');
 						}
 					} else {
+						this.chatHistories[userid].push(text);
```
Zarel: Why is this here? Shouldn't `/ignore` and `/unignore` be added to the histories?

urkerab: I actually added line 299 before I knew that line 282 existed, and then removed line 282 as part of a cleanup sweep without realising what I'd done. I've moved it back (so to speak).

Zarel:
At some point, we should refactor this so the main menu actually contains objects representing each PM box.

PR #482, opened 2015-09-23
Allow spectators to switch sides

urkerab:
Apparently it's been suggested before, and it's definitely been suggested since, but the suggestion I found was http://www.smogon.com/forums/posts/5343275

Note that if you switch side while the battle is animating it will skip to the end; I couldn't work out how to skip to the current turn without restarting the background music.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2647,6 +2647,7 @@ var Battle = (function () {
 			if (this.turnCallback) this.turnCallback(this);
 			if (this.fastForward > -1 && turnnum >= this.fastForward) {
 				this.fastForwardOff();
+				if (this.endCallback) this.endCallback(this);
 			}
 			return;
 		}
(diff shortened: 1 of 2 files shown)


Joimer:
Additionally: Spectators seeing "your" side and "enemy" side is a little weird. Maybe it's about time we add player's name for spectators?

DanUgore:
There's a thought

Zarel:
I've been considering changing that phrasing, but "your" and "the opposing" kind of make it clear which one is which without having to look at the names of the players.

Zarel:
As for switching sides: I think skipping to the current turn is worth restarting the background music.

urkerab:
Well, I thought the only problem I had with skipping to the current turn was the background music, but but I'm struggling with the controls disappearing and I can't get them back.

urkerab:
Whether the controls appeared depends on whether the turn has started or not. If it hasn't started then it happens to work because you get a turn end notification. The best I could do was to fake a turn end notification when the fast forward finished. (Obviously the original version always worked because it fast forwarded to the end of the turn rather than the beginning.)

Also the new FontAwesome has a better icon for Switch Sides, which is neat.

PR #483, opened 2015-09-23
Firefox has magic button overflow

urkerab:
Firefox has this deliberately weird behaviour in that its buttons have 2px of magic padding - the padding is only there if it can make room for it.

This means that long names in the user list don't get the padding, thus misaligning them as seen here (line added for clarity):
screen shot 2015-09-22 at 17 02 35

Maybe there was a time when these needed to be buttons but it doesn't seem to be necessary now so I just changed them to be div elements instead as that seemed simplest.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -1446,6 +1446,10 @@ a.ilink:hover {
 	text-align: left;
 	cursor: pointer;
 }
+/* Remove Firefox's magic padding */
+.userlist li button::-moz-focus-inner {
+	padding: 0;
+}
 .userlist li button:hover {
 	background: #dce4ec;
 }


Zarel:
They're buttons because they're buttons. Their purpose is for you to click on them.

Can't you just hardcode a padding for all buttons? Seems like that would work better.

Zarel:
Specifically, just specify an explicit padding to the CSS for userlist buttons?

urkerab:
Firefox adds this extra optional padding via a special ::-moz-focus-inner pseudoelement, so you would have to turn it off with a Firefox-specific rule, and then optionally add your own padding to the button itself.

Zarel:
Let's just override that, then.

urkerab:
OK I've turned off the magic padding with Firefox-specific CSS. (The other approach would have been to specify a large right padding on the username so that it always overflows, thus triggering Firefox's quirky padding collapsing.)

PR #490, opened 2015-09-27
Fix base power preview for Punishment and Stored Power

urkerab:
The base power preview calculations for Punishment and Stored Power are incorrect:
- The multiplier should not consider any evasion or accuracy boosts
- The multiplier is the sum of the boosts, rather than the count
- Punishment is limited to 200 base power

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1326,9 +1326,10 @@
 				var boosts = Object.keys(target.boosts);
 				var multiply = 0;
 				for (var i = 0; i < boosts.length; i++) {
-					if (target.boosts[boosts[i]] && target.boosts[boosts[i]] > 0) multiply++;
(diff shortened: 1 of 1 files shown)


Slayer95:
> The multiplier should not consider any evasion or accuracy boosts

That's is not consistent with how the sim calculates the Base Power.

> Punishment is limited to 200 base power

This doesn't seem to be implemented in the sim, even though the description says so.

@Marty-D

Marty-D:
@urkerab, your first point is wrong, but the others are right.
@Slayer95, [what are you talking about](https://github.com/Zarel/Pokemon-Showdown/commit/c86660ec4b2f60422b36c1ebd20a6f444022de4f)???

urkerab:
I've reinstated evasion and accuracy boosts.

PR #491, opened 2015-09-27
Fix base power preview for item-dependent moves

urkerab:
`getMoveBasePower` needs to know the volatiles and boosts for some moves and the stats and item for other moves. Unfortunately there is no single point of storage for this information; the volatiles and boosts are stored on the active Pokémon while the stats and item are sent by the server for each decision. (The only time this is not true is while spectating or animating in which case the client is still sometimes aware of an item (e.g. due to Frisk, Trick etc.) on an active Pokémon.) This means that, except in those specific cases, the moves Acrobatics, Fling and Natural Gift display incorrect information.
Fortunately there is a way of getting from the active Pokémon to the server-sent Pokémon, so we can pass the active Pokémon to `getMoveBasePower` and it can then look up the server-sent Pokémon.
The problem also affects `getMoveType` which needs to check the volatiles to see whether the Pokémon has been Embargoed but the server-sent Pokémon to see whether it has an item. The code which updates the controls is also affected in this case. (Is it supposed to show the computed type for moves such as Judgment?)

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -425,7 +425,7 @@
 					if (move.id === 'Struggle' || move.id === 'Recharge') pp = '&ndash;';
 					if (move.id === 'Recharge') move.type = '&ndash;';
 					if (name.substr(0, 12) === 'Hidden Power') name = 'Hidden Power';
-					var moveType = this.getMoveType(move, switchables[pos]);
+					var moveType = this.getMoveType(move, this.battle.mySide.active[pos]);
 					if (moveData.disabled) {
 						movebuttons += '<button disabled="disabled"' + this.tooltipAttrs(moveData.move, 'move') + '>';
 						hasDisabled = true;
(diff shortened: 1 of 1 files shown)


js/client-battle.js
```diff
@@ -425,7 +425,7 @@
 					if (move.id === 'Struggle' || move.id === 'Recharge') pp = '&ndash;';
 					if (move.id === 'Recharge') move.type = '&ndash;';
 					if (name.substr(0, 12) === 'Hidden Power') name = 'Hidden Power';
-					var moveType = this.getMoveType(move, switchables[pos]);
+					var moveType = this.getMoveType(move, this.battle.mySide.active[pos]);
```
Zarel: Why aren't we using `switchables` here?

Zarel:
This is merged now, but I still want to know why we changed from using `switchables` there.

js/client-battle.js
urkerab: `getMoveType` needs the active Pokémon to check whether it is under the effect of Embargo.

PR #493, opened 2015-09-29
Implement |unlink|roomhide| for room-specific message suppression

ascriptmaster:
For https://github.com/Zarel/Pokemon-Showdown/pull/2104

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1118,6 +1118,10 @@
 					if (!$messages.length) break;
 					$messages.find('a').contents().unwrap();
 					if (row[2]) {
+						if (row[1] === 'roomhide') {
+							$messages = this.$chat.find('.chatmessage-' + user);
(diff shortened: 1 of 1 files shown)


Zarel:
hm

ascriptmaster:
But what does "hm" mean, good sir

Zarel:
I'm not sure this is ideal/necessary.

ascriptmaster:
Why not? If we're going to allow messages to be hidden after a user is room banned, shouldn't it only hide messages from that room?

Zarel:
If a user deserves to have their messages hidden, they're a spammer and they should be locked and have their messages hidden everywhere.

This is straight-up a trust problem and we really need to find a way to deal with it.

Zarel:
But I don't have a better solution so I guess this is fine.

PR #496, opened 2015-09-30
Make side rooms overflow left on the tab bar

urkerab:
This implements the following forum suggestion http://www.smogon.com/forums/posts/6447708 so that tabs open just above the side room as normal but if they would overflow then they get pushed left until they hit the main room tabs, at which point the dropdown activates in the normal way.
The 309 comes from the 165 in the line before plus the 144 from eleven lines above.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -1831,18 +1831,29 @@
 				sideBuf += '<li><a class="button' + (curId === 'rooms' || curSideId === 'rooms' ? ' cur' : '') + '" href="' + app.root + 'rooms"><i class="fa fa-plus" style="margin:7px auto -6px auto"></i> <span>&nbsp;</span></a></li>';
 			}
 			if (atLeastOne) buf += '</ul>';
+			var margin = 0;
 			if (sideBuf) {
 				if (app.curSideRoom) {
-					buf += '<ul class="siderooms" style="float:none;margin-left:' + (app.curSideRoom.leftWidth - 144) + 'px">' + sideBuf + '</ul>';
+					margin = app.curSideRoom.leftWidth - 144;
+					buf += '<ul class="siderooms" style="float:none;margin-left:' + margin + 'px">' + sideBuf + '</ul>';
 				} else {
 					buf += '<ul>' + sideBuf + '</ul>';
 				}
 			}
 			this.$tabbar.html(buf);
-			var $lastLi = this.$tabbar.children().last().children().last();
+			var $lastUl = this.$tabbar.children().last();
+			var $lastLi = $lastUl.children().last();
 			var offset = $lastLi.offset();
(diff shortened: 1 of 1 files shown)


Zarel:
You're missing a pixel here:

image

Slayer95:
Awesome!

urkerab:
There's a sneaky extra -1px margin on the a.button which the existing code wasn't taking into account (so if you resized your window carefully enough you could make the right border disappear).

js/client.js
```diff
@@ -1839,10 +1839,17 @@
...
+			var $lastLi = $lastUl.children().last();
 			var offset = $lastLi.offset();
 			var width = $lastLi.outerWidth();
-			if (offset.top >= 37 || offset.left + width > $(window).width() - 165) {
+			// 166 here is the difference between the .maintabbar's right margin and the a.button's right margin.
+			if (app.curSideRoom && offset.left + width > $(window).width() - 166) {
+				var margin = $(window).width() + app.curSideRoom.leftWidth - offset.left - width - 310;
+				$lastUl.css("margin-left", margin + "px");
+				offset = $lastLi.offset();
+			}
```
Zarel: Should you be adding ` else { $lastUl.css('margin-left', 0); }` here?

js/client.js
```diff
@@ -1839,10 +1839,17 @@
...
-			var $lastLi = this.$tabbar.children().last().children().last();
+			var $lastUl = this.$tabbar.children().last();
+			var $lastLi = $lastUl.children().last();
 			var offset = $lastLi.offset();
 			var width = $lastLi.outerWidth();
-			if (offset.top >= 37 || offset.left + width > $(window).width() - 165) {
+			// 166 here is the difference between the .maintabbar's right margin and the a.button's right margin.
+			if (app.curSideRoom && offset.left + width > $(window).width() - 166) {
+				var margin = $(window).width() + app.curSideRoom.leftWidth - offset.left - width - 310;
+				$lastUl.css("margin-left", margin + "px");
```
Zarel: Also, not a huge deal, but in theory our standard is to use `'` and not `"` here.

js/client.js
urkerab: If the tabs fit (left + width + 166 < window) then I shouldn't have to change the margin.

js/client.js
urkerab: Sorry, I'm not used to single-quoted strings. I'll change them and update the PR later.

js/client.js
Zarel: if you've changed the margin before, though, you should change it back...

urkerab:
I tweaked the margin calculation so hopefully this way should be clearer.

PR #497, opened 2015-10-02
Allow tabs to be reordered with Shift+Arrow

urkerab:
I've probably overengineered this, but whatever.
I've given each room an order, which dictates where it gets inserted into the room order array, but could possibly be used in future to simplify the tab creation code because you can group the rooms by the upper nibble of the order.
Someone should look at the autojoin code because I can't test that in the test client.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -359,8 +359,8 @@
 				app.supports['rooms'] = true;
 			}
 
-			this.topbar = new Topbar({el: $('#header')});
 			this.addRoom('');
+			this.topbar = new Topbar({el: $('#header')});
 			if (!this.down && $(window).width() >= 916) {
 				if (document.location.hostname === 'play.pokemonshowdown.com') {
 					this.addRoom('rooms', null, true);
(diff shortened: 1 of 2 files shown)


Zarel:
I would rather implement this differently...

Zarel:
This isn't the architecture I planned on for drag-and-drop reordering.

Zarel:
Specifically, here is how I would design room reordering:

- `app.roomList` (array of rooms)
- `app.sideRoomList` (array of rooms)

Topbar displays these two arrays in order. Rearranging rooms involves rearranging the array(s) and re-running the topbar renderer.

urkerab:
Well, there's some code duplication here, even if you do a whitespace diff, but I was for instance able to remove the atLeastOne variable. It occurs to me that I've forgotten to fix the overflow popup.

js/client.js
```diff
@@ -1563,13 +1608,11 @@
...
 			var autojoins = [];
 			var autojoinCount = 0;
-			for (var i in this.rooms) {
-				if (!this.rooms[i]) continue;
-				if (this.rooms[i].type !== 'chat' || i === 'lobby') {
-					continue;
-				}
-				autojoins.push(this.rooms[i].id.indexOf('-') >= 0 ? this.rooms[i].id : (this.rooms[i].title || this.rooms[i].id));
-				if (i === 'staff' || i === 'upperstaff') continue;
+			for (var i = 0; i < this.sideRoomList.length; i++) {
```
Zarel: Loop through `roomList` as well; we'll be supporting both room types in either side.

js/client.js
```diff
@@ -1764,56 +1806,25 @@
...
-							}
-						}
+			if (app.roomList.length) buf += '<ul>';
+			for (var i = 0; i < app.roomList.length; i++) {
+				var room = app.roomList[i];
+				if (room.notifications) notificationCount++;
+				var id = room.id;
+				var formatid = id.split('-')[1];
+				var name = Tools.escapeHTML(room.title);
+				if (!name) {
```
Zarel: Just use `renderRoomTab` instead of all this silliness.

js/client.js
urkerab: I'm not sure what you mean by that; I wrote the code assuming that only `BattleRoom`s would appear in `roomList` and only `ChatRoom`s would appear in `sideRoomList`.

js/client.js
urkerab: `renderRoomTab` is almost perfect except it can't deal with the Rooms room, which must be rendered if the app supports it even if it doesn't exist yet. (I thought about changing it to take an id but that would cause problems in the Teambuilder and Ladder case.)

js/client.js
urkerab: OK, so it turns out that `renderRoomTab` makes that point moot, and the code doesn't actually care any more. I just need to know what list has preference for autojoins.

Zarel: I mean that you should write your code assuming any kind of room can be in any list.

Zarel: I'd say add both lists to autojoins, skipping battle rooms for now. Just go left to right.

urkerab:
Now removing all the cur(Side)Room[Left|Right] baggage. Incidentally this fixes a bug with the client whereby if you don't have a screen big enough to show a chat room and battle at the same time, the battle thinks you need to press left to go to the rightmost chat room; with this change the battle now thinks you need to press right to go to the leftmost chat room. Also fixes a bug when you have so little room you can't even show the main menu with a chat room whereby you can't reorder rooms. Also fixes a bug whereby the placeholder doesn't show the right arrow key rooms after a reorder. Also fixes some bugs in the autojoins code.

Zarel:
Was going to ask for these to be separate commits, but not too important. Looks good.

PR #498, opened 2015-10-03
Allow tabs to be reordered with Shift+Arrow

urkerab:
This is a hackier version by comparison with PR #497, but less intrusive.


Freigeist:
Would be nice if the order would be save in autojoin

Zarel:
Way too hacky. Let's discuss more in the original pullreq.

PR #499, opened 2015-10-03
Tone down excessive notifications

Slayer95:
- System chat messages (|c|~|) no longer arise subtle notifications
- Do not notify again for the same chat messages on reload: last message date is stored in preferences for each room.


Joimer:
This is very cool.

DanUgore:
^

Slayer95:
Pretty please? Or would you prefer something like ``Tools.roomPrefs(roomid:String, key:String, value:Object)`` ? o.o

Zarel:
One thing we can do is support both `|c||` and `|c|~|` and have them behave differently. Maybe one notifies and one doesn't.

Zarel:
I'd argue that there are very few system notifications people actually wouldn't want to ignore.

Slayer95:
Sounds good

Slayer95:
It seems that's already the case lol. We support any name which maps to an empty ID.
In this PR, specifically ``|c|~|`` is made not to arise subtle notifications.

Zarel:
I'm going to merge this without proofreading like I usually do; we can clean up the storage syntax later, I guess.

PR #501, opened 2015-10-04
Show the Rooms tab in the test client

urkerab:
The client doesn't update the tab bar when the server sends it the rooms, so you don't see the rooms tab unless you do something else that causes the tab bar to update.

Since the client now updates the tab bar automatically, you don't need to manually support the rooms tab, but I can remove that hunk if you prefer.


Zarel:
I think at this point, all servers support rooms, so we can go ahead and set that.

PR #502, opened 2015-10-04
Don't let room names overflow the tablist popup

urkerab:
If you're, for instance, watching an Anything Goes battle between allaboutthatpanda and Sammichinmyhead (I kid you not), you'll find that the text doesn't fit in the tablist popup.


Zarel:
Screenshot?

urkerab:
Screenshot as requested for completeness.
pr502

PR #504, opened 2015-10-08
Show the pretty format name in the battle tab

urkerab:
Tabs for battles currently show the format id and the players. I thought it would look slightly nicer if we used the format name instead. Example:

    _______________________
    |      ounomega      X|
    |junzheng5 vs. OVOOHKO|
    _______________________
    |    OU (no Mega)    X|
    |junzheng5 vs. OVOOHKO|

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -1893,7 +1893,7 @@
 							name = '(empty room)';
 						}
 					}
-					return buf + '<i class="text">' + formatid + '</i><span>' + name + '</span></a><a class="closebutton" href="' + app.root + id + '"><i class="fa fa-times-circle"></i></a></li>';
+					return buf + '<i class="text">' + Tools.escapeFormat(formatid) + '</i><span>' + name + '</span></a><a class="closebutton" href="' + app.root + id + '"><i class="fa fa-times-circle"></i></a></li>';
 				} else {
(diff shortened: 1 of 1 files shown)


js/client.js
```diff
@@ -1893,7 +1893,7 @@
 							name = '(empty room)';
 						}
 					}
-					return buf + '<i class="text">' + formatid + '</i><span>' + name + '</span></a><a class="closebutton" href="' + app.root + id + '"><i class="fa fa-times-circle"></i></a></li>';
+					return buf + '<i class="text">' + format + '</i><span>' + name + '</span></a><a class="closebutton" href="' + app.root + id + '"><i class="fa fa-times-circle"></i></a></li>';
```
Slayer95: Escape on render

urkerab: So you'd prefer me to inline the `Tools.escapeFormat(id.split('-')[1])` directly?

Slayer95: Yup, but I myself would keep ``formatid``'s variable around for readibility.

Zarel: I assume Slayer95 means `Tools.escapeFormat(formatid)`. That's what I'd prefer here.

Zarel: Apparently `format` is already defined but unused? We can get rid of that.

urkerab:
(sorry for the delay, was working on PR #507)

PR #505, opened 2015-10-08
Add support for <style> tags in caja

lc-guy:
The only danger \<style\> tags can introduce (vulnerabilities in very old browsers) doesn't matter since PS! doesn't even support those anyway.


Zarel:
There are plenty of modern vulnerabilities involved. Phishing is a big one.

kotarou3:
Could possibly work if we use scoped styles (https://github.com/PM5544/scoped-polyfill) or sandbox it inside a iframe

PR #506, opened 2015-10-08
Update chatbox focus text to reflect reality

urkerab:
The Left/Right room navigation was changed so that it always loops through all rooms (including the Rooms room) even when the window is wide enough to tab each set of rooms separately. This updates the chatbox focus text to match.

That code uses the roomList.concat(sideRoomList) trick to create a master list of all chat rooms. While I was there I let the scope creep and applied the trick to the focusNextRoom code too. And then for good measure I implemented a request that the rooms should wrap around so that Right on the Rooms room should focus the leftmost room and vice versa, because that actually ended up further simplifying the code.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -59,9 +59,7 @@
 
 		focusText: function () {
 			if (this.$chatbox) {
-				var rooms = app.roomList;
-				if (this === app.curSideRoom) rooms = app.sideRoomList;
-				else if (!app.curSideRoom) rooms = rooms.concat(app.sideRoomList);
+				var rooms = app.roomList.concat(app.sideRoomList);
 				var roomIndex = rooms.indexOf(this);
 				var roomLeft = rooms[roomIndex - 1];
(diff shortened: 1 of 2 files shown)


js/client-chat.js
```diff
@@ -59,9 +59,7 @@
 
 		focusText: function () {
 			if (this.$chatbox) {
-				var rooms = app.roomList;
-				if (this === app.curSideRoom) rooms = app.sideRoomList;
-				else if (!app.curSideRoom) rooms = rooms.concat(app.sideRoomList);
+				var rooms = app.roomList.concat(app.sideRoomList);
```
Zarel: Should we concat `rooms` to this?

urkerab: Sorry, I'm not sure what you're asking.

Zarel: Oh well, it's not a big deal.

CrestfallPS:
ty for adding wrap as I requested urkerab \o/

PR #511, opened 2015-10-14
Add missing commas in data/graphics.js

Joimer:
Those missing commas make battle joining crash in a lot of browsers.


Slayer95:
@Zarel

Joimer:
or @monsanto I think he can update Client as well

PR #512, opened 2015-10-17
Omit forfeit dialog when leaving instant replay.

Koushien:
If user finished a battle, leaving the room during an instant
replay would open an unneeded forfeit dialog.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2124,6 +2124,7 @@ var Battle = (function () {
 		this.sides = [];
 		this.lastMove = '';
 		this.gen = 6;
+		this.replay = 0;
 
 		this.frameElem = frame;
 		this.logFrameElem = logFrame;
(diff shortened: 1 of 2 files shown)


Zarel:
There is definitely a better way to do this...

Slayer95:
Basically set the property ``ended`` of the <i>room</i> to ``true`` once ``battle.ended`` is first set ``true``. Then check the property of the <i>room</i> rather than of the battle.

I was looking into it before, but I desisted because the only implementation I <i>liked</i> required getters/setters...

Koushien:
I tried that approach, but it would mess with the fast forward behavior. On my setup, the buttons would even disappear after being pressed.
I'll play with it some more.

Slayer95:
Fixed as of https://github.com/Zarel/Pokemon-Showdown-Client/commit/d304e8ac0f1add26ba4650322ee0fce3b3363736

PR #513, opened 2015-10-17
Correct Zorua dimensions

QuiteQuiet:


Marty-D:
normal - normal (49x69)
shiny - shiny (47x63)

Whether this is merged or not, both of these images should be the same size. It looks like they were taken from different angles though.

QuiteQuiet:
Oh, I knew I was forgetting something when I did this. The option is always to pick one and remake the other to match it by size.

I don't mind closing this and making a new non-shiny model honestly, but I should've checked that first :c

Zarel:
Please go ahead, that would be best.

PR #515, opened 2015-10-18
Highlight the current user in elimination tournaments

urkerab:
The original idea was to draw a dashed border for the team but as it turns out you get it on the battles you win too.

### DIFF
--- a/js/client-chat-tournament.js
+++ b/js/client-chat-tournament.js
@@ -554,6 +554,7 @@
 				if (!data.rootNode)
 					return;
 
+				var name = app.user.get('name');
 				var nodeSize = {
 					width: 150, height: 20,
 					radius: 5,
@@ -635,6 +636,7 @@
 							elem.attr('y', -nodeSize.smallRealHeight / 2).attr('height', nodeSize.smallRealHeight);
 						else
 							elem.attr('y', -nodeSize.realHeight / 2).attr('height', nodeSize.realHeight);
+						if (node.team === name) elem.attr('stroke-dasharray', '5,5');
 					});
 				nodeGroup.each(function (node) {
 					var elem = d3.select(this);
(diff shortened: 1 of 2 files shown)


sirDonovan:
Do you have a screenshot of this?

urkerab:
This is a mockup I created using developer tools (I did join a tournament to test this but didn't think to take a screenshot at the time).
capture

kotarou3:
hm, it doesn't stand out very much. Maybe tint it with a bit of yellow?

Zarel:
blah, those colors aren't very good. How are the colors chosen, anyway? What do green and red mean?

kotarou3:
Green means the bolded person won, while red means lost

urkerab:
We can tweak the colours of course... for this simulated screenshot I adjusted the colours so that they would all have the same apparent luminance (Y).
tournament

Zarel:
I generally avoid having green and red as the only hues, because of Christmas implications. I usually do red and gray, or red and blue, or gray and green, or something like that.

Zarel:
Also, if the color coding is who won, how is who's bolded determined?

ascriptmaster:
Bolded seems to be who won that round. Red/Green seems to be whether the winner of that round won the next round. At least, that's what seems to be the case.

Zarel:
ooooooooh

Zarel:
Maybe we should get rid of the color coding? The colors are kinda ugly and it's not like it's clear anyway.

kotarou3:
I'm a terrible designer - do what you want

urkerab:
I asked around and people seem to like the colours.

For this version I also made the player's name italic. (This also affects round robin, because I wanted to be consistent.) With this change, I would be happy to revert the colours back to the original, since now it doesn't matter so much that the borders aren't so visible.

Zarel:
I don't like the tour design in general but it's not really a problem with this change specifically.

Your color change is good and should be kept.

urkerab:
So, which options would you like me to keep, the colour change, the border change and/or the text decoration change?

Zarel:
Use the colors from your simulated screenshot. For everything else, just use your own judgment.

urkerab:
OK, so the current diff has the colour changes and the dotted border, but not the italic font.

PR #518, opened 2015-10-25
Update teams on deletions, undeletions, and drops

### LINKED ISSUE #443: Team select image does not update properly
I found a bug on the client that has to do with deleting teams. When a team is deleted, the index for the team selection stays the same, but the image previewing the team does not update. So say you have the following teams:
0 | Team A
1 | Team B <-
2 | Team C
In the "look for a battle" team selection, it is currently selecting team B. Then you delete team A and the selection is as follows:
0 | Team B
1 | Team C <-
However, the image displayed is still that of team B. So you can search for a battle thinking you are using team B, but you end up using team C. The image updates only after a battle is found, or the matchmaking is canceled.

augmt:
Fixes #443.

Also: dragging and dropping teams did not update `showdown_teams` in localStorage. It does so now.

### DIFF
--- a/js/client-chat-tournament.js
+++ b/js/client-chat-tournament.js
@@ -208,13 +208,7 @@
 				return;
 			}
 
-			var teamIndex = undefined;
-			if (!forceFormatChange && this.$teamSelect.children().val()) {
-				teamIndex = parseInt(this.$teamSelect.children().val());
-				if (isNaN(teamIndex)) teamIndex = undefined;
-			}
-
-			this.$teamSelect.html(app.rooms[''].renderTeams(this.info.format, teamIndex));
+			this.$teamSelect.html(app.rooms[''].renderTeams.call(this, this.info.format));
 			this.$teamSelect.children().data('type', 'teamSelect');
 			this.$teamSelect.children().attr('name', 'tournamentButton');
 			this.$teamSelect.show();
@@ -266,6 +260,7 @@
 					var type = data[1];
 					this.room.$chat.append("<div class=\"notice tournament-message-create\">" + format + " " + Tools.escapeHTML(type) + " Tournament created.</div>");
 					this.room.notifyOnce("Tournament created", "Room: " + this.room.title + "\nFormat: " + format + "\nType: " + type, 'tournament-create');
+					this.curTeamIndex = 0;
 					this.updateTeams();
 					break;
 
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -561,10 +561,9 @@
 			var self = this;
 
 			this.$('button[name=team]').each(function (i, el) {
-				var val = el.value;
-				if (val === 'random') return;
+				if (el.value === 'random') return;
 				var format = $(el).closest('form').find('button[name=format]').val();
-				$(el).replaceWith(self.renderTeams(format, val));
+				$(el).replaceWith(self.renderTeams(format));
 			});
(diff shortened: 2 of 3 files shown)


Zarel:
This fixes some things, but not #443 as far as I can tell.

js/client-teambuilder.js
```diff
@@ -237,8 +237,7 @@
 			i = +i;
 			this.deletedTeamLoc = i;
 			this.deletedTeam = teams.splice(i, 1)[0];
-			Storage.deleteTeam(this.deletedTeam);
-			this.update();
+			this.saveTeams();
```
Zarel: This breaks the desktop client.

augmt:
Fixed the desktop issue, hopefully.

I don't see how this fails to fix #443 though.

Zarel:
The problem with 443 is that team indexes don't update when you rearrange teams. They still don't.

augmt:
The code's been updated to adjust `curTeamIndex` accordingly. `TournamentBox` now has its own `curTeamIndex` as well.

Zarel:
Can you use `for..in` instead of `_.each`? `_.each` doesn't seem to add anything here except slow down the code?

Also can you split the team index adjustment into a separate commit?

augmt:
Done and split into 3 commits.

js/client-chat-tournament.js
```diff
@@ -208,13 +208,7 @@
...
 			}
 
-			var teamIndex = undefined;
-			if (!forceFormatChange && this.$teamSelect.children().val()) {
-				teamIndex = parseInt(this.$teamSelect.children().val());
-				if (isNaN(teamIndex)) teamIndex = undefined;
-			}
-
-			this.$teamSelect.html(app.rooms[''].renderTeams(this.info.format, teamIndex));
+			this.$teamSelect.html(app.rooms[''].renderTeams.call(this, this.info.format));
```
urkerab: Unfortunately this gets called for every join and leave, which means that there's no point selecting a team until after the tournament starts.

js/client-mainmenu.js
```diff
@@ -896,8 +895,12 @@
 			var formatid = this.sourceEl.closest('form').find('button[name=format]').val();
 			i = +i;
 			this.sourceEl.val(i).html(TeamPopup.renderTeam(i));
-			app.rooms[''].curTeamIndex = i;
-			app.rooms[''].curTeamFormat = formatid;
+			if (this.sourceEl[0].offsetParent.className === 'mainmenuwrapper') {
+				app.rooms[''].curTeamIndex = i;
+				app.rooms[''].curTeamFormat = formatid;
+			} else if (this.sourceEl[0].offsetParent.className === 'tournament-box active') {
+				app.curSideRoom.tournamentBox.curTeamIndex = i;
```
urkerab: Tournaments don't have to be created in side rooms; your window might not be large enough for a side room, or you may have shifted the tab to the main side.

js/client-chat-tournament.js
augmt: I do find calling `updateTeams` on every join and leave superfluous but having some time to select a team before the tournament starts seems better than not.

How about we make it so `updateTeams` is only called whenever the user, instead of simply anyone, joins or leaves a tournament.

urkerab: I just picked join/leave because someone complained about it in chat. It may affect other callers as well.

PR #520, opened 2015-10-26
Disallow invalid /challenges

augmt:
Disallow `/challenge` from allowing users to challenge their own account to a battle or challenge other users to battle formats that don't exist.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -874,6 +874,14 @@
 				this.add('This player does not exist or is not online.');
 				return;
 			}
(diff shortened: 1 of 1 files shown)


Irraquated:
https://github.com/Zarel/Pokemon-Showdown-Client/pull/414

augmt:
Fair enough. Should using `/challenge` to issue challenges in invalid formats be overlooked too?

Slayer95:
I don't think the client should check that.

``/challenge`` is not a command you should be manually using anyway.

PR #522, opened 2015-10-28
Show items that have been consumed

urkerab:
This was a request made to me in a PM. When an item is removed, the nature of the removal is saved and displayed with the item in the tooltip.
I also made some more items show up in the tooltip such as leftovers and mega stones.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -181,6 +181,7 @@ var Pokemon = (function () {
 		this.ability = '';
 		this.baseAbility = '';
 		this.item = '';
+		this.itemEffect = '';
 		this.species = species;
 		this.fainted = false;
 		this.zerohp = false;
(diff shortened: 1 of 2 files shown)


js/battle.js
```diff
@@ -4027,12 +4035,14 @@ var Battle = (function () {
 				var item = Tools.getItem(args[2]);
 				var effect = Tools.getEffect(kwargs.from);
 				var ofpoke = this.getPokemon(kwargs.of);
-				poke.item = '';
+				poke.item = item.name;
```
Slayer95: ``Pokemon#item`` is no longer accurate?!

We should look into another way to implement this... The existence of item history shouldn't imply that there is no item (history shouldn't be limited to removal but it should also include item transfer).

Zarel: I agree with Slayer95 here.

urkerab: That depends on your definition of accurate; prior to this change the tooltip couldn't distinguish between an unknown item or a removed item, but I agree that using the same property might be confusing and I'll think of another way to implement this.

Out of interest, in what way would the history from prior to the transfer of an item be useful?

Zarel: This is simple enough to fix, just have `.itemEffect` be a an array `[prevItem, effect]`.

`.itemEffect` is, incidentally, a horrible name for it. `.removedItem` is probably better.

PR #523, opened 2015-10-28
Fix Team Import/Export barely visible > 640px width

forsureitsme:
Yeah, you could still drag the resizer down, but this make it less worse.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -2469,6 +2469,7 @@ a.ilink.yours {
 }
 .teamwrapper.scaled {
 	width: 640px;
+	height: 100%;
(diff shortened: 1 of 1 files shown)


Zarel:
Wait, which browser is this?

forsureitsme:
Bug occurs in Chrome 46.0.2490.80 m (64-bit) and Firefox 41.0.2, both latest versions.
I'm not reenabling IE on my Win8 just to test this.

Zarel:
That's fine, it's just weird because I thought I'd fixed this already.

PR #524, opened 2015-10-28
Track the previous item in the battle tooltip

urkerab:
I'm assuming this is what @Slayer95 wanted in PR #522.

I had to change the way itemEffect was set otherwise the display for Trick and Switcheroo was wrong.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -182,6 +182,8 @@ var Pokemon = (function () {
 		this.baseAbility = '';
 		this.item = '';
 		this.itemEffect = '';
+		this.prevItem = '';
+		this.prevItemEffect = '';
 		this.species = species;
 		this.fainted = false;
 		this.zerohp = false;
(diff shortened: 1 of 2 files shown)


Zarel:
`.itemEffect` is still a horrible name. `previousItemEffect` maybe.

Also we might as well be consistent with server and use `prev` instead of `previous`.

Slayer95:
I simply meant something like ``Draco Plate (Received with Trick)``, but I guess the more info the better.

urkerab:
Item: Choice Scarf (tricked; was Leftovers)

Also includes the renames as requested.

Zarel:
Do we really need to track `.itemEffect` as well...?

Slayer95:
I'm not a fan of it, but I suppose it's relevant for Recycle...

urkerab:
It's the best I could do within these constraints:
- If a Pokémon lost an item, then the item it lost should be shown (implemented by previous PR)
- The reason that it lost the item should also be shown (implemented by previous PR)
- `.item` should only contain an item that the Pokémon is actually holding (from comment on previous PR)
- The reason that it came into possession of the item should also be shown (from comment on previous PR)

Zarel:
> The reason that it came into possession of the item should also be shown (from comment on previous PR)

Yeah, that was the part I was questioning. I don't really see a comment to that effect. Oh well, it's probably not a bad feature.

PR #527, opened 2015-10-30
Allow tab completion to work in battles

urkerab:
Console rooms know about autocompletion but only chat rooms maintain a user list. This makes the battle maintain a dummy user list (thus the necessity to prepend `' '` to the user before adding it to the list).

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2169,6 +2169,7 @@ var Battle = (function () {
 		// external
 		this.resumeButton = this.play;
 
+		this.users = {};
 		this.preloadCache = {};
 
 		this.preloadEffects();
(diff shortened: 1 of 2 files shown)


js/battle.js
```diff
@@ -5679,12 +5680,14 @@ var Battle = (function () {
...
 		case 'join':
 		case 'j':
+			this.users[toUserid(args[1])] = ' ' + args[1];
 			if (!this.ignoreSpects) {
 				this.log('<div class="chat"><small>' + Tools.escapeHTML(args[1]) + ' joined.</small></div>', preempt);
 			}
 			break;
 		case 'leave':
 		case 'l':
+			delete this.users[toUserid(args[1])];
```
Zarel: Do we necessarily need to do this?

Also do we have rename events in battles?

Zarel: Is the battle object the right place to keep track of this? Maybe we should retain a reference to the parent room object.

Zarel: I think we already reference the containing room in other parts of battle.js

Slayer95: We don't have any reference to the room from the battle.

Zarel: We don't hold a reference, but we do weird stuff with `$.closest` iirc.

urkerab: To answer your questions in order:
1. No, you don't need to delete old users, but the chat rooms do (a room mod asked me about this because he wanted to tab complete hit and run spammers)
2. Battles don't seem to have rename events.
3. I made the room take a reference to the battle because not all battles have rooms, i.e. replays, so it was easier that way around.

Zarel: Yes, that's my point, though, there's no need to keep track of users if the battle doesn't have a room attached.

urkerab: I think I see what you mean now, but seeing as you've already merged this, I'll create a new PR.

PR #528, opened 2015-10-30
Tournaments: support showing users who have joined

sirDonovan:
During signups in elimination tournaments, show the number of users who
have joined in place of the empty bracket

Screenshot: http://i.imgur.com/dg2xE1s.png

### DIFF
--- a/js/client-chat-tournament.js
+++ b/js/client-chat-tournament.js
@@ -551,8 +551,14 @@
 
 		TournamentBox.prototype.generateBracket = function (data) {
 			if (data.type === 'tree') {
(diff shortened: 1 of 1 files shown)


TheImmortal:
Why not show the users as well? Like by clicking that.

Zarel:
Can you just show a list of rooms?

Zarel:
Can you just show a list of all the users? Yeah, what TI said.

sirDonovan:
Can we do both? http://i.imgur.com/HK2HKTw.png

Some people may only need the number while others may want to see individual users.

PR #529, opened 2015-10-30
Support capital letters in <<room>> links

sirDonovan:


Zarel:
<<room>> links are roomids, capital letters are not allowed in roomids.

sirDonovan:
I'm aware of that, but many users are not and try to use the feature with capital letters (even some staff have asked about it).

Zarel:
The problem is that if they try to use room titles, they'll have lots of other problems because room titles are not unique, nor are room links deducible from room titles.

Zarel:
Honestly, supporting titles in `/join` already annoys me. :|

PR #531, opened 2015-10-31
Don't track users in replays

urkerab:
As requested in PR #527 this moves the onus for enabling user tracking from the battle to the room.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2174,7 +2174,6 @@ var Battle = (function () {
 		// external
 		this.resumeButton = this.play;
 
-		this.users = {};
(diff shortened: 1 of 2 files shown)


Zarel:
Maybe just set `this.room`?

Zarel:
Better idea, set `this.roomid` in the battle.

I distrust the garbage collector. :P

urkerab:
So you want something like `if (this.roomid) delete app.rooms[this.roomid].users[userid]?`

Zarel:
Yes.

urkerab:
Done.

PR #532, opened 2015-10-31
Fix issue with team select image not updating.

### LINKED ISSUE #443: Team select image does not update properly

18vshanmugam:
Some minor workarounds to fix #443. Now, the teams will update whenever the team images on the teambuilder update, so including team swaps and deletions. Also, only calling updateTeams() did not  work when I tested it so I added the code to update the teams inline on the .on() function. Finally, I added a new message for when the teamIndex is over the teams size, so that users weren't confronted with a "Corrupted team" message when they deleted their last team while the index was pointing at it.

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -74,7 +74,17 @@
 			app.on('init:formats', this.updateFormats, this);
 			this.updateFormats();
 
-			app.user.on('saveteams', this.updateTeams, this);
+			app.user.on('saveteams', function(){
+        this.updateTeams;
+        var $searchForm = $('.mainmenu button.big').closest('form');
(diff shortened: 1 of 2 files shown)


Zarel:
#518 renders this irrelevant, doesn't it?

PR #533, opened 2015-11-01
Prevent move tooltips from covering the buttons

Marty-D:
Clicking moves with long descriptions was next to impossible on mobile
devices after 1114cb3, so the short move descriptions are now used
instead on smaller resolutions.

*Edit*: If https://github.com/Zarel/Pokemon-Showdown-Client/pull/534 allows for clicking through tooltips on touch screens, feel free to close this since it would be unnecessary.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1065,7 +1065,12 @@
 					"sound": "Has no effect on Pokemon with the Ability Soundproof."
 				};
 				if (move.desc) {
-					text += '<p class="section">' + move.desc;
+					text += '<p class="section">';
+					if (move.shortDesc && ($(window).width() < 640 || $(window).height() < 580)) {
+						text += move.shortDesc;
+					} else {
+						text += move.desc;
+					}
 					for (var i in move.flags) {
 						if (i === 'distance' && move.target !== 'any') continue;
 						text += " " + flags[i];


Zarel:
Many long descriptions offer useful details. How many of them are so long they don't fit on phones? I'd rather just rewrite those to be shorter.

Marty-D:
I didn't really want to resort to this either, since there should be no reason to remove information, but I couldn't think of anything else. No one reads the move tooltips anyway, or we'd get less bug reports about things that are clearly explained in them.

I have no idea how many are too long, because [it varies by device](https://scontent-cdg2-1.xx.fbcdn.net/hphotos-xat1/v/t1.0-0/s600x600/12195924_10206367001512020_1821772023203406977_n.jpg?oh=cb77e57342fbe1e50b49c9ccb60d1764&oe=56C25811). In that image, we'd have to remove nearly half of Substitute's description just so it could fit above the button cleanly, which is totally unreasonable, and frankly a waste of time to find and fix other long cases beyond Pledge moves, Mirror Move, and whatever else. Considering the size of that screen, it would probably be any move that doesn't have a completely basic mechanic.

What's the problem with #534 anyway? On phones the issue isn't readability, because the tooltip doesn't actually appear unless you hold the button or are attempting to click something that the tooltip would cover. If that PR allows a single click to go through a tooltip that would otherwise cover the button, that's probably the best solution to this. You can always click anywhere else to remove the tooltip if you held it down anyway whenever you want to see the move buttons more clearly again.

How many people do you think are actually aware of the tooltip features on mobile, considering we've never had anyone complain about the top part of them being cut off before? No one seems to care unless it's hindering their freedom to click on things.

CrestfallPS:
@Zarel just today since urkerabs change Help room had over 6 reports. All issues only with the Substitute description. I highly doubt pple dedicated enough to play on mobile don't know how Sub works. Theyd rather have the move be clickable than care about a lenghty detailed description.

Zarel:
> Theyd rather have the move be clickable than care about a lenghty detailed description.

They were already clickable as of my first comment on this pullreq; I reverted urkerab's changes.

> What's the problem with #534 anyway? On phones the issue isn't readability, because the tooltip doesn't actually appear unless you hold the button or are attempting to click something that the tooltip would cover.

Makes it hard to see what you're doing if the area you're doing things in is covered. It's just bad UX in general. It's not like people are going to read the tooltip either way. Pushing the tooltip offscreen is probably better overall, but detecting when it overflows and replacing the description with the shortDesc would probably be even better, if that were an option.

Zarel:
Either way, I don't think this approach is right, either, so closing.

PR #534, opened 2015-11-01
Tooltips should be ignored by the mouse

urkerab:
I've noticed that when a tooltip gets very big then the mouse can get confused if it overlaps the trigger area. By using this slightly non-standard but apparently well supported style we can cause the tooltips to be ignored allowing you to mouse over (and indeed click through if necessary) the entire trigger area even if the tooltip overlaps it.
This may also improve the behaviour on touch screens but I don't have one to try it with.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -1864,6 +1864,7 @@ a.ilink.yours {
 	left: 100px;
 	text-align: left;
 	color: black;
+	pointer-events: none;
 }
 #tooltipwrapper .tooltipinner {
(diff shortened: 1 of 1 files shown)


Zarel:
This addresses one problem, but not really the other problems such as the readability issue when the tooltip covers the button. See my commit comment on 1114cb3.

Zarel:
In theory, there's nothing wrong with doing this... I'm going to merge this and probably regret it later.

PR #536, opened 2015-11-03
Flip tooltips that don't fit above the element

urkerab:
Unlike PR #525 this seeks to flip the tooltip to appear below the element being hovered thus preventing it from interfering with the element itself. Doesn't attempt to solve the problem of the tooltip for Substitute being larger than the screen on a mobile phone.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1234,6 +1234,15 @@
 				break;
 			}
 			$('#tooltipwrapper').html(text).appendTo(document.body);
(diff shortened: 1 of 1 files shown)


Zarel:
What problem is this intended to solve? Can you give a screenshot?

urkerab:
pr536
The top half shows the current placement of the tooltip. The bottom half shows where it would get flipped to. As you can see the flip avoids the tooltip trigger area.

PR #540, opened 2015-11-05
Add /ignore to the user popup

sirDonovan:
(Updated)

Add options to use <code>/ignore</code> and <code>/rank</code> from the user popup. <code>/rank</code> would mostly be used in battles (I have a related commit planned for when this PR is finalized).

Screenshots: <a href="http://i.imgur.com/Zwgyp22.png" target="_blank">1</a>, <a href="http://i.imgur.com/b0EKXa3.png" target="_blank">2</a>, <a href="http://i.imgur.com/rTFBgeP.png" target="_blank">3</a>

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -2445,11 +2445,13 @@
 			}
 			buf += '</div>';
 
+			buf += '<p class="buttonbar">';
 			if (userid === app.user.get('userid') || !app.user.get('named')) {
-				buf += '<p class="buttonbar"><button disabled>Challenge</button> <button disabled>Chat</button></p>';
+				buf += '<button disabled>Challenge</button> <button disabled>Chat</button>';
 			} else {
-				buf += '<p class="buttonbar"><button name="challenge">Challenge</button> <button name="pm">Chat</button></p>';
+				buf += '<button name="challenge">Challenge</button> <button name="pm">Chat</button> <button name="userOptions">\u2026</button>';
 			}
+			buf += '</p>';
 
 			this.$el.html(buf);
 		},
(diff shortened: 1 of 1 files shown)


Zarel:
Can you just make it a dropdown?

sirDonovan:
Updated.

Slayer95:
Hmm... I don't think that UI is any good :/

Zarel:
We have dropdowns for rearranging pokemon in teams, for selecting formats, and for selecting teams. I sort of expected it to be modeled off that.

Zarel:
On second thought, this isn't a bad idea, either...

Zarel:
I don't like the "Rank" button, though. Clicking on the username already shows a user page with ranks.

Zarel:
Let's use this character: `⋯` for a "More" menu.

js/client.js
```diff
@@ -2475,10 +2478,60 @@
 			this.close();
 			app.focusRoom('');
 			app.rooms[''].focusPM(this.data.name);
+		},
+		toggleOptions: function () {
+			this.moreOptions = !this.moreOptions;
+			if (this.moreOptions) {
```
Zarel: This is redundant. `app.addPopup` will already close a popup if it's already open (see how the top-right gear menu is implemented). You don't need to keep track of `this.moreOptions` at all.

sirDonovan:
<code>⋯</code> is slightly too big for the button (<a href="http://i.imgur.com/rGAs7PZ.png" target="_blank">screenshot</a>). If that's not acceptable, maybe we can use &hellip; or the gear icon like in the top-right menu?

My idea for the "Rank" button was to serve the requests for an easier way to see your opponents' ranks in battles (a future commit would make the names and/or avatars in the side panels clickable). Being able to see them directly in the battle chat, as well as in the specific tiers, would be more convenient than going to another page.

Zarel:
`fa-caret-down` is an option, although I think I like `...`. Something to make it clear that it's for "etc" "less important options".

I think seeing rank inline is only slightly easier than opening a new page. I'd rather not add clutter for it.

js/client.js
```diff
@@ -2475,10 +2477,54 @@
...
+					app.addPopupMessage(buf);
+					return this.update();
+				}
+				room.add(buf);
+			}
+			app.dismissPopups();
+		},
+		rank: function () {
+			var room = (app.curRoom && app.curRoom.parseCommand ? app.curRoom : app.curSideRoom);
+			if (!room || !room.parseCommand) return app.addPopupMessage("You must join a room to use this command.");
```
Zarel: See, this is just ugly, clicking on the username doesn't have weird gotchas like this.

sirDonovan:
I updated the button to <code>...</code> and removed /rank. Screenshots: <a href="http://i.imgur.com/Myv6wNq.png" target="_blank">1</a>, <a href="http://i.imgur.com/WZkfH8g.png" target="_blank">2</a>

PR #541, opened 2015-11-05
Pass roomid when adding TeamPopup from TournamentBox

augmt:
Addresses [this issue](https://github.com/Zarel/Pokemon-Showdown-Client/pull/518#discussion_r43911754).

### DIFF
--- a/js/client-chat-tournament.js
+++ b/js/client-chat-tournament.js
@@ -763,7 +763,7 @@
 		};
 
 		TournamentBox.prototype.teamSelect = function (team, button) {
-			app.addPopup(TeamPopup, {team: team, format: this.info.format, sourceEl: button});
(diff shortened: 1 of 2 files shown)


Zarel:
This somehow seems wrong to me.

Zarel:
`curTeamIndex` in general seems like a pretty ugly hack

Zarel:
I suppose I got nothin' for this, though.

Zarel:
Okay, so I think the best way is to drop the `curTeamIndex` stuff entirely. The value of the team selection dropdown button in the DOM is the canonical team index.

When updating team locations (deleting teams, inserting teams at the top, rearranging teams), we instead just use jQuery to find all team selection dropdown buttons and update their values and contents accordingly. If any of the dropdowns are open when this happens (in theory impossible so feel free to neglect this case), we can force-close them.

PR #543, opened 2015-11-07
Make user popups use the usernames color for names

panpawn:
This makes the user popups less generic and more unique imo

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -2418,7 +2418,7 @@
 
 			var buf = '<div class="userdetails">';
 			if (avatar) buf += '<img class="trainersprite' + (userid === ownUserid ? ' yours' : '') + '" src="' + Tools.resolveAvatar(avatar) + '" />';
-			buf += '<strong><a href="//pokemonshowdown.com/users/' + userid + '" target="_blank">' + Tools.escapeHTML(name) + '</a></strong><br />';
+			buf += '<strong><a href="//pokemonshowdown.com/users/' + userid + '" target="_blank" title="(profile)" style="' + hashColor(toUserid(name)) + '">' + Tools.escapeHTML(name) + '</a></strong><br />';
 			buf += '<small>' + (group || '&nbsp;') + '</small>';
 			if (data.rooms) {
 				var battlebuf = '';


panpawn:
<a href="http://prntscr.com/8zxg29">Preview screen cap</a>

Zarel:
I don't know if this looks better. What are everyone else's thoughts?

Zarel:
It also no longer goes purple for visited userpages.

Slayer95:
I was not a fan of the previous link style.
I think this would be an improvement, if we got rid of the underlining as well.
Of course, then its being a link would be pretty much an easter egg, but that's probably a feature only relevant to Staff. Marking visited user pages, in that line, shouldn't be needed imo.

sirDonovan:
I agree with @Slayer95.

panpawn:
My followup question would be: how often does the general user use a users rank link in a user popup?  If it's commonly used, then I think not underlying it would take away from that feature as not as many <i>new</i> users would recognize that as a feature.  If it is not commonly used, then perhaps we should remove the underline.

Zarel:
> Of course, then its being a link would be pretty much an easter egg, but that's probably a feature only relevant to Staff. Marking visited user pages, in that line, shouldn't be needed imo.

The username being a link is one of the easiest ways to see a user's ladder rank, which is relevant to non-staff.

AustinXII:
I agree with @Zarel, people also use the link constantly to see when they registered which is very relevant to normal users since it's a good way to see when they will be autoconfirmed. Changing the underline could also cause a lot of confusion which may affect the help room.

Slayer95:
Do they? "One week from your register date" should be a pretty straightforward calculation for new users. It's not like the user page displays exactly ``"Will be autoconfirmed at DD-MM-YY"``.

Regarding ladder ranks, it's to be noted that they only show the main server ranks. Not sure how to take that, but it's something to keep in mind.

If you still assert that this is a useful feature (do you have some ``referer`` data, @Zarel), then I am all for keeping status quo.

AustinXII:
@Slayer95 you would think so, but people ask a lot in the help chat about autoconfirmed and how much longer until they get it.

panpawn:
I think that the underline is being overthought.  Right now, there isn't a rank button, so being able to get a users rank by clicking the users username is now a familiar feature, so wether there is or is not statistical data on how often and by what group of users that feature gets used by, I don't think removing it is needed or the UI would benefit from it, in my opinion.

panpawn:
bump

Zarel:
I'm going to say no. Blue + underline make it clear that it's a link. And the username you click on to open the popup is already colored.

panpawn:
@Zarel - I understand that your main concern is that it will be unclear that the name is still a link, so since this topic has been discussed more, would it be better if the link had a mouseover that said "Click to see [user]'s rating / registered date" or something along those lines, to make it clearer that it's a link still?  (If we were to make these changes along this line)

Zarel:
Well, I'd make the mouseover just say "(profile)" or something, but yes.

panpawn:
And should the underline stay?

panpawn:
Updated to keep the underline and have a mouseover that says "(profile)".  If the underline is to go, I will re-update this.

Zarel:
Let's see what people think of this.

panpawn:
Now that colors have been changed since this, would this be something to re-consider (a second time)?

panpawn:
I remember the issue with this was with _really really_ light names, and names have been darkened iirc

panpawn:
Also, idea: what if we made the username a button, like the other click-able things on the usercard.  This would make it clear that it does something :o?

Zarel:
There were a lot of issues, but in general people just didn't like them and thought they were confusing.

panpawn:
So, is that a no to the button idea?

Zarel:
I think it would look ugly.

PR #544, opened 2015-11-07
Allow import a new pokemon to the team before selecting one

lucasfcosta:
When selecting which pokemon to add to the team it was not possible to import a pokemon from text.

How it was before:
Before

How it is now:
After

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -589,10 +589,11 @@
 			var buf = '<li value="' + i + '">';
 			if (!set.species) {
 				if (this.deletedSet) {
-					buf += '<div class="setmenu setmenu-left"><button name="undeleteSet"><i class="fa fa-undo"></i> Undo Delete</button></div>';
+					buf += '<div class="setmenu"><button name="undeleteSet"><i class="fa fa-undo"></i> Undo Delete</button><button name="importSet"><i class="fa fa-upload"></i>Import</button></div>';
+				} else {
+					buf += '<div class="setmenu"><button name="importSet"><i class="fa fa-upload"></i>Import</button></div>';
 				}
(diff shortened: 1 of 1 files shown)


js/client-teambuilder.js
```diff
@@ -592,6 +592,7 @@
 					buf += '<div class="setmenu setmenu-left"><button name="undeleteSet"><i class="fa fa-undo"></i> Undo Delete</button></div>';
 				}
 				buf += '<div class="setchart"><div class="setcol setcol-icon" style="background-image:url(' + Tools.resourcePrefix + 'sprites/bw/0.png);"><span class="itemicon"></span><div class="setcell setcell-pokemon"><label>Pokemon</label><input type="text" name="pokemon" class="chartinput" value="" /></div></div></div>';
+				buf += '<div class="setmenu"><button name="importSet"><i class="fa fa-upload"></i>Import</button></div>';
```
Zarel: This conflicts with the `setmenu` for deletedSet.

Which you may have been more likely to notice if you put the setmenu at the top of the `li` where it usually is.

lucasfcosta: I just tested this here and it seems the `setmenu` for deletedSet always gets displayed on the left, while the import is always displayed on the right.

I may have misunderstood what you said, but I wasn't able to spot any bugs with this.

Zarel: Oh, my bad. Still, though, this would be much more readable if you put it in the same setmenu, instead of creating a new setmenu for it.

Zarel: Oh, wait, they're differently-aligned setmenus. Still, putting them next to each other is probably ideal.

lucasfcosta:
As you wish, @Zarel, they're side to side now.
Side to side buttons

Zarel:
I didn't mean it like that...

Zarel:
I meant in the code, like this: 13d69757a3a

lucasfcosta:
@Zarel , ah, ok then, srry

Issue #412, opened 2015-07-05
Filter list of battles by rating


Ketsuban:
I like watching other people battle on Showdown, but I tire of picking a battle at random from the hundred or more going on in OU at any given moment only to be greeted with noncompetitive players or people who still think Electivire is worth using. Would it be possible to provide a way to filter by rating so I can limit myself to watching players who actually want to be competent?

Zarel:
Yes, this is planned.

PR #547, opened 2015-11-11
Implement Filter by Minimum elo - Untested

18vshanmugam:
Added the feature requested in #412. As of now, is untested because I can't test around the Same-origin popup closing the BattleListPopup when I try to retrieve elo from the ladder.

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -944,6 +944,7 @@
 				if (curSection) buf += '</optgroup>';
 			}
 			buf += '</select></p>';
+			buf += '<p><label>Minimum Elo Ranking:</label><textarea cols="4" rows="1" style = "vertical-align:bottom" class="elo"></textarea><button class="changeelo" name="changeelo">Set</button></p>'
 			buf += '<div class="list"><p>Loading...</p></div>';
 			buf += '</div>';
 			this.$el.html(buf);
@@ -956,9 +957,11 @@
 		},
 		events: {
 			'click .ilink': 'clickLink',
-			'change select': 'changeFormat'
+			'change select': 'changeFormat',
+			'click .changeelo': 'setMinElo'
 		},
 		format: '',
+		minElo: 0,
 		changeFormat: function (e) {
 			this.format = e.currentTarget.value;
 			this.update();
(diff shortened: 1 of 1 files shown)


Slayer95:
This should be implemented server-side...

18vshanmugam:
(https://cloud.githubusercontent.com/assets/9061831/11081564/c23023da-87eb-11e5-8822-5989a2e621ed.PNG)

Sorry, just figured it would be a client side thing because the issue was posted here, but on second thought it does look pretty server intensive

Slayer95:
Maybe an slider? The lower bound is always ``1000``. The upper bound could be by default ``2000``, and possibly customizable with a ``/query range`` command or something like that.

18vshanmugam:
hmmm, im not that experienced in js/animations beyond the basics, I probably wouldn't be able to do that... sorry about that

Slayer95:
You may check the implementation of the sounds menu: https://github.com/Zarel/Pokemon-Showdown-Client/blob/900809631caee92c4efd483c823ab189977b30eb/js/client.js#L2731

I'd suggest using steps of 200 Elo. The server could also classify battles according to 200-Elo brackets, and retrieve the requested room list cheaply.

18vshanmugam:
alright ill look at that, thanks so much!

Zarel:
Why slider? Sliders (and textboxes for that matter) are bad UX. Just have "sort by Elo" and "sort by time". Shouldn't be too much of a problem in practice.

Slayer95:
I find it better pragmatically.

If we actually sorted the battles by rating, what will happen is that everyone will flock to the highest rated battle. Grouping the battles by rating brackets, on the other hand, allows more variety and decreases the audience pressure exerted on high ladderers. This approach is furthermore kinder on the server.

18vshanmugam:
For now, I implemented it as a slider  sliderminelo. (I'm planning on adding a message telling the user to refresh if the minElo doesn't equal the slider's value) I figured for actually sorting/filtering out matches by elo, it would be helpful for there to be another server message that adds the elos of each match to the roomdata; would that be the correct way to implement it?

AustinXII:
@18vshanmugam you should make the limit higher than 2000 ELO since random battles already surpasses that and a few other tiers are getting close such as OU and AG

Slayer95:
Consider that setting the bar at 2000 will imply that the user will only see battles between idarki, Ocredin, and iSmooths (Random Battle). That's... not very useful.
2000 is an appropriate default, and it can be extended in the future with data provided by the server.

AustinXII:
I suppose, although that number of people will only increase especially since that there are multiple people almost 2000+.

Zarel:
> If we actually sorted the battles by rating, what will happen is that everyone will flock to the highest rated battle.

This is true, but I'd rather not give a bad UI just for that reason...

Slayer95:
Ok, so... Most of the concerns I have seen with regards to sliders are focused in so-called content carrousels.

There is some banner blindness somehow put forward too, which shouldn't be an issue with sensible defaults. Also people who want to use this feature will certainly not be blind to it.

Now, regarding numeric sliders, I've found this quote, which probably explains why you see it as bad design

> A slider is a good choice when you know that users think of the value as a relative quantity, not a numeric value. For example, users think about setting their audio volume to low or medium—not about setting the value to 2 or 5. — Microsoft

In this case, however, the brackets-based input is effectively a relative quantity. If we furthermore hidden the number and displayed ``Min rating: Amateur / Medium / High / Very High``, I think that would be a great compromise.

Brackets of 250 Elo instead by default I suppose with bounds and width customizable by the server. Set "Medium" as the default. "Amateur" is the current available feature, and the proposed "Sort by Date" (I'm sure you can think of a better word).
And if High / Very High lists are empty, fall back to the next lower with some visual notice.

Zarel:
I would accept radio buttons as a compromise. Sliders just suck in nearly all cases.

Zarel:
Anyway, this isn't the right way to implement anyway, so I'm closing. I'll implement something for this use-case myself later.

PR #548, opened 2015-11-11
Past generation adaptation for the teambuilder

Joimer:
The following set of changes update several parts of the teambuilder to allow it to give a better experience to past generation players.

First, Teambuilder is now able to get the generation from the format name, given they must follow the genN scheme.

Then, we are going to hide irrelevant stats from past gens, hiding ability, hiding item in gen 1, and last but not least, we are going to not to show future Pokémon in past gens.

This is just the first step of a fully functional multi-generation teambuilder.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -621,14 +621,16 @@
 				'N': '&mdash;'
 			};
 			buf += '<span class="detailcell detailcell-first"><label>Level</label>' + (set.level || 100) + '</span>';
-			buf += '<span class="detailcell"><label>Gender</label>' + GenderChart[template.gender || set.gender || 'N'] + '</span>';
-			buf += '<span class="detailcell"><label>Happiness</label>' + (typeof set.happiness === 'number' ? set.happiness : 255) + '</span>';
-			buf += '<span class="detailcell"><label>Shiny</label>' + (set.shiny ? 'Yes' : 'No') + '</span>';
+			if (this.curTeam.gen > 1) {
+				buf += '<span class="detailcell"><label>Gender</label>' + GenderChart[template.gender || set.gender || 'N'] + '</span>';
+				buf += '<span class="detailcell"><label>Happiness</label>' + (typeof set.happiness === 'number' ? set.happiness : 255) + '</span>';
+				buf += '<span class="detailcell"><label>Shiny</label>' + (set.shiny ? 'Yes' : 'No') + '</span>';
+			}
 
 			buf += '</button></div>';
 			buf += '</div><div class="setrow">';
-			buf += '<div class="setcell setcell-item"><label>Item</label><input type="text" name="item" class="chartinput" value="' + Tools.escapeHTML(set.item) + '" /></div>';
-			buf += '<div class="setcell setcell-ability"><label>Ability</label><input type="text" name="ability" class="chartinput" value="' + Tools.escapeHTML(set.ability) + '" /></div>';
+			if (this.curTeam.gen > 1) buf += '<div class="setcell setcell-item"><label>Item</label><input type="text" name="item" class="chartinput" value="' + Tools.escapeHTML(set.item) + '" /></div>';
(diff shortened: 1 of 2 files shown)


Slayer95:
> and last but not least, we are going to not to show future Pokémon in past gens.

Can you add an exception for Custom Game?

Slayer95:
Also I don't think this plays well with Gen-NEXT

Joimer:
I can add an exception for custom gens, yeah, even though I still don't get that.
Gen next defaults to gen 6.

Joimer:
I've fixed the tree and added more and better changes. Adding custom game exception is too much of work for something that's going to be used... never, so I suggest we take the same route as we do with level 1000: Text edit the team to get future gen mons in your past gen games.

TheImmortal:
You can't select Custom Game as your format either as it isn't a Ladder. You'd have to name it via Import/Export. I agree with not adding an exception to it. I also don't get why future Pokemon can be used in it.

js/client-teambuilder.js
```diff
@@ -1486,28 +1508,30 @@
 			var set = this.curSet;
 			var template = Tools.getTemplate(set.species);
 			if (!set) return;
-			buf += '<h3>Details</h3>';
-			buf += '<form class="detailsform">';
+				buf += '<h3>Details</h3>';
```
Zarel: Indentation?

Zarel:
Incidentally, I was going to do a scratch-rewrite of utilichart.js soon.

js/client-teambuilder.js
Joimer: Uh, weird. Fixed.

Joimer:
That's nice, it's been a bit confusing at first to work with this.

Zarel:
Yeah, it massively sucks. To be exact, I was going to replace it with `search.js`.

Joimer:
Cool. We can still get this in in the meantime though, right?

Zarel:
Yeah, sure.

Zarel:
Are you going to handle the by-gen data loader? You should probably run its architecture by me first.

Zarel:
I have an architecture in mind and I'm going to reject any other architecture that isn't superior.

Joimer:
Yeah, sure. I have to think about it, as it is the next step for old gens in the client.

Slayer95:
I'd suggest taking advantage of ``prefetch``, which is currently in Chrome Beta channel iirc. idk of its support in other browsers.

Joimer:
Well, my main idea was to have in client precached old gen data, only for the relevant changes, as that will never change. We can lazy load it when it's needed.

Zarel:
Lazy loading is a good idea, although if we can store it in a lightweight way, I wouldn't mind just loading it all at once.

Zarel:
Storing it in arrays was an interesting idea... although, seriously, let's move to `search.js` before doing anything _too_ fancy.

Joimer:
Well, formats data (for ordering the tiers) might get a bit heavy, as it's a map with hundreds of entries. As for moving to search.js, will you do it, or should any of us move the utilichart changes there and move it to teambuilder first?

btw Doubles formats data would be very cool to have for the teambuilder as well.

Zarel:
Exactly, that's why I'm saying to take tiers out of formats data. We can autogenerate tiers as arrays.

Joimer:
On both server and client or just client for ease of access and less memory usage?

Zarel:
Just client.

Zarel:
I was actually going to compile pretty much all of `data` on client.

Joimer:
That would be really useful, tbh. It would also make it easier to manage it for several parts of the client.

Zarel:
Really, `JSON.stringify` already removes functions, which in theory should be plenty of savings in, for instance, `moves.js`.

Then we can realize further gains if we filter out other things the client doesn't use.

Slayer95:
FWIW, you may extract the stuff the client uses from here
http://hastebin.com/exukinigof.tex

Never got around making it a standalone tool. It's intended to be run from an *offline* PS server (it will mutate all the data in ``Tools``).

PR #549, opened 2015-11-11
Fix HP display for fainted Pokemon

urkerab:
When choosing an ally to switch to you get a tooltip for a fainted Pokémon which shows the HP as `NaN%`. This replaces that with `(fainted)`, which resembles the tooltip for the sidebar.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1122,10 +1122,14 @@
 					text += 'Types unknown';
 				}
 				text += '</h2>';
-				var exacthp = '';
-				if (pokemon.maxhp != 100 && pokemon.maxhp != 1000 && pokemon.maxhp != 48) exacthp = ' (' + pokemon.hp + '/' + pokemon.maxhp + ')';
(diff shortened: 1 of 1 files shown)


js/client-battle.js
```diff
@@ -1122,10 +1122,13 @@
 					text += 'Types unknown';
 				}
 				text += '</h2>';
-				var exacthp = '';
-				if (pokemon.maxhp != 100 && pokemon.maxhp != 1000 && pokemon.maxhp != 48) exacthp = ' (' + pokemon.hp + '/' + pokemon.maxhp + ')';
-				if (pokemon.maxhp == 48 && isActive) exacthp = ' <small>(' + pokemon.hp + '/' + pokemon.maxhp + ' pixels)</small>';
-				text += '<p>HP: ' + pokemon.hpDisplay() + exacthp + (pokemon.status ? ' <span class="status ' + pokemon.status + '">' + pokemon.status.toUpperCase() + '</span>' : '') + '</p>';
+				if (pokemon.fainted) text += '<p>HP: (fainted)</p>';
+				else {
```
Zarel: we require `} else {`.

Zarel: i.e. if there's an `else` block, there must be an `if` block.

urkerab:
Done.

PR #550, opened 2015-11-12
Teambuilder should ignore EV limits prior to Generation 3

urkerab:
I had to move the place where `ignoreEVLimits` was set so that it would pick up changes in the format correctly. I also fixed a bug whereby if you manually delete an EV value you can then slide any of the other EVs as much as you like.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -109,6 +109,7 @@
 		update: function () {
 			teams = Storage.teams;
 			if (this.curTeam) {
+				this.ignoreEVLimits = (this.curTeam.gen < 3);
 				if (this.curSet) {
 					return this.updateSetView();
 				}
@@ -1415,7 +1416,7 @@
 
 				// cap
 				if (val > 252) val = 252;
-				if (val < 0) val = 0;
+				if (val < 0 || isNaN(val)) val = 0;
 
 				if (set.evs[stat] !== val || natureChange) {
 					set.evs[stat] = val;
@@ -2149,7 +2150,6 @@
 				}
 			}
 
-			this.ignoreEVLimits = (this.curTeam.format === 'classichackmons');
 			if (this.curTeam && this.ignoreEVLimits) {
 				evs = {hp:252, atk:252, def:252, spa:252, spd:252, spe:252};
 				if (hasMove['gyroball'] || hasMove['trickroom']) delete evs.spe;
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1219,9 +1219,14 @@
 				delete guessedEVs.minusStat;
 				buf += ' </small><button name="setStatFormGuesses">' + role + ': ';
 				for (var i in guessedEVs) {
-					if (guessedEVs[i]) buf += '' + guessedEVs[i] + ' ' + BattleStatNames[i] + ' / ';
+					if (guessedEVs[i]) {
+						var statName = this.curTeam.gen === 1 && i === 'spa' ? 'Spc' : BattleStatNames[i];
+						buf += '' + guessedEVs[i] + ' ' + statName + ' / ';
+					}
 				}
-				buf += ' (+' + BattleStatNames[guessedPlus] + ', -' + BattleStatNames[guessedMinus] + ')</button><small> (<a target="_blank" href="' + this.smogdexLink(template) + '">Smogon&nbsp;analysis</a>)</small></p>';
+				if (guessedPlus && guessedMinus) buf += ' (+' + BattleStatNames[guessedPlus] + ', -' + BattleStatNames[guessedMinus] + ')';
+				else buf = buf.slice(0, -3);
+				buf += '</button><small> (<a target="_blank" href="' + this.smogdexLink(template) + '">Smogon&nbsp;analysis</a>)</small></p>';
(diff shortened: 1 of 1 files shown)


urkerab:
Now also adding EV "suggestions" for Generations 1 and 2.

js/client-teambuilder.js
```diff
@@ -109,6 +109,7 @@
 		update: function () {
 			teams = Storage.teams;
 			if (this.curTeam) {
+				this.ignoreEVLimits = (this.curTeam.gen < 3) || (this.curTeam.format === 'classichackmons');
```
TheImmortal: You can remove the Classic Hackmons as it hasn't existed for a very long time, and won't be available in the future.

urkerab: Do we ignore EV limits anywhere else e.g. Custom Game?

TheImmortal: Well Custom Game isn't a ladder so you can't select it as your format in the teambuilder (unless you do it in import/export). To answer the question, I believe Custom Game does ignore EV limits.

urkerab: Ah, so no teambuilder formats then. I've updated the branch, assuming `git rebase` hasn't destroyed anything again.

Zarel:
I don't understand what all this is. Shouldn't this be a one-liner? If gen's 1 or 2, return maxed EVs?

js/client-teambuilder.js
```diff
@@ -109,6 +109,7 @@
 		update: function () {
 			teams = Storage.teams;
 			if (this.curTeam) {
+				this.ignoreEVLimits = (this.curTeam.gen < 3);
 				if (this.curSet) {
```
urkerab: This is the "one line" for the "ignore EV limits" part of the PR.

js/client-teambuilder.js
```diff
@@ -1415,7 +1421,7 @@
 
 				// cap
 				if (val > 252) val = 252;
-				if (val < 0) val = 0;
+				if (val < 0 || isNaN(val)) val = 0;
```
urkerab: This fixes a bug I noticed trying to test the one line.

js/client-teambuilder.js
```diff
@@ -2149,10 +2155,11 @@
 				}
 			}
 
-			this.ignoreEVLimits = (this.curTeam.format === 'classichackmons');
```
urkerab: This is the original position of the line, except that it doesn't actually work if you do it here.

js/client-teambuilder.js
```diff
@@ -1218,9 +1219,14 @@
...
+					if (guessedEVs[i]) {
+						var statName = this.curTeam.gen === 1 && i === 'spa' ? 'Spc' : BattleStatNames[i];
+						buf += '' + guessedEVs[i] + ' ' + statName + ' / ';
+					}
 				}
-				buf += ' (+' + BattleStatNames[guessedPlus] + ', -' + BattleStatNames[guessedMinus] + ')</button><small> (<a target="_blank" href="' + this.smogdexLink(template) + '">Smogon&nbsp;analysis</a>)</small></p>';
+				if (guessedPlus && guessedMinus) buf += ' (+' + BattleStatNames[guessedPlus] + ', -' + BattleStatNames[guessedMinus] + ')';
+				else buf = buf.slice(0, -3);
+				buf += '</button><small> (<a target="_blank" href="' + this.smogdexLink(template) + '">Smogon&nbsp;analysis</a>)</small></p>';
 				//buf += ' <small>(' + role + ' | bulk: phys ' + Math.round(this.moveCount.physicalBulk/1000) + ' + spec ' + Math.round(this.moveCount.specialBulk/1000) + ' = ' + Math.round(this.moveCount.bulk/1000) + ')</small>';
```
urkerab: This makes the prefill button display correctly.

js/client-teambuilder.js
```diff
@@ -1860,7 +1866,7 @@
 			var itemid = toId(set.item);
 			var abilityid = toId(set.ability);
 
-			if (set.moves.length < 4 && template.id !== 'unown' && template.id !== 'ditto') return '?';
+			if (set.moves.length < 4 && template.id !== 'unown' && template.id !== 'ditto' && this.curTeam.gen > 2) return '?';
```
urkerab: This allows the teambuilder to prefill the EVs without needing 4 moves.

js/client-teambuilder.js
```diff
@@ -2149,10 +2155,11 @@
 				}
 			}
 
-			this.ignoreEVLimits = (this.curTeam.format === 'classichackmons');
 			if (this.curTeam && this.ignoreEVLimits) {
 				evs = {hp:252, atk:252, def:252, spa:252, spd:252, spe:252};
 				if (hasMove['gyroball'] || hasMove['trickroom']) delete evs.spe;
+				if (this.curTeam.gen === 1) delete evs.spd;
```
urkerab: This avoids trying to prefill EVs that don't exist.

js/client-teambuilder.js
```diff
@@ -2149,10 +2155,11 @@
...
+				if (this.curTeam.gen < 3) return evs;
```
urkerab: And finally this is the one-liner for the "prefill max EVs" part of the PR.

js/client-teambuilder.js
Zarel: Okay, fair enough, but what happened to Hackmons?

PR #552, opened 2015-11-12
Automatically drag EVs appropriately for LC

urkerab:
When choosing EVs for LC (or even VGC) this makes the sliders snap to those positions that actually change the resulting stat.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1442,6 +1442,9 @@
 			var set = this.curSet;
 			if (!set) return;
 			val = +val;
+			var originalVal = val;
+			var result = this.getStat(stat, set, val);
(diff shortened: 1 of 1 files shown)


js/client-teambuilder.js
```diff
@@ -1442,6 +1442,9 @@
 			var set = this.curSet;
 			if (!set) return;
 			val = +val;
+			var originalVal = val;
+			var result = this.getStat(stat, set, val);
+			while (val && this.getStat(stat, set, val - 4) == result) val -= 4;
```
Zarel: Um, what happens if `val` is `0`?

Zarel: oh, `val &&`

PR #553, opened 2015-11-14
A C Script

Taskforce-Zero:


Taskforce-Zero:
ok

TheImmortal:
ok

xfix:
You should print new line at end of program, otherwise the prompt will appear just after the printed text. Also, you should include `<stdio.h>` in order to use I/O functions.

Please read the style guide that is yet to written because submitting a pull request again.

Zarel:
@TheImmortal, @xfix, Can we not do all these semi-joke responses?

@Taskforce-Zero, seriously, what are you trying to do here? This is the repository for the client for the Pokémon Showdown battle simulator, it's not the place to submit random C code.

xfix:
I sorta would like to have a style guide however for a client so it could be pointed to, oh well.

PR #554, opened 2015-11-16
Item + Old Gen teambuilding update

Enigami:
Fixed Speed UI glitch when modifying EVs in Gen 1.
Adding Pokemon in Gen 1/2 sets EVs to 252 and IVs to 15 instead of 0/31.
252 EVs now does not display in the stat graph for Gen 1/2, when EVs are 0 they display (since EVs in the GB games are usually 252, with 0 being the exception)
Hidden Power now sets proper IVs for Gen1/2.
Pokemon sprite loads by generation, if there is no image it loads a (?) image.
Moves are now categorized Physical/Special by type in Gen1/2/3.
Items are now separated by Usable Item and Item like Moves, with 'smart' detection of certain situational items.

Perhaps EVs/IVs for Gen 1 & 2 on the server could be changed so it no longer locks them?


Zarel:
Your commit tree is messy. What's with the pull request merging? Can you rebase on top of master and get rid of things like that?

Slayer95:
Also, don't edit stuff in the ``data`` folder with server-side equivalents. They should be modified in Zarel/Pokemon-Showdown instead

PR #555, opened 2015-11-16
Fixed Item & Old-gen teambuilding upgrades + UI fixes

Enigami:
Fixed Speed UI glitch when modifying EVs in Gen 1.
Adding Pokemon in Gen 1/2 sets EVs to 252 and IVs to 15 instead of 0/31.
252 EVs now does not display in the stat graph for Gen 1/2, when EVs are 0 they display (since EVs in the GB games are usually 252, with 0 being the exception)
Hidden Power now sets proper IVs for Gen1/2.
Pokemon sprite loads by generation, if there is no image it loads a (?) image.
Moves are now categorized Physical/Special by type in Gen1/2/3.
Items are now separated by 'Items' and 'Uncompetitive Items', with 'smart' detection of certain situational items.

Edit:
Forgot to add a couple things, made a couple tweaks to Suggested Spread for Gen 1 & 2, and removed 'Remaining EVs' UI for Gen 1 & 2, since there is no purpose for it.

(database changes moved to server pull request)

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -1112,8 +1112,29 @@ var Tools = {
 		return 'background:transparent url(' + Tools.resourcePrefix + 'sprites/bwicons-sheet.png?g6) no-repeat scroll -' + left + 'px -' + top + 'px' + fainted;
 	},
 
-	getTeambuilderSprite: function (pokemon) {
+	getTeambuilderSprite: function (pokemon, generation) {
 		if (!pokemon) return '';
+		if (!generation) var generation = 6;
+		switch (generation) {
+			default:
+				var gen = 'bw';
+				break;
+			case 1:
+				var gen = 'rby';
+				break;
+			case 2:
+				var gen = 'gsc';
+				break;
+			case 3:
+				var gen = 'rse';
+				break;
+			case 4:
+				var gen = 'dpp';
+				break;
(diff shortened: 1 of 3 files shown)


Enigami:
Making some revisions, closing temporarily.

Enigami:
Changed 'Usable Items/Items' to 'Items/Uncompetitive Items' so that it doesn't look like 'advice', and added more 'smart' conditional item checking.

Slayer95:
I think I'd prefer a blatant advicing over a seemingly arbitrary classification. Thoughts?

Zarel:
Please don't create new pull requests when we ask you to make changes. Do you know how to rebase?

Enigami:
The idea was to make it easier to pick out items when teambuilding, but everyone seems to have a problem with it. Maybe instead I should change it so that items that are even remotely viable are left alone, and only separate items with no competitive purpose?

@Zarel Yeah, I got it figured out so that's not going to happen again.

Zarel:
I think "Usable Items" is a good approach to get the completely worthless items out of the way for a shorter list to scroll through. Like what we do with moves.

Zarel:
This is fixed by a wide variety of teambuilder changes, including `3621eb45`.

PR #556, opened 2015-11-18
Allow data URIs in <img> tags

bumbadadabum:


js/battledata.js
```diff
@@ -603,7 +603,19 @@ var Tools = {
 					}
 				}
 			}
+			var dataUri = '';
+			if (tagName === 'img') {
+				for (var i = 0; i < attribs.length - 1; i += 2) {
+					if (attribs[i] === 'src' && attribs[i + 1].startsWith('data:image/')) {
```
Zarel: The oldest browser we support is Safari 4, which is an ES3 browser.

`.startsWith` is ES6, which isn't supported until Safari 9.

heck, it's probably not even compatible with like

PR #559, opened 2015-11-21
Support disabling autoscroll when hovering some elements

Slayer95:
Useful for in-chat GUIs, such as the one used by polls.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -168,7 +168,9 @@
 			this.noAutoScroll = true;
 		},
 		getAutoScroll: function () {
-			if (this.noAutoScroll) return false;
+			return !this.noAutoScroll && this.getBottomScrolled();
+		},
+		getBottomScrolled: function () {
 			return (this.$chatFrame.scrollTop() + 60 >= this.$chat.height() - this.$chatFrame.height());
 		},
 		clear: function () {
@@ -953,6 +955,12 @@
 			textbox.value = value;
 			textbox.setSelectionRange(start, end);
 			return true;
+		},
+		scrollHandler: function () {
+			if (!app.focused || this !== app.curRoom && this !== app.curSideRoom || !this.getBottomScrolled()) return;
+			this.dismissNotification()
+			this.$chatFrame.off('scroll', this.scrollHandler);
+			this.scrollHandler = null;
 		}
 	});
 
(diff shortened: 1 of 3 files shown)


Zarel:
Should we re-enable autoscroll when mousing out of said elements?

Slayer95:
I guess... Why not?

js/client-chat.js
```diff
@@ -1022,8 +1036,11 @@
 				}
 			}
 			if (userlist) this.addRow(userlist);
-			if (autoscroll) {
+			if (this.getAutoScroll()) {
 				this.$chatFrame.scrollTop(this.$chat.height());
+			} else if (!this.scrollHandler) {
+				this.scrollHandler = _.bind(ConsoleRoom.scrollHandler, this);
+				this.$chatFrame.on('scroll', this.scrollHandler);
```
Zarel: Can you explain what this is doing?

Zarel: The rest makes sense to me, but this is kind of confusing. Why are you attaching an extra scroll handler here?

Slayer95: This is the only scroll handler, and it's intended to clear subtle notifications once the user scrolls down to the bottom of the chat log.

Zarel:
> > Should we re-enable autoscroll when mousing out of said elements?
>
> I guess... Why not? o.o

Sorry, I think this was a misinterpretation. I mean, should we automatically scroll to the bottom when mousing out of said elements?

js/client-chat.js
Zarel: Oh, I see. Maybe add comments to the top of `scrollHandler` describing that behavior?

Slayer95:
Oh. That would be handy for short scrolling distances.
For other circumstances, such behavior may block users' attemps to read old logs (which is a very annoying thing Facebook itself does).
Actually, even for "short distances" (<em>to be defined</em>), that would be problematic if moving the pointer upwards from a "no-autoscroll" area scrolled the chat log so that the pointer got back there. So maybe only if the user exited the element from the bottom?

js/client-chat.js
Slayer95: Fine

Zarel:
> For other circumstances, such behavior may block users' attemps to read old logs (which is a very annoying thing Facebook itself does).

We should probably have a "There are new messages" bar at the bottom of chat, instead, in these cases.

> Actually, even for "short distances" (<em>to be defined</em>), that would be problematic if moving the pointer upwards from a "no-autoscroll" area scrolled the chat log so that the pointer got back there.

I don't consider it problematic. Scrolling up should be one step (no current need to animate it), and then whether or not you're in a no-autoscroll area again afterwards isn't exactly a huge problem.

Slayer95:
> We should probably have a "There are new messages" bar at the bottom of chat, instead, in these cases.

I considered this initially, but then I thought we might as well use the subtle notifications system for this.

Zarel:
Let's do both. At the very least, an orange line at the bottom of the chat log.

Zarel:
poke?

Slayer95:
On the radar!

Zarel:
re-poke?

Slayer95:
This doesn't work on mobile.

Zarel:
I mean, it's still useful on desktop... :|

PR #561, opened 2015-11-30
Rejoining a done battle checks the wrong flag

urkerab:
To avoid the forfeit popup appearing when you rejoin a battle that you already played, `init` tries to check whether the battle already ended, however it checks the wrong flag so you get prompted anyway.


Zarel:
Good catch, but the variable should probably be renamed `.ended` anyway.

Zarel:
See e33c3b268635

PR #563, opened 2015-12-01
Add gen 6 substitute graphic.

Layell:
https://i.imgur.com/IoFY3Sz.png


Slayer95:
Hmmm ``0`` is reserved for Missingno.
Maybe we should put these in ``data/graphics.js``?

Zarel:
Yeah, `num: 0` shouldn't be used for substitute, but it's not a huge deal; we can correct it later.

PR #566, opened 2015-12-03
Disable links in code tag

MonoHearted:
This is an annoyance when arrays within arrays are used, for example:

```javascript
[[1,2,3],[1,2,3]]
```

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -365,130 +365,133 @@ var Tools = {
 
 		var options = Tools.prefs('chatformatting') || {};
 
-		// ``code``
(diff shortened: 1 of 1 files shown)


MonoHearted:
This won't work, closing for fixing

MonoHearted:
For this particular PR, I was only looking at links in code tags. Should all formatting be removed?

Italics:
```
__init__
```

Bold:
```
5**8**9
```

bumbadadabum:
Given that I suggested to remove all formatting, yes I do think that. After all, most [code] tags on forums are primarily used to display formatted text without formatting it.

PR #567, opened 2015-12-03
Remove all formatting within code tag

MonoHearted:
see #566.

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -434,134 +434,139 @@ var Tools = {
 
 		var options = Tools.prefs('chatformatting') || {};
 
-		if(!(/\`\`([^< ](?:[^<`]*?[^< ])??)\`\`/g.test(str))) {
-			// ~~strikethrough~~
-			str = str.replace(/\~\~([^< ](?:[^<]*?[^< ])??)\~\~/g,
-				options.hidestrikethrough ? '$1' : '<s>$1</s>');
-			// <<roomid>>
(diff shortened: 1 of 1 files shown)


MonoHearted:
well, this won't work as is, I'll see if I can find another way to do it.

kotarou3:
At a certain point, I think it would be better to use a proper parser rather than trying to purely regex this

ascriptmaster:
This doesn't feel like the right choice, to abort all formatting immediately. What if someone sends the following message:

    ``some code`` **also some other formatting**

Zarel:
I've been meaning to stop using regex this entire time...

kotarou3:
Could we just use a markdown parser instead? Maybe slightly modified if you still want to support the `__` for italics formatting

Joimer:
I am.
Also, don't forget to squash the commits into a single one.

Zarel:
Markdown is a poor language for end-user input; Reddit gets so many unreadable posts because of it. Github and Stack Overflow subtly change it, and they're websites for developers, which is the only reason it works.

We have a lot of PS-specific features, anyway. I don't think there's much a Markdown parser can really help with.

kotarou3:
It's something you could base the parser off, rather than rolling your own.
Keep the bits of formatting you want, remove the ones you don't, etc

Zarel:
Closing – superseded by #931

PR #568, opened 2015-12-04
Fix Tooltip dimensions

Ridaz:
https://my.mixtape.moe/lciyek.png

This is what's happening so far


Zarel:
???

Also, markdown in commit summaries?

Zarel:
But seriously, why does this happen?

Ridaz:
Looks like it was my fault the tooltip went haywire. Sorry.

PR #571, opened 2015-12-09
Added Gem-boosted powers to move description

hvk3:
Small change to client UI.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1105,7 +1105,7 @@
 
 			// cache movelist ref
 			var speciesid = toId(this.curSet.species);
-			var g6 = (this.curTeam.format && (this.curTeam.format === 'vgc2014' || this.curTeam.format === 'vgc2015'));
+			var g6 = (this.curTeam.format && this.curTeam.format === 'vgc2016');
 			this.applyMovelist(g6, speciesid);
 
 			this.$chart.html('<em>Loading ' + this.curChartType + '...</em>');
@@ -1121,7 +1121,7 @@
 		updateChartDelayed: function () {
 			// cache movelist ref
 			var speciesid = toId(this.curSet.species);
-			var g6 = (this.curTeam.format && (this.curTeam.format === 'vgc2014' || this.curTeam.format === 'vgc2015'));
+			var g6 = (this.curTeam.format && this.curTeam.format === 'vgc2016');
 			this.applyMovelist(g6, speciesid);
 
 			var self = this;


js/client-battle.js
```diff
@@ -1332,6 +1332,30 @@
...
 			var thereIsWeather = (this.battle.weather in {'sunnyday': 1, 'desolateland': 1, 'raindance': 1, 'primordialsea': 1, 'sandstorm': 1, 'hail':1});
+			if (myPokemon.item) {
+				var item = Tools.getItem(myPokemon.item).name;
+				var pos = this.choice.choices.length;
+				var moveType = this.getMoveType(move, this.battle.mySide.active[pos]);
+				var itemInfo = item.split(' ');
+				var name = move.name;
+				if (itemInfo[1] == 'Gem') {
+					if (moveType == itemInfo[0]) {
+						if (basePower) {
```
Slayer95: Use ``&&``

js/client-battle.js
```diff
@@ -1332,6 +1332,30 @@
...
+								if (this.battle.gen == 6) {
+									basePower *= 1.3;
+								}
+								else if (this.battle.gen == 5) {
+									basePower *= 1.5;
+								}
+								var temp = item.concat(')');
+								var comment = ' (Boosted by ';
+								basePowerComment = comment.concat(temp);
+							}
```
Slayer95: Use ``+`` instead of ``concat``

js/client-battle.js
```diff
@@ -1332,6 +1332,30 @@
...
+				if (itemInfo[1] == 'Gem') {
+					if (moveType == itemInfo[0]) {
+						if (basePower) {
+							if (name != 'Struggle' || name != 'Water Pledge' || name != 'Grass Pledge' || name != 'Fire Pledge' || name != 'Fling' ) {
+								if (this.battle.gen == 6) {
+									basePower *= 1.3;
+								}
+								else if (this.battle.gen == 5) {
+									basePower *= 1.5;
+								}
```
Slayer95: A more robust variant would be:
``basePower *= this.battle.gen >= 6 ? 1.3 : 1.5``

Thus handling the likely possibility of gems not changing its effect in the future, and supporting gems in DPP and below (Custom Game).

Slayer95:
@hvk3, you can run ``npm test`` in your computer before pushing to do the same checks as Travis.

Zarel:
This seems fine this time, but can you squash your commits?

js/client-battle.js
```diff
@@ -1332,6 +1332,19 @@
...
+				var item = Tools.getItem(myPokemon.item).name;
+				var pos = this.choice.choices.length;
+				var moveType = this.getMoveType(move, this.battle.mySide.active[pos]);
+				var itemInfo = item.split(' ');
+				var name = move.name;
+				if (itemInfo[1] == 'Gem' && moveType == itemInfo[0] && basePower && (name != 'Struggle' || name != 'Water Pledge' || name != 'Grass Pledge' || name != 'Fire Pledge' || name != 'Fling')) {
+					basePower *= this.battle.gen >= 6 ? 1.3 : 1.5;
+					var temp = item + ')';
+					var comment = ' (Boosted by ';
+					basePowerComment = comment + temp;
```
Zarel: This should probably just all be one line:

    basePowerComment += ' (Boosted by ' + item + ')';

js/client-battle.js
```diff
@@ -1332,6 +1332,19 @@
 			var basePower = move.basePower;
 			var basePowerComment = '';
 			var thereIsWeather = (this.battle.weather in {'sunnyday': 1, 'desolateland': 1, 'raindance': 1, 'primordialsea': 1, 'sandstorm': 1, 'hail':1});
+			if (myPokemon.item) {
+				var item = Tools.getItem(myPokemon.item).name;
+				var pos = this.choice.choices.length;
+				var moveType = this.getMoveType(move, this.battle.mySide.active[pos]);
+				var itemInfo = item.split(' ');
```
Zarel: `splitItemName` might be a better name for this variable.

js/client-battle.js
```diff
@@ -1332,6 +1332,19 @@
 			var basePower = move.basePower;
 			var basePowerComment = '';
 			var thereIsWeather = (this.battle.weather in {'sunnyday': 1, 'desolateland': 1, 'raindance': 1, 'primordialsea': 1, 'sandstorm': 1, 'hail':1});
+			if (myPokemon.item) {
+				var item = Tools.getItem(myPokemon.item).name;
```
Zarel: `itemName` might be a better name for this variable.

urkerab:
You can also detect whether an item is a Gem by using `Tools.getItem(myPokemon.item).isGem` but unfortunately there doesn't seem to be an equivalent `onGem` property for the type of the Gem (it's effectively part of the onUseItem code for each Gem).

Zarel:
You asked on PS:

> hvk3: are 4 commits fine, or should I reduce them to 1 or 2?

You can have as many commits as you want as long as they aren't separate changes, or fixes to commits in your pull request.

In this case, I don't understand why you need more than one. You're only making one change, so you only need one commit for that.

js/client-battle.js
```diff
@@ -1332,6 +1332,17 @@
...
 			var thereIsWeather = (this.battle.weather in {'sunnyday': 1, 'desolateland': 1, 'raindance': 1, 'primordialsea': 1, 'sandstorm': 1, 'hail':1});
+			if (myPokemon.item) {
+				var itemName = Tools.getItem(myPokemon.item).name;
+				var pos = this.choice.choices.length;
+				var moveType = this.getMoveType(move, this.battle.mySide.active[pos]);
+				var splitItemName = itemName.split(' ');
+				var moveName = move.name;
+				if (splitItemName[1] == 'Gem' && moveType == splitItemName[0] && basePower && (moveName != 'Struggle' || moveName != 'Water Pledge' || moveName != 'Grass Pledge' || moveName != 'Fire Pledge' || moveName != 'Fling')) {
+					basePower *= this.battle.gen >= 6 ? 1.3 : 1.5;
+					basePowerComment += ' (Boosted by ' + item + ')';
```
urkerab: Note that some moves subsequently overwrite the basePowerComment.

js/client-battle.js
```diff
@@ -1332,6 +1332,17 @@
 			var basePower = move.basePower;
 			var basePowerComment = '';
 			var thereIsWeather = (this.battle.weather in {'sunnyday': 1, 'desolateland': 1, 'raindance': 1, 'primordialsea': 1, 'sandstorm': 1, 'hail':1});
+			if (myPokemon.item) {
```
urkerab: I assume Gems should be ignored during Embargo/Magic Room?

urkerab:
So, if you're going to do Gems, why not Plates or other power-boosting items?

hvk3:
Sure thing. Let me complete this and I'll start on the remaining.

On Wed, Dec 9, 2015 at 4:16 PM, urkerab <notifications@github.com> wrote:

> So, if you're going to do Gems, why not Plates or other power-boosting
> items?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Zarel/Pokemon-Showdown-Client/pull/571#issuecomment-163185793>
> .
>



-- 
Regards
Harshvardhan

js/client-battle.js
Zarel: We should probably just refactor those to `+=` as well

hvk3: Trying it out.

Zarel:
Can you rebase out the merge commit?

Zarel:
And squash this down to one commit? With this commit summary:

`Add Gem-boosted powers to move description`

PR #574, opened 2015-12-10
Added gem-boosted powers to move description

hvk3:
A small change to ui

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1332,6 +1332,17 @@
 			var basePower = move.basePower;
 			var basePowerComment = '';
 			var thereIsWeather = (this.battle.weather in {'sunnyday': 1, 'desolateland': 1, 'raindance': 1, 'primordialsea': 1, 'sandstorm': 1, 'hail':1});
+			if (myPokemon.item && !this.battle.hasPseudoWeather('Magic Room') && (!pokemon.volatiles || !pokemon.volatiles['embargo'])) {
+				var itemName = Tools.getItem(myPokemon.item).name;
+				var pos = this.choice.choices.length;
+				var moveType = this.getMoveType(move, this.battle.mySide.active[pos]);
+				var splitItemName = itemName.split(' ');
+				var moveName = move.name;
+				if (splitItemName[1] == 'Gem' && moveType == splitItemName[0] && basePower && (moveName != 'Struggle' || moveName != 'Water Pledge' || moveName != 'Grass Pledge' || moveName != 'Fire Pledge' || moveName != 'Fling')) {
+					basePower *= this.battle.gen >= 6 ? 1.3 : 1.5;
+					basePowerComment += ' (Boosted by ' + itemName + ')';
+				}
+			}
 			if (move.id === 'acrobatics') {
 				if (!myPokemon.item) {
 					basePower *= 2;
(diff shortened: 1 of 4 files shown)


panpawn:
Next time, try to rebase instead of closing / opening a new pull request

Zarel:
I even asked you not to create a new pull request when we talked about it the other day...

hvk3:
Added support for all items

Zarel:
@Joimer, thoughts?

I think we might want this to be its own file, if it's getting to be this complicated.

Joimer:
Yeah, it's getting pretty big and that makes the file less readable, and it's already pretty big. How about a file for tooltip information processing?

Zarel:
Full support

hvk3:
What should be done next?

Marty-D:
Couldn't you just move [this block of code](https://github.com/Zarel/Pokemon-Showdown-Client/pull/574/files#diff-bca2828a0ba7901d7d7c95b095afd014R1335) to below [this line](https://github.com/Zarel/Pokemon-Showdown-Client/pull/574/files#diff-bca2828a0ba7901d7d7c95b095afd014R1913), and remove all the duplication in between? Am I missing something? This could also probably be done more simply by including the check for `basePower` in the first check instead of everywhere else, using switch statements, etc.

hvk3:
The point is that some of the moves don't have base powers initially, and were added as an exception before I started. So I had to duplicate some of the stuff.

Marty-D:
Does that matter? They will all have `basePower` defined by the time it gets to the return statement, which is where this should be done so duplication becomes unnecessary.

hvk3:
Okay, so moves with basePower can be handled as you said. There'll be some duplication for moves with ranges then, I guess.

hvk3:
Removed redundancy from code

hvk3:
Corrected electroball range issue as well

hvk3:
Moved code to client-battle-tooltips.js

hvk3:
Modifications according to Quite Quiet and E4 Flint

Zarel:
Can you give an MIT license? Post:

> I, __, hereby license all my contributions to the Pokémon Showdown project, past and future, under the MIT license.

hvk3:
I, Harshvardhan Kalra, hereby license all my contributions to the Pokémon Showdown project, past and future, under the MIT license.

js/client-battle.js
```diff
@@ -234,7 +234,7 @@
 
 			if (this.battle.ended) {
 
-				var replayDownloadButton = '<a href="//replay.pokemonshowdown.com/" class="button replayDownloadButton" style="float:right;padding:2px 6px"><i class="fa fa-download"></i> Download replay</a>';
+				var replayDownloadButton = '<span style="float:right;"><a href="//replay.pokemonshowdown.com/" class="button replayDownloadButton" style="padding:2px 6px"><i class="fa fa-download"></i> Download replay</a><br /><br /><button name="saveReplay"><i class="fa fa-upload"></i> Upload and share replay</button></span>';
```
Zarel: What's going on here? Why was this changed?

Zarel: Well, at least it merges cleanly, but ideally you should rebase on latest master for a cleaner blame history.

hvk3:
Some commit that i had not merged possibly

hvk3:
Small corrections

Slayer95:
@Zarel, the new file wasn't added to the index.
@hvk3 should be sending a follow-up PR soon, but this is a heads-up so that the client isn't left in a bad state after pulling the changes.

Slayer95:
lol I'm slow. Merged.

PR #575, opened 2015-12-11
Make menu buttons change color according to background

Joimer:
This commit introduces a new library, color thief:
https://github.com/lokesh/color-thief/
When a background image is available, we try to detect the main color of the background, and use it to
change the menu colors so they fit with the background image.

I've used a function so I had it available in both in App and the Popups, but I'm no backbone expert, so if there's a better way to do this please do tell me.


js/client.js
```diff
@@ -123,6 +123,47 @@
 		return roomid.replace(/[^a-zA-Z0-9-]+/g, '').toLowerCase();
 	};
 
+	var changeButtonColors = this.changeButtonColors = function () {
```
Slayer95: This is a new global, too

Zarel: We might as well put it in main menu...

Zarel:
I'd suggest `app.setBackground` and put the button colors there.

Zarel:
But I'm planning on massively refactoring background saving/loading, so this isn't a big deal.

.eslintrc.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
 		// Generic global variables
 		"Config": false, "Chart": false, "soundManager": false, "Storage": false, "Tools": false,
 		"app": false, "toId": false, "toRoomid": false, "toUserid": false, "toName": false, "hashColor": false, "MD5": false,
-		"ChatHistory": false, "Topbar": false, "UserList": false,
+		"ChatHistory": false, "Topbar": false, "UserList": false, "ColorThief": false,
```
Slayer95: There is a section above for library-related globals

Zarel:
I ended up rewriting most of this, but thanks for showing me the way!

211da6e8b1

PR #580, opened 2015-12-18
Avoid changing border width or padding on hover

urkerab:
Cosmetic issue whereby if you're not at 100% zoom then moving pixels from padding to border causes layout to get recalculated. In my case it causes the wastebasket icon for the "Delete" button (but not the text!) to move when the edit button is hovered at 90% zoom.

I noticed the problem with the teamlist button but I changed the other hover too just in case. (I didn't find any other hovers which changed the border or padding.)

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -1033,7 +1033,7 @@ p.or:after {
 	font-size: 8pt;
 }
 .roomlist a.ilink:hover {
-	border: 1px solid #8899AA;
+	border-color: #8899AA;
 	background: #F1F4F9;
 	color: #224466;
 	text-decoration: none;
@@ -1789,7 +1789,7 @@ a.ilink.yours {
     font-weight: bold;
 }
 .megaevo:hover {
-    border: 1px solid #888;
+    border-color: #888;
     background: #E5E5E5;
     color: black;
 }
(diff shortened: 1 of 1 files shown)


Zarel:
Yeah, I do it this way by habit because IE7 doesn't support transparent borders, but I don't think we support any browsers that have trouble with transparent borders, so this approach is much better.

style/client.css
```diff
@@ -2301,7 +2301,6 @@ a.ilink.yours {
 .setmenu button:hover,
 .teamlist button:hover {
 	border: 1px solid #888888;
```
Zarel: just do `border-color:`

style/client.css
```diff
@@ -2724,20 +2723,19 @@ a.ilink.yours {
...
 	border-radius: 4px;
-	padding: 1px 1px 1px 5px;
+	padding: 0px 0px 0px 4px;
 	margin: 0 5px 1px 5px;
 	background: transparent;
 	cursor: pointer;
 }
 .teambuilder-results .result a:hover {
-	padding: 0px 0px 0px 4px;
 	border: 1px solid #D8D8D8;
```
Zarel: ditto

style/client.css
```diff
@@ -2724,20 +2723,19 @@ a.ilink.yours {
...
 	background: #F8F8F8;
 }
 .teambuilder-results .firstresult a,
 .teambuilder-results .firstresult a:hover {
-	padding: 0px 0px 0px 4px;
 	border: 1px solid #CCCCCC;
```
Zarel: ditto

urkerab: Actually that one won't work because there's a non-hover rule.

PR #582, opened 2015-12-19
Change the tournament autostart off message

Teremiare:
Made to better match the change to setautostart in Zarel/Pokemon-Showdown#2326
Changed `timeout` to `timer` in the autodq message too to make it a bit clearer.

### DIFF
--- a/js/client-chat-tournament.js
+++ b/js/client-chat-tournament.js
@@ -310,9 +310,9 @@
 
 				case 'autodq':
 					if (data[0] === 'off') {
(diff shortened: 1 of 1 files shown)


Zarel:
This really doesn't need its own pullreq, although I guess if it has one, it's not too big of a deal. You still need to make the commit message something that matches what it does, though - "Update client-chat-tournament.js" is not an acceptable commit summary.

Teremiare:
My bad about the commit message, I thought I changed it but somehow forgot to.

PR #583, opened 2015-12-20
Properly display effectiveness of spread moves

urkerab:
The current code displays all the damage animations at once. Unfortunately damage messages come in two parts: actions for any effectiveness messages and hidden actions for the actual damage. This means that they now display out of order; first you see any effectiveness messages for all targets, then their damage messages.
The easiest way I could see to remedy this was to make the remaining hidden actions display immediately instead of collecting up for display at the end of the minor (most of them already do this). In the case of a damage minor, any previous effectiveness messages are also shown and then cleared for the next target of the spread move.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -3253,7 +3253,6 @@ var Battle = (function () {
 	};
 	Battle.prototype.runMinor = function (args, kwargs, preempt, nextArgs, nextKwargs) {
 		var actions = '';
-		var hiddenactions = '';
 		var minors = this.minorQueue;
 		if (this.multiHitMove && minors.length) {
 			var lastMinor = minors[minors.length - 1];
@@ -3315,7 +3314,7 @@ var Battle = (function () {
 						actions += "" + poke.getName() + " was hurt by poison!";
 						break;
 					case 'lifeorb':
-						hiddenactions += "" + poke.getName() + " lost some of its HP!";
+						this.message('', '<small>' + poke.getName() + ' lost some of its HP!</small>');
 						break;
 					case 'recoil':
 						actions += "" + poke.getName() + " is damaged by the recoil!";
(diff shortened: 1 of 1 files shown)


Slayer95:
``Unfortunately damage messages come in two parts: actions for any effectiveness messages and hidden actions for the actual damage``

This should be fixed imo.

Zarel:
> Unfortunately damage messages come in two parts: actions for any effectiveness messages and hidden actions for the actual damage.

What's wrong with this behavior?

> This means that they now display out of order

Why can't you fix this without changing the hidden action behavior?

> This should be fixed imo.

What's wrong with this behavior?

urkerab:
> > Unfortunately damage messages come in two parts: actions for any effectiveness messages and hidden actions for the actual damage.
>
> What's wrong with this behavior?

I'm not trying to suggest that the behaviour itself is wrong, just that the change didn't allow for it.

> >This means that they now display out of order
>
> Why can't you fix this without changing the hidden action behavior?

Sorry, I'm not sure what you're trying to ask here. Let me give you an example: Charizard uses Heat Wave on Mega Kangaskhan and Aegislash . It happens to crit against Kangaskhan, and it's also super-effective on Aegislash. Since you merged the animations together, all four minors run in the same call:

    |-crit|p1a:
    |-damage|p1a:
    |-supereffective|p1b:
    |-damage|p1b:

Processing the minors results in `actions` becoming `A critical hit! It's supereffective! ` while `hiddenActions` becomes

    Kangaskhan lost 10% of its health!
    Aegislash lost 100% of its health!

The actions and hiddenActions are then concatenated and displayed, but by this point it's too late to do anything about the mismatch of actions with their hiddenActions. This is why I changed all the hiddenActions to output immediately and not concatenate. (The actions still need to concatenate in the case of a supereffective or resisted critical hit.)

Zarel:
Okay, now I'm just confused by Slayer95's comment.

Slayer95:
I was (and still am) under the assumption that the client doesn't display the right amount of effectiveness messages in Doubles/Triples due to a less than ideal protocol.
While that specific issue is not the one here, if we could fix them all together it could we great.

Zarel:
> I was (and still am) under the assumption that the client doesn't display the right amount of effectiveness messages in Doubles/Triples due to a less than ideal protocol.

I'm not entirely sure what you mean, but I'm like 80% sure you're wrong (based on my guess of what you mean).

PR #586, opened 2015-12-24
Change the teambuilder clipboard design

Teremiare:
Moved the pokemon down and the clear button up so that it better fits with the new folders.
Current clipboard looks like this:
https://i.gyazo.com/bcf2ab5e17da803fff25c76a27db641a.png
https://i.gyazo.com/7767f32bd25633761a56aff95e8d0470.png
The changed one looks like this:
https://i.gyazo.com/36b612d75826ad821c8f65576c8abb7d.png
https://i.gyazo.com/a8ad0f372a3ce024f7f7c7d240972fc6.png

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -2584,10 +2584,11 @@ a.ilink.yours {
 
 .teambuilder-clipboard-container {
 	height: 52px;
-	max-width: 560px;
+	max-width: 455px;
 	font-size: 8pt;
 	padding: 4px;
 	position: relative;
+	margin: auto;
 
 	border: 1px solid #FFAA00;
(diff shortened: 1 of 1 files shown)


Zarel:
Looks better, but let's have both the "Clipboard" text and the "Clear clipboard" buttons at the bottom?

Teremiare:
So something like this?

Zarel:
Yeah, that looks great. How does that look with the Paste button?

Teremiare:
Made the container smaller, it's also not as close to the scroll bar with this.

Zarel:
Should probably also use new icons for this...

PR #589, opened 2015-12-26
Show exact HP for my active side

urkerab:
When choosing your move, if you hover over your own active Pokémon, you see your exact ability, item and stats, but not your exact HP, as mentioned in http://www.smogon.com/forums/posts/6570673 (you do see your exact HP if you can switch as then you can hover over the switch choices).

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1144,6 +1144,8 @@
 				text += '</h2>';
 				if (pokemon.fainted) {
 					text += '<p>HP: (fainted)</p>';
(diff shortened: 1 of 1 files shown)


ascriptmaster:
Don't we have the count to the top right of the selectable moves? Maybe we should make it more prominent.

Zarel:
The only reason this is necessary is because of a bug in the `|split` code. I'd kind of prefer if that was fixed instead.

Zarel:
Fixed: 3bfdb16bcb

PR #596, opened 2015-12-30
Tooltip item power fixes

urkerab:
I've split this up into 11 parts otherwise it just looks like I rewrote the code, which was not the intention when I first started reading PR #594.


hvk3:
Thanks urkerab. I honestly had no idea that my code could be improved this significantly.

Zarel:
Really, though, if BattleTooltips isn't going to have a constructor, it should be a global.

PR #598, opened 2015-12-30
Properly calculate DVs when calculating stats in Gen1/2

urkerab:
The Teambuilder uses 31 IVs by default however this needs to map to 15 DVs rather than 15.5 as it currently does. For example, Zapdos should have a maximum speed of 298 in Gen2, but 299 in Gen3.


Zarel:
Can you rebase?

Zarel:
Also name the variable `iv` instead of `ivs`; there's only one.

PR #600, opened 2016-01-01
Redundant checks in teambuilder

hvk3:


Zarel:
Damn it, TI, this is wrong. :|

Zarel:
```
			if (stats.spa > stats.atk && moveCount['Special'] > 1) offenseBias = 'Special';
			else if (stats.atk > stats.spa && moveCount['Physical'] > 1) offenseBias = 'Physical';
			else if (moveCount['Special'] > moveCount['Physical']) offenseBias = 'Special';
```

was supposed to be this :|

TheImmortal:
Both lines looks exactly the same to me...?

Zarel:
Yes, but obviously they weren't intended to be the same.

Zarel:
Now I'm going to get merge conflicts when I push the change that fixes it correctly...

hvk3:
Should I make the changes?

hvk3:
#Zarel?

Zarel:
Feel free.

PR #603, opened 2016-01-06
Prevent users from ignoring themselves

Teremiare:
Didn't even know it was possible until I saw this in the Help room:
[20:10:19] Madisaur: How do you un-ignore someone?
[20:10:24] Combeenation: /unignore (username)
[20:10:42] Madisaur: Thanks. Accidently ignored myself. It's rather frustrating.


Zarel:
What about while changing names?

PR #604, opened 2016-01-07
New Aura Sphere animation

Ridaz:


data/graphics.js
```diff
@@ -1552,6 +1552,58 @@ var BattleMoveAnims = {
 			}, 'accel', 'explode');
 		}
 	},
+	aurasphere: {
+		anim: function(battle, args) {
+			var attacker = args[0];
+			var defender = args[1];
+
+			battle.backgroundEffect('#cc9900',700,.6);
```
Zarel: You see all this red? It's because you used `.6` instead of `0.6`

ascriptmaster:
Do you have footage of this new version?

Ridaz:
That's the old one. @Zarel said it looked good without the extra graphics so, according to the code; it looks like this:

Zarel:
Oh, I just meant I wanted to know whether or not you added any extra graphics files to `fx/`. I'm fine with or without the extra effects.

PR #605, opened 2016-01-08
Fix pokemon forms appearing twice in search

Teremiare:


js/search.js
```diff
@@ -155,7 +155,7 @@
 		var queryAlias;
 		if (query in BattleAliases) {
 			queryAlias = toId(BattleAliases[query]);
-			if (query === 'pika' || query === 'bee' || query === 'sab') searchPasses.shift();
+			if (queryAlias.search(query) === 0) searchPasses.shift();
```
Zarel: ...no

Zarel: This removes `zard` as being an alias of `charizard`, which is bad.

Zarel: Oh, wait, `=== 0`

Zarel: Why not slice and compare? In theory that should be faster than `.search`

Zarel: What's the difference between `.search` and `.indexOf`?

Zarel: Also, `searchPasses.shift()` removes the non-alias search. This makes it impossible to find "Sablenite" on `sab`. We should probably do it the other way around.

Teremiare: I have tested it and it worked for every alias as far as I could see. Also I guess .indexOf would be better here, gonna change that.

Teremiare:
Ah yeah, the item thing is a problem

js/search.js
Teremiare: Ah yeah, the item thing is a problem

Teremiare:
Now it only shifts on pokemon. Also changed it to indexOf as there was no need to use search there.

js/search.js
Zarel: Yes, it is.

js/search.js
```diff
@@ -155,7 +155,7 @@
 		var queryAlias;
 		if (query in BattleAliases) {
 			queryAlias = toId(BattleAliases[query]);
-			if (query === 'pika' || query === 'bee' || query === 'sab') searchPasses.shift();
+			if (queryAlias.indexOf(query) === 0 && qType === 'pokemon') searchPasses.shift();
```
Zarel: Didn't I say slice and compare would be better than either search or indexOf? :|

Zarel: Also, are you sure this doesn't lead to any other conflicts?

Zarel: You fixed the item conflict but not the other ones... might be easier to skip the alias rather than skipping the original query.

Zarel:
This still isn't skipping the alias...

Teremiare:
Now it just ignores the alias if the name starts with it instead of skipping the original query

PR #607, opened 2016-01-12
Improve the way usernames look on user popups

bumbadadabum:
schermafdruk van 2016-01-12 18 57 15
This removes the ugly default html look of the usernames, and removes purple usernames

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -2297,7 +2297,7 @@
 
 			var buf = '<div class="userdetails">';
 			if (avatar) buf += '<img class="trainersprite' + (userid === ownUserid ? ' yours' : '') + '" src="' + Tools.resolveAvatar(avatar) + '" />';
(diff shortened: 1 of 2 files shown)


panpawn:
This is an improvement imo

Zarel:
Purple usernames are intentional. -_-

Zarel:
As is the "default HTML look". It's intentionally made very obviously a link.

bumbadadabum:
Whether it's intentional or not it's ugly.

Zarel:
I'd be fine with the color scheme change if it was at least underlined.

The way it currently looks, the name doesn't look different enough from the rank. Bringing back the underline should be enough, but if not, we might want to consider more separation, too.

bumbadadabum:
welcome back underline

Zarel:
In the end, after a Staff discussion and some thought, I don't think there's enough consensus to support this.

I still don't like this color. I'm still open to some sort of change since there's definitely interest, but I think if we want to do something like this, we might as well use panpawn's pullreq that uses usercolor.

PR #614, opened 2016-01-15
Don't notify for old tournament creation messages

sirDonovan:
(Updated)

### DIFF
--- a/js/client-chat-tournament.js
+++ b/js/client-chat-tournament.js
@@ -301,6 +301,7 @@
 					break;
 
 				case 'start':
+					this.room.closeNotification('tournament-create');
 					if (!this.info.isJoined) {
 						this.toggleBoxVisibility(false);
 					} else if (this.info.format.substr(0, 4) === 'gen5' && !Tools.loadedSpriteData['bw']) {
@@ -471,6 +472,7 @@
 					// Fallthrough
 
 				case 'forceend':
+					this.room.closeNotification('tournament-create');
 					this.info = {};
 					this.updates = {};
 					this.savedBracketPosition = {};


Zarel:
This is way overcomplicated.

You can just test to see if the message was in the init message.

sirDonovan:
Updated.

js/client-chat-tournament.js
```diff
@@ -266,7 +266,7 @@
 					var formatName = window.BattleFormats && BattleFormats[data[0]] ? BattleFormats[data[0]].name : data[0];
 					var type = data[1];
 					this.room.$chat.append("<div class=\"notice tournament-message-create\">" + Tools.escapeHTML(formatName) + " " + Tools.escapeHTML(type) + " Tournament created.</div>");
-					this.room.notifyOnce("Tournament created", "Room: " + this.room.title + "\nFormat: " + formatName + "\nType: " + type, 'tournament-create');
+					if (!onInit) this.room.notifyOnce("Tournament created", "Room: " + this.room.title + "\nFormat: " + formatName + "\nType: " + type, 'tournament-create');
```
Zarel: Should the room not even be orange?

Should it depend on whether or not the tournament has started?

sirDonovan: The start message won't be parsed before this one so we can't base it on that, but a subtle notification on init sounds fine to me.

Zarel: Well, it'd be a subtle notification on init either way.

I mean a normal notification, just without the desktop notification part.

sirDonovan: There's still a chance that the tournament has already started or ended, which we can't check, so I think keeping <code>subtleNotifyOnce()</code> might be best.

Zarel: In theory, we can, it'd just be more code.

I think the simplest approach is, if the tournament has already started, remove the notification.

Except clearing the `notifyOnce` notification would also clear highlights, so maybe we should also track a `notifyOncePriority` which is `1` for tournaments and `2` for highlights, and only clear it if it's `1`.

sirDonovan: That could work. I'll see what I can do.

sirDonovan:
Updated. Highlights aren't cleared if you pass in the notification's tag, so we don't need to add <code>notifyOncePriority</code>.

PR #615, opened 2016-01-15
Move Animations Update #2

Ridaz:
Claw attack: 
claw attack

Bubble Beam:
bubble beam

Water Pulse:
water pulse

Fusion Flare:
fusion flare

Fusion Bolt:
fusion bolt


ascriptmaster:
Why does fusion flare have the shockwave in the ground for its animation but not fusion bolt?

Ridaz:
Sort of slipped my mind. I'll add that in the next commit

PR #616, opened 2016-01-15
Fix animations

Ridaz:
Claw Attack:
claw attack

Fusion Bolt:
fusion bolt


Zarel:
Imperative mood: Use `Fix` in the commit summary, not `Fixing`.

ascriptmaster:
The commit name itself hasn't changed, though

PR #619, opened 2016-01-20
Update ignore list on rename

Teremiare:
You'll now unignore your own alts on updateuser if you for some reason ignored them and ignore the new name of someone you've already ignored when they change it.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -798,6 +798,11 @@
 					}
 				}
 				return;
+			} else if (data.substr(0, 3) === '|N|') {
+				var names = data.substr(1).split('|');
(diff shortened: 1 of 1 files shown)


js/client.js
```diff
@@ -853,6 +858,9 @@
 				if (named) {
 					this.trigger('init:choosename');
 				}
+				if (app.ignore[toUserid(name)]) {
+					delete app.ignore[toUserid(name)];
+				}
```
Zarel: This may be better off inside `setPersistentName`

Zarel: wait nvm

PR #621, opened 2016-01-22
Sacred & Secret Sword Animation Update

Ridaz:
No longer a "sword that claws"
sacred sword

### DIFF
--- a/data/graphics.js
+++ b/data/graphics.js
@@ -3644,9 +3644,6 @@ var BattleMoveAnims = {
 	furycutter: {
 		anim: BattleOtherAnims.clawattack.anim
 	},
-	sacredsword: {
-		anim: BattleOtherAnims.clawattack.anim
-	},
 	scratch: {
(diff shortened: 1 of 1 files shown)


Zarel:
Why not a single slash? We have a slash particle...

ascriptmaster:
We don't actually have a slash particle at this moment. I don't know which animation you're looking at but it's probably achieved with fast moving white particles.

Zarel:
I was actually talking about the claw particle, which you would have known if you'd checked in the Dev room at all. <_<

Zarel:
Which in hindsight I kind of understand why you're not using it, but seriously, I don't like this look. A bunch of balls doesn't say "slash" to me.

Ridaz:
Sacred Sword:
sacred sword

Secret Sword:
secret sword

ascriptmaster:
My understanding from the pull request initial comment was that we were specifically taking out the claw graphic, hence my assumption that we weren't going to use that as our "slash" particle

PR #622, opened 2016-01-23
Update Move Animations Round 4

Ridaz:
Scald:
scald

Hydro Pump:
hydro pump

Sacred Sword:
sacred sword

### DIFF
--- a/data/graphics.js
+++ b/data/graphics.js
@@ -996,6 +996,54 @@ var BattleOtherAnims = {
 			}, 'ballistic2Under');
 		}
 	},
+	hydroshot: {
+		anim: function(battle, args) {
+			var attacker = args[0];
+			var defender = args[1];
+
+			battle.showEffect('waterwisp', {
+				x: attacker.x,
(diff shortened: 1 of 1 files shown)


data/graphics.js
```diff
@@ -10009,11 +10058,60 @@ var BattleMoveAnims = {
 			}, 'decel', 'explode');
 		}
 	},
+	hydroshot: {
```
ascriptmaster: If you're going to do something like this as a base point for moves, since it isn't a move itself, it needs to go in BattleOtherAnims

ascriptmaster: Otherwise this should be "named" as one of the moves that use this animation, which I think is the preferred option right now. Unless you also update scald and hydro pump to call this hydro shot animation

Ridaz: The idea was to use Hydro Pump for this animation, but when I was trying to put it in, errors would show up, dunno why. that's why I just c/p'd it. If I actually get it to work, I'll update it

PR #623, opened 2016-01-23
Add suppress interstice option

Teremiare:


CrestfallPS:
Not Zarel, but when I brought up doing this last year it was rejected by multiple tech upper staff. Let's hope things have changed!

Zarel:
I would call this 'skip interstice' and not 'hide interstice'.

Actually, I wouldn't use "interstice" at all. Something more like "Don't warn for untrusted links".

@CrestfallPS who rejected it and why?

Teremiare:
So something like that?

panpawn:
As long as this isn't enabled by default, I see no issue with it?  The prompt is very clear so users will generally know what it means before clicking it and finding out later.

PR #624, opened 2016-01-23
Put volume control on a loudness scale rather than amplitude

kotarou3:


js/battle.js
```diff
@@ -139,16 +139,22 @@ var BattleSoundLibrary = (function () {
 			if (this.bgm) this.bgm.play();
 		}
 	};
+
+	function loudnessPercentToAmplitudePercent(loudnessPercent) {
+		// 10 dB is perceived as approximately twice as loud
+		var decibels = 10 * Math.log(loudnessPercent / 100) / Math.log(2);
+		return Math.pow(10, decibels / 20) * 100;
```
urkerab: Out of interest, another way of calculating this is `Math.pow(loudnessPercent / 100, 1 / Math.log10(4)) * 100`.

kotarou3: Yes, I know, but I kept it unsimplified so people can see the intent to the calculation :p

PR #629, opened 2016-01-28
Update Dark Mode error-message

psnsVGC:
Make error-message dark red to make it more noticeable old colour: http://prntscr.com/9vqc5k new colour: http://prntscr.com/9vqcwj


Zarel:
New color is way too saturated. Also please don't make multiple dark mode pullreqs; a single one is enough.

PR #634, opened 2016-01-31
Shrink assets

liangricha:
Ran [ImageOptim](https://imageoptim.com/) on this repo with the highest settings and got around 400 KB of savings without any loss of quality.


Zarel:
@Marty-D

Marty-D:
Well, just looking at the very first image in the diff shows an obvious difference in brightness, and on further inspection all the PNGs appear to have changed contrast in some way. The JPGs on the other hand look the same, so if you could remove the PNG changes (or compress them in a way that is actually lossless), that would be appreciated.

liangricha:
Hmm weird, I compared the two images with ImageMagick, and it found no difference. As you noted, there is a visual difference in the browser, especially the first image of the diff. I've also compiled Zopfli (which ImageOptim almost always chooses) locally, tested the png compression it uses on that icon, and got the darker one as a result. My wild guess (not an image/browser expert or anything) is that the original images may have metadata that causes some of them to be displayed differently in certain settings? 

I'll take the PNGs for now and see what's up.

liangricha:
@Zarel @Marty-D updated the PR to only include JPEGs.

Zarel:
https://imageoptim.com/color-profiles.html

This is the documentation you're looking for.

Zarel:
I use sRGB for all my images, so I don't understand why those colors changed during the conversion process...

Zarel:
It seems to me like we don't want color profiles in our PNGs anyway, so I'm going to accept ImageOptim's PNG files.

kotarou3:
More likely is it's the gamma information in the PNGs

Zarel:
Or that. Either way we shouldn't be embedding nonstandard gamma in the first place. :|

liangricha:
Sorry for the delay! I'm updated the PR.

kotarou3:
Embedded gamma is part of the PNG standard. Support is just spotty

liangricha:
So we can:

- Remove the gamma metadata from the images, changing them in certain settings, but making the experience consistent across all.
- Try the image optimization while keeping the metadata. This will result in larger files, but we won't change the current behavior.

Zarel:
I think we can accept these. I don't think they look bad, and consistent display across more rendering engines is probably a good idea.

@Marty-D?

Marty-D:
I agree they don't look bad, I was just curious why they changed and assumed it was loss of quality. Since it turns out to be an improvement, I'm good with this.

I'll leave it to you to merge, @Zarel; thanks for being thorough, @liangricha! :+1:

liangricha:
Thanks folks! :smiley:

PR #635, opened 2016-01-31
Use subtle class for unlinking in battles

Teremiare:


Zarel:
Can you update the text itself, too?

Teremiare:
Updated to what chatrooms use

PR #636, opened 2016-01-31
Don't let spectators chat in battles when they're ignoring spectators

Teremiare:


Zarel:
I think it might be better just not to let you send messages at all if you're ignoring spectators and you're not a player...

Teremiare:
Well the reasoning for this would be if you wanna ignore all other spectators but still talk to one or both players

Zarel:
Yes, and if you say something that everyone else has already said?

If you want to talk, you should be able to see the context you're talking in.

Teremiare:
This would only let the room players send messages when ignoring spectators, not too sure about it though as I would prefer if it didn't block global auth. (And I need to make a better commit summary for it)

Slayer95:
If auth need to act officially, they shouldn't be ignoring spectators in the first placer imo

Teremiare:
I guess that's true

js/client-chat.js
```diff
@@ -93,7 +93,9 @@
 				this.tabComplete.reset();
 				this.chatHistory.push(text);
 				text = this.parseCommand(text);
-				if (text) {
+				if (this.id.slice(0, 7) === 'battle-' && this.battle.ignoreSpects && app.user.get('userid') !== this.battle.p1.id && app.user.get('userid') !== this.battle.p2.id) {
```
Zarel: `if (this.battle &&` is a more straightforward check here...

Teremiare: I was remembering it causing issues when it actually was `room.battle` that was the problem and ended up not making that change, so yeah...

Teremiare:
Fixed

js/client-chat.js
```diff
@@ -93,7 +93,9 @@
 				this.tabComplete.reset();
 				this.chatHistory.push(text);
 				text = this.parseCommand(text);
-				if (text) {
+				if (this.battle && this.battle.ignoreSpects && app.user.get('userid') !== this.battle.p1.id && app.user.get('userid') !== this.battle.p2.id) {
+					this.add('You can\'t chat in this battle as you\'re currently ignoring spectators');
```
kotarou3: Double quotes please

Also, this can be trivially bypassed as it's a client restriction...

Zarel: It's supposed to be trivially bypassed; `ignoreSpects` is a client-side setting in the first place.

It's more to prevent accidentally speaking while it's on, and being confused why you don't get a response.

PR #637, opened 2016-02-01
Change Ignore opponent to Ignore Players

hvk3:


hvk3:
As discussed in the chat room.

PR #639, opened 2016-02-05
"Online Only" main menu button tweaks

liangricha:
disable clicks on the "onlineonly" buttons in the main menu in addition to
adding the disabled css class - also initialize them as disabled


Zarel:
Your commit summary should be less than 50 characters long, to prevent this:

image

liangricha:
Updated!

Zarel:
I'd consider showing error messages in `roomlist` and `finduser` instead of failing silently when clicked.

PR #641, opened 2016-02-05
Move .dark .error-reply to battle.css

panpawn:


panpawn:
> n.b. this is entirely the wrong place for it; it should be in battle.css
>The dark version of any style goes in the same file as the regular version of any style.
- Zarel

Zarel:
I already did this when I improved the color. Pure red is way too much of an eyesore. :|

panpawn:
Oh, ops.  Sorry again!

PR #643, opened 2016-02-06
Update Move Animations Round 5

Ridaz:
Primal Reversion:
primal

Mega Evolution:
mega evo

Dark Void:
dark void

Some slashing animations + Heal Pulse:
slash

Improved Sacred & Secret Sword:
sacred sword

X-Scissor:
xscissor

Tri-Attack:
triattack

Light of Ruin:
light of ruin

Refresh:
refresh

Geomancy
geomancy

Drill Run:
drill run


ascriptmaster:
Wow, did you make these images yourself?

Ridaz:
Crap I forgot to add comments where I got the mega symbol and the slash

Ridaz:
I didn't add credits to the rainbow image because I don't know / forgot where Marty got it from

data/graphics.js
```diff
@@ -142,6 +142,30 @@ var BattleEffects = {
...
+	},
+	alpha: { // Ripped from Pokemon Global Link
+		url: 'alpha.png',
+		w: 80, h: 80
+	},
+	omega: { // Ripped from Pokemon Global Link
+		url: 'omega.png',
+		w: 80, h: 80
+	},
+	mega: { // http://orig05.deviantart.net/4ec3/f/2015/249/0/7/mega_evolution_templates_by_leparagon-d98ke9q.png
```
Zarel: Source is actually http://leparagon.deviantart.com/art/Mega-Evolution-templates-558584126

Don't just hotlink, that gives zero license information.

Zarel: Also you should ask for a specific license. "For anyone who wants to use them" is a good sign but legally we'd ideally need specific permission about what we can and can't do with it.

Ridaz:
Replaced the mega wisp with the one ripped from in-game

js/battle.js
```diff
@@ -799,7 +799,7 @@ var Sprite = (function () {
 	Sprite.prototype.behind = function (offset) {
 		return this.z + (this.isBackSprite ? -1 : 1) * offset;
 	};
-	Sprite.prototype.animTransform = function (species) {
+	Sprite.prototype.animTransform = function (species, isMega, isPrimal) {
```
Zarel: Adding two params is weird... Maybe just pass `template`.

Zarel:
Why do you need a mega wisp anyway? It seems like it's only used in one animation and looks fine without it...?

Ridaz:
Without the mega wips I just feel it feels lacking on the Mega Evolution transformation, as in, losing that detail that people associate with Mega Evolution, without needing to look at the log (since it shows up and disappears very quickly).
Besides, we don't even have a Mega Evolution symbol next to the Pokémons name, to help with that association. Think of the new people joining Pokémon Showdown. They are going to do a transition from playing In-game, to this. I want them to feel more comfortable with this transition visually, so I feel like that wisp will help with that.
Furthermore, a considerable about of Pokémon Mega Evolve, so, they're going to use this animation

Zarel:
Like, I'd think `iceball`, `mistball`, `rainbow` would be enough. Do you really need `mega` to make it look good?

ascriptmaster:
I'll be honest, I didn't even notice the mega graphic at all until it was pointed out just now. I don't think we need it.

Ridaz:
@Zarel, yes.
@ascriptmaster, I updated it to make it more noticeable around the time Zarel made that comment. Are you sure a person wouldn't notice it until pointed out?

Ridaz:
Now it looks like this:
mega evo1

Zarel:
Well, we could add a mega symbol next to the name. I'd consider that a better solution than the `mega` wisp, tbh.

I do think it looks better with the mega wisp, but not so much better to be worth adding an extra sprite.

PR #644, opened 2016-02-07
Genders can be hacked - client changes

hvk3:
Reference : https://github.com/Zarel/Pokemon-Showdown/pull/2413

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1047,7 +1047,7 @@
 			};
 			buf += '<span class="detailcell detailcell-first"><label>Level</label>' + (set.level || 100) + '</span>';
 			if (this.curTeam.gen > 1) {
-				buf += '<span class="detailcell"><label>Gender</label>' + GenderChart[template.gender || set.gender || 'N'] + '</span>';
+				buf += '<span class="detailcell"><label>Gender</label>' + GenderChart[set.gender || template.gender || 'N'] + '</span>';
 				buf += '<span class="detailcell"><label>Happiness</label>' + (typeof set.happiness === 'number' ? set.happiness : 255) + '</span>';
 				buf += '<span class="detailcell"><label>Shiny</label>' + (set.shiny ? 'Yes' : 'No') + '</span>';
 			}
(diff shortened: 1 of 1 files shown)


js/client-teambuilder.js
```diff
@@ -1047,7 +1047,11 @@
 			};
 			buf += '<span class="detailcell detailcell-first"><label>Level</label>' + (set.level || 100) + '</span>';
 			if (this.curTeam.gen > 1) {
-				buf += '<span class="detailcell"><label>Gender</label>' + GenderChart[template.gender || set.gender || 'N'] + '</span>';
+				if (this.curTeam.format !== 'balancedhackmons') {
+					buf += '<span class="detailcell"><label>Gender</label>' + GenderChart[template.gender || set.gender || 'N'] + '</span>';
+				} else {
+					buf += '<span class="detailcell"><label>Gender</label>' + GenderChart[set.gender || 'N' || template.gender] + '</span>';
```
Zarel: Should `'N'` go before `template.gender`? I think it should be the other way around...

Zarel: As in, we should support explicitly setting `set.gender` to `N`.

hvk3:
Is it fine otherwise? I'll make the required change later today.

js/client-teambuilder.js
```diff
@@ -2059,7 +2063,7 @@
 
 			if (this.curTeam.gen > 1) {
 				buf += '<div class="formrow"><label class="formlabel">Gender:</label><div>';
-				if (template.gender) {
+				if (template.gender && this.curTeam.format !== 'balancedhackmons') {
 					var genderTable = {'M': "Male", 'F': "Female", 'N': "Genderless"};
 					buf += genderTable[template.gender];
 				} else {
```
Zarel: `N` means "genderless" and not "random" in Hackmons, that needs to be fixed too.

js/client-teambuilder.js
Zarel: Okay, we do. Next up, `set.gender` should be reset if necessary when changing templates, and `set.gender` should be in front no matter. So this should always be `set.gender || template.gender || 'N'`.

Note that `set.gender || 'N' || template.gender` just means `set.gender || 'N'`, since `'N'` is never falsy.

js/client-teambuilder.js
hvk3: Zarel, could you elaborate this last point about 'N'? As in, what needs to be fixed?

hvk3:
Updated pr. @Zarel, could you explain the 'N' portion once?

Zarel:
This line:

	buf += '<label><input type="radio" name="gender" value="N"' + (!set.gender ? ' checked' : '') + ' /> Random</label>';

should be replaced by this line

	buf += '<label><input type="radio" name="gender" value="N"' + (set.gender === 'N' ? ' checked' : '') + ' /> Genderless</label>';

in Hackmons

Zarel:
Really, a full implementation would be a bit more complicated, but this should be enough for now.

hvk3:
Updated.

js/client-teambuilder.js
```diff
@@ -2059,13 +2059,13 @@
 
 			if (this.curTeam.gen > 1) {
 				buf += '<div class="formrow"><label class="formlabel">Gender:</label><div>';
-				if (template.gender) {
+				if (template.gender && this.curTeam.format !== 'balancedhackmons' && this.curTeam.format !== 'anythinggoes') {
```
Zarel: Why `anythinggoes`? `anythinggoes` should not be in this list.

js/client-teambuilder.js
```diff
@@ -2059,13 +2059,13 @@
...
 					var genderTable = {'M': "Male", 'F': "Female", 'N': "Genderless"};
 					buf += genderTable[template.gender];
 				} else {
 					buf += '<label><input type="radio" name="gender" value="M"' + (set.gender === 'M' ? ' checked' : '') + ' /> Male</label> ';
 					buf += '<label><input type="radio" name="gender" value="F"' + (set.gender === 'F' ? ' checked' : '') + ' /> Female</label> ';
-					buf += '<label><input type="radio" name="gender" value="N"' + (!set.gender ? ' checked' : '') + ' /> Random</label>';
+					buf += '<label><input type="radio" name="gender" value="N"' + (set.gender === 'N' ? ' checked' : '') + ' /> Genderless</label>';
```
Zarel: It should only be "Genderless" if the format is Balanced Hackmons

hvk3:
Made appropriate change

js/client-teambuilder.js
Zarel: You still haven't changed this.

hvk3:
Hope this is fine.

Zarel:
I believe so.

PR #646, opened 2016-02-09
Add tournament message for AlreadyDisqualified

Teremiare:


Zarel:
Is this message already used or planned to be used in the server?

Teremiare:
It's already in use.

PR #650, opened 2016-02-21
Bring string utility functions to parity with server counterparts

Slayer95:
- toName is updated to deal with zalgo and LTR text.
- toId is now infinite-recursion-safe, and equivalent to toUserid.
- Tools.escapeHTML and Tools.sanitizeHTML no longer crash for some input.
- Tools.getString is made available as getString

(We actually need more inter-branch PRs)

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -107,25 +107,34 @@ function hashColor(name) {
 	return colorCache[name];
 }
 
+function getString(str) {
+	if (typeof str === 'string' || typeof str === 'number') return '' + str;
+	return '';
+}
+
 function toId(text) {
-	text = text || '';
-	if (typeof text === 'number') text = '' + text;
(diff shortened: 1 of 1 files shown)


Zarel:
The rest is good, but I'm not sure `getString` is necessary?

Zarel:
The worst those can do is crash iirc, and we trust servers at least not to send data of the wrong type over...

Slayer95:
``getString`` is at least needed internally by other functions to prevent crashes (I assert that the server shouldn't be able to crash the client by sending bad (TM) data. Making it private would require some closures that I'd actually find harder to justify.

I suspect, nevertheless that we actually do need to use toString to avoid crashes, in the context of jQuery#text calls.

Zarel:
Fair.

PR #651, opened 2016-02-21
Remember when Team Preview showed that there was no item

urkerab:
I had to rearrange the code slightly so that the item defaults to 'None' if we have an item effect of any sort (previously this only happened when the previous item was tricked or knocked off).

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -366,14 +366,16 @@ var BattleTooltips = (function () {
 			var item = '';
 			var itemEffect = pokemon.itemEffect || '';
 			if (pokemon.prevItem) {
-				item = 'None';
 				if (itemEffect) itemEffect += '; ';
 				var prevItem = Tools.getItem(pokemon.prevItem).name;
 				itemEffect += pokemon.prevItemEffect ? prevItem + ' was ' + pokemon.prevItemEffect : 'was ' + prevItem;
 			}
-			if (pokemon.item) item = Tools.getItem(pokemon.item).name;
-			if (itemEffect) itemEffect = ' (' + itemEffect + ')';
-			if (item) text += '<p>Item: ' + item + itemEffect + '</p>';
+			if (itemEffect) {
+				item = '<p class="itemless">Item:';
+				itemEffect = ' (' + itemEffect + ')';
+			}
+			if (pokemon.item) item = '<p>Item: ' + Tools.getItem(pokemon.item).name;
+			if (item) text += item + itemEffect + '</p>';
 
 			if (template.baseStats) {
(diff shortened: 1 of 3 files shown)


Zarel:
I'd suggest "(None)" instead of "None". "None" looks too much like an item named "None" at a glance.

urkerab:
(Done)

js/client-battle-tooltips.js
```diff
@@ -366,13 +366,15 @@ var BattleTooltips = (function () {
...
 			var itemEffect = pokemon.itemEffect || '';
 			if (pokemon.prevItem) {
-				item = 'None';
 				if (itemEffect) itemEffect += '; ';
 				var prevItem = Tools.getItem(pokemon.prevItem).name;
 				itemEffect += pokemon.prevItemEffect ? prevItem + ' was ' + pokemon.prevItemEffect : 'was ' + prevItem;
 			}
+			if (itemEffect) {
+				item = '(None)';
+				itemEffect = ' (' + itemEffect + ')';
```
Zarel: `(None) (Team Preview)` looks kind of odd.

Maybe it should just be `(None)`?

Zarel: Or, `(none)`.

Although in general we should probably have `(exists)` and `(none)` look more different. Suggestions?

urkerab: Latest suggestion:
Item: (previewed)
Item: &mdash; (previewed)
Item: &mdash; (Leftovers was knocked off)

Zarel: I still don't think that's very clear.

Zarel:
Maybe icons could help make this clearer?

urkerab:
I'm no good with graphics (I got lucky with Switch Sides when you upgraded Font Awesome), so I'm giving it one last try with text:
Item: (previewed)
<s>Item:</s> (previewed)
<s>Item:</s> (Leftovers was knocked off)

Zarel:
Ehh... maybe gray out when items are removed.

urkerab:
OK, the Item paragraph is now grey if there is no item.

Zarel:
In the end, I think this doesn't really look any better. I think "exists" is clearer than "previewed".

urkerab:
> I think "exists" is clearer than "previewed".

What about the case when the item doesn't exist, which was after all the whole point of this PR?

Zarel:
I think leaving it blank is clear enough...

js/client-battle-tooltips.js
```diff
@@ -366,14 +366,16 @@ var BattleTooltips = (function () {
 			var item = '';
```
urkerab: Ah, so I could have just changed this to `'None'` (plus associated code cleanup) in the first place...

Zarel:
The main problem was that, "(exists)" and "(none)" are both words wrapped in parens, which is bad because they're opposites.

Leaving the item field blank for no item is a lot more recognizable, imo.

Issue #658, opened 2016-02-28
[Lint error] Duplicate identifier 'sandtomb' in data/graphics.js


mizdra:
In data/graphics.js, ``BattleMoveAnims.sandtomb`` properties are written at [line 7186](https://github.com/Zarel/Pokemon-Showdown-Client/blob/335794b8cd4c59089e0532e754222cb79578d4d2/data/graphics.js#L7186) and [line 10901](https://github.com/Zarel/Pokemon-Showdown-Client/blob/335794b8cd4c59089e0532e754222cb79578d4d2/data/graphics.js#L10901).

The former isn't used because the latter override the former in this case.
So I think one of those should be deleted.

An example is below:
```js
var obj = {
    a: 10,
    b: 20,
    a: 30
};
console.log(obj.a); // => 30
```

A replay is below:
- <http://replay.pokemonshowdown.com/1v1-337041825>

Zarel:
@Ridaz, which one is the intended Sand Tomb animation? The other should be deleted.

Zarel:
Also: Isn't this the sort of thing `eslint` is supposed to pick up? @Slayer95.

ascriptmaster:
Ridaz wasn't involved in this one, the error apparently traces back over a year ago. `git blame` traces the duplication to #269, when I initially implemented the animation, meaning I accidentally duplicated it in the commit, and because the diff file was too big to show on the web client, nobody noticed the duplication.

Because they were implemented simultaneously on accident, the latter animation is the "intended" one.

Zarel:
Fixed in 9cd878d5f, but I'm leaving this issue open for my question to @Slayer95.

mizdra:
Thank you for fixing. :smiley:

> but I'm leaving this issue open for my question to @Slayer95.
I understand.

Slayer95:
@Zarel, the ``data`` folder isn't covered yet.

Talking about which, we probably should add most of it to ``.gitignore``

Zarel:
It's not added to `.gitignore` for the sole reason that that would make testing client changes a lot more annoying.

Slayer95:
@Zarel, can't we just use the code from ``play.pokemonshowdown.com/data`` in ``testclient.html``?
It's not like those files are editted directly with any frequency, except for the few files remaining in source control after #660

Zarel:
[...]

PR #660, opened 2016-02-28
Update build system: clean up and turn on linting of data

Slayer95:


Zarel:
...I'd rather you just linted client data, rather than requiring a checkout and linting all data.

Zarel:
@Slayer95, to move the discussion from #658 here:

> can't we just use the code from play.pokemonshowdown.com/data in testclient.html?
It's not like those files are editted directly with any frequency, except for the few files remaining in source control after #660

I've had to develop PS offline a few times (seriously, nothing better to do when the US's third-world internet goes down), and for that I'm very thankful that the teambuilder fully works offline besides the images.

Slayer95:
I'd bet you had access to the data files from a server checkout available then, since it's the sort of edge case long-time PS devs would meet (hey I have too!). Then, the procedural difference would be copying those files over to the client directory, and running the index-generator script (maybe we can add an ``--offline`` command line flag so that ``git pull`` isn't run?)

Zarel:
I sometimes do, but the point is that it allows people to develop by editing a few files and opening testclient, instead of having to learn a build system.

README.md
```diff
@@ -25,6 +25,15 @@ You can make and test client changes simply by opening `testclient.html`.
 This will allow you to test changes to the client without setting up your
 own login server.
 
+However, you will have to retrieve the latest data files first, which are stored
+only in the [Pokémon Showdown server][3] repository.
+
+```bash
+git clone https://github.com/Zarel/Pokemon-Showdown.git data/Pokemon-Showdown --depth=1
+cd githooks
+node build-indexes
```
Zarel: The intended use is `node githooks/build-indexes` which should probably be renamed to something like `build/build` anyway

Zarel: Ideally, though, we don't teach people how to use the console at all because that's annoying. :|

Zarel: There should definitely be no `cd`ing going on.

Zarel: Because `cd` changes state, and state is bad.

Slayer95:
Well, again, for newbies and general usage, using the resources from the live PS server should be good enough, and that removes the need for any build-system usage.

Getting the test client to work offline can still be done, with just a few extra steps done once in a blue moon, as described above (I forgot the search & replace for ´´https://play.pokemonshowdown.com/data/´´ -> ´´data/´´)

Removed ´´cd´´

Zarel:
Yes, but that means that you have to make changes to use it offline. Like, changes that dirty your Git checkout. Which sucks...

Slayer95:
``testclient-offline.html`` y/y? ``data`` is ignored by Git anyway

Zarel:
I... guess...

But you'd also have to tell people to use `testclient-offline.html` if they're testing changes to the data build script or to data itself... what's wrong with the current approach?

Slayer95:
The problem is the usage of outdated data by default. It defeats the point of usage of Github if we plan to store outdated source files.

In that line, this new approach should work (https://github.com/Zarel/Pokemon-Showdown-Client/commit/2698e318400a0eededdab94e98ae6c57c283c6d8). No console, no extra instructions :)

Slayer95:
Just noted that PSdex still uses ``learnsets-g6.js``. Is that planned to remain as is? (If that's the case, I've got to remove db1f9fe3d8)

Zarel:
Okay, that's... fair.

.eslintrc.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
 		"Battle": true, "Pokemon": true, "BattleSound": true, "BattleTooltips": true,
 		"BattleAbilities": false, "BattleAliases": false, "BattleBackdrops": false, "BattleBackdropsFour": false, "BattleBackdropsThree": false, "BattleEffects": false,
 		"BattleFormats": false, "BattleFormatsData": false, "BattleLearnsets": false, "BattleItems": false, "BattleMoveAnims": false, "BattleMovedex": false, "BattleNatures": false, 
-		"BattleOtherAnims": false,  "BattlePokedex": false,"BattlePokemonSprites": false, "BattlePokemonSpritesBW": false, "BattleSearchCountIndex": false, "BattleSearchIndex": false,
+		"BattleOtherAnims": false,  "BattlePokedex": false,"BattlePokemonSprites": true, "BattlePokemonSpritesBW": true, "BattleSearchCountIndex": false, "BattleSearchIndex": false,
```
Zarel: What does `true` mean here?

Zarel:
Yes, we still need learnsets-g6 unless you want to generate learnset lists some other way (e.g. caching)

js/replay-embed.js
```diff
@@ -7,8 +7,8 @@ document.write('<script src="https://play.pokemonshowdown.com/js/lib/html-saniti
 document.write('<script src="https://play.pokemonshowdown.com/js/lib/soundmanager2-nodebug-jsmin.js"></script>');
 document.write('<script src="https://play.pokemonshowdown.com/js/config.js?a6"></script>');
 document.write('<script src="https://play.pokemonshowdown.com/js/battledata.js?a6"></script>');
-document.write('<script src="https://play.pokemonshowdown.com/data/pokedex-mini.js?a6"></script>');
-document.write('<script src="https://play.pokemonshowdown.com/data/pokedex-mini-bw.js?a6"></script>');
+document.write('<script src="https://play.pokemonshowdown.com/data/pokedex-mini.min.js?a6"></script>');
+document.write('<script src="https://play.pokemonshowdown.com/data/pokedex-mini-bw.min.js?a6"></script>');
```
Zarel: > `.min.js`

Honestly, if we're doing this, we should just combine them into one file. They're not big enough to be worth the on-demand loading.

.eslintrc.js
Slayer95: That the variable is modified in one of the linted files.
Really, it means that the linter is able to figure out that it's a global assign for only these cases, as the rest of them are via the ``exports`` alias for ``window``.
We might as well use ``exports`` for ``BattlePokemonSprites*``...

js/replay-embed.js
Slayer95: I don't think that's a valid trade-off, as old-gen data is almost never needed.
However, since the rest of past-gen support is already done differently, it's pointless to keep the load fragmented.
I'll move the merge to a different PR.

Slayer95:
Ok, this should be good to go?

Zarel:
```
Updating file `data/learnsets-g6`... NEW ENTRY at learnsets-g6.js: Crucibelle.
NEW ENTRY at learnsets-g6.js: Pikachu-Cosplay.
NEW ENTRY at learnsets-g6.js: Floette-Eternal.
/var/www/html/play.pokemonshowdown.com/githooks/update:154
                newLearnsetsG6[speciesid] = {learnset: newLearnset};
                ^

ReferenceError: newLearnsetsG6 is not defined
```

:|

Zarel:
I don't have internet right now (because Comcast – don't you love America's third-world internet?) so it's your fault I can't work on the client right now. D:

Zarel:
I'm using this for now, please advise if something else needs to be done:

```diff
diff --git a/githooks/update b/githooks/update
index fc212b2..75a670f 100755
--- a/githooks/update
+++ b/githooks/update
@@ -103,6 +103,9 @@ function updateLearnsets (callback) {
        const Pokedex = Tools.data.Pokedex;
        const Learnsets = Tools.data.Learnsets;

+       var newLearnsetsG6 = {};
+       var oldLearnsetsG6;
+
        try {
                oldLearnsetsG6 = require(path.join(rootDir, 'data', 'learnsets-g6.js')).BattleLearnsets;
        } catch (err) {
```

While you're fixing this, could you move this from `update` to `build-indexes`?

Zarel:
```
/var/www/html/play.pokemonshowdown.com/data/abilities.js
  0:0  warning  File ignored because of your .eslintignore file. Use --no-ignore to override
```

Why is this a warning?

Slayer95:
It has been a while since the ``learnsets-g6`` script has been run apparently...
Its build process broke upstream with the automatization of all the data (except ``learnsets.js`` which remained stale), so I couldn't test it in the right conditions.

I'll check it. And next goes linting ``githooks`` to avoid that kind of errors...

RE: That warning; it's because ESLint is dumb... The file is included in the glob pattern, but ESLint claims to be confused because it's also in ``.eslintignore``.

Slayer95:
Yea, you fixed it right....

PR #661, opened 2016-02-29
Fix animations

Ridaz:
- Rock Slide doesn't animate properly in Doubles / Triples, made a workaround animation;
- Cut wasn't animated. It uses Slash's animation now;
- Remove extra Jump Kick;
- High Jump Kick doesn't look like it's going to miss every time now (hopefully).

Rock Slide:
rock slide trips

High Jump Kick currently:
high jump kick

High Jump Kick after:
high jump kick
high jump miss

### DIFF
--- a/data/graphics.js
+++ b/data/graphics.js
@@ -3993,6 +3993,9 @@ var BattleMoveAnims = {
 	scratch: {
 		anim: BattleOtherAnims.slashattack.anim
 	},
+	cut: {
+		anim: BattleOtherAnims.slashattack.anim
+	},
 	slash: {
 		anim: BattleOtherAnims.slashattack.anim
 	},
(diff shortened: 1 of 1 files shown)


Zarel:
What was the problem with Rock Slide?

Ridaz:
http://replay.pokemonshowdown.com/smogontriples-337148149

Ridaz:
Spread moves just don't animate properly. It seems the centre point of all mons animation is the last Mon that comes out
I think #662 starts some work

Zarel:
I don't understand what your workaround animation changes, though, even looking at both of them...

Ridaz:
It's not apparent in singles, but in doubles / triples, the defender is projected to the right in a ridiculous way, when it should be a simple flinch. The work around (that is not exactly a work around, but at least it doesn't look as dumb) is just making the defender flinch by only moving in the y axis

Ridaz:
There's also the fact only 1 of the defenders / attackers animate, when all of them should animate.

I'll record some gifs explaining my points when I get home

data/graphics.js
```diff
@@ -9656,9 +9642,8 @@ var BattleMoveAnims = {
 			});
 			defender.delay(200);
 			defender.anim({
-				x: defender.leftof(+5),
```
Zarel: Here's your problem. You're using `leftof`. Why? In theory it should only be forwards and backwards movement... what's the intent of the effect?

Ridaz: I wanted to make the defender flinch in different directions

Ridaz: That still doesn't excuse the defenders animation behaviour

Zarel: Hm, maybe we should add movement to the flinch animation.

Ridaz: As in `case: flinch` or when it's hit by the rocks?

Zarel: We already have a flinch animation (the ping thing) but I meant to add pokemon movement to it.

PR #662, opened 2016-02-29
Support animating several Pokemon in spread moves

QuiteQuiet:
Currently spread moves in doubles / triples only animate one target and without doing something here, that won't change. Though, I don't know if this is the best way to accomplish this, or even if it's the best place to put it.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -3219,7 +3219,23 @@ var Battle = (function () {
 			if (kwargs.prepare || kwargs.anim === 'prepare') {
 				this.prepareMove(pokemon, move, target);
 			} else if (!kwargs.notarget) {
-				(kwargs.anim ? Tools.getMove(kwargs.anim) : move).anim(this, [pokemon.sprite, target.sprite]);
+				var usedMove = kwargs.anim ? Tools.getMove(kwargs.anim) : move;
+				if (kwargs.spread) {
+					var targets = [pokemon.sprite];
+					var hitPokemon = kwargs.spread.split(',');
+					if (hitPokemon[0] !== '.') {
+						for (var i = 0; i < hitPokemon.length; i++) {
+							targets.push(this.getPokemon(hitPokemon[i] + ': ?').sprite);
+						}
+					} else {
+						// if hitPokemon[0] === '.' then no target was hit by the attack
+						targets.push(target.side.missedPokemon.sprite);
+					}
+
+					usedMove.anim(this, targets);
+				} else {
+					usedMove.anim(this, [pokemon.sprite, target.sprite]);
+				}
 			}
 		}
 		pokemon.lastmove = move.id;


Zarel:
We need some way of seeing what parts hit and what parts miss...

Zarel:
A lot of very common spread moves in doubles (Origin Pulse, Precipice Blades, Rock Slide, Heat Wave) are less than 100% accurate.

QuiteQuiet:
If it matters, the games animate spread moves like they hit, even at the times they missed, or was protected against. Any misses and the like are announced before the animation, so there's still perfectly clear what happened, though.

Even so, would something like a look-ahead in the activity queue and check which Pokemon it hit work? Or would that mess up something else?

Zarel:
Would work but isn't ideal.

Really, we might just want to pass `isSpread` into the animation, and let the animation handle the rest.

QuiteQuiet:
How would passing `isSpread` to the animation help it know which targets it missed or not? 

And, wouldn't that still mean the entire target side needs to be passed to `anim`?

Zarel:
No, that's more about having it animate as a whole.

On second thought, passing a list of targets is probably better. I kind of like this approach.

js/battle.js
```diff
@@ -3219,7 +3219,16 @@ var Battle = (function () {
 			if (kwargs.prepare || kwargs.anim === 'prepare') {
 				this.prepareMove(pokemon, move, target);
 			} else if (!kwargs.notarget) {
-				(kwargs.anim ? Tools.getMove(kwargs.anim) : move).anim(this, [pokemon.sprite, target.sprite]);
+				move = kwargs.anim ? Tools.getMove(kwargs.anim) : move;
```
Zarel: `getMove` is unavailable (or, to be exact, provides no useful information) in replays.

We should pass `[spread]` or something from the server for this, instead of reading `move.target`.

ascriptmaster: All `kwargs` are sent up from the server, as usual.

It's not being used right now, but here's an old Seasonal that used it:

https://github.com/Zarel/Pokemon-Showdown/blob/a8ba012e1e7f9ae14010473a16b453ee6be65e51/config/formats.js#L665-L675

Zarel: `moves.js` isn't available in replays, that's the entire point.

`kwargs.anim` just replaces the move animation with another move's animation.

QuiteQuiet: Am I misunderstanding what you mean with
> `getMove` is unavailable (or, to be exact, provides no useful information) in replays.

then, since it's usage here is the same as before? 

Though, I probably shouldn't overwrite `move` when I look at it now.

Zarel: `getMove` just returns placeholder information in replays. It tells you the ID and gives you the string you passed in back out as a name. Basically, you can't read information from `moves.js` when you're in a replay, such as the move target.

js/battle.js
```diff
@@ -3219,7 +3219,16 @@ var Battle = (function () {
...
+				if (target.side.active.length > 1 && (move.target === 'allAdjacent' || move.target === 'allAdjacentFoes')) {
```
Zarel: Basically, this should just be `if (kwargs.spread) {`

And then on the server we should be doing `attrLastMove('[spread]')`

QuiteQuiet: Would the correct way to do that be adding this to all the spread moves in `moves.js` on the server?
`onPrepareHit: function () { this.attrLastMove('[spread]'); },`

I have little knowledge of how the server works, but from looking at examples that should make the server send `[spread]` when a spread move is used, right? Or is there another place where it fits better?

Zarel: No, that is the least correct way to do it in the history of ways to do it.

Do it in scripts.js where the spread/non-spread-ness is read.

QuiteQuiet:
Updated for `kwargs.spread` (`kwargs.spread === "."` here, but that doesn't matter does it?)

I also changed so it didn't overwrite `move`, because I was worried that altered something further along.

Zarel:
We should probably send a list of hits or something.

js/battle.js
```diff
@@ -3219,7 +3219,16 @@ var Battle = (function () {
 			if (kwargs.prepare || kwargs.anim === 'prepare') {
 				this.prepareMove(pokemon, move, target);
 			} else if (!kwargs.notarget) {
-				(kwargs.anim ? Tools.getMove(kwargs.anim) : move).anim(this, [pokemon.sprite, target.sprite]);
+				var usedMove = kwargs.anim ? Tools.getMove(kwargs.anim) : move;
+				if (kwargs.spread) {
+					var targets = [pokemon.sprite].concat(target.side.active.map(function (active) { return active.sprite; }));
+					if (target.side.active.length > 2 && target.side.active.indexOf(target) !== 1) {
+						targets.splice((target.side.active.indexOf(target) ? 1 : 3), 1);
```
Zarel: `.splice` is the entirely wrong way to do it, it's really slow. Just construct it with `.push`

QuiteQuiet:
Updated for https://github.com/Zarel/Pokemon-Showdown/commit/20877368a3f57a7322e1cea00fee467a0b961d5e though I'm not sure if looping through active pokemon and testing if they're in the list is the best way to go about this.

js/battle.js
```diff
@@ -3219,7 +3219,28 @@ var Battle = (function () {
 			if (kwargs.prepare || kwargs.anim === 'prepare') {
 				this.prepareMove(pokemon, move, target);
 			} else if (!kwargs.notarget) {
-				(kwargs.anim ? Tools.getMove(kwargs.anim) : move).anim(this, [pokemon.sprite, target.sprite]);
+				var usedMove = kwargs.anim ? Tools.getMove(kwargs.anim) : move;
+				if (kwargs.spread) {
+					var targets = [pokemon.sprite];
+					var hitPokemon = kwargs.spread;
+					var locId;
+					for (var i = 0; i < target.side.active.length; i++) {
```
Zarel: It seems like it makes more sense split `hitPokemon` and loop through it.

QuiteQuiet: I suppose, though that still mean you have to loop through both sides for every entry and make the location doesn't it. It looks like that would just move the iteration from `indexOf` to here.

Zarel: It would change the algorithm from O(n^2) to O(n) because it wouldn't use `indexOf` anymore.

Zarel: You can use `getPokemon(entry + ': ?')` if you have trouble figuring out how to get it without looping

QuiteQuiet:
I didn't know that worked.

The last thing that bothers me is that in some situations, with moves like Surf your only target might be on your own side rather than the opponents. I don't know how the animations will be affected by this.

Zarel:
They'll work the same way as if a single-target move targets your own side.

PR #663, opened 2016-03-01
Remove trailing commas

Ridaz:


Slayer95:
@Ridaz, there are at least a dozen. They are removed in my PR though.

Ridaz:
Well then.

PR #665, opened 2016-03-05
Add link visited to dark mode

Teremiare:
Looks like this


Zarel:
Pretty unreadable.

Teremiare:
This should be clearer

PR #669, opened 2016-03-14
Show target of effectiveness of spread moves

urkerab:
When watching a doubles battle on a small screen, you don't get to see the chat, but you do get to see the "It's super effective!" or "It's not very effective..." messages. However for a spread move it's unclear whom each effectiveness refers to. My understanding is that in the case of a spread move the cart identifies each effectiveness.
Before:
before
After:
after

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2279,6 +2279,7 @@ var Battle = (function () {
 
 	Battle.prototype.resultWaiting = false;
 	Battle.prototype.multiHitMove = null;
+	Battle.prototype.activeMoveIsSpread = null;
 
 	// callback
 	Battle.prototype.faintCallback = null;
@@ -2422,6 +2423,7 @@ var Battle = (function () {
 		// activity queue state
 		this.animationDelay = 0;
 		this.multiHitMove = null;
+		this.activeMoveIsSpread = null;
 		this.activityStep = 0;
 		this.activityDelay = 0;
 		this.activityAfter = null;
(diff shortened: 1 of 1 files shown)


TheImmortal:
You are correct! However, there are is only on dot and not three. In addition, it's the same for critical hits so can you add that as well here?

Reference: https://youtu.be/ckoydlkmEEU?t=264

js/battle.js
```diff
@@ -2279,6 +2279,7 @@ var Battle = (function () {
 
 	Battle.prototype.resultWaiting = false;
 	Battle.prototype.multiHitMove = null;
+	Battle.prototype.spreadMove = null;
```
Zarel: We might want to track `activeMove` specifically, and clear it on `|`.

Zarel: either way, I'd recommend this be called `activeMoveIsSpread`

urkerab:
* Changed `...` into a single dot.
* Applied change to critical hits.
* Renamed member.

TheImmortal:
It's only single dot for spread. In singles, it's three dots.

urkerab:
Sorry, I misunderstood. Restored the extra dots for non-spread resisted moves.

TheImmortal:
Did you? I don't see it.

js/battle.js
```diff
@@ -3787,21 +3790,21 @@ var Battle = (function () {
...
-				actions += "It's super effective! ";
+				actions += "It's super effective" + (this.activeMoveIsSpread ? " on " + poke.getLowerName() : "") + "! ";
 				break;
 
 			case '-resisted':
 				var poke = this.getPokemon(args[1]);
 				for (var j = 1; !poke && j < 10; j++) poke = this.getPokemon(minors[i + j][0][1]);
 				if (poke) this.resultAnim(poke, 'Resisted', 'neutral');
-				actions += "It's not very effective... ";
+				actions += "It's not very effective" + (this.activeMoveIsSpread ? " on " + poke.getLowerName() : "") + ". ";
```
TheImmortal: Should be ``(this.activeMoveIsSpread ? " on " + poke.getLowerName() : "..")`` if I'm not mistaken

urkerab: Sorry, I keep forgetting to `git add` my changes.

PR #679, opened 2016-03-26
Add Fur Coat to modified stats tooltip

Teremiare:


Marty-D:
Good catch, thanks! There's no need to floor the result, though.

Teremiare:
Oh right, didn't really think about that as I basically just c/p it

Marty-D:
:+1:

PR #680, opened 2016-03-27
Better touch handling on mobile devices

unimonkiez:
First of all, love the game, thanks for bringing Pokemon to the web!
Playing it at school, at work, even some before sleep for a long time!

One thing that bothered me for long is how it runs on mobile.
Works fast and all but buttons are a nightmare, tooltip stays until you click somewhere else, on iOS I find myself to click the button 20 times until it registers.

Not long ago I saw this on Github finally and decided to fix this today.

Anyway I changed the tooltip attributes from mouse events to touch events.
When you long press (more than 300ms) the tooltip shows, and when you short press (shorter than 300ms) the normal click event is called (which is kind of a hack, you are welcome to change).

Wished you made this more of a standalone project, pain to debug on release site.

Have a nice day and hope to see this in the site really soon!

***Update:***
*Before*:
- Focus stays sometimes when clicking, so you must click outside

- Sometimes it just clicks when wanting to see the tooltip

*After*:
- **You either see the tooltip or click, can't be both.**
- **disappears right after lifting the finger**

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -53,21 +53,14 @@ var BattleTooltips = (function () {
 
 		// On first tap start counting
 		if (elem._numOfTaps === 1) {
-			setTimeout(function () {
-				// When tap is undefined means `_onTouchEnd` deleted it by the end of the time, perform the action
-				if (elem._numOfTaps === undefined) {
-					// short tap
-					elem._isCalledManually = true;
-					elem.click();
-				} else {
-					// When tap is not defined open the tooltip until `_onTouchEnd` will close it
-					elem._toolTipIsOpen = true;
-					BattleTooltips.showTooltipFor(roomid, thing, type, elem, ownHeight);
-				}
+			// Timeout will be canceled by `_handleTouchEndFor` if a short tap have occurred
+			elem._timeout = setTimeout(function () {
+				elem._toolTipIsOpen = true;
+				BattleTooltips.showTooltipFor(roomid, thing, type, elem, ownHeight);
 			}, touchDelayForTap);
 		}
 	};
-	BattleTooltips._handleTouchEndFor = function (event, elem) {
+	BattleTooltips._handleTouchLeaveFor = function (event, elem) {
 		// Prevent default on touch events to block mouse events when touch is used
 		event.preventDefault();
 
@@ -80,6 +73,34 @@ var BattleTooltips = (function () {
 
 		// If tooltip is open and the last finger left, close the tooptip
 		if (elem._toolTipIsOpen && elem._numOfTaps === undefined) {
+			// Delete variable
+			delete elem._toolTipIsOpen;
+			BattleTooltips.hideTooltip();
+		}
+	};
+	BattleTooltips._handleTouchEndFor = function (event, elem) {
+		// Save bool for firing the action before changing the variables
+		// If tooltip is not opened and the last finger left,
+		// meaning the timeout in `_handleTouchStartFor` wasn't fired, fire the action
+		var shouldFireAction = (elem._toolTipIsOpen === undefined && elem._numOfTaps === 1);
+
+		// Decrease taps and close tooltip when needed
+		BattleTooltips._handleTouchLeaveFor(event, elem);
+
+		// When firing the action also cancel the timeout in `_handleTouchStartFor`
+		if (shouldFireAction) {
+			clearTimeout(elem._timeout);
+			delete elem._timeout;
+
+			elem._isCalledManually = true;
+			elem.click();
+		}
+	};
+	BattleTooltips._handleTouchCancelFor = function (event, elem) {
+		if (elem._numOfTaps !== undefined) {
+			delete elem._numOfTaps;
+		}
+		if (elem._toolTipIsOpen) {
 			// Delete variable
 			delete elem._toolTipIsOpen;
 			BattleTooltips.hideTooltip();
@@ -105,7 +126,8 @@ var BattleTooltips = (function () {
 		return ' onclick="BattleTooltips._handleClickFor(this, event)"' +
 		' ontouchstart="BattleTooltips._handleTouchStartFor(event, \'' + roomid + '\', \'' + Tools.escapeHTML('' + thing, true) + '\',\'' + type + '\', this, ' + (ownHeight ? 'true' : 'false') + ')"' +
 		' ontouchend="BattleTooltips._handleTouchEndFor(event, this)"' +
-		' ontouchleave="BattleTooltips._handleTouchEndFor(event, this)"' +
+		' ontouchleave="BattleTooltips._handleTouchLeaveFor(event, this)"' +
+		' ontouchcancel="BattleTooltips._handleTouchCancelFor(event, this)"' +
 		' onmouseover="BattleTooltips.showTooltipFor(\'' + roomid + '\', \'' + Tools.escapeHTML('' + thing, true) + '\',\'' + type + '\', this, ' + (ownHeight ? 'true' : 'false') + ')"' +
 		' onmouseout="BattleTooltips.hideTooltip()"' +
 		' onmouseup="BattleTooltips._handleMouseUpFor(this)"';
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -61,12 +61,12 @@ var BattleTooltips = (function () {
 		// Prevent default on touch events to block mouse events when touch is used
 		e.preventDefault();
 
-		// If tooltip is open and the last finger left, close the tooptip
+		// If tooltip is open and the last finger just left, close the tooptip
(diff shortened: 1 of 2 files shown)


js/client-battle-tooltips.js
```diff
@@ -37,11 +37,71 @@ var BattleTooltips = (function () {
 	};
 
 	// tooltips
-	BattleTooltips.prototype.tooltipAttrs = function (thing, type, ownHeight) {
-		var roomid = this.room.id;
-		return ' onmouseover="BattleTooltips.showTooltipFor(\'' + roomid + '\', \'' + Tools.escapeHTML('' + thing, true) + '\',\'' + type + '\', this, ' + (ownHeight ? 'true' : 'false') + ')" ontouchstart="BattleTooltips.showTooltipFor(\'' + roomid + '\', \'' + Tools.escapeHTML('' + thing, true) + '\',\'' + type + '\', this, ' + (ownHeight ? 'true' : 'false') + ')" onmouseout="BattleTooltips.hideTooltip()" onmouseup="BattleTooltips.hideTooltip()"';
-	};
+	if (navigator.userAgent.match(/(Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini)/)) {
+		// If user is using a mobile phone or tablet, use touch events
```
Zarel: Can't we use both, instead of branching here?

That also fixes the user agent matching, which, while sometimes necessary, isn't really necessary here.

unimonkiez:
@Zarel Added some gifs to the PR

js/client-battle-tooltips.js
unimonkiez: Maybe.. I thought my phone showed the toolbar because it picked up some of the mouse events, only now to see that it has `ontouchstart`, and the click closes the tooltip

js/client-battle-tooltips.js
```diff
@@ -37,11 +37,71 @@ var BattleTooltips = (function () {
...
+			elem._numOfTaps = elem._numOfTaps !== undefined ? (elem._numOfTaps + 1) : 1;
+
+			// On first tap start counting
+			if (elem._numOfTaps === 1) {
+				setTimeout(function () {
+					// When tap is undefined means `_onTouchEnd` deleted it by the end of the time, perform the action
+					if (elem._numOfTaps === undefined) {
+						// short tap
+						elem._isCalledFromTocuh = true;
+						elem.click();
```
unimonkiez: @Zarel Actually because of the code here I can't share onclick event on both devices, unless there is another way to call the action for the button right over here.

js/client-battle-tooltips.js
```diff
@@ -37,11 +37,71 @@ var BattleTooltips = (function () {
...
+			} else {
+				delete elem._isCalledFromTocuh;
+			}
+		};
+		BattleTooltips.prototype.tooltipAttrs = function (thing, type, ownHeight) {
+			var roomid = this.room.id;
+			return ' onclick="BattleTooltips._onClick(this, event)" ontouchstart="BattleTooltips._onTouchStart(\'' + roomid + '\', \'' + Tools.escapeHTML('' + thing, true) + '\',\'' + type + '\', this, ' + (ownHeight ? 'true' : 'false') + ')" ontouchend="BattleTooltips._onTouchEnd(this)" ontouchleave="BattleTooltips._onTouchEnd(this)"';
+		};
+	} else {
+		// If user is using a desktop, use mouse events
```
Slayer95: What about touch-screen monitors?

unimonkiez: Yeah you are right, better support both (I read [this](http://www.html5rocks.com/en/mobile/touchandmouse/) too).

Might add a variable like I did with `_isCalledFromTocuh` that in `_onTouchStart` to `onMouseOver`, unless there is a better way to call that action.

js/client-battle-tooltips.js
TheImmortal: ``/Mobi/.test(navigator.userAgent)`` is the recommended way to check for mobile devices.

https://developer.mozilla.org/en-US/docs/Browser_detection_using_the_user_agent

unimonkiez:
@Zarel Did the requested change, pretty happy with the result except the hack for the action.
Also found an old laptop with touch monitor, works great both touch and mouse.

unimonkiez:
@Zarel Alright done with the changes, just made touch bit faster and removed the changes I made to the css because `event.preventDefault` makes the text unselectable anyway

js/client-battle-tooltips.js
```diff
@@ -35,13 +35,92 @@ var BattleTooltips = (function () {
 	BattleTooltips.hideTooltip = function () {
 		$('#tooltipwrapper').html('');
 	};
-
 	// tooltips
+	// Touch delay, pressing finger more than that time will cause the tooltip to open.
+	// Shorter time will cause the button to click
+	var touchDelayForTap = 300; // ms
```
Zarel: These constants are generally allcaps.

I'd suggest `var LONG_TAP_DELAY`

js/client-battle-tooltips.js
```diff
@@ -35,13 +35,92 @@ var BattleTooltips = (function () {
...
+
+	// Each time a finger presses the button this function will be callled
+	// First finger starts the counter, and when last fingers leaves time is checked
+	BattleTooltips._handleTouchStartFor = function (event, roomid, thing, type, elem, ownHeight) {
+		// Prevent default on touch events to block mouse events when touch is used
+		event.preventDefault();
+
+		// Saving number of Touch start happends (number of fingers on teh button,
+	  // though most of the time will be 1, don't won't a nasty bug like tooltip stuck open)
```
Zarel: weird indentation

js/client-battle-tooltips.js
```diff
@@ -35,13 +35,92 @@ var BattleTooltips = (function () {
...
 	// tooltips
+	// Touch delay, pressing finger more than that time will cause the tooltip to open.
+	// Shorter time will cause the button to click
+	var touchDelayForTap = 300; // ms
+
+	// Each time a finger presses the button this function will be callled
+	// First finger starts the counter, and when last fingers leaves time is checked
+	BattleTooltips._handleTouchStartFor = function (event, roomid, thing, type, elem, ownHeight) {
+		// Prevent default on touch events to block mouse events when touch is used
+		event.preventDefault();
```
Zarel: Convention is to use `e` and not `event` for this variable name.

js/client-battle-tooltips.js
Zarel: Is 300ms a normal delay for a longpress?

js/client-battle-tooltips.js
```diff
@@ -35,13 +35,92 @@ var BattleTooltips = (function () {
...
+
+		// Saving number of Touch start happends (number of fingers on teh button,
+	  // though most of the time will be 1, don't won't a nasty bug like tooltip stuck open)
+		// Starts by 1 and increases each time this func is called
+		elem._numOfTouches = event.touches.length;
```
Zarel: A tooltip stuck open isn't really a huge issue, you can tap the tooltip to close it.

I'd prefer not to attach anything directly to an element. Something something extending native objects. Put it in the dataset if you really need it.

js/client-battle-tooltips.js
```diff
@@ -35,13 +35,92 @@ var BattleTooltips = (function () {
...
+		event.preventDefault();
+
+		elem._numOfTouches = event.touches.length;
+		// Delete variable when last finger left, no need for memory
+		if (elem._numOfTouches === 0) {
+			delete elem._numOfTouches;
+		}
+
+		// If tooltip is open and the last finger left, close the tooptip
+		if (elem._toolTipIsOpen && elem._numOfTouches === undefined) {
```
Zarel: I wouldn't consider `._toolTipIsOpen` the best way to track an open tooltip. Since there should be only one open tooltip at a time, we should just track the open tooltip globally.

js/client-battle-tooltips.js
unimonkiez: 300ms is like everything I seen around.
500ms is long press in android tho

Zarel: Compromise at 350ms I guess

unimonkiez:
@Zarel Made the changes, tell me if there is anything more, would really like to have this on the game :smile:

unimonkiez:
?

Zarel:
There are things I'd do differently, but it'd probably be easier to merge + refactor than to leave this sitting forever. I can merge this.

Zarel:
Can I get an MIT license from you? Repeat after me:

> I, [name], hereby dedicate all my contributions to the Pokémon Showdown project, past and future, under the MIT license, reproduced at https://opensource.org/licenses/MIT

unimonkiez:
Sure, i'll see whatever you change, always good to learn.

*I, Unimonkiez (Yuval Saraf), hereby dedicate all my contributions to the Pokémon Showdown project, past and future, under the MIT license, reproduced at https://opensource.org/licenses/MIT*

PR #681, opened 2016-04-03
Updated tooltips for Mummy activation

hvk3:


hvk3:
Whenever Mummy activates, the opponent's ability becomes visible. This pr aims to handle that.

js/battle.js
```diff
@@ -5234,8 +5234,10 @@ var Battle = (function () {
 					this.resultAnim(ofpoke, ability.name, 'ability');
 					this.animationDelay += 700;
 					this.message('', "<small>[" + ofpoke.getName(true) + "'s " + ability.name + "!]</small>");
+					ofpoke.markAbility(ability.name);
 					this.resultAnim(ofpoke, 'Mummy', 'ability');
 					this.message('', "<small>[" + ofpoke.getName(true) + "'s Mummy!]</small>");
+					ofpoke.markAbility('Mummy', effect.id && !kwargs.fail);
```
Zarel: What's this line doing? Specifically the `effect.id && !kwargs.fail` part

hvk3: Didn't need it. Updated.

Marty-D: ??? You definitely did need it; fixed in https://github.com/Zarel/Pokemon-Showdown-Client/commit/8496d25b162cb4ae612f39d38aacfad16120d0e0. That is what I meant by proper use of `isNotBase`. Otherwise you're overwriting what you just did above.

PR #683, opened 2016-04-04
Make show and hide joins configuration per-room

Joimer:
Hidejoins hides all joins. Showjoins shows all joins.
Using showjoins room adds that room as an exception to hidejoins.
Using hidejoins room removes the exception.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -561,12 +561,29 @@
 				return false;
 
 			case 'showjoins':
-				this.add('Join/leave messages: ON');
-				Tools.prefs('showjoins', true);
+				if (target) {
+					var showJoins = Tools.prefs('showroomjoins') || {};
+					var room = toId(target);
+					showJoins[room] = true;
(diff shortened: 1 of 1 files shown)


Zarel:
Why not just a threshold?

Joimer:
This is more configurable, I want joins shown all the time in staff, but never in lobby, for example.

Zarel:
a threshold of, say, 100 users would do the same thing.

Joimer:
No, it does not. We want to be able to configure exceptions for specific rooms so I can get joins in those, not see all joins in all small rooms. There is no advantage whatsoever into pushing joins on small rooms to the face of all users instead of letting them opt-in in some rooms (someone even may want to shwojoins in Lobby).

Zarel:
Oh, hm, this just a client thing.

I still maintain that the server should also have a threshold of 100 users instead of defaulting to off.

Joimer:
Oh, for server that's a whole different issue, I agree.

PR #652, opened 2016-02-22
Updates towards gen6 battle backgrounds.

The-Run:

PR #686, opened 2016-04-09
Add XY battle backgrounds

QuiteQuiet:
I assume this was what https://github.com/Zarel/Pokemon-Showdown-Client/pull/652 was working towards, but I saw nothing from this after the backgrounds were uploaded. 

@Zarel I recall you mentioning you didn't want some of them, but I don't think you specified which ones.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,7 +17,7 @@ module.exports = {
 
 		// Battle stuff
 		"Battle": true, "Pokemon": true, "BattleSound": true, "BattleTooltips": true,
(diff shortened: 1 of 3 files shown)


Slayer95:
@QuiteQuiet, the build fails because you need to register ``BattleBackdropsFive`` in https://github.com/Zarel/Pokemon-Showdown-Client/blob/master/.eslintrc.js#L20

QuiteQuiet:
Updated. I had no idea that was a thing that had to be done :(

PR #689, opened 2016-04-10
bwgfx will set backgrounds to gen 5 backgrounds

Layell:
I see no reason why we shouldn't do this.


js/battle.js
```diff
@@ -2378,6 +2378,7 @@ var Battle = (function () {
 			else if (this.gen <= 4) this.backdropImage = 'fx/' + BattleBackdropsFour[Math.floor(Math.random() * BattleBackdropsFour.length)];
 			else if (this.gen <= 5) this.backdropImage = 'fx/' + BattleBackdropsFive[Math.floor(Math.random() * BattleBackdropsFive.length)];
 		}
+		if (this.gen <=6 && Tools.prefs('bwgfx')) this.backdropImage = 'fx/' + BattleBackdropsFive[Math.floor(Math.random() * BattleBackdropsFive.length)];
```
TheImmortal: missing space after =

QuiteQuiet: Does `this.gen <= 6` even need to be included? It will always be true anyway (at least until gen 7).

PR #690, opened 2016-04-11
Add Life Orb to tooltips

hvk3:
Since we already have life orb activation in -damage, this adds the item to the tooltips.


Zarel:
Can you make it more general, like how abilities are detected?

hvk3:
Perhaps a function like markItem?

Zarel:
Would be nice, but currently unnecessary.

hvk3:
Then what changes should I make?

Zarel:
I'll just make them myself.

PR #691, opened 2016-04-14
Updated tooltip for Truant activation

hvk3:


Zarel:
The ability should be detected outside of the message. See how abilities are usually marked.

hvk3:
Isn't this the only place where Truant is being handled?

Zarel:
...

hvk3:
Umm?

Zarel:
It's easier just to show you if that explanation wasn't enough...

PR #693, opened 2016-04-19
Improve greentext color in Dark Mode

Ridaz:
Nevermind. Running into trouble rebasing. Starting all over again


Zarel:
You do not need to open a new pull request to start all over again.

I've blocked people from the repo for repeatedly making the same pull request. If you can't figure out how to fix it, _ask for help_. Don't make a new pull request.

Ridaz:
Yeah, you're right. My bad. I'll try so it doesn't happen again

PR #698, opened 2016-04-25
Make subtle buttons more visible in dark mode

asgdf:
While this isn't a huge issue, the "(x lines from y hidden)" message can be pretty hard to read in front of a blueish background (e.g. http://i.imgur.com/yUxRvRY.png). I changed it to look like this: http://i.imgur.com/MlXvtq2.png
Something I'm not really sure about is where to place the new rule in the css file - I put it in the dark mode/chat section since that seemed the most fitting.

### DIFF
--- a/style/battle.css
+++ b/style/battle.css
@@ -875,3 +875,27 @@ License: GPLv2
 	font-size:9pt;
 	content: '...';
 }
+a.subtle {
+	text-decoration: none;
+	color: black;
+}
+a.subtle:hover {
+	text-decoration: underline;
+	color: #6688CC;
+}
+button.subtle {
+	border: 0;
+	background: transparent;
+	padding: 0;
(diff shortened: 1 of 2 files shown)


Zarel:
It goes in chat formatting in `battle.css`, in the chat styling section, right under where the normal `.subtle` styling is.

At least I hope that's where the normal `.subtle` styling is. If it's not there, it should be moved there.

asgdf:
Both `button.subtle` and `a.subtle` are in `client.css` (which is why I put it there as well). I assume all of them should be moved? Can I just add that as a second commit to this, or would you prefer a seperate pullreq?

Zarel:
Replace the commit (`commit --amend` or squash).

Yes, those should all be moved to chat styling in `battle.css`

PR #699, opened 2016-04-27
Allow /hidejoins [room] to override /showjoins

asgdf:


asgdf:
I'd greatly appreciate feedback beyond a thumbs down :(

I don't think this is a superfluous change, because excluding single rooms from showjoins seems like a decent feature for people who only want to exclude lobby or something. I could not think of a better way to do this than to introduce another preference. Most of it is reused existing code, so I assume it can't be that wrong. As it turns out, some of the explicit setting of preferences is unneeded because it's working with a reference of the prefs object anyway, but I went with "better safe than sorry". I suppose the conditional there is kind of unwieldy and confusing? Where'd I mess up here?

js/client-chat.js
```diff
@@ -1309,7 +1316,7 @@
 				this.userList.add(userid);
 				return;
 			}
-			if (silent && (!Tools.prefs('showroomjoins') || !Tools.prefs('showroomjoins')[this.id]) && !Tools.prefs('showjoins')) return;
+			if (silent && (((!Tools.prefs('showroomjoins') || !Tools.prefs('showroomjoins')[this.id]) && !Tools.prefs('showjoins')) || (Tools.prefs('hideroomjoins') && Tools.prefs('hideroomjoins')[this.id]))) return;
```
Zarel: This should probably be split into multiple lines

asgdf:
Split the condition into two lines for better readability

js/client-chat.js
```diff
@@ -563,22 +563,29 @@
...
 			case 'hidejoins':
 				if (target) {
 					var showJoins = Tools.prefs('showroomjoins') || {};
+					var hideJoins = Tools.prefs('hideroomjoins') || {};
 					var room = toId(target);
+					hideJoins[room] = true;
 					delete showJoins[room];
 					this.add('Join/leave messages on room ' + room + ': HIDDEN');
 					Tools.prefs('showroomjoins', showJoins);
+					Tools.prefs('showroomjoins', hideJoins);
```
Zarel: um

asgdf: My bad, fixed. That's what I get for improperly copypasting.

Zarel:
This isn't really my ideal solution to this problem, but I don't know what is. I think this is probably a better solution to doing nothing, anyway, so I'm willing to merge this.

@Slayer95, thoughts?

Slayer95:
In order to avoid getting a ridiculous amount of preference entries over a (half-deprecated) setting:

All joins shown
`` Tools.prefs('showjoins', {global: 0})``
All joins hidden
`` Tools.prefs('showjoins', {global: 1}); ``
All joins shown, except a few rooms
``Tools.prefs('showjoins', {global: 1, [room1]: 0, [room2]: 0});``
No joins shown, except a few rooms
``Tools.prefs('showjoins', {global: 0, [room1]: 1, [room2]: 1});``

Default: all joins hidden

As for backwards compat, the transition can be handled as a ``Storage.whenPrefsLoaded`` handler, but I wouldn't worry if the data was just wiped tbh.

asgdf:
Ok so I finally found time to get back to this, and (hopefully) implemented it along the lines of what slayer had in mind. Minus the idea with different room namespaces for different servers, since I'm not really sure how to approach that one.

Slayer95:
Looks good. Note that ``0/1`` is advantageous over ``false/true`` due to storage size concerns.

As for server namespacing, if you are willing to tackle it, it would involve prepending ``Config.server.id + '-'`` to the ``showjoins`` keys.

asgdf:
After some trying, I've decided to implement the server namespacing with nested objects instead of adding the server id to the keys directly. This way
-there is no ugly string concatenation every time you want to access a property
-the condition is shorter/more readable
-it's easier to only modify the showjoins of a single server
It's only tested with the main server though, since I couldn't get testclient to connect to other servers.

js/client-chat.js
```diff
@@ -561,29 +561,32 @@
...
+				Tools.prefs('showjoins', showjoins);
 				return false;
 			case 'hidejoins':
+				var showjoins = Tools.prefs('showjoins') || {};
+				var serverShowjoins = showjoins[Config.server.id] || {};
 				if (target) {
-					var showJoins = Tools.prefs('showroomjoins') || {};
 					var room = toId(target);
-					delete showJoins[room];
+					serverShowjoins[room] = 0;
```
Slayer95: We can probably ``delete`` when showjoins are globally disabled?

js/client-chat.js
```diff
@@ -561,29 +561,32 @@
 				return false;
 
 			case 'showjoins':
+				var showjoins = Tools.prefs('showjoins') || {};
+				var serverShowjoins = showjoins[Config.server.id] || {};
 				if (target) {
-					var showJoins = Tools.prefs('showroomjoins') || {};
 					var room = toId(target);
-					showJoins[room] = true;
+					serverShowjoins[room] = 1;
```
Slayer95: Maybe refuse when showjoins are globally enabled?

asgdf:
I assume the goal is to minimize the amount of data written to localstorage? Anyway, did that. Along the same line, I also figured that if the showjoins pref has been undefined so far, we can keep it undefined until it is needed, which shouldn't break anything.

Slayer95:
lgtm

any comments, @Zarel ?

PR #701, opened 2016-04-30
Add namelocked rank to client

Joimer:
Order it between mute and locked, make it not ignore ignores.


Zarel:
They should straight-up not show up on user lists though

Joimer:
Should be the case for locked users, right?

Zarel:
Locked as well as namelocked, but I'm waiting for bumba's refactor.

PR #702, opened 2016-05-01
Support alternate cosmetic models

QuiteQuiet:
Requires https://github.com/Zarel/Pokemon-Showdown/pull/2539 to work.

Looks like: http://imgur.com/a/yFd0C

Works for anything that currently has more than a single set of sprites uploaded (notably not Gastrodon-East's back) and an `otherForms` entry.

I'd love to hear another way to not break the import/pack, the current solution works, but hacky, assumes a bunch of things about what it got, and is slower...

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -982,6 +982,7 @@ var Tools = {
 		if (!template || typeof template === 'string') {
 			var name = template;
 			var id = toId(name);
+			var speciesid = id;
 			if (window.BattleAliases && BattleAliases[id]) {
 				name = BattleAliases[id];
 				id = toId(name);
@@ -1050,6 +1051,20 @@ var Tools = {
 				} else {
 					template.gen = 0;
 				}
+			}
+			if (template.otherForms && template.otherForms.indexOf(speciesid) >= 0) {
+				if (!window.BattlePokedexAltForms) window.BattlePokedexAltForms = {};
+				if (!window.BattlePokedexAltForms[speciesid]) {
+					template = window.BattlePokedexAltForms[speciesid] = $.extend({}, template);
+					var form = speciesid.slice(template.baseSpecies.length);
+					var formid = '-' + form;
+					form = form[0].toUpperCase() + form.slice(1);
+					template.form = form;
+					template.species = template.baseSpecies + (form ? '-' + form : '');
+					template.speciesid = toId(template.species);
+					template.spriteid = toId(template.baseSpecies) + formid;
+				}
+				template = window.BattlePokedexAltForms[speciesid];
 			}
 		}
 		return template;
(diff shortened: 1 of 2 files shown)


Zarel:
We might want to roll this into `getTemplate`. I think that might reduce a lot of this complexity?

QuiteQuiet:
Dunno if this was how you meant, but it did remove some ambiguity at least (I think)? I don't think I can get away without messing with importing/unpacking teams a little, but this feels better than before, at least...

js/search.js
```diff
@@ -927,6 +927,7 @@
 			return this.renderPokemonSortRow();
 		case 'pokemon':
 			var pokemon = BattlePokedex[id];
+			if (!pokemon) pokemon = BattlePokedex[toId(Tools.getTemplate(id).baseSpecies)];
```
QuiteQuiet: This becomes a case only now, since forms aren't entries in `BattlePokedex` unlike everything else but shouldn't break the search.

js/battledata.js
```diff
@@ -1024,6 +1025,18 @@ var Tools = {
...
 				template.formeid = formeid;
 			}
+			if (template.otherForms) {
+				var form = '', formid = '';
+				if (template.otherForms.indexOf(speciesid) >= 0) {
+					form = speciesid.slice(template.baseSpecies.length);
+					formid = '-' + form;
+					form = form[0].toUpperCase() + form.slice(1);
+				}
+				template.form = form;
```
Zarel: Ideally, you're not overwriting data in `BattlePokedex`.

You might want to make another table, and clone templates into that table (shallow-clones are fine here).

js/storage.js
```diff
@@ -765,7 +765,12 @@ Storage.unpackTeam = function (buf) {
 
 		// species
 		j = buf.indexOf('|', i);
-		set.species = Tools.getTemplate(buf.substring(i, j)).name || set.name;
+		var template = Tools.getTemplate(buf.substring(i, j));
+		set.species = template.species || set.name;
+		if (template.form) {
+			set.baseSpecies = template.baseSpecies;
+			set.form = template.form;
+		}
```
Zarel: This should no longer be necessary with the `getTemplate` change

js/battledata.js
QuiteQuiet: Would that be a table of alternative pattern templates, then? And put it in something like `BattlePokedexAltForms`?

Zarel: Yeah, I don't know if that's the best solution, but I think it's a better solution than what you're currently doing.

QuiteQuiet: Unless I'm totally misunderstanding what a shallow colone is, doesn't a shallow clone of the template still modify the original in `BattlePokedex` if you modify something in the clone?

Zarel: A shallow clone is a clone that's only one level deep.

e.g. `obj2 = $.merge({}, obj1)` is the usual way to create a shallow clone.

and then if you do `obj2.num = 2`, it won't change `obj1.num`

but if you do `obj2.point.x = 1`, it will change `obj1.point.x`.

QuiteQuiet:
I think I did? According to the jQuery documentation `$.extend()` is the method to use for Objects, though. (`$.merge()` is for Arrays, apparently) 

I removed a bunch of other code that turns out to be useless now, as well.

Zarel:
Oh, that's what I meant, then.

Zarel:
Should probably mention:

this depends on Zarel/Pokemon-Showdown#2539

Zarel:
This might be best merged in three steps, though:

1. battle.js support for this
2. server support for this
3. teambuilder UI support for this

QuiteQuiet:
So split this in two and put up a separate pullreq for the teambuilder changes?

Zarel:
Yes please

Zarel:
Only the teambuilder changes, keep the battle changes and the team loading/unloading changes together.

QuiteQuiet:
Did. Teambuilder changes are in #704 and the other here.

PR #703, opened 2016-05-01
CAP Crucibelle Model Added

Layell:


Zarel:
For future reference: https://github.com/Zarel/Pokemon-Showdown/blob/master/CONTRIBUTING.md

PR #704, opened 2016-05-04
Add UI support for alternate models

QuiteQuiet:
Depends on https://github.com/Zarel/Pokemon-Showdown-Client/pull/702 and https://github.com/Zarel/Pokemon-Showdown/pull/2539

Looks like: http://imgur.com/a/yFd0C

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -2906,6 +2906,29 @@ a.ilink.yours {
 	width: 95px;
 	text-align: right;
 }
+.changeform {
+	cursor: pointer;
+}
+.changeform i{
+	cursor: pointer;
+	display: block;
+	margin: -12px 0 -6px 0;
+	padding-left: 4px;
+	padding-top: 2px;
+	width: 12px;
+	height: 14px;
+	position: relative;
+	left: 90px;
+	border: 1px solid #569;
+	border-radius: 2px;
+	color: #569;
+	box-shadow: 0 1px 2px rgba(0,0,0,.2);
+}
+.altform {
+	margin-left: 10px;
+	margin-top: 4px;
+}
+
 
 /*********************************************************
  * Popups
@@ -2946,6 +2969,7 @@ a.ilink.yours {
 	margin: 0 -7px;
 }
 .avatarlist button,
+.formlist button,
 .bglist button {
 	display: block;
 	float: left;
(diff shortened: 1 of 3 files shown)


Slayer95:
Maybe make the sprite/model itself clickable?

QuiteQuiet:
My only concern with doing that is how do you show to users that the model is clickable without doing something with the current layout of the Pokemon card? Unless it's obvious, people will miss it.

Slayer95:
Well, the grand majority of users almost never click the ``Details`` menu.
I guess now that we don't autofix shinyness they do a bit more often?

Giving it ``cursor:pointer`` is enough to convey the clickability IMO.
We could also have both UIs...

Zarel:
I would suggest both UIs.

QuiteQuiet:
Added clickable Pokemon icons, but I still don't think it's obvious enough that the icon is clickable even with something like `cursor: pointer` but I don't have any good ideas how to show it...

Slayer95:
If you are willing to, you could add a [caret-down](https://upload.wikimedia.org/wikipedia/commons/thumb/9/9a/Caret_down_font_awesome.svg/512px-Caret_down_font_awesome.svg.png) in the bottom-right corner of the icon, I guess

QuiteQuiet:
Something like this (can be made larger)?

Zarel:
You might want an outline around it, and probably pick a color other than pure black. Maybe something like blue-gray with a white outline?

QuiteQuiet:
Bright borders were really hard to get visible, so I tried a darker version too:

Or something entirely different: (I kind of like this version, to be honest)

Zarel:
That wasn't actually what I meant but that looks pretty good, too.

QuiteQuiet:
Oh you meant a border around the entire sprite?

Is this fine, or do you want to look at the border around the sprite too? I don't think I can make the border around the sprite look good, when it's going to be broken by everything around it...

Zarel:
No, this is better than what I had in mind. Which was a border around the arrow, not a box around the arrow.

Zarel:
although adding a box-shadow anyway might still look good.

QuiteQuiet:
Updated for:

I'm still not sure this is the best colour to use here, to be fair...

style/client.css
```diff
@@ -2906,6 +2906,31 @@ a.ilink.yours {
...
+	padding-top: 2px;
+	width: 12px;
+	height: 14px;
+	position: relative;
+	left: 90px;
+	border: 1px solid #569;
+	border-radius: 2px;
+	color: #569;
+	-webkit-box-shadow: 0 1px 2px rgba(0,0,0,.2);
+	-moz-box-shadow: 0 1px 2px rgba(0,0,0,.2);
```
Slayer95: Do we still need these? @Zarel

I see we are seldom using the prefixed versions for ``box-shadow``

Zarel: I wouldn't use prefixed versions.

QuiteQuiet:
Updated.

PR #705, opened 2016-05-05
Add Facade BP increase

FlyingTypesRule:
Small thing I noticed while battling was that Facade's BP was not shown to increase from 70 to 140, so I quickly added it.


Zarel:
Um, this has a syntax error in it... did you not test it at all? o_o

Also your commit summary needs to match the standards: https://github.com/Zarel/Pokemon-Showdown/blob/master/CONTRIBUTING.md

`Tooltips: Show Facade BP increase` works.

PR #707, opened 2016-05-09
Show ability activation for confusion immunity

urkerab:
When the _Own Tempo_ Ability blocks confusion, the ability does not activate. This happens even if the server sends it (which it doesn't as yet, see Zarel/Pokemon-Showdown#2555). I moved the Ability activation block above the `switch`, and inlined that into the `else if` blocks.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -3865,25 +3865,20 @@ var Battle = (function () {
 				var poke = this.getPokemon(args[1]);
 				var effect = Tools.getEffect(args[2]);
 				var fromeffect = Tools.getEffect(kwargs.from);
-				switch (effect.id) {
-				case 'confusion':
+				if (fromeffect && fromeffect.effectType === 'Ability') {
+					var ofpoke = this.getPokemon(kwargs.of) || poke;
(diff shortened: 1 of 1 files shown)


js/battle.js
```diff
@@ -3865,25 +3865,20 @@ var Battle = (function () {
...
 				var fromeffect = Tools.getEffect(kwargs.from);
-				switch (effect.id) {
-				case 'confusion':
+				if (fromeffect && fromeffect.effectType === 'Ability') {
+					var ofpoke = this.getPokemon(kwargs.of) || poke;
+					this.resultAnim(ofpoke, fromeffect.name, 'ability');
+					this.message('', "<small>[" + ofpoke.getName(true) + "'s " + fromeffect.name + "!]</small>");
+					ofpoke.markAbility(fromeffect.name);
+				}
+				if (effect.id == 'confusion') {
```
Zarel: This should still be a switch, imo. Just to be consistent with everything else.

Zarel:
(Feel free to push that change at your leisure)

PR #713, opened 2016-05-23
Fix active pokemon ability text in tooltip

luiscarloscb:
Currently the ability displayed in the active pokemon tooltip is the base ability of the pokemon, even after Skill Swap or similar skills.

For example, after Skill Swapping No Guard:
Before this change, it incorrectly displays No Guard as the current ability:
before 
After this change, it's more consistent with what the opponent sees:
after

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -386,7 +386,13 @@ var BattleTooltips = (function () {
 		var showOtherSees = isActive;
 		if (myPokemon) {
 			if (this.battle.gen > 2) {
-				text += '<p>Ability: ' + Tools.getAbility(pokemon.ability || myPokemon.baseAbility).name;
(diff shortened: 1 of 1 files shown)


Zarel:
maybe it should be "Synchronize (base: No Guard)"

luiscarloscb:
might as well, will do : )

luiscarloscb:
Updated as recommended to show the current and base ability of the active pokemon, as a bonus it also works on the foe's active pokemon. 
capture

luiscarloscb:
Accidentally deleted the branch, restored, then realized force pushing a quick fix actually prevents me from reopening this >-<, I think I'll just open a new one o-o

Zarel:
in the future, reopen before pushing changes, and you'll be fine, I think

PR #720, opened 2016-05-27
Tooltips: fix OHKO moves accuracy text

luiscarloscb:
Currently OHKO moves show their accuracy as if they were affected by acc boosts/items/abilities (so for example, Hone Claws will increase the shown %).

This change also adds some text regarding how the true accuracy of the moves is calculated, the exact percentage can't be shown since there's no target data consistently available (target in double battles etc.).

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -837,6 +837,10 @@ var BattleTooltips = (function () {
 		}
 		if (!accuracy || accuracy === true) return '&mdash;';
 		if (ability === 'No Guard') return '&mdash; (Boosted by No Guard)';
+		if (move.ohko) {
+			if (this.battle.gen === 1) return accuracy + '% (Will fail if target\'s speed is higher)';
+			return accuracy + '% (Will fail if target\'s level is higher, increases 1% per each level above target)';
+		}
 		if (pokemon.boosts && pokemon.boosts.accuracy) {
 			if (pokemon.boosts.accuracy > 0) {
 				accuracy *= (pokemon.boosts.accuracy + 3) / 3;


js/client-battle-tooltips.js
```diff
@@ -837,6 +837,11 @@ var BattleTooltips = (function () {
 		}
 		if (!accuracy || accuracy === true) return '&mdash;';
 		if (ability === 'No Guard') return '&mdash; (Boosted by No Guard)';
+		if (move.ohko) {
+			if (this.battle.gen === 1)
+				return accuracy + '% (Will fail if target\'s speed is higher)';
```
Zarel: hm, weird that the linter isn't picking this up

Slayer95: The linter is extremely relaxed for the client, as its style is fairly more diverse than server's.

But this is one of the things that aren't seen elsewhere in the client...

luiscarloscb:
removed the newline as requested

js/client-battle-tooltips.js
```diff
@@ -837,6 +837,10 @@ var BattleTooltips = (function () {
 		}
 		if (!accuracy || accuracy === true) return '&mdash;';
 		if (ability === 'No Guard') return '&mdash; (Boosted by No Guard)';
+		if (move.ohko) {
+			if (this.battle.gen === 1) return accuracy + '% (Will fail if target\'s speed is higher)';
```
luiscarloscb: Removed the newline as requested :)

Zarel:
Works. Do I have a license from you?

luiscarloscb:
Nope, not yet, ready to provide one if you post a template

luiscarloscb:
I, luiscarloscb (Luis C. Carrasco Bravo), hereby dedicate all my contributions to the Pokémon Showdown project, past and future, under the MIT license, reproduced at https://opensource.org/licenses/MIT

Zarel:
Works, thanks!

PR #721, opened 2016-05-30
Fix teambuilder problems with alternate forms

QuiteQuiet:
- Visible changeform button in dark mode
Oversight on my part, I forgot to make an alternative for dark mode, and the current colour is practiacally invisible.

- Show all learned moves on alternate forms
The teambuilder right now only shows the pre-evolutions moveset if you pick an alternate form first.

- Stop the changeform click event from triggering when you click the Pokemon input box
Right now, if you click the input box to change pokemon and there's a Pokemon with an alternate sprite selected, the event triggers anyway.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -3115,6 +3115,10 @@ a.ilink.yours {
 	background-color: #AAAAAA;
 	color: black;
 }
+.dark .changeform i {
+	border: 1px solid #bbb;
+	color: #bbb;
+}
 
 .dark .ps-overlay {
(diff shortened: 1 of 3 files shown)


style/client.css
```diff
@@ -3115,6 +3115,10 @@ a.ilink.yours {
 	background-color: #AAAAAA;
 	color: black;
 }
+.dark .changeform i{
```
Zarel: Space after `i`

QuiteQuiet:
Added the space.

PR #724, opened 2016-06-04
Apply highlights in battles as well

asgdf:
Also fix ignoring in battles to not work on staff members.

I hope it's ok to put the highlight stuff in Tools - figured it's the best place for it to be accessible from client-chat.js and battle.js, similarly to parseChatMessage which is already there.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -5975,11 +5975,20 @@ var Battle = (function () {
 			args.shift();
 			args.shift();
 			var message = args.join('|');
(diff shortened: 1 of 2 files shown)


asgdf:
So apparently I managed to duplicate #723 a day after it was issued.Sorry about that, completely missed it :( I also forgot notifications.
So since I was told to finish this at least, I took that code and added some notification behavior on top that resembles that of regular chatrooms more closely. I'll also move the battle chat ignore fix to a separate pullreq.

PR #726, opened 2016-06-06
Tournaments: reduce the number of battle messages

sirDonovan:
Only show battle start, battle end, and disqualify messages to players
by default with an added option to always show these messages.

This was suggested in staff to allow users who don't join a tournament to continue their conversations and not be flooded with messages that don't directly affect them.

### DIFF
--- a/js/client-chat-tournament.js
+++ b/js/client-chat-tournament.js
@@ -435,7 +435,8 @@
 					break;
 
 				case 'battlestart':
-					this.room.$chat.append('<div class="notice tournament-message-battlestart"><a href="' + app.root + toRoomid(data[2]).toLowerCase() + '" class="ilink">' +
+					var roomid = toRoomid(data[2]).toLowerCase();
+					this.room.$chat.append('<div class="notice tournament-message-battlestart tournament-' + roomid + '"><a href="' + app.root + roomid + '" class="ilink">' +
 						"Tournament battle between " + Tools.escapeHTML(data[0]) + " and " + Tools.escapeHTML(data[1]) + " started." +
 						'</a></div>');
 					break;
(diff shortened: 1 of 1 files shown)


TheImmortal:
Speaking of tournament messages, I had an idea a while back but I'm not really sure how feasible it is given my lack of client expertise. What if the "Tournament battle between The Immortal and sirDonavan started." was removed when the battle ended? The link to the battle can be added to the "The Immortal has won the match 3 - 0 against sirDonavan" message. This results in one less message in the chat, which will make a big difference for big room tours, or for fast tours like CC1v1.

I don't see any downsides to this, so if you're able to do it then can you?

asgdf:
I imagine most battles take longer than it takes a room to create a page of chat, given that there is some sort of discussion going on that could be disrupted. Would retroactively hiding these battle start messages actually help to unclutter that discussion, as they'd most likely have scrolled out of view aready for anyone following the chat?
The (small) downside I see is that since chat logs are generally perceived as static, removing earlier lines might be confusing or jarring to new users. Granted, we do that for roombans and locks as well, but these are way rarer, it is explained that and why they happened, and there is an option to unhide the messages.

Zarel:
The others are maybe useless, but people who don't join battles definitely still want to watch battles.

TI's idea is interesting and feasible afaik.

sirDonovan:
How about an option to hide the messages like this?

I like TI's idea too; I can work on that.

sirDonovan:
Updated with a version of TI's idea. This had more of an effect on the flow of conversation than keeping |battleend| as a new message in the chat. A possible argument against it is that you may need to scroll up to see the results of early matches, but it's easier to look on the bracket anyway.

Zarel:
This seems to be combining both ideas. Do we really both?

sirDonovan:
I think the menu option would still be useful for some people and I've talked to others who feel the same.

Zarel:
We already have a "disable tournaments" option...

sirDonovan:
Yes, but that doesn't help those who only want to know whether or not a tournament is in progress and don't care about the individual battles. Or those who would prefer to check the bracket occasionally and not have their chat interrupted by battle messages.

Zarel:
we might merge that all under one option or something

sirDonovan:
I guess we can figure that out in a later PR. This is now only the battlestart/battleend change.

TheImmortal:
I still think my original idea is better. Showing results is good. People know how the tournament is going, who's playing/winning, etc. without having to look at the bracket. Breaking the flow of conversation isn't a problem for every tournament. It's only a problem for fast tournaments. And your proposed change has no or equal impact on fast tournaments as mine does. However, your change harms every other tournament.

sirDonovan:
Breaking the flow of conversation is also a problem for big tournaments, which I find to be more common than fast ones. Your original idea of removing past messages has no effect on this because the same amount of new messages are added to the chat throughout a tournament as with the current code. 

Having up to half the amount of battle messages in the chat with this code is an improvement for any type of tournament. I don't consider possibly needing to look at the bracket or scroll up to see results to be harmful. The bracket has tournament information organized in one place, compared to battle messages that are scattered between regular chat messages.

PR #727, opened 2016-06-07
Allow use of multiple interval operators in hl regex

asgdf:
I.e. `/hl add,abc{1,2}def{2,3}` will no longer lead to the highlights `abc{1,2}def{2` and `3}` being set, instead counting as one.


Zarel:
Wouldn't it be simple just to use `/` to denote regexes?

PR #728, opened 2016-06-15
Prevent empty highlights

QuiteQuiet:
A command like `/highlight add, , , , , , , , , , , , , , , ` currently adds all of them as empty highlights, which means every message that is sent gets highlighted.

You could probably accomplish this by changing the regex too, but I don't know how to do that.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -658,6 +658,7 @@
 					switch (targets[0]) {
 					case 'add':
 						for (var i = 1, len = targets.length; i < len; i++) {
+							if (!targets[i]) continue;
 							if (/[\\^$*+?()|{}[\]]/.test(targets[i])) {
 								// Catch any errors thrown by newly added regular expressions so they don't break the entire highlight list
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -652,9 +652,14 @@
...
 					for (var i = 0, len = targets.length; i < len; i++) {
 						targets[i] = targets[i].replace(/\n/g, '').trim();
+						if (targets[i]) {
+							validTargets.push(targets[i]);
+						}
 					}
+					targets = validTargets;
 					switch (targets[0]) {
 					case 'add':
 						for (var i = 1, len = targets.length; i < len; i++) {
```
Zarel: Why not just do `if (!targets[i]) continue` in this loop?

Wouldn't require an extra allocation and would also use a lot less code.

QuiteQuiet: Because I'm not smart enough to think of the simple solution to things.

PR #729, opened 2016-06-16
Rework /html as inline, old behavior kept as /raw

panpawn:
This makes it so the client can send /html [message] and it can parse it like a chat message with HTML.

This commit is especially helpful to side-servers that sometimes add custom formatting, such as emoticons.

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -515,6 +515,8 @@ var Tools = {
 		case 'error':
 			return '<div class="chat message-error">' + Tools.escapeHTML(target) + '</div>';
 		case 'html':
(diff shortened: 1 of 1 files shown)


Slayer95:
I've verified that ``/html`` is currently not used in the server

Zarel:
It's used in PMs, though. We should probably fix PMs in a multi-step approach.

Zarel:
But okay this is fine.

panpawn:
🎉  🌟  🎆  🎈  🎂  🎉 
Thank you, Zarel!

PR #731, opened 2016-06-18
Don't wait until after the move to activate Protean

urkerab:
I don't know whether this is the best place to add this or the best way of implementing it but it seems to work.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -5919,7 +5919,7 @@ var Battle = (function () {
 			this.spriteElems[k].html(buf + buf2);
 		}
 	};
-	Battle.prototype.runMajor = function (args, kwargs, preempt) {
+	Battle.prototype.runMajor = function (args, kwargs, preempt, nextArgs, nextKwargs) {
 		switch (args[0]) {
 		case 'start':
 			this.teamPreview(false);
@@ -6147,6 +6147,10 @@ var Battle = (function () {
 			poke.sprite.beforeMove();
 			this.useMove(poke, move, poke2, kwargs);
 			poke.sprite.afterMove();
+			if (nextArgs[0] === '-start' && nextArgs[2] === 'typechange' && nextKwargs.from === 'Protean') {
+				this.activityStep++;
+				this.runMinor(nextArgs, nextKwargs, preempt);
+			}
 			break;
 		case 'cant':
 			this.endLastTurn();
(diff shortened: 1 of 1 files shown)


Zarel:
This seems like the wrong place to do it...

urkerab:
So, it seems to work here too, if that's what you were thinking of?

Zarel:
Can you more clearly describe what the problem is and how this commit fixes it?

urkerab:
Before:
Pokemon used **Move**!
(move animates)
<small>[Pokemon's Protean]
Pokemon transformed into the Type type!</small>
After:
Pokemon used **Move**!
<small>[Pokemon's Protean]
Pokemon transformed into the Type type!</small>
(move animates)

urkerab:
Alternative approach: change some of the Protean code:

    this.attrLastMove('[still]');
    this.add('-start', source, 'typechange', type);
    if (!move.flags['charge'] || pokemon.volatiles[move.id]) this.add('-anim', source, move, target);

Zarel:
Give me some time to think about this.

urkerab:
Storm Drain and Lightning Rod's handlers activate before Protean's does, and I think Protean fails if the move has no target, so you can't change that, so Protean may not actually be the next step, and the suggested server-side change is definitely wrong.

Zarel:
This has been long-fixed, and as of 8dc8037fd8c I consider the Protean animation basically perfect.

Zarel:
https://replay.pokemonshowdown.com/oumonotype-82345404

is a good replay file to refer to, if you want to see what Protean currently looks like.

urkerab:
That's... an ancient replay. Try something newer, such as https://replay.pokemonshowdown.com/gen7pu-666651723 and you'll see the move animating before Protean's typechange.

Zarel:
It should probably be fixed to work more like the old replay, then. :[

Zarel:
It's insane to change the client to work around a server bug.

urkerab:
Server bug? Do you want Protean's message to be inserted before the last move? I guess that's doable...

Zarel:
Yes.

Zarel:
Well, actually, it depends basically purely on, how does it work in-game?

```
Pokemon used Move!
[Pokemon's Protean]
Pokemon transformed into the Type type!
(move animates)
```

or

```
[Pokemon's Protean]
Pokemon transformed into the Type type!
Pokemon used Move!
(move animates)
```

PR #732, opened 2016-06-22
Reset sprite when affected by Smack Down

asgdf:
The lastmove check is there in case a levitating mon gets hit by No Guard Smack Down while using Dig or Dive. Considering how contrived that scenario is, I'm not sure if it's worth it. Seems cleaner this way though.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -4761,6 +4761,7 @@ var Battle = (function () {
 					actions += "" + poke.getName() + " fell straight down!";
 					poke.removeVolatile('magnetrise');
 					poke.removeVolatile('telekinesis');
+					if (poke.lastmove === 'fly' || poke.lastmove === 'bounce') poke.sprite.animReset();
 					break;
 				case 'substitute':
 					if (kwargs.damage) {


js/battle.js
```diff
@@ -4761,6 +4761,7 @@ var Battle = (function () {
 					actions += "" + poke.getName() + " fell straight down!";
 					poke.removeVolatile('magnetrise');
 					poke.removeVolatile('telekinesis');
+					if (poke.lastmove === 'fly' || poke.lastmove === 'bounce') poke.sprite.animReset();
```
Zarel: Why `if (poke.lastmove === 'fly' || poke.lastmove === 'bounce') `?

asgdf: "The lastmove check is there in case a levitating mon gets hit by No Guard Smack Down while using Dig or Dive.", as in that case the Dig/Dive anim would also be canceled.

Zarel: I don't understand. It normally shouldn't be?

asgdf: If a Flying-type or levitating mon gets hit with Smack Down, that message is sent. If said mon is currently using Dive or Dig, and still gets hit by Smack Down somehow, the message is still sent (correctly, as the "remove flying type/levitate" effect is still applied).

Zarel: `animReset` is just an animation, not a message, isn't it?

And I still don't understand why we shouldn't `animReset` all the time...

asgdf: With "message", I meant the `|-start|smackdown` that the server sends to the client if Smack Down has hit something flying, that would cause the anim to be reset.
If I use Dive with Latios and it gets hit with Smack Down while diving, this message would be sent, because Smack Down has removed Levitate from Latios. If we unconditionally reset the anim then, Latios will not appear to be diving anymore. This does not reflect the state of the game, as Dive is not canceled, whereas Fly/Bounce would be.

Zarel: oooooh

PR #736, opened 2016-06-29
Prevent roomtabs unnecessarily jumping to the side

asgdf:
Only a situation where a battle is open in both the left and right room shouldn't use the hardcoding, since the right battle wouldn't get enough space otherwise.

See https://github.com/Zarel/Pokemon-Showdown-Client/commit/07a339f62c6d897f01ddeb539d28f05d49189d85#commitcomment-18000443

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -1471,7 +1471,7 @@
 			if (leftWidth < leftMinMain) {
 				leftWidth = leftMinMain;
 			}
-			if (this.curRoom.type === 'battle' && this.sideRoom.type === 'chat') {
+			if (this.curRoom.type === 'battle' && this.sideRoom.type !== 'battle') {
 				// I give up; hardcoding
 				var offset = Math.floor((available - leftMinMain - rightMin) / 2);
 				if (offset > 0) leftWidth += offset;


js/client.js
```diff
@@ -1471,7 +1471,7 @@
 			if (leftWidth < leftMinMain) {
 				leftWidth = leftMinMain;
 			}
-			if (this.curRoom.type === 'battle' && this.sideRoom.type === 'chat') {
+			if (this.curRoom.type === 'battle' && this.sideRoom.type !== 'battle') {
```
Zarel: Nooooo

Teambuilder and ladder have explicit widths. I hope this doesn't mess them up.

asgdf: They do, but they also can't be side rooms, as far as I can tell, which is why I figured this should be fine to avoid a long and unwieldy condition there. To my knowledge, only chats, battles, the rooms tab and the battles tab can be on the right side. I tried to test all possible combinations, and nothing seemed to get broken.

Zarel: Oh, at the moment they can't. That may change at a future date but for now this works, I suppose.

PR #743, opened 2016-07-15
Added recoil check for Reckless to tooltips

hvk3:


hvk3:
Hope I didn't miss out any condition.

PR #747, opened 2016-07-31
Add Room-Specific Highlights

submindraikou:
http://www.smogon.com/forums/threads/suggestions.3534365/page-38#post-6916361
Battle room highlights are under one battlehls, and adding a room specific highlight in one battle will add it in all battles.
Commands are roomadd, roomdelete, roomshow, and roomlist
Regular highlight commands now say that youve changed the global highlight list and room highlights say youve changed this rooms highlights (or replacing "this room" with the actual roomname)

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -187,12 +187,12 @@
 		getHighlight: function (message) {
 			var highlights = Tools.prefs('highlights') || {};
 			if (highlights.constructor === 'Array') this.convertHighlights();
-			var thisroom = (this.type === 'battle' ? 'battlehls' : this.id);
-			if (!app.highlightRegExp || !app.highlightRegExp[(Config.server.id + 'globalhls')] || !app.highlightRegExp[(Config.server.id + thisroom)]) {
+			var thisroom = (this.type === 'battle' ? 'battle' : this.id);
+			if (!app.highlightRegExp || !app.highlightRegExp['global'] || !app.highlightRegExp[thisroom]) {
 				try {
 					if (!app.highlightRegExp) app.highlightRegExp = [];
-					if (!app.highlightRegExp['globalhls']) this.updateHighlightRegExp(highlights[(Config.server.id + 'globalhls')], 'globalhls');
-					if (!app.highlightRegExp[thisroom]) this.updateHighlightRegExp(highlights[(Config.server.id + thisroom)], thisroom);
+					if (!app.highlightRegExp['global']) this.updateHighlightRegExp(highlights['global'], 'global');
+					if (!app.highlightRegExp[thisroom]) this.updateHighlightRegExp(highlights[thisroom], thisroom);
 				} catch (e) {
 					// If the expression above is not a regexp, we'll get here.
 					// Don't throw an exception because that would prevent the chat
@@ -204,7 +204,7 @@
 			if (!Tools.prefs('noselfhighlight') && app.user.nameRegExp) {
 				if (app.user.nameRegExp.test(message)) return true;
 			}
-			return ((highlights[(Config.server.id + 'globalhls')].length > 0) && app.highlightRegExp['globalhls'].test(message)) || ((highlights[(Config.server.id + thisroom)].length > 0) && app.highlightRegExp[thisroom].test(message));
+			return ((highlights[(Config.server.id + 'global')].length > 0) && app.highlightRegExp['global'].test(message)) || ((highlights[(Config.server.id + thisroom)].length > 0) && app.highlightRegExp[thisroom].test(message));
 		},
 		updateHighlightRegExp: function (highlights, room) {
 			// Enforce boundary for match sides, if a letter on match side is
@@ -215,7 +215,7 @@
 		},
 		convertHighlights: function () {
 			var highlights = {};
-			highlights[(Config.server.id + 'globalhls')] = Tools.prefs('highlights');
+			highlights[(Config.server.id + 'global')] = Tools.prefs('highlights');
 			Tools.prefs('highlights', highlights);
 		},
 
@@ -668,7 +668,7 @@
 					switch (targets[0]) {
 					case 'add':
 					case 'roomadd':
-						var thisroom = (targets[0] === 'add' ? 'globalhls' : (this.type === 'battle' ? 'battlehls' : this.id));
+						var thisroom = (targets[0] === 'add' ? 'global' : (this.type === 'battle' ? 'battle' : this.id));
 						var hlkey = Config.server.id + thisroom;
 						if (!highlights[hlkey]) highlights[hlkey] = [];
 						var textending = (targets[0] === 'add' ? ' everywhere: ' : (this.type === 'battle' ? ' in all battles: ' : ' in ' + this.id + ': '));
@@ -693,7 +693,7 @@
 						break;
 					case 'delete':
 					case 'roomdelete':
-						var thisroom = (targets[0] === 'delete' ? 'globalhls' : (this.type === 'battle' ? 'battlehls' : this.id));
+						var thisroom = (targets[0] === 'delete' ? 'global' : (this.type === 'battle' ? 'battle' : this.id));
 						var hlkey = Config.server.id + thisroom;
 						if (!highlights[hlkey]) highlights[hlkey] = [];
 						var textending = (targets[0] === 'delete' ? ' everywhere: ' : (this.type === 'battle' ? ' in all battles: ' : ' in ' + this.id + ': '));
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -186,6 +186,8 @@
 
 		getHighlight: function (message) {
 			var highlights = Tools.prefs('highlights') || [];
+			var roomHighlights = (this.type === 'battle' ? Tools.prefs('battlehls') || [] : Tools.prefs(this.id) || []);
```
Zarel: You're storing them directly in `Tools.prefs(roomid)`???

submindraikou: would it be better if i did it as room.id + 'hls' ? ex: lobby would be under lobbyhls

Zarel: no, you probably want it as data in `'highlights'`. You shouldn't take up more than one prefs entry for any one feature.

submindraikou:
I'm actually gonna be gone for maybe a few weeks. If anyone else wants to do this then feel free if not I'll keep doing this when I get back

js/client-chat.js
panpawn: so like, ``Tools.prefs('highlights')[roomid] = [array of highlights];``?

Morfent:
Are the extra commands for room highlights necessary?

submindraikou:
^no, theres a new commit that changes this to run in the same case instead of duplicating code

js/client-chat.js
```diff
@@ -746,28 +722,20 @@
...
+						this.add(textending);
+					} else if (target === 'show' || target === 'list' || target === 'roomshow' || target === 'roomlist') {
+						var thisroom = (target === 'show' || target === 'list' ? 'globalhls' : (this.type === 'battle' ? 'battlehls' : this.id));
+						var textending = (target === 'show' || target === 'list' ? 'urrent global highlight list' : 'urrent highlight list for this room');
 						if (serverHighlights[thisroom].length > 0) {
-							this.add("Current room highlight list: " + serverHighlights[thisroom].join(", "));
+							this.add('C' + textending + ': ' + serverHighlights['globalhls'].join(", "));
 						} else {
-							this.add('Your highlight list in this room is empty.');
+							this.add('Your c' + textending + 'is empty.');
```
panpawn: ... this is weird

if both lines of the options first word is "current", then why are you breaking it up like this..?

js/client-chat.js
```diff
@@ -746,28 +722,20 @@
...
 						Tools.prefs('highlights', allHighlights);
-						this.add("All room highlights cleared" + textending + ".");
-					} else if (target === 'roomshow' || target === 'roomlist') {
+						this.add(textending);
+					} else if (target === 'show' || target === 'list' || target === 'roomshow' || target === 'roomlist') {
+						var thisroom = (target === 'show' || target === 'list' ? 'globalhls' : (this.type === 'battle' ? 'battlehls' : this.id));
+						var textending = (target === 'show' || target === 'list' ? 'urrent global highlight list' : 'urrent highlight list for this room');
 						if (serverHighlights[thisroom].length > 0) {
-							this.add("Current room highlight list: " + serverHighlights[thisroom].join(", "));
+							this.add('C' + textending + ': ' + serverHighlights['globalhls'].join(", "));
```
panpawn: same here

js/client-chat.js
```diff
@@ -657,54 +676,66 @@
...
+						var textending = (targets[0] === 'add' ? ' everywhere: ' : (this.type === 'battle' ? ' in all battles: ' : ' in ' + this.id + ': '));
 						for (var i = 1, len = targets.length; i < len; i++) {
 							if (!targets[i]) continue;
 							if (/[\\^$*+?()|{}[\]]/.test(targets[i])) {
 								// Catch any errors thrown by newly added regular expressions so they don't break the entire highlight list
 								try {
 									new RegExp(targets[i]);
 								} catch (e) {
-									return this.add(e.message.substr(0, 28) === 'Invalid regular expression: ' ? e.message : 'Invalid regular expression: /' + targets[i] + '/: ' + e.message);
+									return this.add(e.message.substr(0, 28) === 'Invalid regular expression: ' ? e.message : 'Invalid regular expression: /' + targets[i] + '/: ' + e.message);;
```
KrisOnline: you only need one semicolon

js/client-chat.js
```diff
@@ -657,54 +676,66 @@
...
 								}
 							}
-							if (highlights.indexOf(targets[i]) > -1) {
-								return this.add(targets[i] + ' is already on your highlights list.');
+							if (serverHighlights[thisroom].indexOf(targets[i]) > -1) {
+								return this.add(targets[i] + ' is already on your' + (targets[0] === 'add' ? ' global highlights list.' : ' highlights list for this room.'));;
```
KrisOnline: same here

submindraikou:
Outside of removing "urrent" from textending and adding it into the response when a user checks their highlights (i was breaking it up before it was late night and i wasnt really thinking about the most readable way to have one capital and the other not) and fixing my typo w/ the double double semicolons is there anything else i should change in what (hopefully) is the last commit?

js/client-chat.js
```diff
@@ -185,10 +185,17 @@
...
 		getHighlight: function (message) {
-			var highlights = Tools.prefs('highlights') || [];
-			if (!app.highlightRegExp) {
+			var allHighlights = Tools.prefs('highlights') || {};
+			if (allHighlights.constructor === 'Array') this.convertHighlights();
+			var thisroom = (this.type === 'battle' ? 'battlehls' : this.id);
+			var serverHighlights = allHighlights[Config.server.id] || {};
+			if (!serverHighlights['globalhls']) serverHighlights['globalhls'] = [];
+			if (!serverHighlights[thisroom]) serverHighlights[thisroom] = [];
+			if (!app.highlightRegExp || !app.highlightRegExp['globalhls'] || !app.highlightRegExp[thisroom]) {
```
Zarel: I don't know if I like how complicated this highlight scheme is.

You can have a `serverid.roomid` key system without needing nested objects, for the record.

js/client-chat.js
```diff
@@ -185,10 +185,14 @@
...
-			var highlights = Tools.prefs('highlights') || [];
-			if (!app.highlightRegExp) {
+			var highlights = Tools.prefs('highlights') || {};
+			if (highlights.constructor === 'Array') this.convertHighlights();
+			var thisroom = (this.type === 'battle' ? 'battlehls' : this.id);
+			if (!app.highlightRegExp || !app.highlightRegExp[(Config.server.id + 'globalhls')] || !app.highlightRegExp[(Config.server.id + thisroom)]) {
 				try {
-					this.updateHighlightRegExp(highlights);
+					if (!app.highlightRegExp) app.highlightRegExp = [];
+					if (!app.highlightRegExp['globalhls']) this.updateHighlightRegExp(highlights[(Config.server.id + 'globalhls')], 'globalhls');
```
Zarel: Why are you calling them `globalhls` and `battlehls`? It seems better to call them `global` and `battle`, since those are reserved names, while `globalhls` and `battlehls` could be actual room IDs.

Zarel: Also `app . highlight regular expressions [ global highlights ]` is redundant anyway.

submindraikou: ^ are you referring to how i update a serverid + global key even though im calling a global key from highlightregexp? if so it will be fixed in the next commit.
would you prefer if i use just the roomid/global/battle as the highlightregexp key or should i use serverid + roomid/global/battle to keep it the same as the actual hls?

Zarel: I'm not sure I understand; feel free to commit your next commit and then we'll see what still needs to be done.

submindraikou:
is there anything else that needs to be changed in this pr?

js/client-chat.js
```diff
@@ -185,10 +185,14 @@
...
-			if (!app.highlightRegExp) {
+			var highlights = Tools.prefs('highlights') || {};
+			if (highlights.constructor === 'Array') this.convertHighlights();
+			var thisroom = (this.type === 'battle' ? 'battle' : this.id);
+			if (!app.highlightRegExp || !app.highlightRegExp['global'] || !app.highlightRegExp[thisroom]) {
 				try {
-					this.updateHighlightRegExp(highlights);
+					if (!app.highlightRegExp) app.highlightRegExp = [];
+					if (!app.highlightRegExp['global']) this.updateHighlightRegExp(highlights['global'], 'global');
+					if (!app.highlightRegExp[thisroom]) this.updateHighlightRegExp(highlights[thisroom], thisroom);
```
Zarel: It seems like these three lines should probably be placed inside `updateHighlightRexExp`

Zarel:
So highlights were previously an array of highlights.

They should now be an object:

```
{
  "global": ["zarel", "zar"],
  "localhost.psim.us:lobby": ["test"]
}
```

Something like this? So like "global" for global highlights (the default) and "server:roomid" otherwise.

panpawn:
>"server:roomid" otherwise.

... what about servers that are on different ports? Like, for example, ``example:8080:lobby``

Zarel:
that'd be `example-8080.psim.us`

panpawn:
... oh, my bad

Zarel:
Although maybe serverid makes more sense.

In which case we can still do `-`.

submindraikou:
for the port should i use Config.server.port or Config.server.httpport?

Zarel:
`Config.server.port`.

js/client-chat.js
```diff
@@ -657,6 +669,11 @@
 					}
 					switch (targets[0]) {
 					case 'add':
+					case 'roomadd':
+						var thisroom = (targets[0] === 'add' ? 'global' : (this.type === 'battle' ? 'battle' : this.id));
+						var hlkey = (thisroom === 'global' ? 'global' : Config.server.id + '-' + Config.server.port + ':' + thisroom);
```
Zarel: `'-' + Config.server.port` is redundant here.

Zarel: You're going to generate things like `127.0.0.1:8001-8001:lobby` and `gold-8000:lobby`.

Zarel: Maybe we should just do `serverid#roomid` or `serverid/roomid`

PR #749, opened 2016-07-31
Add .dark .teambuilder-clipboard-data

panpawn:
This helps to make this property more visible in darkmode

BEFORE:

AFTER:


panpawn:
I would like to note I noticed this issue from a post in the suggestions thread:

PR #750, opened 2016-08-02
Chat: make hidden message reveal staff-only

team1257Simon:
This is in tandem with https://github.com/Zarel/Pokemon-Showdown/pull/2645, and makes regular users unable to unhide text from banned users if the text is cleared using /unlogtext. Staff (moderator and up) retain this ability.


kotarou3:
The client seems the wrong place to do this...

monsanto:
Thanks for the PR but it doesn't seem like there was enough interest at the time, and it's been a really long time since.

PR #751, opened 2016-08-09
Change darkmode disabled button color

panpawn:
This should make disabled button text more readable while in darkmode

BEFORE:

AFTER:

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -88,6 +88,9 @@ button:disabled {
 	cursor: default;
 }
 
+.dark button:disabled {
+	color: #C0C0C0;
+}
 .dark a {
 	color: #4488EE;
 }


Zarel:
This will break it for certain color schemes and fix it for others.

Don't specify a color without specifying a background color.

panpawn:
Closing after discussion about this in the development room.

> %panpawn: Zarel, I'm kind of confused how client issue#751 would break some color schemes - aren't the only color schemes darkmode and not darkmode, or am I misunderstanding what you meant?
#Zarel: panpawn, you're using OS native buttons
%panpawn: oh...
#Zarel: so the buttons will look different depending on what OS you're using
%panpawn: that makes sense
%panpawn: thanks!
#Zarel: tbh I'd just leave as-is, they're not that unreadable
%panpawn: alright, that's fine
%panpawn: if it was an easy fix like that, I just thought it'd be worth it

PR #753, opened 2016-08-10
Update move animations, round 12.8

Ridaz:
Now with Feathers.
- New Feather Dance, Roost animations
- Added Leafage & Solar Blade animations

All changes can be viewed [here](http://www.smogon.com/forums/threads/move-ability-animations-request-thread.3568451/).


ascriptmaster:
\> update 12.5
\> followed by update 12.8

Are we Touhou now?

PR #756, opened 2016-08-13
Export folders

acz13:
I was sleep drunk when I slapped this together so it probably needs some cleaning up but it works...

not that I didn't add import to folder because you can just append it to restore all...

Folder view

Export view

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -146,8 +146,13 @@
 			var buf = '';
 
 			if (this.exportMode) {
-				buf = '<div class="pad"><button name="back"><i class="fa fa-chevron-left"></i> List</button> <button name="saveBackup" class="savebutton"><i class="fa fa-floppy-o"></i> Save</button></div>';
-				buf += '<div class="teamedit"><textarea class="textbox" rows="17">' + Tools.escapeHTML(Storage.exportAllTeams()) + '</textarea></div>';
+				if (this.curFolder) {
+					buf = '<div class="pad"><button name="back"><i class="fa fa-chevron-left"></i> List</button></div>';
+					buf += '<div class="teamedit"><textarea readonly class="textbox" rows="17">' + Tools.escapeHTML(Storage.exportFolder(this.curFolder)) + '</textarea></div>';
+				} else {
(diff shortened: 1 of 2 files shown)


js/storage.js
```diff
@@ -1117,6 +1117,18 @@ Storage.exportAllTeams = function () {
 	}
 	return buf;
 };
+Storage.exportFolder = function (folder) {
+	var buf = '';
+	for (var i = 0, len = Storage.teams.length; i < len; i++) {
+		var team = Storage.teams[i];
+		if ((folder.indexOf("/") && team.folder === folder.slice(0, -1)) || team.format === folder) {
```
Zarel: `folder.indexOf("/")` and `folder.slice(0, -1)` are weird things to do in an inner loop.

Not to mention `folder.indexOf("/")` should probably be `folder.slice(-1) === '/'`

acz13: do you have a better solution?

i thought to use `folder.indexOf("/")` because it's easier lol

acz13:
@Zarel is this good now?

cleaned up messy if statement and squashed commits

js/client-teambuilder.js
```diff
@@ -404,6 +409,8 @@
 
 			if (window.nodewebkit) {
 				buf += '<button name="revealFolder" class="button"><i class="fa fa-folder-open"></i> Reveal teams folder</button> <button name="reloadTeamsFolder" class="button"><i class="fa fa-refresh"></i> Reload teams files</button> <button name="backup" class="button"><i class="fa fa-upload"></i> Backup/Restore all teams</button>';
+			} else if (this.curFolder) {
+				buf += '<button name="backup" class="button"><i class="fa fa-upload"></i> Backup all teams from this folder</button>';
```
Zarel: Maybe still have a `Backup all teams` button under it or something?

Zarel: I guess that would be ambiguous. This is fine.

Zarel:
Sounds good. Can you confirm you want it (and all your future contributions to Pokémon Showdown) released under the MIT license?

acz13:
MIT license is good

acz13:
I just realized that this doesn't work with the (uncategorized) or genN folders... should I remove the button from those or add more code to make it work. I don't see much reason to export those anyways...

Zarel:
I'd say remove the button from those.

PR #757, opened 2016-08-13
Allow the battle engine to manually cure status after -cureteam

urkerab:
Once this is live, Aromatherapy and Heal Bell can be changed to `this.add('-cureteam', '[from] ' + move, '[of] ' + pokemon);` which I'll submit in a separate PR. This allows old replays to continue to display the whole team being cured which is slightly less incorrect than there not being anything cured.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -4206,7 +4206,7 @@ var Battle = (function () {
 				}
 				break;
 
-			case '-cureteam':
+			case '-cureteam': // For old gens when the whole team was always cured
 				var poke = this.getPokemon(args[1]);
 				for (var k = 0; k < poke.side.pokemon.length; k++) {
 					poke.side.pokemon[k].status = '';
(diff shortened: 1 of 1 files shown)


Zarel:
?

Zarel:
What's wrong with the current syntax of passing the pokemon in `args[1]`?

urkerab:
Because that's what old replays do, so if we just stop curing the team, they will appear to break.

Zarel:
Um, why not just `-activate`? If the activation doesn't cure the team, `-cureteam` is entirely the wrong minor action to use.

urkerab:
OK, so then Aromatherapy/Heal Bell can just use this.add('-activate', pokemon, move); ?

Zarel:
Yeah.

urkerab:
Rewrote the comment as the old code is still useful for old gens.

PR #759, opened 2016-08-16
Allow status to be cured silently

urkerab:
Heal Bell can't cure soundproof allies in Generation 3 or 4, so we can't use `-cureteam`, but it doesn't list cured Pokémon, so we need to be able to make it `[silent]`.


Zarel:
We may want to list them in parentheses anyway?

PR #760, opened 2016-08-24
Implement |m:|time|message

panpawn:
This basically does the same thing as ``case '':``, except this has timestamps (if enabled, of course.)

((And, it doesn't highlight you! yay!))

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1141,6 +1141,11 @@
 					this.addChat(row[2], row.slice(3).join('|'), false, msgTime);
 					break;
 
+				case 'm:':
+					var msgTime = this.timeOffset + (parseInt(row[1], 10) || 0);
+					this.$chat.append('<div class="notice">' + ChatRoom.getTimestamp('chat', msgTime) + Tools.escapeHTML(row.slice(2).join('|')) + '</div>');
+					break;
+
 				case 'tc':
 					if (/[a-zA-Z0-9]/.test(row[2].charAt(0))) row[2] = ' ' + row[2];
 					var msgTime = row[1] ? ~~(Date.now() / 1000) - (parseInt(row[1], 10) || 0) : 0;
@@ -1248,6 +1253,7 @@
 					// fallthrough in case of unparsed message
 
 				case '':
+				case 'm':
 					this.$chat.append('<div class="notice">' + Tools.escapeHTML(row.slice(1).join('|')) + '</div>');
 					break;
 


Slayer95:
:| We have ``|c||[MESSAGE]`` for this...

panpawn:
>:| We have |c||[MESSAGE] for this...

That still highlights users, though.

js/client-chat.js
```diff
@@ -1247,6 +1247,11 @@
 					if (!this.tournamentBox.parseMessage(row.slice(1), row[0] === 'tournaments')) break;
 					// fallthrough in case of unparsed message
```
QuiteQuiet: This is going to cause problems when you're changing what the fallthrough is, isn't it?

panpawn: I tested it in this position and it worked fine, however, for readability, moving it up is probably a good idea.

Thanks!

panpawn: Done!

Zarel:
I guess we could have `|:|` or something? Either way, calling it `timestamp` doesn't match the others...

panpawn:
I'll update this soon.

panpawn:
Updated.

(Note: after further discussion, we noticed ``|:|`` is already a thing, so ``|m:|`` is probably the next-best name for it)

panpawn:
o.o...

Morfent:
Don't SockJS messages already include timestamps in them...?

Zarel:
I don't believe so? Either way, this is something else entirely. When you join a room, the roomlog is sent to you all at once. Sending you one line per message would be ridiculous.

Morfent:
It is, but I'm not entirely sure what it's meant to be a timestamp of, so it may not have been much use to us. I wasn't entirely clear what the rationale behind including it in `|c:|` and now `|m:|` was when I asked

Morfent:
Actually, I'll take a look to see if the SockJS could be used, so `|c:|` and `|m:|` could maybe only be something used for the chunk of roomlogs sent on join rather than for all messages. Might get lucky and be able to save a minute amount of bandwidth

Zarel:
...I would rather not complicate the protocol and possibly introduce desyncs to save a minute amount of bandwidth.

Morfent:
Alright then. It ended up turning out to not be useful, since it ended up being a timestamp taken earlier than the one included in chat/these kind of messages. Not sure exactly what it's meant to be used for, but I could look at SockJS's source code in the off chance whatever it's timestamping could be used for something else.

panpawn:
Is there anything I should be doing with this?

Morfent:
I doubt it

Zarel:
The thing is, we don't have too much need for a thing that doesn't highlight.

Way more useful would be a thing that highlights normally, but doesn't highlight a specific person.

Which is why I said something like

    |c:|timestamp| username|/noname text

would probably be a lot better. Or maybe something like `/notify `

Zarel:
I'm actually really liking `/notify `

panpawn:
I'll work on this as soon as I have the time.  Thanks for getting back to me!

Zarel:
(I actually already implemented it myself by the time you said that)

PR #761, opened 2016-08-24
Add 'wall' as an alias to 'announce'

panpawn:
This is a server-side alias, so maybe it should be a client-side alias, too.


Zarel:
??? why?

panpawn:
>/wall doesn't work in PM

But ``/announce`` does

Zarel:
...you can fix that server-side...

PR #762, opened 2016-08-26
Update move animations, round 13

Ridaz:
- Added a new Shell particle
- New Coil, Razor Shell, Shell Smash, Withdraw animations
- New Z-Move activation animation
- Updated Agility & Work Up animations (Work up uses Bulk Up animation)

Need to do:
- Z-Move animations
- Gear move animations
- Add gear particle

Until the todo list isn't complete, don't merge this.

All changes can be viewed [here](http://www.smogon.com/forums/threads/move-ability-animations-request-thread.3568451).


Zarel:
Should we at some point consider not preloading every single particle before a battle? I'm getting the impression that we're starting to amass a lot of particles...

Ridaz:
Zarel please, I wasn't finished.
But yeah, if I want to animate every move/ability properly, a few more particles are going to be added. Not a lot though, hopefully

ascriptmaster:
How would we not preload everything before battle and not have loading errors with images not appearing fast enough?

Zarel:
Certain images only need to be loaded sometimes. We could, for instance, skip the primal particles if there are no Kyogres/Groudons in Team Preview.

PR #768, opened 2016-09-21
Add FormatMod Support for OMs

CamiloTD:
This is my first pull, if im doing something wrong please tell me, what i did is simple, i wanted to make PS to recognize OMs and adjust pokemon stats depending of the format, for testing i created the "Tier Shift" support.

How does it works? 

When update() is called on client-teambuilder.js i call a function named checkFormatMods(), in that function i check for existing mods and i apply it if didnt applied before. If there is no mod for the current format, then delete the current mod.

In search.js functions, instead of using global BattlePokedex, im creating a scope variable that overrides the global one, and i give it the value of the current mod or the global object depending of the situation.

This system is extensible and i'll make it more sophisticated, but... tomorrow.

No more changes.

### DIFF
--- a/js/client-formatmods.js
+++ b/js/client-formatmods.js
@@ -6,7 +6,7 @@
 
 			var keys = Object.keys(BattlePokedex);
 
-			for (var i = 0, key, poke; key < keys.length;) {
+			for (var i = 0, key, poke; i < keys.length;) {
 				key = keys[i++];
 				poke = $.extend(true, {}, BattlePokedex[key]);
 
--- a/js/client-formatmods.js
+++ b/js/client-formatmods.js
@@ -1,49 +1,53 @@
-(function(exports, $){
+(function (exports, $) {
+
 	var FormatMods = exports.FormatMods = ({
-		tiershift: function tierShiftMod () {
+		tiershift: function tierShiftMod() {
 			var mod = {};
 
 			var keys = Object.keys(BattlePokedex);
 
-			for(var i=0, key, poke;key = keys[i++];){
+			for (var i = 0, key, poke; key < keys.length;) {
+				key = keys[i++];
 				poke = $.extend(true, {}, BattlePokedex[key]);
 
-				switch(poke.tier){
-					case "UU":
-					case "BL2": FormatMods._IncreaseStats(poke, 5); break;
+				switch (poke.tier) {
 
-					case "RU":
-					case "BL3": FormatMods._IncreaseStats(poke, 10); break;
+				case "UU":
+				case "BL2": FormatMods._IncreaseStats(poke, 5); break;
 
-					case "NU":
-					case "BL4": FormatMods._IncreaseStats(poke, 15); break;
+				case "RU":
+				case "BL3": FormatMods._IncreaseStats(poke, 10); break;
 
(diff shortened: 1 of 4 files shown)


js/client-teambuilder.js
```diff
@@ -3035,26 +3046,57 @@
...
+				} else if (SRweak >= 2 || hasMove['bellydrum']) {
+					hpDivisibility = 2;
+				} else if (SRweak >= 1 || hasMove['substitute'] || hasMove['transform']) {
+					hpDivisibility = 4;
+				} else {
+					hpDivisibility = 8;
+				}
+
+				if (hpDivisibility) {
+					while (hp < 252 && evTotal < 508 && !(stat % hpDivisibility) !== hpShouldBeDivisible) {
```
Zarel: This should not be a part of your diff. Are you unfamiliar with Git? This part needs to be rebased - I'm sure someone can tell you how to if you have trouble.

js/client-formatmods.js
```diff
@@ -0,0 +1,49 @@
+(function(exports, $){
```
Zarel: This should be `(function (exports, $){`

Which is the first of around 30 things Travis CI is currently telling you is wrong. Do you know how to use `npm test`? That will give you the same errors Travis CI is giving, without needing to manually push and check this page.

js/client-formatmods.js
```diff
@@ -0,0 +1,49 @@
...
+		stats.def = Math.clamp(stats.def + quant, 1, 255);
+		stats.hp  = Math.clamp(stats.hp  + quant, 1, 255);
+		stats.spa = Math.clamp(stats.spa + quant, 1, 255);
+		stats.spd = Math.clamp(stats.spd + quant, 1, 255);
+		stats.spe = Math.clamp(stats.spe + quant, 1, 255);
+	}
+
+	Math.clamp = function clamp(num, min, max){
+		return num < min? min : num > max? max : num;
+	}
```
Zarel: You should probably not attach new functions to native objects. That's why we dump this in Tools.

Zarel:
In the end, though, my biggest feedback is that you should probably just use the build script for this.

Zarel:
Specifically, that you should edit the already-existing `./githooks/build-indexes` script to include OM mods, rather than this messier system.

js/client-teambuilder.js
AustinXII: Adding a check for magic guard here would be nice

js/client-formatmods.js
CamiloTD: Jaja okey, gonna look for some global Tool Object

js/client-teambuilder.js
CamiloTD: Yeah... i never used Github, this is my first commit, so i use Sublime Text and then paste the code, idk what could be happening

Zarel: Yeah, you can't just paste the code from a different version, you have to use the same version.

Zarel: Search for a guide on rebasing or ask us for help.

Zarel:
This has been unmaintained for around half a year and does not implement things the way we need them; closing.

PR #773, opened 2016-10-02
Port /reply to the client

bumbadadabum:


js/client-chat.js
```diff
@@ -857,6 +857,13 @@
 				}
 				return text; // Send the /avatar command through to the server.
 
+			case 'reply':
```
Zarel: The replacement should be immediate upon typing `/r ` or `/reply `, it shouldn't wait until you press Enter.

bumbadadabum:
Makes sense, I'll change that.

Op 2 okt. 2016 22:53 schreef "Guangcong Luo" <notifications@github.com>:
>
> @Zarel requested changes on this pull request.
>
> ________________________________
>
> In js/client-chat.js:
>
> > @@ -857,6 +857,13 @@ } return text; // Send the /avatar command through
to the server. + case 'reply':
>
> The replacement should be immediate upon typing /r  or /reply , it
shouldn't wait until you press Enter.

js/client-chat.js
```diff
@@ -148,6 +148,8 @@
 					e.preventDefault();
 					e.stopPropagation();
 				}
+			} else if (app.user.lastPM && textbox.value === '/reply' && e.keyCode === 32) { // '/reply ' is being written
```
panpawn: Are we dropping the ``/r`` alias?

bumbadadabum: Now we don't anymore.

js/client-chat.js
```diff
@@ -148,6 +148,11 @@
 					e.preventDefault();
 					e.stopPropagation();
 				}
+			} else if (app.user.lastPM && (textbox.value === '/reply' || textbox.value === '/r') && e.keyCode === 32) { // '/reply ' is being written
+				e.preventDefault();
+				e.stopPropagation();
+				var val = textbox.value = '/pm ' + app.user.lastPM + ', ';
```
```
var val = '/pm ' + app.user.lastPM + ', ';
textbox.value = val;
```

Don't use fancy tricks on client; you don't know what could cause a deoptimization :|

kotarou3: That isn't even a fancy trick. That's like a common thing in basically any imperative language

Zarel: UI responsiveness is not IO bound.

This is a common thing in basically any imperative language, yes, but I don't really want to take chances that any engine interprets it as a DOM read.

PR #778, opened 2016-10-05
Move Join other room up

acz13:
Discussion from dev room


panpawn:
Screen cap?

Zarel:
The phrasing "other" implies that it comes after the list of rooms. If you put it elsewhere, it probably needs a different name.

I also don't think "next to Hide" is a good place for it. Maybe I could see it above "Official chat rooms"?

Zarel:
Maybe a straight-up "Join chatroom by name" box

acz13:
@Zarel do you mean to the left of the Meloetta or where?

panpawn:
He means above where it says "Official chat rooms" (big bold letters above the Lobby in that screen cap), and having it be a text box for the user to use, I believe

Zarel:
Yes, what panpawn said.

acz13:
Okay... I guess I'll do that when I get home

though honestly it's really quick someone else could probs do it

Zarel:
...apparently no one did?

acz13:
Lol,  
I'll start it now since it's the weekend

XD

Zarel:
...?

PR #780, opened 2016-10-07
Mark damage-weakening berries as having been eaten

urkerab:
Damage-weakening berries send two berry events, but the second one clears the reason from the first, so we have to set it again.


Zarel:
There should only be one `-enditem` event. If there are two, the second should be `-activate`.

PR #781, opened 2016-10-11
Darkmode: Change roomtab color scheme from default

panpawn:
Screencap:

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -122,6 +122,37 @@ button:disabled {
 .dark .textbox:focus {
 	background: #111111;
 }
+.dark .roomtab.button.cur.closable {
+	background: #636363;
+	border-color: #111;
+	color: #F9F9F9;
+	text-shadow: 1px 1px 0 #111;
+}
+.dark .roomtab.button.closable,
+.dark .roomtab.button.cur,
+.dark .roomtab.button {
+	background: #3A3A3A;
+	border-color: #111;
+	border-bottom: thick solid #3A3A3A;
+	color: #F9F9F9;
+	text-shadow: 1px 1px 0 #111;
+}
+.dark .roomtab.button:hover {
+	background: #484848;
+}
+.dark .tabbar a.button.subtle-notifying {
+	color: #6BACC5;
+}
+.dark .tabbar a.button.notifying {
+	background: #6BACC5;
+	border-color: #6BACC5;
+	color: #000;
+	text-shadow: none;
+}
+.dark .maintabbarbottom {
+	background: #636363;
+	border-color: #111;
+}
 
 /* .dark button {
 	background: #777777;


panpawn:
Updated Screencap:

Zarel:
I don't think that necessarily looks better. The active tab should be at least the same color as the bar below it.

I'm also not sure I like that color scheme. Looks a bit too "Halloween"...

panpawn:
>The active tab should be at least the same color as the bar below it.

It is. (see second screen cap; "Studio" is the active room)

panpawn:
>I'm also not sure I like that color scheme.

Is it specifically with the orange?  Would you prefer me to look into a different color combination; perhaps blue and black?

Zarel:
There's a gap between the tab and the bar under it, which shouldn't be there. Compare the current:

image

I think a more yellowish orange would make it look less Halloweeny. Also consider toning down saturation/brightness.

WallyTheBully:
Updating this part of the darkmode CSS may be a good idea, however a linear-gradient might be better with other colors

panpawn:
I will be looking into updating this shortly.  Thank you for your feedback thus far!

panpawn:
>however a linear-gradient might be better with other colors 

I'm not sure I agree about that for this part of the CSS, trying to remain as close to the opposite of what not-darkmode looks like as possible for this, specifically... as in, color wise, if that makes sense?  What do other people think about this?

Here is my updated screen cap:


Does that look better?

Specifically, I addressed the gap you were talking about that shouldn't of been there, and I also revised the color to be more of "yellowish orange".

Feedback much appreciated!

panpawn:
It was pointed out to me that the highlighted room border looked off, so I went ahead and fixed that:

Zarel:
It needs to be a _lot_ more yellowish.

You still have that weird black line between the tab and the separator.

panpawn:
>You still have that weird black line between the tab and the separator.

I was toggling between this darkmode and lightmode, and lightmode has a faint gray bar between the separator, much like this faint black bar.  I'm confused :|

panpawn:
@Zarel - it looks different on different browsers.  On Firefox (which is what I'm using), lightmode room tabs look like this:

Are you on Chrome?

panpawn:
Updated screen cap:

panpawn:
Alternate idea, since this this "more yellow" is getting cringe-worthy:

This uses the blue from the PS logo, so that it looks more complimentary to the rest of PS (kind of).

Thoughts?  Should I go back on the [yellow brick road](https://s-media-cache-ak0.pinimg.com/originals/ea/b0/7f/eab07f7969e91626f0acd8f8f295f8af.jpg)?

(PS - this is updated to use the previous posts yellow - **NOT** this blue!)

Zarel:
Blue works, although cool colors aren't great for a dark theme, but I guess it's worth a try.

First, though, that blue needs to be lightened and desaturated. Right now you don't have enough contrast.

panpawn:
Updated with:

Zarel:
Should definitely be desaturated. And you still haven't fixed the border issue.

panpawn:
>And you still haven't fixed the border issue.

... I'm not sure I understand what you mean by that :|

Zarel:
> @Zarel - it looks different on different browsers

This actually has nothing to do with different browsers, this is all about zoom level. Set zoom level to 100% or 200% on either Chrome or Firefox, and it'll look like my screenshot.

panpawn:
ohh, okay, I just tried to see what it looks like in 100% zoom level for me, here's what it looks like:

panpawn:
Updated screen cap:

Zarel:
Still need you to desaturate it...

panpawn:
Updated screen cap:

panpawn:
Is that better?

Zarel:
It's better, but maybe a bit darker?

panpawn:
Updated with:

panpawn:
Is that good?

Zarel:
Better. Possibly good enough.

panpawn:
Is it... good enough?

Zarel:
*exaggerated sigh*

I guess it's better than what we currently have...

Zarel:
And the reason I left it on hold for so long is that there's a lot of minor things wrong with it.

I fixed these: 0529b95ffe560b0

1. active tabs without close buttons weren't being styled correctly (they were displaying like inactive tabs)
2. active tabs without close buttons were having their styles overridden on hover
3. the bottom bar color was weird

Still broken:

1. no active states: background color should be different when mouse is down

and some other stuff

panpawn:
Also, now broken is hovering over tabs that you've been highlighted in.

PR #782, opened 2016-10-12
Support |c|USER|/logstrong MESSAGE

panpawn:
Basically the same thing as `/log`, the only difference is that it's bold.


Zarel:
...why?

Zarel:
Especially since `/log **message**` should already do whatever you're trying to do...

bumbadadabum:
Also, if you're going to do this (and I don't think you should), don't call
it "logstrong"

PR #783, opened 2016-10-14
Allow users to ignore nicknames

QuiteQuiet:
A bunch of people have requested this, iirc, because they don't want to see their opponent's nicknames but aren't interested in ignoring them entirely.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -27,7 +27,7 @@
 
 			this.$chat = this.$chatFrame.find('.inner');
 
-			this.$options = this.battle.optionsElem.html('<div style="padding-top: 3px; text-align: right"><label style="font-size: 8pt; padding: 3px 5px"><input type="checkbox" name="ignorespects" /> Ignore Spectators</label> <label style="font-size: 8pt; padding: 3px 5px"><input type="checkbox" name="ignoreopp" /> Ignore Players</label></div>');
+			this.$options = this.battle.optionsElem.html('<div style="padding-top: 3px; text-align: right"><label style="font-size: 8pt; padding: 3px 5px"><input type="checkbox" name="ignorespects" /> Ignore Spectators</label> <label style="font-size: 8pt; padding: 3px 5px"><input type="checkbox" name="ignoreopp" /> Ignore Players</label> <label style="font-size: 8pt; padding: 3px 5px"><input type="checkbox" name="ignorenicks" /> Ignore Nicknames</label></div>');
 
 			this.battle.customCallback = _.bind(this.updateControls, this);
 			this.battle.endCallback = _.bind(this.updateControls, this);
@@ -38,6 +38,7 @@
 		},
 		events: {
 			'change input[name=ignorespects]': 'toggleIgnoreSpects',
+			'change input[name=ignorenicks]': 'toggleIgnoreNicks',
 			'change input[name=ignoreopp]': 'toggleIgnoreOpponent',
 			'click .replayDownloadButton': 'clickReplayDownloadButton'
 		},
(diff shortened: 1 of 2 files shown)


Zarel:
I don't think we have space for three checkboxes and a userlist. How does this look?

QuiteQuiet:
Best case | Worst case 


How it looks really depends on how wide the battle log is right now, but that's true for its current state too with just two checkboxes. Ideally they wouldn't be part of the battle log at all, I think.

Zarel:
The top left area is reserved for a userlist...

I think we need to get a better idea of what kind of nicknames are being used that people don't like. I don't think this is necessary.

QuiteQuiet:
My fairly standard laptop resolution (1366x768) already can't fit a userlist in the space that is left over from the Ignore Spectators / Ignore Opponent buttons unless I close any open chat rooms (http://puu.sh/rITcC.jpg) and I'd imagine everyone else with relatively small monitors are already in this situation.

From what I recall there was complaints about things like japaneese nicknames, smileys, and emoticons being used in nicknames. Nothing harmful, just things they don't want to see when they battle.

Zarel:
That seems like it belongs in an Options menu. You'd want it to be turned on before the battle starts, and you'd want it to apply to multiple battles?

QuiteQuiet:
I guess that's a somewhat safe assumption. Though, I don't really know where you'd put another options menu to be honest. I guess as a submenu in the existing options menu would work, but then it'd be nowhere near the other "ignore" options. Alternative would be to put Ignore Opponent/Ignore Spectators in a menu too and add this in there too.

Zarel:
We could replace the checkboxes with an Ignore menu. "In this battle" section, then an "All battles" section.

QuiteQuiet:
Something like this?

js/client-battle.js
```diff
@@ -1110,4 +1098,44 @@
...
+		toggleNicknames: function (side) {
+			for (var i = 0; i < side.active.length; i++) {
+				side.active[i].statbarElem.html(side.getStatbarHTML(side.active[i], true));
+				side.updateStatbar(side.active[i], true, true);
+			}
+			side.updateSidebar();
+		},
+	});
+
+}).call(this, jQuery);
```
Zarel: .

QuiteQuiet: That's exactly like it is currently too, so what's wrong with it? It just moved down ~30 lines :|

Zarel: No, it currently has a trailing linebreak, which you lost.

Trailing linebreaks are bad form; we should probably update eslint to warn about them.

Zarel: image
That's what the red symbol at the end is warning about.

Zarel: For some reason that symbol isn't showing up in GitHub's "Conversation" tab, but you should be able to tell by the fact that it's green and not white that you changed _something_

js/client-battle.js
```diff
@@ -1105,4 +1098,44 @@
...
+		},
+		toggleIgnoreOpponent: function (e) {
+			this.battle.ignoreOpponent = !!e.currentTarget.checked;
+			this.battle.add('Opponent ' + (this.battle.ignoreOpponent ? '' : 'no longer ') + 'ignored.');
+		},
+		toggleNicknames: function (side) {
+			for (var i = 0; i < side.active.length; i++) {
+				side.active[i].statbarElem.html(side.getStatbarHTML(side.active[i], true));
+				side.updateStatbar(side.active[i], true, true);
+			}
```
Zarel: Should you also update the log? That sounds difficult but it might be possible.

Zarel: Well, those all _should_ update the log. Especially Ignore Spectators.

QuiteQuiet: Do you want it included in this pullreq, or separate?

Zarel: Separate is fine.

PR #784, opened 2016-10-17
Update battle log when ignoring something

QuiteQuiet:
Continuation from the suggestion in https://github.com/Zarel/Pokemon-Showdown-Client/pull/783. Battle chat now updates when someone clicks Ignore Opponent / Spectators / Nicknames and hides what the user doesn't want to see.

Hiding nicknames is about as efficient as I can get it, I think. I've done my best to keep the html structure (and hover-over things) intact but I may have missed something. It's not the most elegant method but it seems to work fine...

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -524,18 +524,21 @@ var Pokemon = (function () {
 			this.baseAbility = ability;
 		}
 	};
+	Pokemon.prototype.htmlName = function () {
+		return '<span class="battle-nickname' + (this.side.n === 0 ? '' : '-foe') + '" title="' + this.species + '">' + Tools.escapeHTML(this.name) + '</span>';
+	};
 	Pokemon.prototype.getName = function (shortName) {
 		if (this.side.n === 0) {
-			return Tools.escapeHTML(this.name);
+			return this.htmlName();
 		} else {
-			return (shortName ? "Opposing " : "The opposing ") + (this.side.battle.ignoreOpponent || this.side.battle.ignoreNicks ? this.species : Tools.escapeHTML(this.name));
+			return (shortName ? "Opposing " : "The opposing ") + this.htmlName();
 		}
 	};
 	Pokemon.prototype.getLowerName = function (shortName) {
 		if (this.side.n === 0) {
-			return Tools.escapeHTML(this.name);
+			return this.htmlName();
 		} else {
-			return (shortName ? "opposing " : "the opposing ") + (this.side.battle.ignoreOpponent || this.side.battle.ignoreNicks ? this.species : Tools.escapeHTML(this.name));
+			return (shortName ? "opposing " : "the opposing ") + this.htmlName();
 		}
 	};
 	Pokemon.prototype.getTitle = function () {
@@ -553,7 +556,7 @@ var Pokemon = (function () {
 			if (plaintext) {
 				name += ' (' + this.species + ')';
 			} else {
-				name += ' <small>(' + this.species + ')</small>';
+				name = '<span class="battle-nickname' + (this.side && this.side.n === 0 ? '' : '-foe') + '" title="' + this.species + '">' + name + ' <small>(' + this.species + ')</small>' + '</span>';
 			}
 		}
 		if (plaintext) {
@@ -2475,6 +2478,11 @@ var Battle = (function () {
 
 		if (this.mySide) this.mySide.reset();
 		if (this.yourSide) this.yourSide.reset();
+
+		if (this.ignoreNicks) {
+			var $log = $('.battle-log .inner');
+			if ($log.length) $log.addClass('hidenicks');
+		}
 
 		// activity queue state
 		this.animationDelay = 0;
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1117,6 +1117,13 @@
 		toggleIgnoreSpects: function (e) {
 			this.battle.ignoreSpects = !!e.currentTarget.checked;
 			this.battle.add('Spectators ' + (this.battle.ignoreSpects ? '' : 'no longer ') + 'ignored.');
+			var $messages = $('.battle-log').find('.chat').not('small:contains(\u2605)');
+			if (!$messages.length) return;
+			if (this.battle.ignoreSpects) {
+				$messages.hide();
+			} else {
+				$messages.show();
+			}
 		},
 		toggleIgnoreNicks: function (e) {
 			this.battle.ignoreNicks = !!e.currentTarget.checked;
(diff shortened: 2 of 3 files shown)


js/client-battle.js
```diff
@@ -1117,20 +1117,61 @@
 		toggleIgnoreSpects: function (e) {
 			this.battle.ignoreSpects = !!e.currentTarget.checked;
 			this.battle.add('Spectators ' + (this.battle.ignoreSpects ? '' : 'no longer ') + 'ignored.');
+			var $messages = $('.battle-log').find('[class*="chatmessage-"]').not('.chatmessage-' + this.battle.p1.id + ', .chatmessage-' + this.battle.p2.id);
```
Zarel: This breaks if a player changes name.

Better to filter by whether the rank is star.

Zarel:
The nickname trick is pretty genius, but I'm pretty sure it can break in certain situations...

QuiteQuiet:
Hm, the only case I can think of where this would break is if there were more than one pokemon of the same species that had nicknames where one is the end of the other; two Abra's nicknamed `xyz` and `abcxyz` for example.

Multiple pokemon of the same species in general seems to be the problem actually...

Zarel:
Anyway, we can do it like

<span class="pokemon" title="[species]">[nickname]</span>

which would let us show/hide nicknames in CSS

QuiteQuiet:
Wouldn't you still have to replace text in some locations to make that work? PS doesn't always mention what species the Pokemon has, so hiding the nickname through CSS there just get you an empty space (`The opposing In articulo mortis avoided the attack!`), or the Pokemon in parenthesis (`Player sent out In articulo mortis (Krookodile)!`).

Zarel:
That's what the title attribute is for. You can use CSS to put the title attribute inline.

QuiteQuiet:
Do you have an example? I can't seem to find any explanation for how that works, and it's not something I've seen before either.

Zarel:
CSS
```css
.nickname {
  font-size: 0;
}
.nickname::before {
  content: attr(title);
  font-size: 12pt;
}
```

HTML
```html
<span class="nickname" title="Pikachu">Sparky</span>
```

Zarel:
example: http://codepen.io/anon/pen/KgGRrY

QuiteQuiet:
I'm not entirely done with it yet, but something along these lines? I still need to sort out font size so it's not the same for everything, and this doesn't show `Rotom (Rotom-Wash)` for formes either.

js/battle.js
```diff
@@ -524,18 +524,21 @@ var Pokemon = (function () {
 			this.baseAbility = ability;
 		}
 	};
+	Pokemon.prototype.prepName = function (name) {
```
Zarel: Why are you passing a name at all?

Zarel: `prepName` is also a weird name for this function. `htmlName` might be ideal.

js/battle.js
```diff
@@ -575,6 +578,7 @@ var Pokemon = (function () {
 				}
 			}
 		}
+		if (!plaintext && this.prepName) name = this.prepName(name);
```
Zarel: In what situation would `this.prepName` not exist?

js/client-battle.js
```diff
@@ -1117,20 +1117,43 @@
 		toggleIgnoreSpects: function (e) {
 			this.battle.ignoreSpects = !!e.currentTarget.checked;
 			this.battle.add('Spectators ' + (this.battle.ignoreSpects ? '' : 'no longer ') + 'ignored.');
+			var $messages = $('.battle-log').find('[class*="chatmessage-"]').not('small:contains(\u2605)');
```
Zarel: `[class*="chatmessage-"]` is a complex selector; why is it necessary?

js/client-battle.js
```diff
@@ -1117,20 +1117,43 @@
...
 			this.battle.ignoreNicks = !!e.currentTarget.checked;
 			Tools.prefs('ignorenicks', this.battle.ignoreNicks);
 			this.battle.add('Nicknames ' + (this.battle.ignoreNicks ? '' : 'no longer ') + 'ignored.');
 			this.toggleNicknames(this.battle.mySide);
 			this.toggleNicknames(this.battle.yourSide);
+
+			var $log = $('.battle-log .inner');
+			$log = $log.find('span.pokemon');
+			if (this.battle.ignoreNicks) {
+				$log.addClass('hidenick');
```
Zarel: You should probably add the class to the log as a whole, instead of individual nicks.

js/battle.js
QuiteQuiet: Mainly because neither `this.name` or `this.species` would be containing the entire nickname with parenthesis or `<small>` tags (`Nickname <small>(species)</small>`). Passing in the already constructed name seemed better than reconstructing the nickname again.

js/client-battle.js
Zarel: I think that's not a problem. If you're ignoring spectators you probably don't want join/leave spam.

js/battle.js
QuiteQuiet: I guess that's fair. That makes me wonder I should decide what to reconstruct the nickname as, `getName`, `getLowerName`, or `getFullName`?

Zarel: I'm not sure what you're asking?

I'm thinking you should have a `htmlName` function that replaces `prepName` entirely. `getFullName` wouldn't call `prepName` but would need to reimplement wrapping the name in a span.

QuiteQuiet: Something more like this then? I /think/ that's what you meant?

js/battle.js
```diff
@@ -524,18 +524,21 @@ var Pokemon = (function () {
 			this.baseAbility = ability;
 		}
 	};
+	Pokemon.prototype.htmlName = function () {
+		return '<span class="pokemon" title="' + this.species + '">' + (this.side.battle.ignoreOpponent || this.side.battle.ignoreNicks ? this.species : Tools.escapeHTML(this.name)) + '</span>';
```
Zarel: I don't like `class="pokemon"`, you're playing with namespace fire. Something like `battle-nickname` and `battle-nickname-foe`.

js/battle.js
```diff
@@ -553,7 +556,7 @@ var Pokemon = (function () {
 			if (plaintext) {
 				name += ' (' + this.species + ')';
 			} else {
-				name += ' <small>(' + this.species + ')</small>';
+				name = '<span class="pokemon" title="' +  this.species + '">' + name + ' <small>(' + this.species + ')</small>' + '</span>';
```
QuiteQuiet: Wouldn't removing that show the nicknames in tooltips etc. even with the option turned on? They're not part of the battle log so they wouldn't be affected by the CSS that hides it...

QuiteQuiet: That would make Ignore Opponent also ignore nicknames for the entire session wouldn't it? Since you'd keep the `.hidenicks` after each battle ends. I suppose adding it to the specific battle room (`ps-room-opaque`?) would also work right?

Zarel: Tooltips are not part of the room, unfortunately. They're also kind of very messy code... Do whatever you feel is best.

Zarel: You may need to special-case tooltips, but log text should still work as I outlined it.

QuiteQuiet: I did another check before finishing this, and from what I can tell there's two locations that won't be affected by the battle log; tooltips and the icon mouse-over text [here](https://github.com/Zarel/Pokemon-Showdown-Client/blob/master/js/battle.js#L1371). It's not a big problem, but I rather not make a special-case for multiple locations for the same thing, so unless there's an objection to the check staying in `getFullName`, that's what I'll probably end up doing ...

js/client-battle.js
```diff
@@ -1117,20 +1117,42 @@
...
 			this.battle.add('Nicknames ' + (this.battle.ignoreNicks ? '' : 'no longer ') + 'ignored.');
 			this.toggleNicknames(this.battle.mySide);
 			this.toggleNicknames(this.battle.yourSide);
+
+			var $log = $('.battle-log .inner');
+			if (this.battle.ignoreNicks) {
+				$log.addClass('hidenicks');
+			} else {
+				$log.removeClass('hidenicks');
+			}
```
QuiteQuiet: Did this.

js/battle.js
QuiteQuiet: Okay then.

QuiteQuiet:
Updates: Fixed the stuff commented, except the `getFullName` stuff because I'm not comfortable special-casing something in multiple locations rather than keeping it where it is now. Also made sure to add `.hidenicks` when battles start if the `ignoreNicks` setting is set.

QuiteQuiet:
@Zarel do you have any further comments regarding this?

PR #785, opened 2016-10-18
Allow the effect of damage weakening berries to be activated

urkerab:
As mentioned in PR #780.

(I don't know what `this.lastmove = item.id;` is doing here; it dates back to the creation of `battle.js`.)

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -4350,11 +4350,9 @@ var Battle = (function () {
 					poke.prevItemEffect = 'eaten';
 					if (!this.fastForward) BattleOtherAnims.consume.anim(this, [poke.sprite]);
 					actions += '' + poke.getName() + ' ate its ' + item.name + '!';
-					this.lastmove = item.id;
-				} else if (kwargs.weaken) {
+				} else if (kwargs.weaken) { // Deprecated, now uses -activate
 					poke.prevItemEffect = 'eaten';
 					actions += 'The ' + item.name + ' weakened the damage to ' + poke.getLowerName() + '!';
-					this.lastmove = item.id;
 				} else if (effect.id) switch (effect.id) {
 				case 'fling':
 					poke.prevItemEffect = 'flung';
@@ -5409,6 +5407,8 @@ var Battle = (function () {
 					if (kwargs.broken) { // for custom moves that break protection
 						this.resultAnim(poke, 'Protection broken', 'bad');
 						actions += "It broke through " + poke.getLowerName() + "'s protection!";
+					} else if (kwargs.weaken) { // for damage-reducing berries
+						actions += 'The ' + effect.name + ' weakened the damage to ' + poke.getLowerName() + '!';
 					} else if (effect.effectType !== 'Ability') {
 						actions += "" + poke.getName() + "'s " + effect.name + " activated!";
 					}


Zarel:
What are the two messages displayed?

urkerab:
As part of the code for actually eating the berry (shared with other Berries such as the Lansat berry) the display is '(Pokemon) ate its (Berry)!'. Then for a damage-weakening berry you get the secondary activation, currently reusing the `-enditem` notification, 'The (Berry) weakened the damage to (Pokemon)!'

Zarel:
How does it look in game?

Zarel:
The fear is, of course, `-activate` will bring the item back since it implies the item exists. We'll probably need a check for that.

Zarel:
Apparently the first message doesn't appear. So `pokemon ate berry` should be in parens?

urkerab:
> The fear is, of course, -activate will bring the item back since it implies the item exists. We'll probably need a check for that.

Well, Custap Berry and Leppa Berry already use -activate, so there's a precedent that the item no longer necessarily exists.

> Apparently the first message doesn't appear. So pokemon ate berry should be in parens?

I wasn't editing the first message, but I can do that anyway if you like.

Zarel:
`-activate` happens before `-enditem`, right? It should be fine then.

urkerab:
No, it happens afterwards: in the case of Custap Berry, it's explicit `if (pokemon.eatItem()) { this.add('-activate', pokemon, 'item: Custap Berry');`; in the case of Leppa Berry it's trickier as the `onEat` event sends the `-activate` minor; in the case of damage-weakening berries, it's the second `-enditem` minor that would change into an `-activate` (so the code would then resemble Custap Berry).

Zarel:
But then how do we support item activations in which the item is still there?

I guess we can use `-item` for that. Do we do that consistently?

urkerab:
> But then how do we support item activations in which the item is still there?
> 
> I guess we can use -item for that.

Well, it's a little confusing, because `-item` is often used for gaining an item via e.g. Trick or Pickup, but it it's already used this way for Air Balloon, so that works.

This would readily apply to the Focus Band and Safety Goggles which currently set the item as part of their custom `-activate` code. (The custom message would need to remain to support replays.)

The Quick Claw currently shares `-activate` with the Custap Berry, and they, plus the Leppa Berry, are the only other items that use `-activate`. (Black Sludge, Leftovers and Shell Bell recovery use `-heal` while all other items use `-enditem`. I don't see Quick Claw being set as an item, so that appears to need fixing. Black Sludge and Sticky Barb damage also appears to need fixing.)

Zarel:
Maybe we should have a separate `-postactivate` for activating after no longer existing?

Zarel:
But honestly, it makes sense for `-item` to be an item activation...

Slayer95:
-activate|item: Whatever|[consume]?

Zarel:
`[consumed]` would be slightly clearer (this happens after `-enditem` which is presumably the consume step) but otherwise that's a pretty good idea.

Zarel:
Or just `-consume` as an item activation that consumes the item.

Zarel:
Or actually literally `-enditem`?

urkerab:
> Or actually literally `-enditem`?

They are already being ended. The problem that #780 worked around is that they're being ended twice.

Zarel:
Yeah, I think you ended up using `[consumed]` as a flag, which I tentatively okayed. What happened to that?

Zarel:
I'd probably prefer `[used]` as a flag, overall.

Zarel:
This will also need to be updated.

PR #786, opened 2016-10-18
Show the effect of Grudge on a move's PP in the tooltip

urkerab:
If you prefer, I could instead pass -64 and tweak the tooltip code to display 0/ in such a case.


Zarel:
That sounds better, yes.

PR #787, opened 2016-10-19
Show the move affected by Grudge as having 0 PP

urkerab:
As mentioned in PR #786. Although, thinking about it, maybe NaN would be less arbitray?


Zarel:
How about `Infinity`?

urkerab:
`-64` is history.

PR #794, opened 2016-11-06
Add coloration to modified stats

Shashwat986:
screen shot 2016-11-06 at 8 52 17 pm

### DIFF
--- a/style/battle.css
+++ b/style/battle.css
@@ -789,6 +789,13 @@ License: GPLv2
 	background: #333;
 	color: #DDD;
 }
+.stat-boosted {
+	color: #119911;
+}
+.stat-lowered {
+	color: #991111;
+	font-weight: normal;
+}
 .battle-log .rated {
 	padding: 3px 4px;
 }
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -709,8 +709,10 @@ var BattleTooltips = (function () {
 		var statText = '';
 		if (myPokemon.stats[statName] === modifiedStats[statName]) {
 			statText += '' + modifiedStats[statName];
+		} else if (myPokemon.stats[statName] > modifiedStats[statName]) {
+			statText += '<b class="stat-lowered">' + modifiedStats[statName] + '</b>';
 		} else {
-			statText += '<b>' + modifiedStats[statName] + '</b>';
+			statText += '<b class="stat-boosted">' + modifiedStats[statName] + '</b>';
 		}
 		return statText;
 	};


bumbadadabum:
I like this.

Zarel:
I like this too, but I have issues with the coloration.

Debuffed stats should probably be gray instead of red.

Partially because I have a dislike of Christmas colors, partially because that's the convention we already use.

Shashwat986:
@Zarel can you give me the RGB codes, or should I try a few out and get back to you. Also, where do we follow that convention? I'll copy that CSS.

Shashwat986:
22aa22b991111l
119911b119911l
22aa22b999l
000000b999999l

Zarel:
The main place you might notice is that resisted hits are gray and super-effective hits are red. We try very hard to never pair red and green.

The games themselves use red for +stat nature and blue for -stat nature, which also makes things annoying. In general, trying to convey things with color is an exercise in frustration.

Unfortunately, gray doesn't actually look great here. Maybe something like an orangeish-gray. Or maybe only redden the dropped stats without bolding them?

Ridaz:
How about using the colors used in this commit? #232

Shashwat986:
119911b991111fwnl
6688aabcc2222l
119911b803500l
119911ba64500l

Shashwat986:
I like the first one in this set of images

Zarel:
Sure, I'm fine with going with that.

Shashwat986:
Code fixed. Here's the final view:
screen shot 2016-11-07 at 5 46 28 pm

style/battle.css
```diff
@@ -789,6 +789,13 @@ License: GPLv2
 	background: #333;
 	color: #DDD;
 }
+.boosttext {
```
Zarel: Okay, time to nitpick class names.

I think `stat-boosted` and `stat-lowered` might be best here.

Shashwat986: Fixed

Zarel:
Can I get a license? Repeat after me:

>I, _ agree to license all of my contributions to the Pokémon Showdown project, past and future, under the MIT license.

Shashwat986:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

I, Shashwat Chandra, hereby license all of my contributions to the Pokémon Showdown project, past and future, under the MIT license.
-----BEGIN PGP SIGNATURE-----
Version: Keybase OpenPGP v2.0.58
Comment: https://keybase.io/crypto

wsBcBAABCgAGBQJYIYCnAAoJENsAyW6JGTRW2/UIALiKtZ0CpUuw9Q8sTCBCjWdE
ArC+cId1AyJtEFoDY9HAfxT8dcf1vRBgk+4Q3b9B69tOoArl+ataONpFozFsfvPf
nhvioS/VT8MmpEhzIXF1MVgdXFQJm9UY4JX5SccXlqhMraJm2kMu8uSGm1sraWs1
HiX9OrYK1RnUz+0h+NxO246npShUv/NF02QJ+jHTOKt6Ok+ZeG32DvyoZJEdsZPT
OXPCuDNjDp2K/cP3TvvxScchl62oojst8fWXJB9YZ3g+pke/tPey6jOskb8WvGfm
eEb0KzUXQNXXHFp6XM3KBTuCslsb7U3twcPfr9kkFmdw22bZE6cqhuIJggwsgP0=
=LFrv
-----END PGP SIGNATURE-----

PR #795, opened 2016-11-06
Autojoin regardless of window width

### LINKED ISSUE #742: Autojoin fails depending on window width...?
I find autojoin works fine when the window's wide enough to display the button list (Battle!, Teambuilder, etc.) at the far left of the page while focused on a room, but if the window is narrow enough to prevent it from displaying, autojoin fails for whatever reason. This has been happening for several months now, but I haven't been able to find any other explanation as to why this happens.

Tested on Firefox 47.0 (x64) and Chromium 51.0.2704.103 (x64).

Morfent:
This fixes autojoin not working because of the window width being too narrow.
This was from before chatrooms besides Lobby were implemente, but nowadays it
depends on the window layout of a desktop giving the browser enough width,
isn't consistent with autojoining via routing or server-side autojoin, and for
mobile users makes the function dependent on the screen resolution and whether
or not the phone's in profile or landscape mode.

Fixes #742

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -376,27 +376,23 @@
 			this.topbar = new Topbar({el: $('#header')});
 			if (this.down) {
 				this.isDisconnected = true;
-			} else if ($(window).width() >= 916) {
-				if (document.location.hostname === 'play.pokemonshowdown.com' || Config.testclient) {
-					this.addRoom('rooms', null, true);
-					Storage.whenPrefsLoaded(function () {
-						var autojoin = (Tools.prefs('autojoin') || '');
-						var autojoinIds = [];
-						if (autojoin) {
-							var autojoins = autojoin.split(',');
-							var roomid;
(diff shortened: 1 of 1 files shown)


js/client.js
```diff
@@ -1632,7 +1628,7 @@
 			for (var i = 0; i < rooms.length; i++) {
 				var room = rooms[i];
 				if (room.type !== 'chat') continue;
-				autojoins.push(room.id.indexOf('-') >= 0 ? room.id : (room.title || room.id));
+				autojoins.push(room.title);
```
Zarel: ?????

wtf is this?!?!?!

Morfent: Dumb assumption. What room ids include hyphens?

panpawn: Groupchats and battle rooms

Zarel: ^

Zarel: Maybe it should just be `toId(room.title) === room.id`

PR #796, opened 2016-11-07
Show player actions while waiting for opponent

QuiteQuiet:
Proposal for showing the moves any player would make while their opponent to move. 

[Example](http://puu.sh/s9nyA.jpg)

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -710,11 +710,65 @@
 			} else {
 				buf += '<p class="timer"><button name="setTimer" value="on"><small>' + "Start timer" + '</small></button></p>';
 			}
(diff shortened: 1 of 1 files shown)


Zarel:
I started work on this but honestly this whole thing needs a pretty major refactor; people have been adding things to it kind of haphazardly...

js/client-battle.js
```diff
@@ -1040,7 +1094,7 @@
 			}
 			this.closeNotification('choice');
 
-			this.choice = {waiting: true};
+			this.choice.waiting = true;
```
Zarel: Ah, yes, this was probably the main change that needed to be made.

PR #798, opened 2016-11-10
Make the battle timer tick every second

urkerab:
This might not be the best way of tracking whether the timer button exists...

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -225,11 +225,14 @@
 			} else if (this.battle.playbackState === 2 || this.battle.playbackState === 3) {
 
 				// battle is playing or paused
-				if (this.side && !this.battleEnded) {
+				if (!this.side) {
+					// spectator
+					this.$controls.html('<p><button class="button" name="instantReplay"><i class="fa fa-undo"></i><br />First turn</button> <button class="button" name="rewindTurn"><i class="fa fa-step-backward"></i><br />Last turn</button><button class="button" name="skipTurn"><i class="fa fa-step-forward"></i><br />Skip turn</button> <button class="button" name="goToEnd"><i class="fa fa-fast-forward"></i><br />Skip to end</button></p><p><button name="switchSides"><i class="fa fa-random"></i> Switch sides</button></p>');
+				} else if (this.battleEnded) {
+					this.$controls.html('<p><button class="button" name="instantReplay"><i class="fa fa-undo"></i><br />First turn</button> <button class="button" name="rewindTurn"><i class="fa fa-step-backward"></i><br />Last turn</button><button class="button" name="skipTurn"><i class="fa fa-step-forward"></i><br />Skip turn</button> <button class="button" name="goToEnd"><i class="fa fa-fast-forward"></i><br />Skip to end</button></p>');
+				} else {
 					// is a player
 					this.$controls.html('<p>' + this.getTimerHTML() + '<button class="button" name="skipTurn"><i class="fa fa-step-forward"></i><br />Skip turn</button> <button class="button" name="goToEnd"><i class="fa fa-fast-forward"></i><br />Skip to end</button></p>');
-				} else {
-					this.$controls.html('<p><button class="button" name="instantReplay"><i class="fa fa-undo"></i><br />First turn</button> <button class="button" name="rewindTurn"><i class="fa fa-step-backward"></i><br />Last turn</button><button class="button" name="skipTurn"><i class="fa fa-step-forward"></i><br />Skip turn</button> <button class="button" name="goToEnd"><i class="fa fa-fast-forward"></i><br />Skip to end</button></p><p><button name="switchSides"><i class="fa fa-random"></i> Switch sides</button></p>');
 				}
 				return;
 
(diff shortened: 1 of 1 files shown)


js/client-battle.js
```diff
@@ -421,8 +421,20 @@
 			}
 			return '<button name="openTimer" class="button timerbutton' + timerTicking + '"><i class="fa fa-hourglass-start"></i> ' + time + '</button>';
 		},
+		timerInterval: 0,
+		timerTick: function () {
+			if (typeof this.battle.kickingInactive == "number") this.battle.kickingInactive--;
```
Zarel: Don't let it fall below `1`...

Zarel: (But, otherwise, I approve of this method of detecting what state the timer is in)

js/client-battle.js
```diff
@@ -421,8 +421,20 @@
...
+			this.updateTimer();
+		},
 		updateTimer: function () {
-			this.$('.timerbutton').replaceWith(this.getTimerHTML());
+			var timerButton = this.$('.timerbutton');
+			if (timerButton.length) {
+				timerButton.replaceWith(this.getTimerHTML());
+				if (!this.timerInterval) this.timerInterval = setInterval(this.timerTick.bind(this), 1000);
+			} else if (timerInterval) {
```
Zarel: I'm assuming this is a typo of `this.timerInterval` and also the cause of the build error.

js/client-battle.js
```diff
@@ -421,8 +421,20 @@
...
 			return '<button name="openTimer" class="button timerbutton' + timerTicking + '"><i class="fa fa-hourglass-start"></i> ' + time + '</button>';
 		},
+		timerInterval: 0,
+		timerTick: function () {
+			if (typeof this.battle.kickingInactive == "number") this.battle.kickingInactive--;
+			this.updateTimer();
+		},
 		updateTimer: function () {
-			this.$('.timerbutton').replaceWith(this.getTimerHTML());
+			var timerButton = this.$('.timerbutton');
```
Zarel: `$timerButton` is the style for jQuery objects

js/client-battle.js
```diff
@@ -421,8 +421,20 @@
...
+			if (timerButton.length) {
+				timerButton.replaceWith(this.getTimerHTML());
+				if (!this.timerInterval) this.timerInterval = setInterval(this.timerTick.bind(this), 1000);
```
Zarel: Are you sure you want to be setting this in `updateTimer`? You should most likely be doing this in `getTimerHTML` itself. `updateTimer` is only called for specific updates.

Zarel:
Also: do you think users would find "every second" distracting?

I'm kind of considering "every 5 seconds".

js/client-battle.js
Zarel: The timeout should also be cleared and reset every time the timer is manually updated. This makes it more resilient to network errors.

panpawn:
I really like this idea!

>Also: do you think users would find "every second" distracting?

imo I don't, as long as it's not like obnoxious and flashing or something like that.

urkerab:
I found a bug in `updateControls` - as the last turn plays out, players get access to the "Switch Sides" button, which confuses the UI, e.g. you can fool it into rematching yourself. As I was touching that code anyway for this version, I fixed that at the same time.

(Can you believe I almost wrote `if (this.battle.kickingInactive > ticking)`... yes it does work, since it evaluates to false if either is undefined.)

js/client-battle.js
```diff
@@ -421,8 +424,21 @@
...
+			if ($timerButton.length) {
+				$timerButton.replaceWith(this.getTimerHTML());
+			} else {
+				this.clearInterval(this.timerInterval);
+				this.timerInterval = 0;
+			}
+		},
+		createTimer: function () {
+			if (this.timerInterval) clearInterval(this.timerInterval);
+			this.timerInterval = setInterval(this.updateTimer.bind(this, true), 1000);
```
Zarel: `bind` is deprecated, please don't use it

Zarel: Also, put all this in `getTimerHTML`, there's no reason to add yet another function to the stack for something this simple.

js/client-battle.js
```diff
@@ -403,9 +406,27 @@
 				break;
 			}
 		},
-		getTimerHTML: function () {
+		timerInterval: 0,
+		getTimerHTML: function (ticking) {
 			var time = 'Timer';
 			var timerTicking = (this.battle.kickingInactive && this.request && !this.request.wait && !(this.choice && this.choice.waiting)) ? ' timerbutton-on' : '';
+
+			if (!ticking) {
```
Zarel: um, `timerTicking` already exists; I hope you're aware?

urkerab: Bad naming maybe; there were other bugs anyway so I renamed it while I was there.

Zarel:
Okay, looks good.

There's still a style issue but I'll fix that myself, lgtm

Zarel:
Honestly, though, the solution to the `switchSides` issue is to make the rematch button smarter, not to prevent the player from seeing the battle from the other side...

PR #799, opened 2016-11-10
Change PS version 0.10.2 -> 0.10.3

panpawn:


Zarel:
Hm, we need to get that into `client.js` somehow, in a way that doesn't involve code duplication.

I guess we insert it as part of one of the build scripts.

panpawn:
uh, which one

Zarel:
`update` works

Lord-Haji:
Z moves r worth a bump to 0.11.4
@panpawn

Zarel:
@Lord-Haji, that's not how version numbers work. It'd be a bump to 0.11.0.

PR #800, opened 2016-11-11
Tweak a couple of messages

urkerab:
I think that Focus Energy is supposed to indicate if it was caused by an item. There was also a grammar error in the case of Confusion being cured by an item.


urkerab:
I'm surprised I didn't copy and paste that line in the first place...

PR #803, opened 2016-11-18
Update Infobox CSS

Lord-Haji:
Infobox comes with curved edge looks nice :D

### DIFF
--- a/style/battle.css
+++ b/style/battle.css
@@ -835,6 +835,7 @@ License: GPLv2
 	padding: 2px 4px;
 }
 .infobox {
+	border-radius: 25px;
 	border: 1px solid #6688AA;
 	padding: 2px 4px;
 }
--- a/style/battle.css
+++ b/style/battle.css
@@ -837,7 +837,7 @@ License: GPLv2
 .infobox {
 	border-radius: 25px;
 	border: 1px solid #6688AA;
-	padding: 2px 4px;
+	padding: 20px;
 }
 .infobox-limited {
 	max-height: 200px;
--- a/style/battle.css
+++ b/style/battle.css
@@ -836,8 +836,8 @@ License: GPLv2
 }
 .infobox {
 	border-radius: 25px;
-	border: 1px solid #6688AA;
-	padding: 20px;
+	border: 2px solid #6688AA;
+	padding: 10px;
 }
 .infobox-limited {
 	max-height: 200px;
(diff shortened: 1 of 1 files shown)


Zarel:
......I guess it can be a vote. `25px` is probably _too_ rounded even if rounded looks nice, though...

bumbadadabum:
Can you attach a screenshot comparison?

Lord-Haji:
scrnst
Here's the screen shot
But i do think if rounded edges we should make the border 3px or 4px

bumbadadabum:
If you're doing this you might need to add some padding too.

Lord-Haji:
@bumbadadabum is the border radius and border width perfect ?

bumbadadabum:
Too rounded imo. Also I'm not a fan of this in general but this isn't up to
me.

Lord-Haji:
Update:
newscrnsht
Now it looks like this

kotarou3:
What about multiple lines?

Also, still feels too round to me

Ridaz:
It occupies too much space just for info and like bumba said, it's too round and doesn't blend in with the rest of the PS layout

Lord-Haji:
Updated the Infobox CSS.
Reduced the roundness almost like the PS Battle Buttons.
Also added  the roundness to the:- 
* broadcast-red
* broadcast-blue
* broadcast-green

Lord-Haji:
@Zarel 
broadcast
Final screen shots of broadcast. But i do think we also should think of using some other good looking color hex. NP if its a down vote tho

Lord-Haji:
infobox
Final screen shot of infobox @Zarel

Zarel:
I think those have way too much padding for how often we currently use them.

Lord-Haji:
@Zarel  shall i go back to the ``2px 4px padding`` then ?

Zarel:
Probably. At most one more pixel.

panpawn:
I am not a fan of these changes.

panpawn:
Also, "Infobox"... "box"...

A "box" is defined as: "a container with a flat base and sides, **typically square or rectangular** and having a lid." ([Source](https://www.google.com/search?q=define+box&ie=utf-8&oe=utf-8))

With that being said, the name of this CSS class seems unfit to me if we deviate from this.  But that's just my opinion.  That, and well, I don't think it looks as professional as a true "box".

Zarel:
No updates in three-ish months and most people don't like this in its current state; closing.

PR #807, opened 2016-11-20
Fix issue with adding Type: Null to a team

HoeenCoder:
Because Type: Null has a colon in its name, it cant be added to teams. This is a fix to it. 

Sorry if i forgot something its my first time editing the client.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -2376,7 +2376,8 @@
 				this.search.find('');
 				return;
 			}
-			var val = $(e.currentTarget).data('entry').split(':')[1];
+			var entry = $(e.currentTarget).data('entry');
+			var val = entry.slice(entry.indexOf("|") + 1);
 			if (this.curChartType === 'move' && e.currentTarget.className === 'cur') {
 				// clicked a move, remove it if we already have it
 				var $emptyEl;
@@ -2421,7 +2422,8 @@
 					this.search.find('');
 					return;
 				}
-				var val = $firstResult.data('entry').split(':')[1];
+				var entry = $firstResult.data('entry');
+				var val = entry.slice(entry.indexOf("|") + 1);
 				this.chartSet(val, true);
 				return;
 			} else if (e.keyCode === 38) { // up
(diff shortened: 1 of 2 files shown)


HoeenCoder:
nano isn't working for me (ill fix it tommorow) so i cant squash atm. Sorry about that!

Zarel:
I'm going to recommend just replacing `:` with `|` here

HoeenCoder:
That would work, but how much refactoring would be needed?

Zarel:
not too much

kotarou3:
Now to wait for GF to use `[` and `|` in the game...

HoeenCoder:
Sorry the rebase took so long, but its done now.
Also @kotarou3 thats one of the reasons I used this setup instead of suggestions a refactor.

Zarel:
Holy code duplication, though. Make a temp variable called `entry` if you're going to do it this way.

HoeenCoder:
Sure!

Zarel:
But also I'd still prefer `|`

HoeenCoder:
I could refactor this in the future, but it will take some time. And if GF decides to use `|` were back here.

Zarel:
We could do both.

HoeenCoder:
Just to be clear: refactor from `:` to `|` and keep these edits.

If so should i work on that and push to this PR? Or make a new one and merge this?

Zarel:
Sure.

js/client-teambuilder.js
```diff
@@ -2376,7 +2376,8 @@
 				this.search.find('');
 				return;
 			}
-			var val = $(e.currentTarget).data('entry').split(':')[1];
+			var entry = $(e.currentTarget).data('entry');
+			var val = $(e.currentTarget).data('entry').slice(entry.indexOf(":") + 1);
```
Zarel: ...you know, when I said "holy duplicate code", I meant that this should be

    entry.slice(entry.indexOf(":") + 1)

I think you missed the "duplicate code" part.

HoeenCoder: Oops my bad

Zarel:
Okay but Dex probably needs to be updated too; lemme see

Zarel:
nvm

HoeenCoder:
One less thing to fix lol

PR #814, opened 2016-11-22
Read gen7doubles info

HoeenCoder:
Teambuilder currently crashes when building a team for a Gen 7 Doubles format. 

This should (in theory) fix it, but i was unable to test this as i don't know how to make the githook run.

I apologize for my ignorance.


Sw1fterOnPS:
yeah that looks like it should work, vgc was having the same problem as this previously

Zarel:
This conflicts with an uncommitted change; sorry. I'll fix it myself.

HoeenCoder:
Alright thanks zarel

PR #821, opened 2016-11-26
Fix Smeargles learnsets

HoeenCoder:
Smeragle should only be able to learn Non-Z moves from the gen selected and back.

Because `BattleMovedex[i]` does not contain a `.gen` property, I used `Tools.getMove(i);`

The `.isZ` check is for gen 7, because Smeargle cannot sketch a Z-move to learn it.

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -910,6 +910,27 @@
 				moves = [];
 				for (var i in BattleMovedex) {
 					if (i === 'chatter' || i === 'magikarpsrevenge' || i === 'paleowave' || i === 'shadowstrike') continue;
+					if (!BattleMovedex[i].gen) {
+						if (BattleMovedex[i].num >= 622) {
+							BattleMovedex[i].gen = 7;
+						} else if (BattleMovedex[i].num >= 560) {
+							BattleMovedex[i].gen = 6;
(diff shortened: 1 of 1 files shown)


Zarel:
You should look at the source for how `Tools.getMove` checks the gen, and copy that. I'd prefer not to use `Tools.getMove` in an inner loop.

HoeenCoder:
@Zarel I noticed this is still open, do you still want/need this? If its not setup right I can modify it.

Zarel:
You need to make a new comment after doing updates I ask, or I won't be notified of them.

Anyway, blah, this is annoying amounts of code to be duplicated... :/

Zarel:
Honestly, it'd probably be better without the cache, too... I'll deal with that myself I guess.

HoeenCoder:
Oh sorry Zarel! From my past commits you got notified when I pushed a change so I assumed that you would get it here too. I will remember that in the future.

Zarel:
To be exact, it notifies the Development room. If I'm not in the Development room when you push a change, I won't see it. It's always best practice to leave a comment.

HoeenCoder:
I actually assumed it was github notifying you, but since you probably have different settings than me with so much going on.

PR #823, opened 2016-11-28
Update testclient.html

Lord-Haji:


KewlStatics:
Huh?

panpawn:
... please look at the diff _before_ actually submitting a pull request.

If it was not what you had intended, then simply do not submit it.  That solves the several instances of this that you have already done on separate occasions from happening again, at least, hopefully.

PR #824, opened 2016-11-29
Teambuilder: Show typing in set details

asgdf:
grafik

Since there will most likely always be a part of the users that doesn't know every mon's typing by heart, I think it is a good idea to display the typing right there in the set details, so that people won't have to return to the Pokemon selection menu every time they want to check which types they get STAB on or something. There's some free space there anyway.
 I hope I'm not destroying the aesthetics with this, should the typing be centered above the ability box, or somewhere else entirely? Should there be a label, or is it self-explanatory enough?

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1083,8 +1083,11 @@
 				buf += '<span class="detailcell"><label>Happiness</label>' + (typeof set.happiness === 'number' ? set.happiness : 255) + '</span>';
 				buf += '<span class="detailcell"><label>Shiny</label>' + (set.shiny ? 'Yes' : 'No') + '</span>';
 			}
+			buf += '</button></div></div>';
 
-			buf += '</button>';
+			// item/type icons
+			buf += '<div class="setrow setrow-icons">';
+			buf += '<div class="setcell">';
 			var itemicon = '<span class="itemicon"></span>';
 			if (set.item) {
 				var item = Tools.getItem(set.item);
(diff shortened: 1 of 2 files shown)


js/client-teambuilder.js
```diff
@@ -1083,16 +1083,24 @@
...
 			if (set.item) {
 				var item = Tools.getItem(set.item);
 				itemicon = '<span class="itemicon" style="' + Tools.getItemIcon(item) + '"></span>';
 			}
 			buf += itemicon;
 			buf += '</div>';
-			buf += '</div><div class="setrow">';
+			buf += '<div class="setcell setcell-typeicons">';
+			var template = Tools.getTemplate(set.species);
+			for (var i in template.types) buf += Tools.getTypeIcon(template.types[i]);
```
Slayer95: Note that ``template.types`` is not guaranteed to exist

asgdf: I hadn't considered that, but it doesn't throw an error if it happens. Is it still worth putting a "if (set.species) ..." in there?

Slayer95:
What do you think of aligning the types to the right?
Also some padding between primary & secondary type wouldn't hurt imo.

Zarel:
I'm not sure I like the location in general, but I'm also not sure if I have a better location for it.

I guess right-aligning is at least better than the current location (iow I endorse everything Slayer95 said).

asgdf:
Now it's right-aligned and with 3px of padding between the type icons
grafik

TheImmortal:
What if the bottom of the icons were in-line with the bottom of the 2nd move box? i.e. move icons higher

asgdf:
Like this? Also reduced the padding between the type icons to 1px to be consistent with renderPokemonRow()
 grafik

TheImmortal:
Meh, never mind. I imagined that better. The previous one looks nicer. Carry on.

asgdf:
Any other opinions on this? Currently still looks like this 
grafik
The only alternative I could think of would be something along these lines (quick mockup), with the types in line with the copy/import/move buttons, but it would kinda ruin the "physical tab" aesthetics for the sets imo
grafik

asgdf:
Is this still on the radar?

Zarel:
I'd say move it back up?

asgdf:
By "move it back up", I assume you mean like it was [in this screenshot](https://cloud.githubusercontent.com/assets/5814184/20824379/ab723276-b85c-11e6-8853-87c71d429465.png)?
If so, it looks like there's some disagreement here. Should we do a vote or something?

Zarel:
Oh nvm I got confused about which one was which.

PR #826, opened 2016-11-30
Fix gen 7 mons' battle anims not showing

asgdf:
Currently, animations for gen 7 mons fail to load with the default options, because it tries to get them from sprites/bwani. I assume this was meant to be some kinda workaround from when we didn't have the models, but the commit in question https://github.com/Zarel/Pokemon-Showdown-Client/commit/df7a96c4a7af915111eab1eb547ad14b97fccfce doesn't really explain.
Anyway, I tested this with all combinations of animation options, and none of them showed a broken image.


Zarel:
Yeah, that was a mistake that I didn't really think through.

PR #827, opened 2016-11-30
Display [Gen 6] and hide [Gen 7]

TheImmortal:
http://imgur.com/a/OUyPX


Zarel:
This is a hack we should probably do more properly at some point...

PR #828, opened 2016-12-01
Allow the server to update psuedo weather backwards compatibly

urkerab:
Rather conveniently from the client's point of view, weather upkeep data is explicitly sent to the client. This means that if a Pokémon faints and one with a weather Ability is switched in, the duration is set to 5 or 8 turns, and does not change at the end of the turn.

Unfortunately terrain Abilities do not have that luxury. PR #808 worked around this but for the first turn only; this means that terrain Abilities that activate on switching in for a fainted Pokémon will display the wrong turn count, as they are immediately changed at the end of the turn.

Of course you have to be careful to distinguish this case from that of a slow VoltTurn. As far as I can tell, there needs to be a message to inform the client that residual upkeep has taken place and it should update pseudo weather. This is the client side handling of that message.

For compatibility with old replays, the previous turn-based upkeep mechanics are used if no upkeep messages are seen. This helpfully also avoids having to send the upkeep message before the first turn.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2288,6 +2288,7 @@ var Battle = (function () {
 		// has playback gotten to the point where a player has won or tied?
 		// affects whether BGM is playing
 		this.ended = false;
+		this.usesUpkeep = false;
 		this.weather = '';
 		this.pseudoWeather = [];
 		this.weatherTimeLeft = 0;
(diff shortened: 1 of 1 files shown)


Zarel:
How do weather abilities fix this?

I think the easiest way might be to move the turn number increment to right after the residual step, so faint-switches happen at the beginning of a turn rather than at the end of a turn.

urkerab:
Weather upkeep data is explicitly sent to the client, so that the correct messages e.g. "The hail is crashing down." can be generated.

Zarel:
Oh, yeah.

Anyway, what I said earlier stands:

I think the easiest way might be to move the turn number increment to right after the residual step, so faint-switches happen at the beginning of a turn rather than at the end of a turn.

Tagging @Marty-D to see if there's anything wrong with doing it that way.

Marty-D:
@Zarel I mean, mechanically it would be wrong but I don't think that really affects anything. The bigger problem is that everyone's used to it after 4-5 years of expected, accurate behaviour, and changing it now would confuse more people than it's worth.

Zarel:
I still feel uncomfortable about this......

PR #829, opened 2016-12-01
Fix tooltips for Random Battle

TheImmortal:


Zarel:
Are you sure `.match` is the right function here? I think `.test` is better

Zarel:
Or, you know, just `indexOf` tbh

TheImmortal:
test has better performance than match, yeah

PR #830, opened 2016-12-03
Don't display side condition ids in replays

urkerab:
Currently things like screens or tailwind are showing in lower case in replays. The server is sending the correct name, but we're turning it into an id. This happens to work in battles because Tools.getMove turns it back into a name, but fails in replays. If we use the original name, we get the right behaviour in replays, and Tools.getMove becomes unnecessary. Example replay: http://replay.pokemonshowdown.com/gen7pokebankdoublesou-488415145

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -1392,8 +1392,9 @@ var Side = (function () {
 			}
 		}
 	};
-	Side.prototype.addSideCondition = function (condition) {
+	Side.prototype.addSideCondition = function (effect) {
 		var elem, curelem;
+		var condition = effect.id;
 		if (this.sideConditions[condition]) {
 			if (condition === 'spikes' || condition === 'toxicspikes') {
 				this.sideConditions[condition][2]++;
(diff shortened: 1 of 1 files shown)


Slayer95:
Maybe we can create and fill ``window.BattleMovedex`` and alike according to the parameters passed to ``Tools.get*`` if the respective dictionary did not exist when first called?

Zarel:
...sanitize..............

Slayer95:
@Zarel, all those calls seem to be safe because they go through ``Tools.getEffect(x).name``

Zarel:
Do they?

Zarel:
We might also just want to cache `getEffect` or something?

Slayer95:
That would work. No RAM concerns?
Alternatively, is there a downside to making ``.name`` titlecase by default (not-found case)?

Zarel:
Or just hardcode a list of pseudo-weather names; we hardcode a list of weather names, after all.

Zarel:
We don't even need to prefill it, we can just fill it as needed...

urkerab:
Weathers are special because server doesn't have their names anyway.

PR #832, opened 2016-12-05
Show the underlying move name when reviewing a player's choice

urkerab:
This mostly affects the move "Return 102".


Zarel:
Honestly, "Return 102" might not be a bad thing to show...

PR #833, opened 2016-12-07
Teambuilder: Create a button which allows users to clone/duplicate their team.

HappyZombies:
Hello, I am requesting a pull request that contains functionality that will allow users to clone their teams.

Currently, if a user wants to have a copy of an existing team (say they want the same team but maybe with different moves, different items, different pokemons, etc.) they must first: Go to the export/import menu, copy the text, go back and create a new team, go back into export/import, paste the team and save. 

The clone button allows a user to instantly clone a team. After a team is cloned, the name of the team will have " Clone of" appended to the beginning of it. See the screenshots below to see how the clone button works. 

clone-button

clone-of-title

_After duplicating the two teams and going back..._
after-cloning

A margin 10px was added to the ".teamlist li" on the CSS. When adding another button next to the edit and delete button, it causes it to go to the next line (which ends up underneath each teamlist). So I pushed all of the tools underneath of the team list display. I am not sure if this looks nice or if it is something you would like, but a review of the positioning of the clone button might need to be discussed. 

Anyhow, thank you for your time and I hope that this feature is implemented soon as it will help a lot of people!

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -608,10 +608,10 @@
 		reloadTeamsFolder: function () {
 			Storage.nwLoadTeams();
 		},
-		
+
 		clone: function (i) {
 			var newTeam = {
-				name: " Clone of "+teams[i].name,
+				name: " Clone of " + teams[i].name,
 				format: teams[i].format,
 				team: teams[i].team,
 				folder: teams[i].folder,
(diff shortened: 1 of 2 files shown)


Slayer95:
Can we do this with Ctrl+DnD? The team controls are supposed to be to the right of the main team button...

HappyZombies:
@Slayer95 Okay, how about Ctrl+DnD on the 'New Team' button and it clones the team ?

Zarel:
There should probably also be a button, though...

Zarel:
Superseded by #1058, which still doesn't look how I'd prefer...

PR #835, opened 2016-12-09
Update font-awesome to v4.7.0

FelixRilling:
Besides new icons, the latest version of Font-Awesome also improves support for IE/Edge transforms. While I'm not completely sure if this update is necessary, it adds a few icons that could be useful, like "pause" or "handshake".


style/font-awesome.css
```diff
@@ -1,13 +1,13 @@
...
+ *  Font Awesome 4.7.0 by @davegandy - http://fontawesome.io - @fontawesome
  *  License - http://fontawesome.io/license (Font: SIL OFL 1.1, CSS: MIT License)
  */
 /* FONT PATH
  * -------------------------- */
 @font-face {
   font-family: 'FontAwesome';
-  src: url('font/fontawesome-webfont.eot?v=4.4.0');
-  src: url('font/fontawesome-webfont.eot?#iefix&v=4.4.0') format('embedded-opentype'), url('font/fontawesome-webfont.woff2?v=4.4.0') format('woff2'), url('font/fontawesome-webfont.woff?v=4.4.0') format('woff'), url('font/fontawesome-webfont.ttf?v=4.4.0') format('truetype'), url('font/fontawesome-webfont.svg?v=4.4.0#fontawesomeregular') format('svg');
+  src: url('./fonts/fontawesome-webfont.eot?v=4.7.0');
```
Zarel: is that `./` part necessary? do you know what it does?

FelixRilling:
The `./` is not necessarily needed, its more of a habit of mine to write same-directory URLs this way. If you prefer dropping it I could change that.

Zarel:
Was there a reason you renamed `font/` to `fonts/`? iirc our directory structure is kinda unusual because I didn't want a top-level fonts directory.

FelixRilling:
There was both a `font` and a `fonts` folder in the [oiginal structure](https://github.com/Zarel/Pokemon-Showdown-Client/tree/master/style) which seemed strange. The Font-awesome default folder name is `fonts`, which is why I used `fonts`.

Zarel:
Okay, that makes sense.

Zarel:
Oh, the commit comment explains: "Work around CloudFlare cache issue".

Caching is a horrible mess... :|

Zarel:
https://github.com/Zarel/Pokemon-Showdown-Client/commit/76bbf1d46985acf781f5eaa2e0e98b4bbc8f8301

Zarel:
This is why I so rarely update Font Awesome. Maybe we should just wait for Font Awesome 5...

FelixRilling:
Seems like a good idea, I agree with you.
Thanks for the quick response.

PR #836, opened 2016-12-09
Update jQuery to v2.2.4

FelixRilling:
There have been quite a few bug fixes between v2.1.4 and v2.2.4, which is why the later should be used by the client. In addition the jQuery plugins have been updated.
No conflicts occurred.


Zarel:
Why are you wholesale replacing jQuery Cookie? Is the new one API-compatible?

FelixRilling:
I updated jQuery Cookie to avoid errors of using an older version of the plugin with a newer version of jQuery. There have been no breaking changes, only bugfixes in jQuery cookie, so its API-compatible.

PR #838, opened 2016-12-11
Add some activation messages for the new Abilities

urkerab:
- Battle Bond
- Disguise
- Power Construct

I copied the use of `-formechange` from Shaymin-Sky's transformation.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -5507,6 +5507,17 @@ var Battle = (function () {
 				case 'sweetveil':
 					actions += '' + ofpoke.getName() + ' surrounded itself with a veil of sweetness!';
 					break;
+				case 'battlebond':
+					actions += '' + poke.getName() + ' became fully charged due to its bond with its Trainer!';
+					break;
+				case 'disguise':
+					actions += 'Its disguise served it as a decoy!';
+					break;
+				case 'powerconstruct':
+					actions += 'You sense the presence of many!';
+					break;
+
+				// weather activations
 				case 'deltastream':
 					actions += "The mysterious strong winds weakened the attack!";
 					break;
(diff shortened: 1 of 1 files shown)


Marty-D:
You'd have to send `-formechange` from the server for all of these then, and their Abilities would be activating twice, which would be slow and ugly. I was actually thinking about combining `-mega`, `-primal`, and these new ones under a `-permaformechange` or something. And since these new ones have two messages each, the Ability activation can be handled by the `-activate` part, which you've already done here. The Shaymin thing was mostly a hack so I could make use of the existing "POKEMON transformed!" default message. :sweat_smile:

Zarel:
`-permaformechange` already exists, it's called `detailschange`...

Marty-D:
@Zarel, what makes `-mega` and `-primal` special then?

Zarel:
No idea, I didn't add them.

Zarel:
Please tell me `detailschange` is still sent for megaing...

Zarel:
Oh, good, it is. A permanent change to a Pokémon's species is a major.

Anyway, as far as I can tell, someone just decided to send separate messages for the mega and primal messages? I'd personally have used `[mega]` and `[primal]` tags on `detailschange`.

Marty-D:
Unfortunately those already exist so we have to keep supporting them... But yeah, those tags are what I was thinking so doing it on `detailschange` instead is fine with me.

urkerab:
We can't do mega/primal on detailschange because of BH Illusion shenanigans, unfortunately.

Zarel:
Can you clarify?

ascriptmaster:
IIRC Illusion is coded in such a way where if you mega evolve while under Illusion, you keep the illusion appearance (until you switch out? forever? idr but it won't break under damage any more because the ability is no longer Illusion)

I don't know if this is the same scenario that's being brought up, though.

Marty-D:
Yeah what ascriptmaster said (until you switch). I forgot it also uses `-mega` and `-primal` for the Illusion message/animation/cry that happens in those situations.

urkerab:
(I wasn't proposing to send the Ability with the `-formechange` minor, but as we're going with `detailschange` anyway, it doesn't matter.)

Zarel:
...I'm going to leave this to Marty to decide...

Marty-D:
This is fine; all future permanent forme changes can follow this model. Thanks @urkerab! :+1:
I'll handle the removal of placeholders server-side.

PR #843, opened 2016-12-22
Improve the appearance of disabled buttons in dark mode

urkerab:
Dark mode needs to override the background, but the grey colour works OK.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -208,6 +208,14 @@ button:disabled {
 	box-shadow: none;
 	color: #F9F9F9;
 }
+.dark .button.disabled,
+.dark .button.disabled:hover,
+.dark .button.disabled:active {
+	background: #555555;
+	border-color: #555555;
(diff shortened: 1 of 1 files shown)


style/client.css
```diff
@@ -208,6 +208,13 @@ button:disabled {
 	box-shadow: none;
 	color: #F9F9F9;
 }
+.dark .button.disabled,
+.dark .button.disabled:hover,
+.dark .button.disabled:active {
+	background: #555555;
+	border-color: #555555;
+	box-shadow: none;
```
Zarel: Even if you're not going to add a new `color`, you should re-set colors whenever you set background colors (and vice versa) for readability reasons.

urkerab:
I don't know why, but GitHub shows your review comment against my new commit that has the `color`...

Zarel:
Yeah, that's normal. Just something you have to live with. :(

PR #844, opened 2016-12-24
Use exact format name

HoeenCoder:
**This is in tandem with:** [Zarel/Pokemon-Showdown#3049](https://github.com/Zarel/Pokemon-Showdown/pull/3049)

If the changes in the above pull request are merged, this pull request is needed to not add `[Gen 7]` in front of gen 7 formats. If there's any issues or concerns don't hesitate to point them out.

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -922,10 +922,7 @@
 					}
 					bufs[curBuf] += '<li><h3>' + Tools.escapeHTML(curSection) + '</li>';
 				}
-				var formatName = Tools.escapeFormat(format.id);
-				if (formatName.charAt(0) !== '[') formatName = '[Gen 6] ' + formatName;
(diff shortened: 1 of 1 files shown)


Zarel:
Unfortunately, this messes up teams in the teambuilder.

Zarel:
The current plan is to have the teambuilder automatically rename teams with a format without a gen tag from `formatid` to `gen6formatid`, and also match them with formats sent from the server without a gen tag. That needs to be the first step before anything else can happen.

HoeenCoder:
So to make sure im getting this (which im not so sure):

`OU => [Gen 6] OU` and client "thinks" `[Gen 6] OU === OU` (when ou is from from server)?

Zarel:
Yes.

Zarel:
That's something I planned to do myself, but if you wanted to do it, it'd save me a lot of work.

PR #847, opened 2017-01-03
Add support for room host rank

panpawn:
This is a rank that if added, could be used for groupchats to solve some potential issues if/when opened back up to the public.


bumbadadabum:
What kind of issues would this solve?

Zarel:
Although perhaps it'd make more sense to rename Player to Host...

panpawn:
... Room Host, or just Host?

Zarel:
Honestly, ideally we'd have Player and Host as the same rank, and then just use context to figure out whether to label it "Player" or "Host".

Zarel:
Or, use

★ Host
☆ Player

panpawn:
Alright, will update this soon

Zarel:
>soon

panpawn:
>Or, use
★ Host
☆ Player

Update for that

panpawn:
Note: this will require a server-side change of the player rank symbol as well

Zarel:
I'm not sure we can do it all in one step...

They should possibly both be "Player" for now.

Or maybe even "Player/Host"...

Zarel:
Note that, as merged, the ranks are currently named

> ★ Player  
☆ Player

We'll have to rename it to Host over multiple steps.

sparkychildcharlie:
It's still possible to "mass invite" using <<room.id>> in PM's even if you're not the room creator iirc.

I think this change (limiting a groupchat to an actual PM groupchat as opposed to a room groupchat) also has implications on users trying to request private rooms, or for users who want a temporary private room.

As mentioned in the Pokémon Showdown section in Smogon Forums for Admin Requests (http://www.smogon.com/forums/threads/names-passwords-rooms-and-servers-contacting-senior-staff.3538721/) they did mention - 
```
Before requesting a private room there must be evidence of an active groupchat that has existed for a reasonable amount of time (about a month at least). We can access groupchat logs, so just tell us which names you've been using for them and how many users you have on average.
```
Such a change may make this "policy" impossible since a lot of groupchats tend to have their own hierarchy within the users in the chat, possibly in preparation to request a permanent private room.

Zarel:
@sparkychildcharlie, I think you commented on the wrong thing? This is just a client pullreq to support more ranks.

PR #849, opened 2017-01-08
Dont mark Gen 7 OU pokemon as banned

HoeenCoder:
Currently, whenever no format or a gen 7 format is selected in teambuilder, pokemon that are OU in gen 7 are incorrectly marked as banned. This is my attempt to fix this. I might be doing this wrong, If I am please do point it out. I was unable to test this as I cannot run the files in `./githooks`, but from what I've seen this should work.


HoeenCoder:
(bumping this in case it was forgotten, my apologies if it was not and I seem impatient, take your time)

HoeenCoder:
I think I should of put `"OU", "Bank-OU", before `"New", "Bank"`.

HoeenCoder:
@Zarel check this out when you get a chance

PR #850, opened 2017-01-10
Dark Mode: Make it easier to read selected teambuilder options 

HoeenCoder:
Its hard to read the info for a move / ability / item / pokemon when its selected in teambuilder, i changed the font color from `#999` to `#FFF` to resolve this.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -3367,8 +3367,18 @@ a.ilink.yours {
 .dark .utilichart .movenamecol {
 	color: #DDD;
 }
+.dark .utilichart .cur .labelcol em,
+.dark .utilichart .cur .widelabelcol em,
+.dark .utilichart .cur .pplabelcol em,
+.dark .utilichart .cur .statcol em,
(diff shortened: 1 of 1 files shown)


style/client.css
```diff
@@ -3367,9 +3367,21 @@ a.ilink.yours {
...
 }
+.dark .utilichart .cur > .labelcol em,
+.dark .utilichart .cur > .widelabelcol em,
+.dark .utilichart .cur > .pplabelcol em,
+.dark .utilichart .cur > .statcol em,
+.dark .utilichart .cur > .bstcol em {
+	color: #FFF;
+	font-size: 7pt;
+	font-style: normal;
+}
```
Zarel: You should not be changing font size and style in a dark theme...

Zarel: Also you should be mirroring the regular selectors, which do not have `>`

HoeenCoder: My bad, I'll fix those.

Issue #848, opened 2017-01-05
Autojoin support for sideservers


panpawn:
When autojoin was introduced, I remember it saying that support for sideservers would become a thing, and it since has not.

I think supporting this for sideservers would be a nice feature.

HoeenCoder:
This would be a great addition, one thing to handle is how do you keep autojoin separate for different servers & main? Obviously different servers have different rooms and I don't think having your autojoin selection change everytime you switch to a different server will help, in fact it would hurt more than it would help.

panpawn:
> one thing to handle is how do you keep autojoin separate for different servers & main?

Easy, one way of doing that is just having a sub-object held with the serverid if it's not main that contains the room ids.

HoeenCoder:
Since the format of `Tools.prefs('autojoin')` (contains autojoin info) is a string with rooms seperated by commas, what about seperating servers with ``]`` and have the server id listed before the rooms?

Example:
```javascript
var autojoin = "smogonuniversity,lobby,development]spacialgaze,lobby,development]gold,lobby,development";
// each server
let servers = autojoin.split(']');
for (let i = 0; i < servers.length; i++) {
  let data = servers[i].split(',');
  // Server ID
  let server = servers.shift();
  // Rooms === servers[i]
}

```

HoeenCoder:
For a working copy, id probably loop till i found the right server, (default to `''` for no autojoin), then `break;` that loop and run the normal autojoin setup with the roomlist from the server

panpawn:
Why not just literally use objects based on serverids if the severid isn't ``showdown``?

HoeenCoder:
I could do that, maybe it would be best to use a format such as

```javascript
var autojoin = {
  showdown: "lobby,tournaments,development",
  spacialgaze: "lobby,development",
  gold: "lobby,development",
};
```
over the current string setup, as its getting a bit messy with having to parse it twice (once to get the server, then the rooms)

panpawn:
sure, something to that effect is what I meant

HoeenCoder:
Ok, that also allows for
```javascript
if (typeof autjoin === 'String') {
	autojoin = {showdown: autjoin};
}
```
which will come in handy transfering over to the new object setup

HoeenCoder:
@panpawn this can be closed now.

PR #851, opened 2017-01-10
Autojoin support for side servers

HoeenCoder:
Discussed in #848

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -374,24 +374,35 @@
 			this.topbar = new Topbar({el: $('#header')});
 			if (this.down) {
 				this.isDisconnected = true;
-			} else if (document.location.hostname === 'play.pokemonshowdown.com' || Config.testclient) {
-				this.addRoom('rooms', null, true);
+			} else {
+				if (document.location.hostname === 'play.pokemonshowdown.com' || Config.testclient) {
+					this.addRoom('rooms', null, true);
+				} else {
+					this.addRoom('lobby', null, true);
+				}
 				Storage.whenPrefsLoaded(function () {
+					if (!Config.server.registered) return;
 					var autojoin = (Tools.prefs('autojoin') || '');
 					var autojoinIds = [];
+					if (typeof autojoin === 'string') {
+						// Use the existing autojoin string for showdown, and an empty string for other servers.
+						if (Config.server.id !== 'showdown') autojoin = '';
+					} else {
+						// If there is not autojoin data for this server, use a empty string.
+						autojoin = autojoin[Config.server.id] || '';
+					}
 					if (autojoin) {
 						var autojoins = autojoin.split(',');
 						for (var i = 0; i < autojoins.length; i++) {
 							var roomid = toRoomid(autojoins[i]);
 							app.addRoom(roomid, null, true, autojoins[i]);
-							if (roomid !== 'staff' && roomid !== 'upperstaff') autojoinIds.push(roomid);
+							if (roomid === 'staff' || roomid === 'upperstaff') continue;
+							if (Config.server.id !== 'showdown' && roomid === 'lobby') continue;
+							autojoinIds.push(roomid);
 						}
 					}
 					app.send('/autojoin ' + autojoinIds.join(','));
 				});
-			} else {
-				this.addRoom('lobby', null, true);
-				this.send('/autojoin');
 			}
 
 			var self = this;
@@ -1636,7 +1647,7 @@
 			document.title = room.title ? room.title + " - Showdown!" : "Showdown!";
 		},
 		updateAutojoin: function () {
-			if (Config.server.id !== 'showdown') return;
+			if (!Config.server.registered) return;
 			var autojoins = [];
 			var autojoinCount = 0;
 			var rooms = this.roomList.concat(this.sideRoomList);
@@ -1644,11 +1655,37 @@
 				var room = rooms[i];
 				if (room.type !== 'chat') continue;
 				autojoins.push(room.id.indexOf('-') >= 0 ? room.id : (room.title || room.id));
-				if (room.id === 'staff' || room.id === 'upperstaff') continue;
+				if (room.id === 'staff' || room.id === 'upperstaff' || (Config.server.id !== 'showdown' && room.id === 'lobby')) continue;
 				autojoinCount++;
 				if (autojoinCount >= 10) break;
 			}
-			Tools.prefs('autojoin', autojoins.join(','));
+			var curAutojoin = (Tools.prefs('autojoin') || '');
+			if (typeof curAutojoin !== 'string') {
+				if (curAutojoin[Config.server.id] === autojoins.join(',')) return;
+				if (!autojoins.length) {
+					delete curAutojoin[Config.server.id];
+					// If the only key left is 'showdown', revert to the string method for storing autojoin.
+					var count = 0;
+					for (var key in curAutojoin) {
+						if (count >= 2) break;
+						count++;
+					}
+					if (count === 1 && curAutojoin.showdown) curAutojoin = curAutojoin.showdown;
+				} else {
+					curAutojoin[Config.server.id] = autojoins.join(',');
+				}
+			} else {
+				if (Config.server.id !== 'showdown') {
+					// Switch to the autojoin object to handle multiple servers
+					curAutojoin = {showdown: curAutojoin};
+					if (!autojoins.length) return;
+					curAutojoin[Config.server.id] = autojoins.join(',');
+				} else {
+					if (curAutojoin === autojoins.join(',')) return;
+					curAutojoin = autojoins.join(',');
+				}
+			}
+			Tools.prefs('autojoin', curAutojoin);
 		},
 
 		/*********************************************************
--- a/js/client.js
+++ b/js/client.js
@@ -1665,12 +1665,13 @@
 				if (!autojoins.length) {
 					delete curAutojoin[Config.server.id];
 					// If the only key left is 'showdown', revert to the string method for storing autojoin.
-					var count = 0;
+					var hasSideServer = false;
 					for (var key in curAutojoin) {
-						if (count >= 2) break;
-						count++;
+						if (key === 'showdown') continue;
+						hasSideServer = true;
+						break;
 					}
-					if (count === 1 && curAutojoin.showdown) curAutojoin = curAutojoin.showdown;
+					if (!hasSideServer) curAutojoin = curAutojoin.showdown || '';
 				} else {
 					curAutojoin[Config.server.id] = autojoins.join(',');
 				}


HoeenCoder:
@Zarel should autojoin require the server to be registered to work?

js/client.js
```diff
@@ -374,11 +374,21 @@
...
+				} else {
+					this.addRoom('lobby', null, true);
+				}
 				Storage.whenPrefsLoaded(function () {
-					var autojoin = (Tools.prefs('autojoin') || '');
+					var autojoin = (Tools.prefs('autojoin') || {});
 					var autojoinIds = [];
+					if (typeof autojoin === 'string') {
+						// Switch to the new autojoin object
+						autojoin = {showdown: autojoin};
```
Zarel: Would rather leave it as a string if there aren't other servers.

HoeenCoder: Reason I did that is so it uses the new object based autojoin system if your cookies are storing the old string based one still. I could run a setup so that if you don't already have a object based setup and its PS main your connecting too, to just use a string if you want.

Zarel: Yes, that's what I was intending to ask for.

HoeenCoder:
uh dont merge this yet i screwed up and its not using strings for main :/

HoeenCoder:
Ok issues were fixed.

js/client.js
```diff
@@ -1636,7 +1651,19 @@
 			document.title = room.title ? room.title + " - Showdown!" : "Showdown!";
 		},
 		updateAutojoin: function () {
-			if (Config.server.id !== 'showdown') return;
+			if (!Config.server.registered) return;
+			var curAutojoin = (Tools.prefs('autojoin'));
+			var useString = false;
+			if (typeof curAutojoin === 'string' || curAutojoin === undefined) {
```
Zarel: This is pretty verbose code.

How many lines can you save with `var curAutojoin = (Tools.prefs('autojoin') || '');`?

js/client.js
```diff
@@ -1636,7 +1651,19 @@
...
+				if (Config.server.id !== 'showdown') {
+					// Switch to the new autojoin object
```
Zarel: "new" is no longer a good descriptor here.

js/client.js
```diff
@@ -1636,7 +1651,19 @@
 			document.title = room.title ? room.title + " - Showdown!" : "Showdown!";
 		},
 		updateAutojoin: function () {
-			if (Config.server.id !== 'showdown') return;
+			if (!Config.server.registered) return;
+			var curAutojoin = (Tools.prefs('autojoin'));
+			var useString = false;
```
Zarel: If you default `useString` to `true`, you can cut a few more lines. But considering you only read it once, you might want to remove it entirely and just use `typeof curAutojoin === 'string'`.

HoeenCoder:
Hows this @Zarel ?

js/client.js
```diff
@@ -374,11 +374,24 @@
...
+					this.addRoom('rooms', null, true);
+				} else {
+					this.addRoom('lobby', null, true);
+				}
 				Storage.whenPrefsLoaded(function () {
+					if (!Config.server.registered) return;
 					var autojoin = (Tools.prefs('autojoin') || '');
 					var autojoinIds = [];
+					if (typeof autojoin === 'string') {
+						if (Config.server.id !== 'showdown') {
```
Zarel: At this point you can combine the two checks with `&&`

Zarel: Also probably have an early exit if `autojoin` is blank or `lobby`

js/client.js
```diff
@@ -1648,7 +1665,12 @@
 				autojoinCount++;
 				if (autojoinCount >= 10) break;
 			}
-			Tools.prefs('autojoin', autojoins.join(','));
+			if (typeof curAutojoin !== 'string') {
+				curAutojoin[Config.server.id] = autojoins.join(',');
```
Zarel: Also probably have an early exit if autojoin is blank or lobby

HoeenCoder: `Tools.prefs('autojoin');` exits without saving the data if there is no change in the roomlist.

HoeenCoder:
I'll be busy for the next few days and might not be able to update this if necessary. Ill try to get to it ASAP though.

HoeenCoder:
I think we both forgot about this @Zarel (or I forgot to mention you), tell me if there is anything else you would like changed.

Zarel:
The entry should be deleted if the contents is just `lobby` or empty.

I don't want to arbitrarily increase the amount of data storage just for visiting a new server.

HoeenCoder:
@Zarel These are the changes made:

- Don't add lobby to the autojoin list on sideservers. (Since its already autojoined by default.)

- Don't create a slot for a server or switch to the object setup if there's nothing to autojoin. yet (ex: lobby)

- If the servers has no rooms to autojoin, remove it.

- If the only server left when using the object setup is `showdown`, switch back to a string.

- Don't update string autojoin if nothing changed. (This was already done for the object setup but now is done for the string setup.)

js/client.js
```diff
@@ -374,24 +374,38 @@
...
+					if (typeof autojoin === 'string') {
+						// Use the existing autojoin string for showdown, and an empty string for other servers.
+						if (Config.server.id !== 'showdown') autojoin = '';
+					} else {
+						// If there is not autojoin data for this server, use a empty string.
+						if (!autojoin[Config.server.id]) {
+							autojoin = '';
+						} else {
+							autojoin = autojoin[Config.server.id];
+						}
```
Zarel: autojoin = autojoin[Config.server.id] || '';

js/client.js
```diff
@@ -374,24 +374,38 @@
...
+					}
 					if (autojoin) {
 						var autojoins = autojoin.split(',');
 						for (var i = 0; i < autojoins.length; i++) {
 							var roomid = toRoomid(autojoins[i]);
 							app.addRoom(roomid, null, true, autojoins[i]);
-							if (roomid !== 'staff' && roomid !== 'upperstaff') autojoinIds.push(roomid);
+							if (roomid === 'staff' || roomid === 'upperstaff' || (Config.server.id !== 'showdown' && roomid === 'lobby')) continue;
```
Zarel: For readability, these should possibly be split into two lines.

js/client.js
```diff
@@ -1636,19 +1650,37 @@
...
+				// Switch to the autojoin object to handle multiple servers
+				curAutojoin = {showdown: curAutojoin};
+			}
 			var autojoins = [];
 			var autojoinCount = 0;
 			var rooms = this.roomList.concat(this.sideRoomList);
 			for (var i = 0; i < rooms.length; i++) {
 				var room = rooms[i];
 				if (room.type !== 'chat') continue;
+				if (room.id === 'staff' || room.id === 'upperstaff' || (Config.server.id !== 'showdown' && room.id === 'lobby')) continue;
```
Zarel: We still want to store order of Staff and Upper Staff. That's why they're below rather than above the autojoin list.

js/client.js
```diff
@@ -1636,19 +1650,37 @@
...
 				autojoinCount++;
 				if (autojoinCount >= 10) break;
 			}
-			Tools.prefs('autojoin', autojoins.join(','));
+			if (typeof curAutojoin !== 'string') {
+				if (curAutojoin[Config.server.id] === autojoins.join(',')) return;
+				if (!autojoins.length) {
+					delete curAutojoin[Config.server.id];
+					// If the only key left is 'showdown', revert to the string method for storing autojoin.
+					if (Object.keys(curAutojoin).length === 1 && Object.keys(curAutojoin)[0] === 'showdown') curAutojoin = curAutojoin.showdown;
```
Zarel: Is `Object.keys` supported by ES3? You should probably keep track of that separately.

js/client.js
HoeenCoder: Whoops!

js/client.js
HoeenCoder: @Zarel which do you prefer (or something else)?
```javascript
let count = 0;
for (let key in curAutojoin) count++;
if (count === 1 && curAutojoin['showdown']) curAutojoin = curAutojoin.showdown;
```
----
Adding a `count` property to the autojoin object that keeps track of the number of servers listed (Id also need a suggestion for a key name that cant be registered as a server name)

Zarel: We could also try dropping support for Safari 4. I hope it's been long enough. :|

Zarel: I'd rather not make that decision now, though.

Right now, there are easier ways to prove `curAutojoin` has more than one element. You can make it O(1) by breaking out of your `for` loop when `count` reaches `2`, for instance.

HoeenCoder:
@Zarel hows this looking?

js/client.js
```diff
@@ -1636,19 +1647,42 @@
 			document.title = room.title ? room.title + " - Showdown!" : "Showdown!";
 		},
 		updateAutojoin: function () {
-			if (Config.server.id !== 'showdown') return;
+			if (!Config.server.registered) return;
+			var curAutojoin = (Tools.prefs('autojoin') || '');
+			if (typeof curAutojoin === 'string' && Config.server.id !== 'showdown') {
+				// Switch to the autojoin object to handle multiple servers
+				curAutojoin = {showdown: curAutojoin};
+			}
```
Zarel: New lines 1651-1655 should be moved down to right above

    if (typeof curAutojoin !== 'string') {

Keeping everything together will make things more readable, and you'll probably be able to see other optimizations better, too.

HoeenCoder:
@Zarel hope its not too late to mention you, sorry if I missed anything ive been in a rush today.

js/client.js
```diff
@@ -1636,19 +1647,45 @@
...
 				if (autojoinCount >= 10) break;
 			}
-			Tools.prefs('autojoin', autojoins.join(','));
+			var curAutojoin = (Tools.prefs('autojoin') || '');
+			if (typeof curAutojoin !== 'string') {
+				if (curAutojoin[Config.server.id] === autojoins.join(',')) return;
+				if (!autojoins.length) {
+					delete curAutojoin[Config.server.id];
+					// If the only key left is 'showdown', revert to the string method for storing autojoin.
+					var count = 0;
```
Zarel: `var hasSideServer = false;`

js/client.js
```diff
@@ -1636,19 +1647,45 @@
...
+					for (var key in curAutojoin) {
+						if (count >= 2) break;
```
```
if (key ==== 'showdown') continue;
hasSideServer = true;
break;
```

js/client.js
```diff
@@ -1636,19 +1647,45 @@
...
+						count++;
+					}
+					if (count === 1 && curAutojoin.showdown) curAutojoin = curAutojoin.showdown;
```
Zarel: `if (!hasSideServer) curAutojoin = curAutojoin.showdown || '';`

HoeenCoder:
@Zarel I wont give up on this, no matter how bad I am at optimizing this.

Zarel:
Don't give up! You're doing great!

Zarel:
Actually, this is everything, thanks!

Zarel:
And don't worry about the optimization stuff too much. No one's good at it at first, it's all about experience.

HoeenCoder:
Thanks @Zarel also thanks for development @ ^_^

PR #858, opened 2017-01-31
Change popup color in darkmode

spekss:
I think the color is too much blue and that's not make a good contrast

preview: 
 new
 new
 old

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -3235,7 +3235,7 @@ a.ilink.yours {
 	border-color: #888;
 
 	box-shadow: 2px 2px 3px rgba(0,0,0,.2);
-}" height
+}
 
 .dark .popupmenu button,
 .dark .bglist button,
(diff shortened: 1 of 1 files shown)


AustinXII:
Idk If it's just my screen but I really don't see a difference

spekss:
The old is more blue that the new

style/client.css
```diff
@@ -3230,12 +3230,12 @@ a.ilink.yours {
...
 
 .dark .ps-popup {
-	background: #0D151E;
+	background: #1e2225;
 	color: #DDD;
 	border-color: #888;
 
 	box-shadow: 2px 2px 3px rgba(0,0,0,.2);
-}
+}" height
```
Zarel: ...

Zarel:
The bigger problem might be to lighten link colors, if contrast is an issue.

spekss:
oups sorry for that I had written it to display the screen, and I have missclick and write it here

Zarel:
This doesn't seem to be necessary and the criticisms don't seem to be handled. Closing.

PR #859, opened 2017-02-01
Update min/max speed calculations for Random Doubles/Triples Battles

ascriptmaster:
In Gen 7, all formats with "Random" in its name will start using the same
EV calculation formula as Random Battle (singles).

Reliant on Zarel/Pokemon-Showdown#3199

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -751,24 +751,28 @@ var BattleTooltips = (function () {
 	// Functions to calculate speed ranges of an opponent.
 	BattleTooltips.prototype.getTemplateMinSpeed = function (template, level) {
 		var baseSpe = template.baseStats['spe'];
-		if (this.battle.gen < 7) {
-			var overrideStats = BattleTeambuilderTable['gen' + this.battle.gen].overrideStats[template.id];
+		var tier = this.battle.tier;
+		var gen = this.battle.gen;
+		if (gen < 7) {
+			var overrideStats = BattleTeambuilderTable['gen' + gen].overrideStats[template.id];
 			if (overrideStats && 'spe' in overrideStats) baseSpe = overrideStats['spe'];
 		}
 
-		var nature = (this.battle.tier.indexOf('Random Battle') >= 0 || this.battle.gen < 3) ? 1 : 0.9;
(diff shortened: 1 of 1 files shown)


ascriptmaster:
Is there a reason we're using `indexOf` instead of `includes`, actually? Old browsers?

Zarel:
Yes, old browsers. We support down to ES3 and `includes` is from ES7.

ascriptmaster:
Geez. What version of IE is that again

ascriptmaster:
Server side is done. Client side should be pushed on next hotpatch or restart.

Zarel:
Actually, IE9 supports all the ES5 we care about. It's Safari 5 (old versions of iOS) that force us down to ES3.

ascriptmaster:
Ah yes, old iPhone technology... RIP

Anyways yeah look this over and if this looks good, merge when you restart the server or hotpatch formats

js/client-battle-tooltips.js
```diff
@@ -751,24 +751,28 @@ var BattleTooltips = (function () {
...
-			var overrideStats = BattleTeambuilderTable['gen' + this.battle.gen].overrideStats[template.id];
+		var tier = this.battle.tier;
+		var gen = this.battle.gen;
+		if (gen < 7) {
+			var overrideStats = BattleTeambuilderTable['gen' + gen].overrideStats[template.id];
 			if (overrideStats && 'spe' in overrideStats) baseSpe = overrideStats['spe'];
 		}
 
-		var nature = (this.battle.tier.indexOf('Random Battle') >= 0 || this.battle.gen < 3) ? 1 : 0.9;
+		var nature = (tier.indexOf('Random Battle') >= 0 || (tier.indexOf('Random') >= 0 && gen >= 7) || gen < 3) ? 1 : 0.9;
```
Zarel: Wait, is the tier name "Random Doubles Battle" rather than "Doubles Random Battle" or something?

ascriptmaster: Seems to be `Random Doubles Battle`

Zarel: I think we should at least detect the presence of both "Random" and "Battle".

ascriptmaster:
Bumpity

PR #860, opened 2017-02-06
Show power of generic Z moves in tooltips

urkerab:
- Not sure how I should clone the move to change its name/category/type
- Not implemented showing Z effect of status moves yet
- Not implemented unusual Z moves such as Z-Weather Ball

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -113,9 +113,10 @@ var BattleTooltips = (function () {
 		var text = '';
 		switch (type) {
 		case 'move':
+		case 'zmove':
 			var move = Tools.getMove(thing);
 			if (!move) return;
-			text = this.showMoveTooltip(move);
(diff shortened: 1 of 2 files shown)


Zarel:
This is a good idea, but ideally we probably want it sent server-side...

urkerab:
I've looked into sending the data server-side and although it's certainly feasible for the request to include the base power and category I can't actually see how to extract that from the tooltip code.

Zarel:
The tooltip code needs to be seriously rewritten, anyway :|

urkerab:
The other benefit of this client-side change is that it will become possible to show the full effect of a status Z-move, which the client isn't really set up to do otherwise, unless you're planning to do that as part of your tooltip rewrite?

Zarel:
You're right that this is an improvement over the current situation......... but I still think tooltips need to be rewritten; they're a mess right now.

PR #864, opened 2017-02-12
fix fail in credit button

spekss:
fix this

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -1095,9 +1095,9 @@
 			buf += '<li><p>[<strong>The Immortal</strong>] <small>&ndash; Development</small></p></li></ul>';
 			buf += '<h2>Retired Staff</h2>';
 			buf += '<ul><li><p><a href="http://meltsner.com/" target="_blank" class="subtle"><strong>Bill Meltsner</strong> [bmelts]</a> <small>&ndash; Development</small></p></li>';
-			buf += '<li><p><a href="https://cathyjf.com/" target="_blank" class="subtle"><strong>Cathy J. Fitzpatrick</strong> [cathyjf]</a> <small>&ndash; Development, Sysadmin</small></p></li></ul>';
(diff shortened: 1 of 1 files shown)


js/client-mainmenu.js
```diff
@@ -1095,9 +1095,9 @@
 			buf += '<li><p>[<strong>The Immortal</strong>] <small>&ndash; Development</small></p></li></ul>';
 			buf += '<h2>Retired Staff</h2>';
 			buf += '<ul><li><p><a href="http://meltsner.com/" target="_blank" class="subtle"><strong>Bill Meltsner</strong> [bmelts]</a> <small>&ndash; Development</small></p></li>';
-			buf += '<li><p><a href="https://cathyjf.com/" target="_blank" class="subtle"><strong>Cathy J. Fitzpatrick</strong> [cathyjf]</a> <small>&ndash; Development, Sysadmin</small></p></li></ul>';
+			buf += '<li><p><a href="https://cathyjf.com/" target="_blank" class="subtle"><strong>Cathy J. Fitzpatrick</strong> [cathyjf]</a> <small>&ndash; Development, Sysadmin</small></p></li>';
 			buf += '<li><p><strong>Hugh Gordon</strong> [V4] <small>&ndash; Research (game mechanics), Development</small></p></li>';
```
Bryan-0: <ul> is needed here.

Slayer95:
We should probably move the ``<ul>`` tags to their own lines.
This happens EVERY time.

js/client-mainmenu.js
spekss: are you sure ? The bug of the screen is in this line there was a `</ul>` at the end, so the lines with V4 and Joim was not in the `<ul>`, but I have add `</ul>` at the end of the joim's line, so all "retired staff" is in the `<ul>` now

spekss:
slayer, do you mean I should put `<ul>` At the beginning of each line? and `</ul>` at the end ?

js/client-mainmenu.js
spekss: I mean, the `<ul>` is open in the line 1097 and end at the line 1100, all line is in the tag, it's not fine ?

Zarel:
In the long term we need to remove this button entirely...

PR #866, opened 2017-02-16
Properly format Pokemon with event abilities in /data

asgdf:
To be specific, to fix this: 
grafik

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -486,22 +486,17 @@ var Tools = {
 			buf += '</span> ';
 			buf += '<span style="float:left;min-height:26px">';
 			if (template.abilities['1']) {
-				buf += '<span class="col twoabilitycol">';
+				buf += '<span class="col twoabilitycol">' + template.abilities['0'] + '<br />' + template.abilities['1'] + '</span>';
 			} else {
-				buf += '<span class="col abilitycol">';
+				buf += '<span class="col abilitycol">' + template.abilities['0'] + '</span>';
 			}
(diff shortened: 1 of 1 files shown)


js/battledata.js
```diff
@@ -486,22 +486,17 @@ var Tools = {
...
-				if (i == 'H') {
-					ability = '</span><span class="col abilitycol"><em>' + (template.unreleasedHidden ? '<s>' + ability + '</s>' : ability) + '</em>';
-				}
-				buf += ability;
+			if (template.abilities['S']) {
+				buf += '<span class="col twoabilitycol' + (template.unreleasedHidden ? ' unreleasedhacol' : '') + '"><em>' + template.abilities['H'] + '<br />' + template.abilities['S'] + '</em></span>';
+			} else if (template.abilities['H']) {
+				buf += '<span class="col abilitycol' + (template.unreleasedHidden ? ' unreleasedhacol' : '') + '"><em>' + template.abilities['H'] + '</em></span>';
+			} else {
+				buf += '<span class="col abilitycol" />';
```
Zarel: This is HTML, not XML; `<span />` is invalid code; you need to write `<span></span>`

Zarel: `<span />` is actually treated as `<span>` (an unclosed tag) by HTML, unfortunately.

asgdf:
TIL, didn't know that. I assumed it worked for all empty elements. Anyway, fixed.

Zarel:
Yeah, I guess it's better this way for explicitness...

PR #869, opened 2017-02-22
Add /clear to PMs

sparkychildcharlie:
In response to [this suggestion that I find useful and interesting][1], I added /clear to the PM box.  This does:
- closes the pm box
- flushes out chat history and the inner HTML.

(PS: Please let me know if you want this as a regular chatroom based command as well, I can __totally__ try to add that too!)

[1]: http://www.smogon.com/forums/threads/suggestions.3534365/page-49#post-7230510

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -326,6 +326,9 @@
 					delete app.ignore[userid];
 					$chat.append('<div class="chat">User ' + userid + ' no longer ignored.</div>');
 				}
+			} else if (text.toLowerCase() === '/clear') {
+				this.closePM(userid);
+				this.$pmBox.find('.pm-window-' + userid).find('.inner').empty();
 			} else {
 				text = ('\n' + text).replace(/\n\n/g, '\n').replace(/\n/g, '\n/pm ' + userid + ', ').substr(1);
 				if (text.length > 80000) {
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -329,6 +329,7 @@
 			} else if (text.toLowerCase() === '/clear') {
 				this.closePM(userid);
 				$chat.empty();
+				delete this.chatHistories[userid];
 			} else {
 				text = ('\n' + text).replace(/\n\n/g, '\n').replace(/\n/g, '\n/pm ' + userid + ', ').substr(1);
 				if (text.length > 80000) {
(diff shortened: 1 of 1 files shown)


js/client-mainmenu.js
```diff
@@ -326,6 +326,9 @@
 					delete app.ignore[userid];
 					$chat.append('<div class="chat">User ' + userid + ' no longer ignored.</div>');
 				}
+			} else if (text.toLowerCase() === '/clear') {
+				this.closePM(userid);
+				this.$pmBox.find('.pm-window-' + userid).find('.inner').empty();
```
Slayer95: This could be just ``$pmWindow.find('.inner').empty()`` afaict.

sparkychildcharlie: ah thanks! Sorry, I'm not as familiar with client... yet!

sparkychildcharlie: (i used ``$chat.empty()``)

Slayer95:
It's not intuitive for ``/clear`` to also close the PM.

> as a regular chatroom based command

To be *clear*, this is already a client-side command for chat rooms, and it doesn't close the room.

sparkychildcharlie:
@Slayer95 would you prefer another name, or for this just to __not__ close the pm?

Slayer95:
Also,

> flushes out chat history

I don't see where this is done? Perhaps you forgot to commit it?

sparkychildcharlie:
doesn't ``$chat.empty()`` do that? This is what is used in the chatroom command /clearpms (of which this is a single instance of what that does) and after using it, reopening the pm box will not have any of the previous proceedings showing.

Or is there something very important I'm overlooking?

Slayer95:
Oh, *that* history. I thought you meant the history that is browsed with Up Arrow / Down Arrow, which ``/clearpms`` doesn't flush either. This would be done with ``delete this.chatHistories[userid]``.

sparkychildcharlie:
I'll wipe that too then i guess.

Zarel:
...no, don't wipe up/down history :|

Zarel:
Also, I thought we agreed not to close the PM?

sparkychildcharlie:
Sorry Zarel, I didn't notice the edit in that post.   This just clears the HTML in the pm box then.

Zarel:
Looks good. Thanks!

PR #870, opened 2017-02-22
Make hidden avatars last through into the next PS session

sparkychildcharlie:
- hardcode those avatars as potential values that will be written into localStorage.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -868,9 +868,12 @@
 
 			case 'avatar':
 				var parts = target.split(',');
-				var avatar = parseInt(parts[0], 10);
+				let avatarString = toId(parts[0]);
+				var avatar = parseInt(avatarString, 10);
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -868,9 +868,12 @@
 
 			case 'avatar':
 				var parts = target.split(',');
-				var avatar = parseInt(parts[0], 10);
+				var avatarString = toId(parts[0]);
+				var avatar = parseInt(avatarString, 10);
 				if (avatar) {
 					Tools.prefs('avatar', avatar);
+				} else if (['bw2elesa', 'teamrocket', 'yellow', 'zinnia', 'clemont'].includes(avatarString)) { // custom avatar exceptions
```
sirDonovan: The client supports down to ES3 so you need to use <code>indexOf</code> here.

PR #872, opened 2017-02-24
Allow formatting in PMs and fix spelling mistake

sparkychildcharlie:
- fix spelling mistake in client-chat man -> main
- use ``message`` instead of ``textContent`` for showing PMs in chat.  It goes through ``Tools.parseMessage()`` in addChat in client-chat anyways when generating the green message in chatroom, so no need to use the just text version of the PM.

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -153,7 +153,7 @@
 			var $lastMessage = $chat.children().last();
 			var textContent = $lastMessage.html().indexOf('<span class="spoiler">') >= 0 ? '(spoiler)' : $lastMessage.children().last().text();
 			if (textContent && app.curSideRoom && app.curSideRoom.addPM && Tools.prefs('inchatpm')) {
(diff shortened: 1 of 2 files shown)


js/client-mainmenu.js
```diff
@@ -153,7 +153,7 @@
 			var $lastMessage = $chat.children().last();
 			var textContent = $lastMessage.html().indexOf('<span class="spoiler">') >= 0 ? '(spoiler)' : $lastMessage.children().last().text();
 			if (textContent && app.curSideRoom && app.curSideRoom.addPM && Tools.prefs('inchatpm')) {
```
Zarel: Hm, do we even need `textContent` and `$lastMessage` at all, if we do it this way?

sparkychildcharlie: ``textContent`` is mainly for the popup in notifications, which without formatting may be cleaner

PR #873, opened 2017-02-25
Add |notify|

HoeenCoder:
Discussed in staff, not decided on yet so shouldn't be merged.

Adds the ability to highlight a user through protocol. This is used for declares.

 - `|notify|message to display when hovering room tab` Highlights the user normally in the room. http://imgur.com/a/wcj1k (note this image is from a global declare test, but global declares do not highlight in all rooms)

 - `|globalnotify|message` Sends a PM box from `~` with the message, accepts HTML, also sends a notification sound.

EDIT: update information to match changes

Paired with server pull request: https://github.com/Zarel/Pokemon-Showdown/pull/3294

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1237,12 +1237,12 @@
 					this.$chat.append('<div class="notice">' + Tools.sanitizeHTML(row.slice(1).join('|')) + '</div>');
 					break;
 
-				case 'globalhighlight':
-				case 'highlight':
+				case 'globalnotify':
+				case 'notify':
 					if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
 						soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
 					}
-					if (row[0] !== 'globalhighlight') this.notifyOnce("Highlighted by a declare", "\"" + row.slice(1).join('|') + "\"", 'highlight');
+					if (row[0] !== 'globalnotify') this.notifyOnce("You have a notification", "\"" + row.slice(1).join('|') + "\"", 'highlight');
 					break;
 
 				case 'error':
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1238,12 +1238,9 @@
 					break;
 
 				case 'globalnotify':
-					if (app.user.lastGlobalNotify !== row.slice(1).join('|')) {
-						app.rooms[''].addPM('~', '/html ' + row.slice(1).join('|'), '' + this.room.users[app.user.changed.userid]);
-						if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
-							soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
-						}
-						app.user.lastGlobalNotify = row.slice(1).join('|');
+					app.rooms[''].addPM('~', '/raw ' + row.slice(1).join('|'), '' + this.users[app.user.changed.userid]);
+					if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
+						soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
 					}
 					break;
 
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1238,9 +1238,12 @@
 					break;
 
 				case 'globalnotify':
-					app.rooms[''].addPM('~', '/html ' + row.slice(1).join('|'), '' + this.room.users[app.user.changed.userid]);
-					if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
-						soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
+					if (app.user.lastGlobalNotify !== row.slice(1).join('|')) {
+						app.rooms[''].addPM('~', '/html ' + row.slice(1).join('|'), '' + this.room.users[app.user.changed.userid]);
+						if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
+							soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
+						}
+						app.user.lastGlobalNotify = row.slice(1).join('|');
 					}
 					break;
 					
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1237,17 +1237,12 @@
 					this.$chat.append('<div class="notice">' + Tools.sanitizeHTML(row.slice(1).join('|')) + '</div>');
 					break;
 
+				case 'globalhighlight':
 				case 'highlight':
-					var soundOnly = row[1] === 'global';
-					this.$chat.append('<div class="notice">' + Tools.sanitizeHTML(row.slice((soundOnly ? 2 : 1)).join('|')) + '</div>');
 					if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
 						soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
 					}
-					if (!soundOnly) {
-						var $lastMessage = this.$chat.children().last();
-						var notifyText = $lastMessage.html().indexOf('<span class="spoiler">') >= 0 ? '(spoiler)' : $lastMessage.children().last().text();
-						this.notifyOnce("Highlighted by a declare", "\"" + notifyText + "\"", 'highlight');
-					}
+					if (row[0] !== 'globalhighlight') this.notifyOnce("Highlighted by a declare", "\"" + Tools.escapeHTML(row.slice(1).join('|')) + "\"", 'highlight');
 					break;
 
 				case 'error':
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1246,7 +1246,7 @@
 						app.user.lastGlobalNotify = row.slice(1).join('|');
 					}
 					break;
-					
+
 				case 'notify':
 					if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
 						soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1242,7 +1242,7 @@
 					if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
 						soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
 					}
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -1237,6 +1237,19 @@
 					this.$chat.append('<div class="notice">' + Tools.sanitizeHTML(row.slice(1).join('|')) + '</div>');
 					break;
 
+				case 'highlight':
+					var soundOnly = row[1] === 'global';
+					this.$chat.append('<div class="notice">' + Tools.sanitizeHTML(row.slice((soundOnly ? 2 : 1)).join('|')) + '</div>');
```
Slayer95: Can we have this independent from chat?

HoeenCoder: You'll have to be more specific, I don't understand what you mean by that.

Slayer95: For ``|highlight|`` to only notify. No chat message.

js/client-chat.js
```diff
@@ -1237,6 +1237,16 @@
 					this.$chat.append('<div class="notice">' + Tools.sanitizeHTML(row.slice(1).join('|')) + '</div>');
 					break;
 
+				case 'globalhighlight':
+				case 'highlight':
+					if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
+						soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
+					}
+					if (row[0] !== 'globalhighlight') {
+						this.notifyOnce("Highlighted by a declare", "\"" + Tools.escapeHTML(row.slice(1).join('|')) + "\"", 'highlight');
```
Slayer95: ``notifyOnce`` doesn't parse HTML, so no need to escape it.

Morfent:
Should users be able to configure whether these messages highlight them, if only for consistency with the other ways they can get highlighted?

Zarel:
I would rather not do this this way.

Actually, we need to refactor global declares probably.

They should be handled very largely client-side. They also need to appear when the user isn't in any rooms, probably as a PM. We already implement PMs to/from `~`

sparkychildcharlie:
I personally really love how this works.  It would work well in tandem with room games that don't have a method of highlighting users when it's their turn (like tournaments).  Having an extra protocol "tag" that hl's people could make it easier to notify people when it's their turn or when a game starts.

Zarel:
I'm still recommending that the global declare be refactored.

The regular declare can stay, but the message still needs to be changed.

Zarel:
You often have a temptation to write messages to users the way you name variables. Don't. "Highlight" doesn't mean anything to the user.

You should actually do it the other way around. Name variables the way you'd write messages to users. So don't call it `highlight`, call it `notify`.

HoeenCoder:
Ok I haven't done the global declare refactor yet because I wanted to ask a few questions about it. 

I know that you want it to show up in a PM as well as rooms, but what should the PM name be (I haven't checked if its possible to send an unnammed PM yet if so ignore this Q). 

I've considered replacing the server side implementation almost completly, the issue is declares wouldn't stick around on the server rooms since the message isn't stored on the server that way. I think we should just stick with our current setup + a PM and notification sound for it.

Zarel:
The PM should be from `~`.

...I should probably actually implement the PM facility first, shouldn't I?

HoeenCoder:
If you haven't, yes. I haven't gotten a chance to research client side PM code yet.

sparkychildcharlie:
I know this isn't my pull request but i'd like to (bump).  

While it's established that this isn't effective for global declares, I'd still like to request for this (the room notify one at least) to be implemented, as it would help notify users immensely with certain games in rooms, such as uno, scavengers and other potential activities, in the same way tours will spawn a notification.

Zarel:
New idea: PM yourself!!!!

Zarel:
Getting a PM from yourself might be kind of awkward, but honestly PMs from the server work the same way.

HoeenCoder:
Actually @Zarel what about a PM from `~`? ex: `user.send('|pm|~|' + user.group + user.name + '|/html <div class="broadcast-blue"><b>' + message + '</b></div>');` Only possible issue is the semicolon is colored.

Zarel:
Yeah, in hindsight, that's probably better.

HoeenCoder:
I tracked the last global declare in ``app.user``, if theres a better place to track it, or a better way to prevent spam from globalnotify pms, feel free to correct me.

HoeenCoder:
Hold off on merging this, for some reason globalnotify isn't getting to the client, will work more on this when I get some more time tommorow.

sparkychildcharlie:
If I'm not mistaken, PMs starting with ``/html`` will show the user's name as opposed to ``/raw`` which simply shows the html and nothing more.  I think ``/raw`` will be preferable in this, specially since a lot of side servers generally use HTML styling in their global declares.

HoeenCoder:
@Zarel tested and ready for review.

js/client-chat.js
```diff
@@ -1237,6 +1237,20 @@
...
+					if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
+						soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
+					}
+					break;
+
+				case 'notify':
+					if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
+						soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
+					}
+					this.notifyOnce("You have a notification", "\"" + row.slice(1).join('|') + "\"", 'highlight');
```
Zarel: `You have a notification` is pretty vague here.

Zarel: "The staff in `roomid` made a declaration"?

HoeenCoder: I did that because as @sparkychildcharlie suggested it can also be used for games, ect. (UNO)

Zarel: Maybe `|html|[html]` and `|html-notify|[notification message]|[html]`

HoeenCoder: Not sure what you mean by that in the context of this review box

Zarel: Oh, wait, nevermind, I see, you're making `|notify|` a separate message. I'd say just pass both the title and message into `|notify|`.

HoeenCoder: Yeah i seperated notify and globalnotify since they function differently now. One Hls you in a room, the other dosent HL in rooms. 

Currently when /declare is used, it sends `|notify|${Chat.escapeHTML(target)}` to the client, the client then takes whatever is after `|notify|` and puts that as the notification reason. I'm not sure what you mean by "pass both the title and message into `|notify|`"

sparkychildcharlie: I think Zarel means something like ``|notify|title|message``

Zarel: ^

HoeenCoder: Oh sure, sorry about the confusion. I'll start on that.

js/client-chat.js
```diff
@@ -1237,6 +1237,20 @@
 					this.$chat.append('<div class="notice">' + Tools.sanitizeHTML(row.slice(1).join('|')) + '</div>');
 					break;
 
+				case 'globalnotify':
+					app.rooms[''].addPM('~', '/raw ' + row.slice(1).join('|'), '' + this.users[app.user.changed.userid]);
```
Zarel: This one's inconsistent now, anyway.

Just make global declares show up as PMs server-side.

js/client-chat.js
```diff
@@ -1237,6 +1237,20 @@
...
+					if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
+						soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
```
Zarel: PMs already make bleeps, don't they?

js/client-chat.js
```diff
@@ -1237,6 +1237,13 @@
 					this.$chat.append('<div class="notice">' + Tools.sanitizeHTML(row.slice(1).join('|')) + '</div>');
 					break;
 
+				case 'notify':
+					if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
+						soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
+					}
+					this.notifyOnce(row[1], "\"" + row.slice(2).join('|') + "\"", 'highlight');
```
Zarel: Do we need the `\"...\"`?

HoeenCoder: Other notifications (read highlights) have "quotes" around the notification message, as seen here https://github.com/Zarel/Pokemon-Showdown-Client/blob/master/js/client-chat.js#L1493

Zarel: Highlights have quotes because the message is from another user, not from PS itself. Messages from PS itself do not have quotes.

Declares, on the other hand, are intentionally designed to look like they're from PS itself.

Zarel: Like, you see the grammatical difference between:

`Sam said, "hi sup"`

`The game is ending soon`

Zarel: A declare is intended to be a message like the latter two: They're formatted to be official server announcements, rather than `/wall`, which is formatted to be more like a message from staff.

Zarel: Incidentally: maybe `/wall` should highlight. That _is_ what its name is supposed to do.

HoeenCoder: Alright, thanks for the clarification, and for /wall i think it shouldn't mass HL, its constantly used to grab attention in a more subtle manner. An example is the host of a mafia game (assuming they are % or up), they use /wall to announce the player list, deaths, anything of importance really. But non-players don't want to be Hled by that tbh, it would become annoying quickly.

panpawn:
I think having the title for most stuff start with the roomid, like, ``<lobby> Modchat was turned on`` would be nice.  (Obviously global declares would be an exception)

HoeenCoder:
I dont think requiring the client to add the roomid is nessecary, If its wanted the server can send that.

panpawn:
>I dont think requiring the client to add the roomid is nessecary, If its wanted the server can send that.
Yeah, true.

js/client-chat.js
```diff
@@ -1237,6 +1237,13 @@
 					this.$chat.append('<div class="notice">' + Tools.sanitizeHTML(row.slice(1).join('|')) + '</div>');
 					break;
 
+				case 'notify':
+					if (!Tools.prefs('mute') && Tools.prefs('notifvolume')) {
+						soundManager.getSoundById('notif').setVolume(Tools.prefs('notifvolume')).play();
+					}
+					this.notifyOnce(row[1], row.slice(2).join('|'), 'highlight');
```
Zarel: Why isn't the sound thingy part of `notifyOnce`, anyway?

PR #874, opened 2017-02-27
Add private message ignoring

sparkychildcharlie:
- Allow the ability to ignore PMs and challenges from users only, but still see their messages in chat.
> For anyone who has been staff, they've probably dealt with users in PMs who are very persistent, and this is a method to stop getting notifications from them, while still be able to see their chatroom messages for moderation purposes.

- Add new commands, and help entries - /pmignore, /pmunignore, /pmignorelist
- Add additional button in menu for users to press to ignore the target's PMs only.

Referenced in [a suggestion on smogon](http://www.smogon.com/forums/threads/suggestions.3534365/page-49#post-7249203) by SparksBlade.  I've had this "feature" via tampermonkey script for more than a year now, and it's certainly something that's very useful.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -517,6 +517,43 @@
 				}
 				return false;
 
+			case 'pmignore':
+				if (!target) {
+					this.parseCommand('/help pmignore');
+					return false;
+				}
+				if (toUserid(target) === app.user.get('userid')) {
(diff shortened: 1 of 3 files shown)


sparkychildcharlie:
OK, i'll PR this to the server repo tomorrow then... i guess this can be closed then!

PR #875, opened 2017-02-28
Fix ignoring in battles

asgdf:
This fixes two things:
1. the auth symbol change for players which caused player messages to always go through in battles, regardless of "ignore opponent" and /ignore
2. the automatic message hiding/unhiding when clicking "Ignore spectators", which used to hide _all_ chat messages. Now it only hides actual chat messages and joins/leaves, but not the details row of /dt or notes like "Spectatoers ignored.".


Zarel:
Both need to be supported.

Zarel:
Can you make a pullreq to support the old symbol as well?

PR #878, opened 2017-03-03
Teambuilder: support set notes

asgdf:
Allowing users to attach notes to themselves on certain sets is a feature that [I've](http://www.smogon.com/forums/threads/3495758/page-66#post-6944260) [often](http://www.smogon.com/forums/threads/3495758/page-63#post-6858968) [seen](http://www.smogon.com/forums/threads/3495758/page-61#post-6828633) [requ](http://www.smogon.com/forums/threads3495758/page-58#post-6770459)[ested](http://www.smogon.com/forums/threads/3495758/page-67#post-6944269), so here we go.  This way, people can conveniently write down the reasons for spreads and weird set decisions.

I initially went for "team notes", but that really wouldn't have meshed well with the way exporting/importing/packing teams is implemented, set-specific notes worked better.

There's now an additional button next to move/copy/delete: 
grafik
Clicking it will append the notes textbox to the bottom of the set like this: 
grafik
The textbox changes height automatically when lines are added/removed (no scrollbars!), and is hidden entirely when focus is lost while it's empty. Notes are also not displayed in the set view (with the move/item/stat selection) due to lack of space.

In packed teams, the set notes are appended to the end of each pokemon, every note (line) separated by a |. This has the unfortunate side effect of preventing anything else from being reasonably appended with that separator, because of the possible variable amount of comments for each mon. I'd have used \n to separate comment lines of the same pokemon, but those are also already in use for separating teams.

In exported teams, set notes look like this (importing them works as well of course):
grafik

Team notes are not sent to the server, they're cut out beforehand.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -34,6 +34,8 @@
 			'change input.teamnameedit': 'teamNameChange',
 			'click button.formatselect': 'selectFormat',
 			'change input[name=nickname]': 'nicknameChange',
+			'input .setnotes textarea': 'setNotesChange',
+			'blur .setnotes textarea': 'setNotesBlur',
 
 			// details
 			'change .detailsform input': 'detailsChange',
@@ -1025,7 +1027,7 @@
 					if (this.curSetList.length < 6 && this.deletedSet && i === this.deletedSetLoc) {
 						buf += '<li><button name="undeleteSet"><i class="fa fa-undo"></i> Undo Delete</button></li>';
 					}
-					buf += this.renderSet(this.curSetList[i], i);
+					buf += this.renderSet(this.curSetList[i], i, true);
 				}
 				if (this.deletedSet && i === this.deletedSetLoc) {
 					buf += '<li><button name="undeleteSet"><i class="fa fa-undo"></i> Undo Delete</button></li>';
@@ -1040,10 +1042,14 @@
 				buf += '</div>';
 			}
 			this.$el.html('<div class="teamwrapper">' + buf + '</div>');
+			var setNotesTextboxes = $('.setnotes textarea');
+			for (var i = 0; i < setNotesTextboxes.length; i++) {
+				this.autoResizeSetNotes(setNotesTextboxes[i]);
+			}
 			this.$(".teamedit textarea").focus().select();
 			if ($(window).width() < 640) this.show();
 		},
-		renderSet: function (set, i) {
+		renderSet: function (set, i, showNotes) {
 			var template = Tools.getTemplate(set.species);
 			var buf = '<li value="' + i + '">';
 			if (!set.species) {
@@ -1055,11 +1061,12 @@
 				buf += '</li>';
 				return buf;
 			}
-			buf += '<div class="setmenu"><button name="copySet"><i class="fa fa-files-o"></i>Copy</button> <button name="importSet"><i class="fa fa-upload"></i>Import/Export</button> <button name="moveSet"><i class="fa fa-arrows"></i>Move</button> <button name="deleteSet"><i class="fa fa-trash"></i>Delete</button></div>';
+			buf += '<div class="setmenu">' + (showNotes ? '<button name="showSetNotes"><i class="fa fa-plus"></i>Add set notes</button>' : '') + '<button name="copySet"><i class="fa fa-files-o"></i>Copy</button> <button name="importSet"><i class="fa fa-upload"></i>Import/Export</button> <button name="moveSet"><i class="fa fa-arrows"></i>Move</button> <button name="deleteSet"><i class="fa fa-trash"></i>Delete</button></div>';
 			buf += '<div class="setchart-nickname">';
 			buf += '<label>Nickname</label><input type="text" name="nickname" class="textbox" value="' + Tools.escapeHTML(set.name || '') + '" placeholder="' + Tools.escapeHTML(template.baseSpecies) + '" />';
 			buf += '</div>';
-			buf += '<div class="setchart" style="' + Tools.getTeambuilderSprite(set, this.curTeam.gen) + ';">';
+			buf += '<div class="setchart">';
+			buf += '<div class="setdata" style="' + Tools.getTeambuilderSprite(set, this.curTeam.gen) + ';">';
 
 			// icon
 			buf += '<div class="setcol setcol-icon">';
@@ -1143,7 +1150,9 @@
 			}
 			buf += '</button></div></div>';
 
-			buf += '</div></li>';
+			buf += '</div>';
+			if (showNotes) buf += '<div class="setnotes"><label>Set notes</label><textarea class="textbox" rows="1">' + (set.notes ? set.notes.join('\n') : '') + '</textarea></div>';
+			buf += '</li>';
 			return buf;
 		},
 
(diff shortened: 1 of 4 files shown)


asgdf:
Any opinions on this?

Zarel:
I don't consider `|` the right separator here.

Zarel:
I don't think `|` is the right separator here...

Zarel:
Packed teams have a very restricted charset, basically alphanumeric characters, `|`, and `,`. So any other character works as a separator, I think. You should probably have a different character you can convert to/from a linebreak, too. Maybe Tab?

asgdf:
I was mostly worried about limiting the character choice for the notes themselves. Tabs should work well enough, since I assume most browsers will make tab cycle the focus through selectable elements instead of producing a \t anyway, I'll still prevent users from using it so they're not confused if their copy-pasted tabs turn into newlines.
So, changed that, and also brought fastUnpackTeam() up to date which I had forgotten about before.

js/client-teambuilder.js
```diff
@@ -1245,6 +1254,39 @@
 			this.save();
 		},
 
+		setNotesBlur: function (e) {
+			var textarea = $(e.currentTarget);
+			if (!$.trim(textarea.val())) textarea.parent().hide();
+			if (textarea.val().indexOf('|') >= 0 || textarea.val().indexOf(']') >= 0 || textarea.val().indexOf('\t') >= 0) {
+				app.addPopupMessage("Unfortunately, set notes can't contain tabs or the characters | and ], since they're used for storing teams.");
```
Zarel: That appears to be false? It should not be hard to support `|` or `]` in set notes. Tabs, though, should be autoconverted into four spaces.

asgdf: Allowing either | or ] would indeed be possible, however it would make it difficult to ever add more elements to the packed set unless yet another separator would be used for that.
Allowing both | and ] would not be possible, as far as I can tell, because then I can't know where the next set starts.

Zarel: ...oh, this is set comments, not team comments.

Making it team comments would let you use those characters.

Either way, you can support both by using Tab as a delimiter.

i.e. doing `.replace(/\t(.*)\t/g, '')` before passing it to the packed team decoder.

Zarel: ...This reminds me of the main reason we don't support set comments, though: They'd massively increase the size of teams, and make them expire from `localStorage` more often.

Zarel: (Which is why we moved to packed teams in the first place...)

asgdf: Right, the branch is still called team-notes because that's what I initially tried, I switched because set notes were way more convenient to implement regarding exporting/importing.

> doing .replace(/\t(.*)\t/g, '') before passing it to the packed team decoder.

I'm not sure I understand. Do you mean doing that before passing the packed team to the server with `/utm`? That would indeed be a good solution for that case. But the packed team needs to be decoded client-side for the teambuilder as well, and just erasing the set notes is not a good idea there.

asgdf: Wait what, browsers expire their localStorage for reaching the massive size of a couple kilobytes or so? The same size could easily be reached by having many teams, couldn't it?

Zarel: The actual threshold is undocumented, but more data generally makes it expire faster. Set notes could easily 5x the amount of data stored in localStorage, which is very worrying.

asgdf: I have to admit, this is the first I hear of it expiring on its own. The specification says 
>User agents should expire data from the local storage areas only for security reasons or when requested to do so by the user.

Is there some security reason behind this?
Is there some other way to store these notes? I assume not, but I don't want this (very useful) feature to get shelved over this :(

asgdf: I also think a fivefold increase in localStorage size is highly unreasonable to expect. That would require like 2300 characters worth of set notes for every single team. A doubling in size sounds more likely, and that'd only be the case for people annotating every team.
Anyway, why not just make the teambuilder warn to back up teams once a certain size is reached? I'd rather have people be able to choose to risk a possible expiration than deny it to everyone.

Zarel: I can't tell which spec you're reading. The older specs don't seem to mention expiry at all.

But the places you should be looking are straight from the horse's mouth; here's Chrome:

https://developers.google.com/web/updates/2016/06/persistent-storage

> When storage on the local machine is running tight (“under storage pressure”), user agents automatically clear storage to make more available space. [...] the Storage spec defines two different modes for storage for a given domain - “best effort” and “persistent”. The default mode, of course, is “best effort”. Storage for a domain that is “best effort” (aka “not persistent”) can be cleared automatically, without interrupting or asking the user.

And here's Firefox:

https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Browser_storage_limits_and_eviction_criteria

> When the available disk space is filled up, the quota manager will start clearing out data based on a LRU policy — the least recently used origin will be deleted first, then the next one, until the browser is no longer over the limit.

Zarel: Anyway, I did find info on persistent storage, and enabled it in b3b2eea2f21

Perhaps we should only support set notes if persistent storage is on?

asgdf: I had been looking at [W3's spec](https://www.w3.org/TR/webstorage/#the-localstorage-attribute), guess I should have looked at the browsers' implementations instead.
>Perhaps we should only support set notes if persistent storage is on?

It seems that only newer chrome and opera versions support StorageManager.persist() so far, so that'd require waiting until it's more widely adopted I suppose.
Also, I'm not sure if I'm reading this correctly, but my impression is that local storage would only get cleared if the machine is running out of hard drive space. Are <200KB of team data really going to be relevant when it comes to that?

Zarel: No, but the user might be going to other websites that store more data, and data is cleared on an LRU policy, not a "biggest users cleared first" policy. There's a reason most users aren't affected, only a small number of users.

And this is, of course, much more of a problem on mobile.

Hm, I think overall this is probably not going to be a huge problem. At some point, I want to move to server-side database, anyway.

asgdf: If the main problem is _other_ sites storing large amounts of data locally, then the increase in team size from set notes is unlikely to be what pushes it over the storage limit, isn't it? Unless that limit was only a megabyte or two to begin with.
>Hm, I think overall this is probably not going to be a huge problem

Does that mean this PR can proceed?

Zarel: Yes, you're right, which is why I said "Hm, I think overall this is probably not going to be a huge problem."

I'll put it on hold for a moment pending some other work, but yes, this is good overall.

asgdf:
Any updates on this?

asgdf:
Bump

Zarel:
Acknowledged; please hold

asgdf:
Bump²
(hope I'm not being too annoying with my bumps every couple of weeks, I don't know if there's something specific holding it back)

Zarel:
My main issue is that bigger pull requests take longer to code review, and also make me more nervous about merging, so they tend to sit open for quite a while longer.

Especially any pull request that stores data means that it introduces a maintenance burden, which needs to be evaluated very carefully, which is why pull requests that store data are always the ones that take forever... Sorry, no way around that.

In conclusion: Yes, keep bumping; we have not yet worked out a better solution to this problem.

asgdf:
Bump

asgdf:
Bump (anything I can do to make it easier to review?)

Zarel:
Sorry, this will need to be put off a bit logner

Armarx:
bump (ik this isn't my PR but its a really nice feature)

Zarel:
Oh, this needs support for desktop client, which means being able to import/export set notes as well. Which we should probably do, anyway.

The sensible thing to do would be to make it so any paragraph of text would be a set note, but you'd have to detect that a paragraph isn't a pokemon name. That might be hard.

asgdf:
This feature was pretty cool maybe I shouldn't abandon it

>Oh, this needs support for desktop client, which means being able to import/export set notes as well. Which we should probably do, anyway.

It's been a while, but I thought I already supported importing/exporting set notes with this? Is there any desktop-specific code I'd need to touch?

>The sensible thing to do would be to make it so any paragraph of text would be a set note, but you'd have to detect that a paragraph isn't a pokemon name. That might be hard.

That would also prevent people from putting actual importables in their set notes, like a previous version of the set or something. It sounds kind of unnecessarily complex to me with no real upsides.

Zarel:
Am really sorry, but committing this as-is is probably a bad idea. This is a highly-requested feature, but I want to do it properly after thinking through the storage representation, which needs to happen after the client React refactor which will also replace how data storage works entirely.

monsanto:
This seems like a cool idea but the code is out of date and I think the developer has moved on since. I have no intent to merge this as-is or work on this feature but if another developer wanted to make a proposal for this functionality I wouldn't reject it.

asgdf:
I'm still tangentially around, are the local storage size concerns Zarel mentioned back then still relevant?

monsanto:
I'll reopen then, glad to hear you're still around.  [SO](https://stackoverflow.com/a/2989317) suggests a limit in the megabytes per origin. I can't imagine text ever hitting this limit, but maybe a player with a lot of teams could let us know how much local storage they're using.

PartMan7:
I'm currently at 3710520 characters (3.7 MB) in compressed form, but I haven't made teams for like two whole generations so maybe like 5-10x that for other people?

monsanto:
That's intense, I would have expected something in the range of kilobytes

PR #879, opened 2017-03-04
Fix some dark mode visibility issues

asgdf:
This changes this: 
grafik 
to this:
grafik
The color attribute is repeated in the :hover rule because otherwise `.dark .popupmenu button:hover` would overwrite it (with black). Not sure if there's a more elegant way to solve this.

It also changes this:
grafik
to this:
grafik

Lastly, this also changes subtle links to not be black in dark mode by inheriting the regular text color. Zarel mentioned fixing that in the dev room a while ago (and it is indeed fixed on the site), but apparently that change was never actually committed.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -178,7 +178,8 @@ button:disabled {
 }
 */
 .dark .button,
-.dark .tabbar a.button {
+.dark .tabbar a.button,
+.dark .popupmenu button.button {
 	background: #484848;
 	box-shadow: inset 0 3px 4px rgba(255, 255, 255, 0.25);
 	border-color: #A9A9A9;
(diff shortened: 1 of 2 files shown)


style/client.css
```diff
@@ -178,17 +178,20 @@ button:disabled {
 }
 */
 .dark .button,
-.dark .tabbar a.button {
+.dark .tabbar a.button,
+.dark .popupmenu button.button {
```
Zarel: So, why is this necessary?

asgdf: Because otherwise the buttons look like in the top picture, based on the '.button, .popupmenu button.button' rule on line 330.

Zarel: Because `.popupmenu button` has more specificity than `.button`... ugh

Zarel: I'm unhappy with specificity rules :(

asgdf: I guess another class could be added to all the other popupmenu buttons, then the rules for them can be made more specific to not include these two buttons?

asgdf:
Okay, now I'm not adding any rules that are just there for the specifity. In fact, now there are less of them than there were before.

Zarel:
No, I think the old version was better...

asgdf:
Aw, okay, I reverted it back to the old one. What made it worse though?

Zarel:
I don't like the idea of adding a class to every menubutton ever.

asgdf:
Is there a better way to do it than this one, then?

Zarel:
I don't have any better ideas, sadly.

PR #885, opened 2017-03-18
Allow users to filter by gen in /rank

MacChaeger:
This is mostly useful for people wanting to see their general standing but not wanting to see old ladders that they don't play any more.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -749,7 +749,7 @@
 				var formats = {};
 				var gens = {};
 				for (var i = 1, len = targets.length; i < len; i++) {
-					if (targets[i].length === 4 && targets[i].indexOf('gen') === 0) {
+					if (targets[i].length === 4 && targets[i].startsWith('gen')) {
 						gens[targets[i]] = 1;
 					} else {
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -747,8 +747,13 @@
 				var targets = target.split(',');
 				var formatTargeting = false;
 				var formats = {};
+				var gens = {};
 				for (var i = 1, len = targets.length; i < len; i++) {
-					formats[toId(targets[i])] = 1;
+					if (targets[i].length === 4 && targets[i].indexOf('gen') === 0) {
```
Zarel: `.startsWith()`

Zarel: Why do people use search functions to test for things in predictable locations? "Does a string start with X" should be O(1), not O(n).........

MacChaeger: I'll be honest, I missed that when I was looking through javascript string functions.

Zarel: For the record, `string.substr(0, fragment.length) === fragment` is the usual way to test if a string starts with a fragment in basically every programming language ever.

PR #886, opened 2017-03-20
Update /data

Spandamn:
This part of the /data code is no longer needed (https://github.com/Zarel/Pokemon-Showdown/pull/3378)


Zarel:
You realize not all PS servers are up-to-date? This code can be dropped in like, a year.

Spandamn:
Oh yeah. Didn't think about that.

PR #887, opened 2017-03-22
Fix format/gen specific /rank

asgdf:
Also trim targets before parsing so spaces around commas don't throw it off.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -377,7 +377,10 @@
 			switch (cmd.toLowerCase()) {
 			case 'chall':
 			case 'challenge':
-				var targets = target.split(',').map($.trim);
+				var targets = target.split(',');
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -744,7 +744,7 @@
 				if (app.localLadder) return text;
 				if (!target) target = app.user.get('userid');
 
-				var targets = target.split(',');
+				var targets = target.split(',').map($.trim);
```
Zarel: For the record, `map` is an ES5 feature. As is `trim`. We use `$.trim` because client does not have ES5 support. So we can't use `map`, either.

Zarel: Just trim for the final iteration.

asgdf:
Ok I removed the use of `map()`. `/challenge` had also been using it so I replaced it there as well.

PR #889, opened 2017-03-24
Replay: Bug Fix "Open" Button

AlainDuhHero:
Previously the Open Button always failed to Open the Replay

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -2429,9 +2429,10 @@
 		},
 		initialize: function (data) {
 			var buf = '';
+			var replay_url = "http://replay.pokemonshowdown.com/" + data.id;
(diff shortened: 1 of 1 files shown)


urkerab:
It looks as if the `submit` function was intended to do this, but I don't know whether it ever actually had access to the `id` at that point.

Zarel:
Wrapping an `<a>` tag around a `<button>` is not actually supported functionality, and I have worries that this won't work. Screwing with the `submit` function may be a better idea.

Zarel:
The main problem is popup blockers. This solution does fix the popup blocker issue, but it introduces other situations where selecting the button won't open the replay. I'll just remove the "Open" button completely.

PR #891, opened 2017-03-29
Teambuilder: Don't invest in unused stats in BH

MacChaeger:
(or any format without the 510 EV limit)

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -2850,6 +2850,9 @@
 				if (move.category === 'Status') {
 					if (move.id === 'batonpass' || move.id === 'healingwish' || move.id === 'lunardance') {
 						moveCount['Support']++;
+					} else if (move.id === 'metronome' || move.id === 'assist' || move.id === 'copycat' || move.id === 'mefirst') {
+						moveCount['Physical'] += 0.5;
+						moveCount['Special'] += 0.5;
 					} else if (move.id === 'naturepower') {
 						moveCount['Special']++;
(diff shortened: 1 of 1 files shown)


TheImmortal:
This doesn't exactly work because of Copycat and other similar moves

MacChaeger:
Hm, I thought those moves counted as physical attacks, but I'll look at this more closely when I next have a chance.

js/client-teambuilder.js
```diff
@@ -3135,6 +3138,8 @@
 
 			if (this.curTeam && this.ignoreEVLimits) {
 				evs = {hp:252, atk:252, def:252, spa:252, spd:252, spe:252};
+				if (!moveCount['PhysicalAttack']) delete evs.atk;
+				if (!moveCount['SpecialAttack']) delete evs.spa;
```
Zarel: Okay, this one is unnecessary.

MacChaeger: It's not strictly necessary for what I put it there to do (stop the builder from assuming that all bh sets are mixed), but I put it here instead of where the minusStat is determined because Power Split is a thing, if a very rare one.

MacChaeger: It also makes import/export versions of the team a little tidier.

Zarel: Oh, Power Split makes sense.

MacChaeger:
Is there something here that still needs fixing?

js/client-teambuilder.js
Zarel: LOL, you forgot about Gen 1, where this removes the Special stat entirely.

PR #892, opened 2017-03-30
Prevent ghost new message notifications after refreshing

asgdf:
By ghost "new message" notifications, I mean that sometimes, after refreshing, rooms will notify you of new messages, but there aren't any that you hadn't already read. This was caused by the time offset between server and client, which can change by a second every now and then. When it increased by a second on refresh, the latest message in the room will seem to have been sent a second after the latest recorded date, causing the notifcation.
This fixes that by using server timestamps for logtimes.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1455,15 +1455,17 @@
 			var lastMessageDates = Tools.prefs('logtimes') || (Tools.prefs('logtimes', {}), Tools.prefs('logtimes'));
 			if (!lastMessageDates[Config.server.id]) lastMessageDates[Config.server.id] = {};
 			var lastMessageDate = lastMessageDates[Config.server.id][this.id] || 0;
-			var mayNotify = msgTime > lastMessageDate && userid !== app.user.get('userid');
+			// because the time offset to the server can vary slightly, subtract it to not have it affect comparisons between dates
+			var serverMsgTime = msgTime - (this.timeOffset || 0);
+			var mayNotify = serverMsgTime > lastMessageDate && userid !== app.user.get('userid');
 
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -1455,15 +1455,17 @@
 			var lastMessageDates = Tools.prefs('logtimes') || (Tools.prefs('logtimes', {}), Tools.prefs('logtimes'));
 			if (!lastMessageDates[Config.server.id]) lastMessageDates[Config.server.id] = {};
 			var lastMessageDate = lastMessageDates[Config.server.id][this.id] || 0;
-			var mayNotify = msgTime > lastMessageDate && userid !== app.user.get('userid');
+			// because the time offset to the server can vary slightly, subtract it to not have it affect comparisons between dates
+			var serverMsgTime = msgTime - ~~(this.timeOffset);
```
Zarel: ...`~~` isn't conformant to PS coding standards. Is this supposed to be `parseInt` or `Math.floor` or something?

asgdf: It was used twice elsewhere in client-chat (L1171 and 1181), and I figured it was a quick way of making sure that nothing breaks when this.timeOffset is undefined. Can it be assumed that it'll always be defined (i.e. that the server will always send the time offset message first)?

Zarel: Oh, `~~` seems to round closer-to-0, cast to Number, and turn `NaN` into 0.

Sometimes things like this slip through, but my personal preference is to use `(this.timeOffset || 0)` if the worry is that it might be `undefined`.

asgdf:
Right, I had kinda forgotten about `|| 0`, that does look way more readable. Changed.

PR #895, opened 2017-04-07
Add IV Options

submindraikou:
Adds an option to allow Trick Room to change Speed IVs to 0 since it seemed like some people still wanted this in the Removal of Trick Room = Auto 0 Speed IVs thread, and also adds an option to disallow teambuilder changing IVs (except for past gens hidden power) (requested by thimo).

Currently it finds doubles formats by searching `this.curTeam.format` for "doubles" or "vgc" (Surrounding code just searched format for tiers to find level caps)
I also duplicate `AutoIVPopup` and I'm not sure if thats the best way to use a popup (or if it can be somewhere else where client-teambuilder and client-topbar can access it)
The button to access the options popup is just at the bottom of the options and bottom of iv tab in teambuilder (imgur.com/a/OWx2l this is a outdated picture but the buttons are in the same place), can the buttons stay in the same place or should they be moved elsewhere?

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -2072,6 +2072,9 @@
 			});
 			this.suppressSliderCallback = false;
 		},
+		autoivpopup: function () {
+			app.addPopup(window.AutoIVPopup);
+		},
 		setStatFormGuesses: function () {
 			this.updateStatForm(true);
 		},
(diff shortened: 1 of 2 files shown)


Zarel:
These really shouldn't be options...

PR #897, opened 2017-04-11
Remove redundancy

Ghoulean:
Ability is already parsed at line 1043


Zarel:
Thanks!

Do we have the same problem in the server?

PR #898, opened 2017-04-11
Update Mix and Mega COIL

KrisOnline:
http://www.smogon.com/forums/threads/m-m-suspect-3-pheromosa.3600833/


KrisOnline:
@TheImmortal

KrisOnline:
Cool

PR #900, opened 2017-04-18
Add option to open new battle rooms on the right

asgdf:
I think this might be a useful option for people who would like having battles and PMs or the teambuilder open side by side.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -9,6 +9,8 @@
 		initialize: function (data) {
 			this.me = {};
 
+			this.isSideRoom = Tools.prefs('rightpanelbattles');
+
 			this.$el.addClass('ps-room-opaque').html('<div class="battle">Battle is here</div><div class="foehint"></div><div class="battle-log"></div><div class="battle-log-add">Connecting...</div><div class="battle-controls"></div><button class="battle-chat-toggle button" name="showChat"><i class="fa fa-caret-left"></i> Chat</button>');
 
 			this.$battle = this.$el.find('.battle');
@@ -1237,18 +1239,21 @@
 	var BattleOptionsPopup = this.BattleOptionsPopup = Popup.extend({
 		initialize: function (data) {
 			this.battle = data.battle;
+			var rightPanelBattlesPossible = (MainMenuRoom.prototype.bestWidth + BattleRoom.prototype.minWidth < $(window).width());
 			var buf = '<p><strong>In this battle</strong></p>';
 			buf += '<p><label class="optlabel"><input type="checkbox" name="ignorespects"' + (this.battle.ignoreSpects ? ' checked' : '') + '/> Ignore Spectators</label></p>';
 			buf += '<p><label class="optlabel"><input type="checkbox" name="ignoreopp"' + (this.battle.ignoreOpponent ? ' checked' : '') + '/> Ignore Opponent</label></p>';
 			buf += '<p><strong>All battles</strong></p>';
 			buf += '<p><label class="optlabel"><input type="checkbox" name="ignorenicks"' + (Tools.prefs('ignorenicks') ? ' checked' : '') + ' /> Ignore nicknames</label></p>';
+			if (rightPanelBattlesPossible) buf += '<p><label class="optlabel"><input type="checkbox" name="rightpanelbattles"' + (Tools.prefs('rightpanelbattles') ? ' checked' : '') + ' /> Open new battles on the right side</label></p>';
 			buf += '<p><button name="close">Close</button></p>';
 			this.$el.html(buf);
 		},
 		events: {
 			'change input[name=ignorespects]': 'toggleIgnoreSpects',
 			'change input[name=ignorenicks]': 'toggleIgnoreNicks',
-			'change input[name=ignoreopp]': 'toggleIgnoreOpponent'
+			'change input[name=ignoreopp]': 'toggleIgnoreOpponent',
+			'change input[name=rightpanelbattles]': 'toggleRightPanelBattles'
 		},
 		toggleIgnoreSpects: function (e) {
 			this.battle.ignoreSpects = !!e.currentTarget.checked;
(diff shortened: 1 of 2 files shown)


js/client-battle.js
```diff
@@ -9,6 +9,8 @@
 		initialize: function (data) {
 			this.me = {};
 
+			this.isSideRoom = Tools.prefs('righthandbattles');
```
Zarel: `rightpanelbattles`, maybe? "hand" seems like a nonsequitur, even though I know what you're going for.

js/client.js
```diff
@@ -1325,8 +1325,8 @@
 				if (this.curSideRoom === oldRoom) this.curSideRoom = room;
 				if (this.sideRoom === oldRoom) this.sideRoom = room;
 			}
-			if (type === BattleRoom) this.roomList.push(room);
-			if (type === ChatRoom || type === BattlesRoom) this.sideRoomList.push(room);
+			if (type === BattleRoom && !room.isSideRoom) this.roomList.push(room);
+			if (type === ChatRoom || type === BattlesRoom || (type === BattleRoom && room.isSideRoom)) this.sideRoomList.push(room);
```
Zarel: Maybe these should just be `!room.isSideRoom` and `room.isSideRoom`?

js/client-battle.js
```diff
@@ -1242,13 +1244,15 @@
 			buf += '<p><label class="optlabel"><input type="checkbox" name="ignoreopp"' + (this.battle.ignoreOpponent ? ' checked' : '') + '/> Ignore Opponent</label></p>';
 			buf += '<p><strong>All battles</strong></p>';
 			buf += '<p><label class="optlabel"><input type="checkbox" name="ignorenicks"' + (Tools.prefs('ignorenicks') ? ' checked' : '') + ' /> Ignore nicknames</label></p>';
+			buf += '<p><label class="optlabel"><input type="checkbox" name="righthandbattles"' + (Tools.prefs('righthandbattles') ? ' checked' : '') + ' /> Open new battles on the right side</label></p>';
```
Zarel: Possibly only display this if the screen has enough room for two-panel stuff?

asgdf: As far as I can tell, if there's not enough space, it'll display the newly joined room/batte in full width. Because of that, the behavior with and without this option enabled is basically identical in that case, except for the position of the room tab in the top bar. Is it really worth disabling the option for that?

js/client.js
asgdf: I've tried that, but it turns out that the Teambuilder, Ladder, Home and Rooms rooms are appended to the room lists separately, so they need to be excluded here or they'll get duplicate tabs.
I suppose there could be a "room is freely movable" flag for room objects, which includes only those room types that can be added to the room list in this way?

js/client-battle.js
asgdf: I suppose that name would be clearer, changed it to "rightpanelbattles".

js/client.js
Zarel: You could just... not append them separately?

js/client-battle.js
Zarel: Will be slightly less confusing on e.g. mobile, where there aren't two panels.

js/client.js
asgdf: Actually, what I said wasn't quite right in retrospect, what I mean is that those rooms are not added to the room lists at all. The tabs are added to the tab bar separately, since Home/Ladder/Teambuilder have their own section of the tab bar, and the Rooms room is a special case that should always be to the right of all other side room tabs so separate logic is required for that anyway.

js/client-battle.js
asgdf: Another issue: whether there is enough room for a battle on the right depends on what room the left panel is showing at the time of a new battle starting. Which setup should it use to check? Home screen + battle?

js/client.js
Zarel: Oh. You could check those conditions separately?

js/client-battle.js
Zarel: Home screen + battle is a good one.

js/client.js
asgdf: I'm a bit confused about what you mean with that. Do you mean moving the logic for the "home/teambuilder/ladder are separate" and "Rooms is always the last side room" to this function?

asgdf:
So this is how it now looks on mobile, for example. Figured I should add a note for why it's disabled. Disabled checkboxes are kinda hard to differentiate from regular ones in dark mode (in Firefox at least), but I suppose the note helps with that as well.
grafik

js/client.js
Zarel: I mean, test for teambuilder, ladder, home, and rooms, and don't push to roomlists at all.

Otherwise push depending on `isSideRoom`.

Zarel:
I didn't mean to disable it, no point in disabling it at all... I meant to hide it entirely...

asgdf:
Wouldn't people find it confusing if options just disappeared for (to them) no discernible reason?

Zarel:
I think it's much more confusing to have an option that doesn't do what it says it does. Not having the option makes perfect sense.

js/client.js
```diff
@@ -1325,8 +1325,13 @@
 				if (this.curSideRoom === oldRoom) this.curSideRoom = room;
 				if (this.sideRoom === oldRoom) this.sideRoom = room;
 			}
-			if (type === BattleRoom) this.roomList.push(room);
-			if (type === ChatRoom || type === BattlesRoom) this.sideRoomList.push(room);
+			if ([MainMenuRoom, TeambuilderRoom, LadderRoom, RoomsRoom].indexOf(type) < 0) {
```
Zarel: Might want to do this by roomid rather than type?

asgdf: Right, sorry, I just adopted that from the previous code without thinking about it. Changed.

asgdf:
I meant I preferred disabling to hiding in this situation. Anyway, it's being hidden now.

PR #901, opened 2017-04-20
Fix linking undefined when not linking anything

panpawn:
Currently, if you do [[]] on PS, it will link as if you had done [[undefined]].

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -622,6 +622,9 @@ var Tools = {
 			});
 			// [[blah]]
 			str = str.replace(/\[\[(?![< ])(?:(?:(youtube|yt|wiki)\: ?)?([^<`]*?[^< ])?)\]\]/g, function (match, p1, p2) {
+				if (match === '[[]]') {
+					return (p1 ? p1 : '') + match + (p2 ? p2 : '');
+				}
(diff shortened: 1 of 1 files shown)


bumbadadabum:
Yeah my bad was going to get to this eventually but never did.

Zarel:
We'll be rewriting this soon, but this might as well be a stopgap.

js/battledata.js
```diff
@@ -622,6 +622,9 @@ var Tools = {
 			});
 			// [[blah]]
 			str = str.replace(/\[\[(?![< ])(?:(?:(youtube|yt|wiki)\: ?)?([^<`]*?[^< ])?)\]\]/g, function (match, p1, p2) {
+				if (match === '[[]]') {
+					return (p1 ? p1 : '') + match + (p2 ? p2 : '');
```
Zarel: Oh, btw, this isn't standard PS notation.

We use `(p1 || '')` instead of `(p1 ? p1 : '')`

Zarel: (I didn't really review before merging because this function is being replaced soon)

panpawn: Yeah, I seem to always forget this convention, my bad

PR #902, opened 2017-04-22
Fix dark mode style regressions

asgdf:
1. subtly notifying roomtab title turned white on mouseover (I accidently broke this in #879)
2. bold text in modchat/modjoin boxes [turned blue](https://cloud.githubusercontent.com/assets/5814184/25301690/fa23b102-272d-11e7-8414-37a9a3ea36c9.png) in the battle chat since they now use ``<strong>`` as well

### DIFF
--- a/style/battle.css
+++ b/style/battle.css
@@ -828,6 +828,13 @@ License: GPLv2
 	/* Readable link colour */
 	color: #DDEEFF;
 }
+.battle-log .inner .chat .broadcast-green strong,
+.battle-log .inner .chat .broadcast-blue strong,
+.battle-log .inner .chat .broadcast-red strong,
+.battle-log .inner .chat .message-announce strong {
+	/* readable bold text in broadcast boxes in battles */
+	color: white;
+}
 .broadcast-green {
 	background-color: #559955;
 	color: white;
--- a/style/client.css
+++ b/style/client.css
@@ -187,7 +187,10 @@ button:disabled {
 	color: #F9F9F9;
 }
 .dark .button:hover,
-.dark .tabbar a.button:hover,
+.dark .tabbar a.button:hover {
+	background: #606060;
+	border-color: #EEEEEE;
+}
 .dark .popupmenu button.button:hover {
 	background: #606060;
 	border-color: #EEEEEE;


style/battle.css
```diff
@@ -630,7 +630,7 @@ License: GPLv2
 	background: #333333;
 	color: #CCCCCC;
 }
-.dark .battle-log .chat strong,
+.dark .battle-log .chat > strong,
```
Zarel: Why is `>` necessary?

Zarel: PS style guide is not to use `>`.

asgdf: Well that explains why it wasn't used anywhere else, but it doesn't seem to be written down anywhere.
Anyway, I used it because it needs to differentiate between ``<strong>`` that's part of the battle ("abc's team:" and updated elo score - should be blue) and ``<strong>`` that's part of a broadcast box, like for modchat (should be white). As far as I can tell, the options are:
1. use really really specific rules to override it (.battle-log .inner .chat .broadcast-red strong, once for every broadcast class)
2. use ``:not`` to exclude broadcast boxes - I couldn't get this to work (and based on its absence in the rest of the code it's probably against the style guidelines anyway)
3. use ``>`` to enforce that only ``<strong>`` tags that are direct children of ``.chat`` become blue, which is only true for the team/elo but not for broadcast boxes
Out of those, 3 seemed the least ugly to me. I suppose it'll have to be option 1?

Zarel: Oh, you definitely don't want to use `>` for that, that's incredibly brittle. (Brittleness is the main reason I don't like `>`)

Zarel: The really specific rules are my usual way? But you can also achieve this by adding CSS classes.

Zarel: How does light mode handle this problem? You should just copy that solution.

Zarel: The CSS order rules are annoying but once you learn them they're not too hard to work around. It's just specificity, ties broken by bottom-first.

asgdf: >How does light mode handle this problem?

It doesn't. It has the same problem, except the blue is slightly lighter.
>The really specific rules are my usual way?  

Well the last I did that you said [you're unhappy with them](https://github.com/Zarel/Pokemon-Showdown-Client/pull/879#discussion_r104277198) so that's why I tried something new this time.
Anyway, did this.

Zarel: I meant that I'm unhappy with CSS not making a good solution possible to this problem, not with your CSS which makes the best of a bad situation.

asgdf:
This should also be good to go now

Zarel:
Yeah, I'm going to personally investigate better solutions to that mess.

Zarel:
Will merge while investigating better solutions

Zarel:
Also I thought I vetoed using `<strong>` in the first place. :|

Issue #906, opened 2017-04-23
Upgrade Backbone/lodash to latest version


FelixRilling:
Backbone.js and lodash should be updated to the latest version to avoid bugs.
There are no breaking changes in those updates.

Zarel:
We don't need _both_ a pullreq and an issue for this... Just the pullreq is fine.

PR #907, opened 2017-04-23
Updated Backbone and lodash #906

FelixRilling:
Updated backbone and lodash to the latest versions.
No breaking changes; No Incompabilities found.
Tested trough the testclient.

See #906


Zarel:
Why rename Backbone?

FelixRilling:
The rename is to avoid browsers using the old cached version.

Zarel:
There's already a `?` there. Our build scripts will automatically cachebust.

Like, notice how if you view source on play.pokemonshowdown.com:

`<script src="//play.pokemonshowdown.com/js/lib/backbone.js?1adb34b9"></script>`

That hash will automatically update if you update the file.

FelixRilling:
Oh, I did not notice that, there is no point for the namechane then.
Would you like me to rename the file back to the initial name then?

Zarel:
Yeah, thanks.

FelixRilling:
Alright, reverted the name.
Sorry for the inconvenience!

Zarel:
Okay, thanks! Can you explain what changed with `lodash`? I noticed you changed `.compat` to `.custom`? That's fine, but I'd like more details on exactly what's in the custom build.

FelixRilling:
Yes, no problem.
Backbone 1.3.x requires lodash 4.x.
lodash comes in [different builds](https://github.com/lodash/lodash/wiki/build-differences). the used build is `lodash-core`, which is the required build for backbone.
when compiling the custom core build, the output filename is `lodash.custom.js`

Zarel:
It looks like the usual name for that is `lodash.core.js` (see https://lodash.com/ ). Does this build differ from the core build in any way? Can you explain what you mean by "custom core build"?

Zarel:
I would prefer to use the official build and name it `lodash.core.js` unless there's a reason to use a custom-compiled core build?

FelixRilling:
It is the official build, the same file as `lodash.core.js`, only created by the lodash-cli tool instead of downloading the version on their website.
The contents are the same.

Zarel:
Okay, I'd prefer for it to be named `lodash.core.js` for consistency reasons.

FelixRilling:
Alright, done.

Zarel:
Yay! Thanks a lot!

PR #908, opened 2017-04-25
Show the effects of all abilities on base power

MacChaeger:
Except the ones dependent on things like turn order. Also not Battery for now.

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -1107,7 +1107,7 @@ var BattleTooltips = (function () {
 		if (!basePower) return basePowerComment;
 
 		// Other ability boosts.
-		let abilityBoost = 0;
+		var abilityBoosted = 0;
 		if (ability === 'Water Bubble' && move.type === 'Water') {
 			abilityBoost = 2;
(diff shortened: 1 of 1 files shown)


MacChaeger:
Let -> var?

Zarel:
This is the client; `let` is not supported by IE9 or any of a number of other common browsers.

js/client-battle-tooltips.js
```diff
@@ -1107,28 +1107,43 @@ var BattleTooltips = (function () {
 		if (!basePower) return basePowerComment;
 
 		// Other ability boosts.
+		var abilityBoosted = 0;
```
Zarel: You mean `abilityBoost`?

Zarel:
In addition to the build error; do you not have access to `npm test`?

MacChaeger:
I had forgotten to have Travis CI automatically test for my client repository as well; that's fixed now.

MacChaeger:
Oh damn i just realized this is wrong

MacChaeger:
Ok I think it's good now unless you _really_ want Battery supported.

Zarel:
Looks good, thanks!

PR #909, opened 2017-04-26
Expand tabcompletion functionality

asgdf:
* skip own username when cycling through candidates
* allow cancelling tabcompletion with Esc
* allow cycling through candidates in reverse in PMs as well
* allow tabcompletion everywhere in a multi-line message

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -134,16 +134,10 @@
 			} else if (e.keyCode === 34) { // Pg Dn key
 				this.$chatFrame.scrollTop(this.$chatFrame.scrollTop() + this.$chatFrame.height() - 60);
 			} else if (e.keyCode === 9 && !e.ctrlKey) { // Tab key
-				if (!e.shiftKey) {
-					if (this.handleTabComplete(this.$chatbox, false)) {
-						e.preventDefault();
-						e.stopPropagation();
-					}
-				} else { // Shift + Tab
-					if (this.handleTabComplete(this.$chatbox, true)) {
-						e.preventDefault();
-						e.stopPropagation();
-					}
+				var reverse = !!e.shiftKey; // Shift+Tab reverses direction
+				if (this.handleTabComplete(this.$chatbox, reverse)) {
+					e.preventDefault();
+					e.stopPropagation();
 				}
 			} else if (e.keyCode === 38 && !e.shiftKey && !e.altKey) { // Up key
 				if (this.chatHistoryUp(this.$chatbox, e)) {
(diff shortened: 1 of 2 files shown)


panpawn:
... but sometimes I tab complete myself on purpose...

asgdf:
In what situations are you tabbing your own name? When testing commands on yourself on a local server?

Zarel:
I also tab-complete myself on purpose

asgdf:
But why? If it's about testing on a local server, what about allowing it only when you're the only user in the room? Or is it worth having a toggle-able option for it?
It's just that for the vast majority of users, it gets in the way more than it helps.

AustinXII:
@asgdf maybe allow staff to still auto complete their name? It's useful for /modlog and for global leaders + to use commands on themselves such as room owner

Zarel:
also sometimes I like to talk about myself and am lazy <_<

asgdf:
What about allowing tabbing your own name when the message starts with ``/`` (i.e. when using a command)?
Zarel, like talking about yourself in third person? Wouldn't ``/me`` cover that?

AustinXII:
Typing /me every time he wants to talk about himself is much more annoying than tab completing

Zarel:
sometimes I tell people that they can see if I'm online with `/user Zarel` in an attempt to get them to stop PMing me gibberish

asgdf:
Hmm, it'd probably be difficult to programmatically detect "telling someone else to use a command on yourself" as a situation where tabcompleting your own name would be acceptable.
So what do we do with this? What do the priorities look like? In which situations is self-tabcompletion an acceptable loss, if at all?

Zarel:
I'd say just always allow self-tabcompletion.

asgdf:
Aw. Well, okay, here you go.

js/client-chat.js
```diff
@@ -134,16 +134,10 @@
...
-					if (this.handleTabComplete(this.$chatbox, false)) {
-						e.preventDefault();
-						e.stopPropagation();
-					}
-				} else { // Shift + Tab
-					if (this.handleTabComplete(this.$chatbox, true)) {
-						e.preventDefault();
-						e.stopPropagation();
-					}
+				var reverse = e.shiftKey; // Shift+Tab reverses direction
```
Zarel: Would cast to boolean with `!!`

asgdf: If you say so, I suppose it can't hurt. I originally didn't because I figured it's not going to be anything other than a bool in the foreseeable future, and even if it were, nothing would break.
But yeah, changed.

Zarel:
Cool thanks!

PR #910, opened 2017-05-02
Add GUI support for requesting timer to stay switched on

asgdf:
The motivation for this is that it's possible to timerstall indefinitely by repeatedly switching the timer off and on in quick succession so the opponent doesn't get the chance to turn it on themself. This was somewhat alleviated by allowing multiple users to request the timer to be on, but most people probably aren't aware that they can use ``/timer on`` to do that, so there should be a button for it.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1312,7 +1312,8 @@
 		initialize: function (data) {
 			this.room = data.room;
 			if (this.room.battle.kickingInactive) {
-				this.$el.html('<p><button name="timerOff"><strong>Stop timer</strong></button></p>');
+				this.$el.html('<p><button name="timerOff"><strong>Stop timer</strong></button></p>' +
+					'<p><button name="timerOn"><strong>Keep timer switched on</strong></button></p>');
 			} else {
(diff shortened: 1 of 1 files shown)


Zarel:
The easier way is to prevent turning the timer on or off more than once every 10 seconds.

Zarel:
The phrasing on that button is incredibly awkward/unclear, anyway..

asgdf:
That was what I had planned at first, actually, but then you did the timer overhaul and allowed multiple users to request the timer at once, so I figured why not use that functionality?
Wouldn't this be a more direct and less intrusive solution to the problem, as opposed to using an arbitrary cooldown that doesn't fully prevent the issue but merely gives users a time window to combat it? What are the advantages of the cooldown-based approach?
I do agree that the phrasing on the second button is kinda eh, but I wanted to keep it short. Not sure how else to phrase it.

Zarel:
The cooldown-based approach prevents spamming the battle window.

asgdf:
Eh, fair point. Guess I'll close this then.

PR #913, opened 2017-05-07
Chat: suppress metadata messages

sparkyneko:
In tandem with https://github.com/Zarel/Pokemon-Showdown/pull/3508


monsanto:
Thanks for the PR but I don't really understand what this is all about and it doesn't seem like there was enough interest at the time to merge.

PR #915, opened 2017-05-08
Show team size in generations without team preview

QuiteQuiet:
Completes https://github.com/Zarel/Pokemon-Showdown/pull/3512

Also backwards compatible for servers where this isn't sent.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -6365,6 +6365,10 @@ var Battle = (function () {
 			this.getSide(args[1]).setName(args[2]);
 			this.getSide(args[1]).setSprite(args[3]);
 			break;
+		case 'teamsize':
+			this.getSide(args[1]).totalPokemon = parseInt(args[2], 10);
+			this.getSide(args[1]).updateSidebar();
+			break;
 		case 'win':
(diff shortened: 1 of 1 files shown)


Zarel:
I'm not sure if this is the best place to support this. It seems entirely different from how Team Preview is sent, which is four `|pokemon|` messages.

Zarel:
In particular, arguments go in order from most to least important, and team size is too important to go after avatar.

Zarel:
I might think a new `|teamsize|` message might be best, but I'm not sure. I'd like to hear others' opinions on this.

QuiteQuiet:
>In particular, arguments go in order from most to least important, and team size is too important to go after avatar.

This sounds like a reasonable enough reason to change it to a message of it's own to be honest. Since I can't think of anywhere else it makes sense to send this information, and sending empty `|pokemon|playerid` messages to notify teamsize seems impractical at best.

QuiteQuiet:
Would `|teamsize|playerid|size` work?

Zarel:
Sounds good.

QuiteQuiet:
Updated both

PR #917, opened 2017-05-10
Dark Mode: Improve credits readability

TheFenderStory:


panpawn:
Screen cap of before/after?

Zarel:
Seems unnecessary when Credits is only an easter-egg on the teambuilder...

TheFenderStory:
I mean your call. I made the pr after clicking into it by chance and not being able to read it which seemed weird. 
image Here's a screen cap if you'd like one

Zarel:
Might be better if we just replaced the Credits button in the teambuilder.

TheFenderStory:
If you let me know what you want done i can take care of that for you. Otherwise feel free to close the pull request.

Zarel:
I mean, the dialog section of the teambuilder is in my voice, so it's really not like anyone else can edit it.

Zarel:
I fixed this by removing the Credits entirely in be0ca58192ef

PR #918, opened 2017-05-13
Client: support custom ranks

sparkychildcharlie:
- add a new ``|authconfig|`` message which will merge the new data with the default groups provided in ``groupDetails``
In tandem with https://github.com/Zarel/Pokemon-Showdown/pull/3529

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1624,7 +1624,7 @@
 			text += '<li' + (this.room.userForm === userid ? ' class="cur"' : '') + ' id="' + this.room.id + '-userlist-user-' + Tools.escapeHTML(userid) + '">';
 			text += '<button class="userbutton username" data-name="' + Tools.escapeHTML(name) + '">';
 			var group = name.charAt(0);
-			var details = GroupDetails[group] || {rank: 0};
+			var details = window.GroupDetails[group] || {rank: 0};
 			text += '<em class="group' + (details.group === 2 ? ' staffgroup' : '') + '">' + Tools.escapeHTML(group) + '</em>';
 			if (details.group === 2) {
 				text += '<strong><em style="' + hashColor(userid) + '">' + Tools.escapeHTML(name.substr(1)) + '</em></strong>';
@@ -1654,8 +1654,8 @@
 		},
 		comparator: function (a, b) {
 			if (a === b) return 0;
-			var aRank = (GroupDetails[this.room.users[a] ? this.room.users[a].substr(0, 1) : app.defaultGroup || ' '] || {order: 6}).order;
-			var bRank = (GroupDetails[this.room.users[b] ? this.room.users[b].substr(0, 1) : app.defaultGroup || ' '] || {order: 6}).order;
+			var aRank = (window.GroupDetails[this.room.users[a] ? this.room.users[a].substr(0, 1) : app.defaultGroup || ' '] || {order: 6}).order;
+			var bRank = (window.GroupDetails[this.room.users[b] ? this.room.users[b].substr(0, 1) : app.defaultGroup || ' '] || {order: 6}).order;
 			if (aRank !== bRank) return aRank - bRank;
 			return (a > b ? 1 : -1);
 		},
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1656,11 +1656,11 @@
 			if (a === b) return 0;
 			var aRank = (
 				Config.groups[(this.room.users[a] ? this.room.users[a].charAt(0) : Config.defaultGroup || ' ')] ||
-				{order: (Config.defaultOrder || 5.5)}
+				{order: (Config.defaultOrder || 10005.5)}
 			).order;
 			var bRank = (
 				Config.groups[(this.room.users[b] ? this.room.users[b].charAt(0) : Config.defaultGroup || ' ')] ||
-				{order: (Config.defaultOrder || 5.5)}
+				{order: (Config.defaultOrder || 10005.5)}
 			).order;
 
 			if (aRank !== bRank) return aRank - bRank;
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1624,11 +1624,11 @@
 			text += '<li' + (this.room.userForm === userid ? ' class="cur"' : '') + ' id="' + this.room.id + '-userlist-user-' + Tools.escapeHTML(userid) + '">';
 			text += '<button class="userbutton username" data-name="' + Tools.escapeHTML(name) + '">';
 			var group = name.charAt(0);
-			var details = window.GroupDetails[group] || {rank: 0};
+			var details = Config.groups[group] || {type: 0};
 			text += '<em class="group' + (details.group === 2 ? ' staffgroup' : '') + '">' + Tools.escapeHTML(group) + '</em>';
-			if (details.group === 2) {
+			if (details.type === 2) {
 				text += '<strong><em style="' + hashColor(userid) + '">' + Tools.escapeHTML(name.substr(1)) + '</em></strong>';
-			} else if (details.group === 1) {
+			} else if (details.type === 1) {
 				text += '<strong style="' + hashColor(userid) + '">' + Tools.escapeHTML(name.substr(1)) + '</strong>';
 			} else {
 				text += '<span style="' + hashColor(userid) + '">' + Tools.escapeHTML(name.substr(1)) + '</span>';
@@ -1654,8 +1654,8 @@
 		},
 		comparator: function (a, b) {
 			if (a === b) return 0;
-			var aRank = (window.GroupDetails[this.room.users[a] ? this.room.users[a].substr(0, 1) : app.defaultGroup || ' '] || {order: 6}).order;
-			var bRank = (window.GroupDetails[this.room.users[b] ? this.room.users[b].substr(0, 1) : app.defaultGroup || ' '] || {order: 6}).order;
+			var aRank = (Config.groups[this.room.users[a] ? this.room.users[a].substr(0, 1) : Config.defaultGroup || ' '] || {order: 6}).order;
+			var bRank = (Config.groups[this.room.users[b] ? this.room.users[b].substr(0, 1) : Config.defaultGroup || ' '] || {order: 6}).order;
 			if (aRank !== bRank) return aRank - bRank;
 			return (a > b ? 1 : -1);
 		},
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -369,7 +369,7 @@
 			var cmd = '';
 			var target = '';
 			var noSpace = false;
-			if (text.substr(0, 2) !== '//' && text.substr(0, 1) === '/') {
+			if (text.substr(0, 2) !== '//' && text.charAt(0) === '/') {
 				var spaceIndex = text.indexOf(' ');
 				if (spaceIndex > 0) {
 					cmd = text.substr(1, spaceIndex - 1);
@@ -1155,7 +1155,7 @@
 		addRow: function (line) {
 			var name, name2, room, action, silent, oldid;
 			if (line && typeof line === 'string') {
-				if (line.substr(0, 1) !== '|') line = '||' + line;
+				if (line.charAt(0) !== '|') line = '||' + line;
 				var row = line.substr(1).split('|');
 				switch (row[0]) {
 				case 'init':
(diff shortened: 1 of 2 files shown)


sparkychildcharlie:
foxie operator

js/client.js
```diff
@@ -869,6 +869,24 @@
 			}
 
 			switch (parts[0]) {
+			case 'authconfig':
```
Zarel: Is `authconfig` the right name for this?

Maybe just `groups` or something. We still officially call them groups, don't we?

`customauth`?

js/client.js
```diff
@@ -869,6 +869,24 @@
...
+
+				var tarRow = data.slice(12, nlIndex).split(/(?:\b|\B)\|(?:\B)/);
+				for (var i = 0; i < tarRow.length; i++) {
+					var entry = tarRow[i];
+					if (!entry) continue;
+
+					var symbol = entry.charAt(0);
+					var rank = entry.slice(1);
+
+					groupDetails[symbol] = rank + ' (' + symbol + ')'; // override the default symbol.
```
Zarel: We should probably do `delete groupDetails[symbol];` first, to support reordering hardcoded ranks.

Zarel: You need `escapeHTML` for this.

Zarel:
This is a really great idea! Thanks!

js/client.js
sparkychildcharlie: The re-ordering happens in client-chat.js in another place.  I'll add more support for reordering and redo parts of this sometime tomorrow.

Zarel: Why? `delete...` is the easier way by far...

Zarel: Oh, I guess `groupDetails` isn't supposed to be an order. Um, the order should also be done in the same place.

sparkychildcharlie:
rank image

sparkychildcharlie:
I've been using ``window.GroupDetails`` so far to keep it global across the different files, and I dont know how "correct" doing this is.  Do you have a suggestion on how to keep it global and unchanged (somehow it reverts to the default sometimes) and still have it accessible across all files if this isnt the right way of doing it?

js/client.js
```diff
@@ -2217,19 +2272,71 @@
...
-		'@': "Moderator (@)",
-		'%': "Driver (%)",
-		'*': "Bot (*)",
-		'\u2606': "Player (\u2606)",
-		'\u2605': "Player (\u2605)",
-		'+': "Voice (+)",
-		'‽': "<span style='color:#777777'>Locked (‽)</span>",
-		'✖': "<span style='color:#777777'>Namelocked (✖)</span>",
-		'!': "<span style='color:#777777'>Muted (!)</span>"
+	var GroupDetails = window.GroupDetails = window.GroupDetails || {
```
Zarel: This should not be a global. `Config.groups`, maybe.

js/client.js
```diff
@@ -2217,19 +2272,71 @@
...
+		'~': {
+			name: "Administrator (~)",
+			group: 2,
```
Zarel: `rank` or `type`, I'd say.

js/client.js
sparkychildcharlie: OMG thank you so much - I forgot completely that there was a global called Config

panpawn:
This is a nice idea :thumb:

js/client.js
```diff
@@ -1011,6 +1021,72 @@
...
 			}
 		},
+		parseGroups: function (groupsList) {
+			var data = null;
+			try {
+				data = JSON.parse(groupsList);
+			} catch (e) {}
+			if (!data) return; // broken JSON - keep default ranks
+
+			var authTiers = [[], [], []]; // we will be splitting them into three tiers, before compiling them back into the main tree
```
Zarel: Why?

js/client.js
```diff
@@ -1011,6 +1021,72 @@
...
+					symbol: '!',
+					data: {
+						name: "<span style='color:#777777'>Muted (!)</span>",
+						type: 0,
+					},
+				},
+				{
+					symbol: '✖',
+					data: {
+						name: "<span style='color:#777777'>Namelocked (✖)</span>",
```
Zarel: Do we still need to support this? Namelock should use `‽` now.

AustinXII: I kind of like having name locks and regular locks having different symbols

panpawn: I mean, it's easy enough to distinguish between them based on if their name starts with "Guest..." as it is imo

js/client.js
sparkychildcharlie:
```
|groups|[{"symbol":">","name":"Meme","type":2},{"symbol":"~","name":"Administrator","type":2},{"symbol":"&","name":"Leader","type":2},{"symbol":"#","name":"Room Owner","type":2},{"symbol":"☆","name":"Player","type":0},{"symbol":"*","name":"Bot","type":0},{"symbol":"@","name":"Moderator","type":1},{"symbol":"%","name":"Driver","type":1},{"symbol":"|","name":"Operator","type":0},{"symbol":"+","name":"Voice","type":0},{"symbol":" ","name":null,"type":0}]
```
The server sends out this order in terms of the ranks (this is the hierarchy of jurisdiction for PS from greatest to least) along with their properties (2: upper staff - root and/or declare but not addhtml, 1: staff - mute but not addhtml, 0: regs, players, bots - the rest)

However the server-sent order shows bot higher than moderator in terms of jurisdiction order which in the userlist is obviously wrong.  By sorting the three auth types (0, 1, 2) within itself and compiling them in the order 2, 1 and 0, the order of ranks would compile itself properly that way, as bot is labelled as a type 0 (not bold or italics)

@Zarel

js/client.js
sparkychildcharlie: Yeah i checked the server code and they don't even use ``✖`` anymore but ``‽`` like Zarel said, so i'm removing this extra group

Zarel: Also, maybe these three should also be sent by the server. Type `-1` or something.

Zarel: Also maybe the types should be strings for clarity. `leadership`, `staff`, `normal`?

sparkychildcharlie: the server could hardcode the symbols for muted and locked and that could cause issues.  Should i make an additional config field to declare those symbols?

sparkychildcharlie: Or would leaving them solely to the client like this be safer

Spandamn:
This is so cool.

js/client.js
Zarel: I mean, you can just add them to the existing config ranks server-side.

Zarel:
In hindsight, we should probably keep the namelock symbol. No harm in it, and some people might be on old versions that still use it.

js/client.js
Zarel: That should probably be handled server-side? The server should send them in the correct order.

js/client.js
sparkychildcharlie: adding them to the ranks serverside would make the ranks promoteable

sparkychildcharlie:
Anyways, I've opted to keep namelock on the default groups object for any servers that still have that old symbol.

I think this should cover everything that has been mentioned - sorry if i missed anything

js/client-chat.js
```diff
@@ -1683,8 +1654,8 @@
 		},
 		comparator: function (a, b) {
 			if (a === b) return 0;
-			var aRank = (this.rankOrder[this.room.users[a] ? this.room.users[a].substr(0, 1) : ' '] || 6);
-			var bRank = (this.rankOrder[this.room.users[b] ? this.room.users[b].substr(0, 1) : ' '] || 6);
+			var aRank = (Config.groups[this.room.users[a] ? this.room.users[a].substr(0, 1) : Config.defaultGroup || ' '] || {order: (Config.defaultOrder || 6)}).order;
```
Zarel: `.substr(0, 1)` should be `.charAt(0)`

(This code was originally from when I kept on forgetting `.charAt` existed)

Zarel: This is also complex enough that it might need to be broken down into multiple lines.

js/client.js
```diff
@@ -1011,6 +1021,33 @@
...
+
+				var symbol = entry.symbol || ' ';
+				var groupName = entry.name;
+				var groupType = entry.type || 'user';
+
+				if (groupType === 'user' && !Config.defaultOrder) Config.defaultOrder = i; // this is where any undeclared groups will be positioned in userlist
+				if (!groupName) Config.defaultGroup = symbol;
+
+				Config.groups[symbol] = {
+					name: groupName ? Tools.escapeHTML(groupName + ' (' + symbol + ')') : null,
```
Zarel: This can crash if `groupName` is `{"toString": "this is not a function"}`

In general if you do JSON you have to be prepared for a lot of b.s. like that...

Zarel: Also to be fair, for the most part crashes aren't a huge problem.

Zarel: Yeah, I think it's safe not to worry about that. It's unexploitable and isn't worth the additional code complexity in dealing with hostile JSON.

js/client.js
```diff
@@ -2199,6 +2236,73 @@
...
+			order: 9,
+		},
+		' ': {
+			type: 0,
+			order: 10,
+		},
+		'!': {
+			name: "<span style='color:#777777'>Muted (!)</span>",
+			type: 0,
+			order: 11,
```
Zarel: Maybe set these default orders for muted/namelocked/locked to, like, 10001, 10002, 10003.

So then they'll stay at the end if they're not overridden.

sparkychildcharlie: in parseGroups, i make it set ``Config.groups`` to ``{}`` first so the order won't cause any issues.  It will rewrite the punishgroups to use the neccesary order as it is getting parsed

js/client.js
sparkychildcharlie: Maybe i should store the data in a seperate object until the entire parsing is done, so any crashes wouldn't mess up the the entire client; it will just use the default ranks if the custom groups aren't completely loaded

sparkychildcharlie:
*playfully bumps* @Zarel 
Awaiting further comments for improvements or the time to make this live :3

js/client.js
Zarel: Why? I think my solution is better.

sparkychildcharlie: That way it removes any potential groups and allows side servers to use them as "custom symbols" instead.  But honestly it wouldn't matter anyways

Zarel: Side servers can still use them as custom symbols by overriding them in config.js.

Zarel:
It's looking pretty good. I think it might be nice to keep all the default groups at 10000+, so they go below custom groups?

AustinXII:
@Zarel I don't like the idea of having all custom group rankings go above the default groups. I've seen 90% of side servers use ranks that go below some default groups and above others, mainly a group in between voice and driver.

js/client.js
sparkychildcharlie: *fat* Zarel said:
```
It's looking pretty good. I think it might be nice to keep all the default groups at 10000+, so they go below custom groups?
```

Oh when i meant custom symbols, I meant something like something user selected that can be bought from the server's shop using serverside bucks (obtained from levelling up, server games, and NOT purchased).  So users could select to use custom symbols like ``>`` if they so choose to identify as an ~~attack helicopter~~ meme.

I'm pretty certain that there are a lot of side servers that have a shop that sell this option for users, so putting them at the very top of the userlist would be quite an inconvenience.

ATM, i've put those symbols at the right between the transition of ☆ and % (0.5 more than % and 0.5 less than ☆, so they will always be sorted amongst themselves without interfering with the other groups)

Zarel:
@AustinXII, I don't think you understand, custom groups include default groups, unless the server removes a default group for whatever reason. The default groups are only there as a fallback.

AustinXII:
@Zarel oh.. yeah I was confused, that makes sense

sparkychildcharlie:
@Zarel I hope i didn't misunderstand you in this last change

js/client.js
```diff
@@ -1011,6 +1021,35 @@
...
+			// process the data and sort into the three auth tiers, 0, 1, and 2
+			for (var i = 0; i < data.length; i++) {
+				var entry = data[i];
+				if (!entry) continue;
+
+				var symbol = entry.symbol || ' ';
+				var groupName = entry.name;
+				var groupType = entry.type || 'user';
+
+				if (groupType === 'user' && !Config.defaultOrder) Config.defaultOrder = i + 0.5; // this is where any undeclared groups will be positioned in userlist
```
Zarel: Maybe we should just set it to `++i`.

Zarel:
Looks mostly ready. I won't say I _like_ JSON here but I guess it's acceptable.

js/client.js
```diff
@@ -869,6 +869,16 @@
 			}
 
 			switch (parts[0]) {
+			case 'groups':
```
Zarel: Wait, is `groups` the name we decided on?

I guess maybe `customgroups`

sparkychildcharlie: yeah that was one of your suggstions from earlier in this PR

sparkychildcharlie: I'll change it

Zarel:
Looks good!

Issue #248, opened 2014-09-16
CSS3 rules are deleted


Kingfisher7:
Posting in the right repository this time I hope. 
Well, I've never been able to use things like linear-gradient(), but other CSS3 things like border-radius and box-shadow are just removed too.
These are things added through custom.css. I think it would add a good amount of design possibilities if we could use these kinds of things.

Dubstepper:
Any idea when this will be fixed?

Zarel:
Probably never, unless you do it yourself. Feel free to do it yourself, though; PS is open-source.

kotarou3:
Note for people potentially wanting to fix this for us, fixing this will involve two parts:
1. Change the client to pull the CSS directly from `server/custom.css` and sanitise it, before loading it onto the page
2. Updating `Tools.sanitizeHtml()` to not filter out all CSS. Best done by switching out the current sanitiser completely with caja

Zarel:
I thought we _were_ using Caja?

kotarou3:
Definitely not, for `custom.css`.

For `Tools.sanitizeHtml()`, though, it does seem we're using caja, but we're not allowing css through for some reason, so that should be fixed

Zarel:
(I don't think a lack of support for something can be called a bug)

Slayer95:
More like incomplete support, but let's not bikeshed the labeling.

Zarel:
Our support for CSS2 is mostly complete.

Zarel:
Fixed in #919

PR #919, opened 2017-05-14
Use Wikimedia's css sanitizer for custom css parsing

supersoniko:
As discussed in pms @Zarel 

 - Full css3 support (old sanitizer didn't have this)
 -  Covers all xss attacked listed at https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet

Main purpose: Allow side servers to have css3 in their custom css

*Nii Sama*

### DIFF
--- a/customcss.php
+++ b/customcss.php
@@ -2,6 +2,13 @@
 
 include '../pokemonshowdown.com/config/servers.inc.php';
 
+function __autoload($class_name) {
+    require_once('lib/css-sanitizer/'.$class_name.'.php');
+}
+
+use Wikimedia\CSS\Parser\Parser;
+use Wikimedia\CSS\Sanitizer\StylesheetSanitizer;
+
 $server = @$_REQUEST['server'];
 if ($server === 'showdown' || $server === 'smogtours') die();
 if (empty($PokemonServers[$server])) {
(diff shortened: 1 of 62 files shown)


panpawn:
There HAS to be an easier way to do this.

This is much too much.

supersoniko:
If you can show me sure, I've done my research and this is the only library that fulfills these requirements.
I don't think its more than htmlpurifier and csstidy library anyway?

panpawn:
See https://github.com/Zarel/Pokemon-Showdown-Client/issues/248

supersoniko:
Anyway it's csstidy that's the problem not htmlpurifier, Zarel said he was good with this library.

panpawn:
I don't see why we should add an entire new library where we don't necessarily have to is all

supersoniko:
More like replacing, since we can remove htmlpurifier and csstidy. I'll look into kota's solution if @Zarel wants that.

Zarel:
If someone _wants_ to port to Caja, that's fine by me, but barring that, I consider MediaWiki sufficiently trustworthy to publish a CSS sanitizer.

Zarel:
(Also, the MediaWiki sanitizer is easier for us to use because it's written in PHP, unlike Caja.)

Zarel:
Also, is our old sanitizer not in Git?

Zarel:
Oh, it's in the other repository. Yeah, this is better.

supersoniko:
We could use a package manager like Composer to easily maintain libraries. I don't know if we use enough php libraries to warrant such a change though. Anyhow that's a different issue than what I'm trying to solve in this PR.

Zarel:
And yes, I validated the CSS sanitizer code, it's unmodified from the official `css-sanitizer-1.0` from Wikimedia.

PR #920, opened 2017-05-14
Fix reference from tools.js to sim/dex.js which exports Tools

Sketch:
tools.js doesn't exist after a fresh clone of Pokemon-Showdown from GitHub, but Pokemon-Showdown's sim/dex.js does and exports the Tools that the Pokemon-Showdown-Client githooks scripts are looking for.

### DIFF
--- a/githooks/build-indexes
+++ b/githooks/build-indexes
@@ -14,7 +14,7 @@ child_process.execSync('git pull', {
 });
 console.log("DONE");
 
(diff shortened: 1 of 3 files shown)


Zarel:
This is fixed server-side. Do you actually use these?

Zarel:
Fixed in d0420138d5d

PR #925, opened 2017-05-22
Tooltips: improve wording when move is weakened by ability

asgdf:
Mostly applies to Rivalry.

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -883,6 +883,10 @@ var BattleTooltips = (function () {
 		return moveType;
 	};
 
+	BattleTooltips.prototype.makePercentageChangeText = function (boost, source) {
+		return ' (&times;' + boost + ' from ' + source + ')';
+	};
+
 	// Gets the current accuracy for a move.
 	BattleTooltips.prototype.getMoveAccuracy = function (move, pokemon) {
 		var pokemonData = this.room.myPokemon[pokemon.slot];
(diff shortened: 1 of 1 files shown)


Zarel:
I like this pullreq more (in terms of code readability), but I like my suggestion in #926 even more.

Zarel:
"×0.75 from Rivalry" is I think better phrasing.

Zarel:
(Discuss.)

urkerab:
Other than "(Technician boosted)", everything else already uses "(Boosted by ...)" or "(Reduced by ...)", so that wording would at least be consistent.

asgdf:
In particular, there's "Boosted by status" and "Boosted by lack of item" that don't quite work with "x2 from [...]", so either we'll have to use something else or be somewhat inconsistent.

Zarel:
"×2 from lack of item"

"×1.5 from status"

Seems fine to me?

asgdf:
But like, does the boost come *from* status, or does it rather come from Venoshock because of the status, for example? Maybe I'm overthinking this.

Zarel:
you can do "from activated Venoshock" and "from activated Unburden"

asgdf:
What about Wake-Up Slap and Smelling Salts? "×2 from activated Smelling Salts"? I think it's a bit unfortunate that the tooltip doesn't actually say what exactly *causes* the boost anymore then.
Would it be okay to have both "×2 from [ability]" and "×2 due to [condition]"?

Zarel:
"×2 from Smelling Salts + slp" or whatever is fine, but I think explaining how moves/abilities activate is maybe beyond the scope of a move tooltip...

asgdf:
Okay. Hope it's a good balance of conciseness and looking sufficiently pretty for end-user output now.

js/client-battle-tooltips.js
```diff
@@ -883,6 +883,10 @@ var BattleTooltips = (function () {
 		return moveType;
 	};
 
+	BattleTooltips.prototype.makePercentageChangeText = function (boost, source) {
+		return ' (×' + boost + ' from ' + source + ')';
```
Zarel: Unicode in client code reeeeeee

Zarel: Seriously, if this field accepts HTML, do `&times;`

Zarel: Otherwise do `\u...`

asgdf:
Woops. Fixed.

Zarel:
I don't know if I _like_ this solution but I don't dislike it.

Good enough, I guess.

PR #926, opened 2017-05-22
Tooltips: Don't call a BP penalty a boost

MacChaeger:


asgdf:
https://github.com/Zarel/Pokemon-Showdown-Client/pull/925
Yours is more concise, but on the other hand "(Rivalry weakened)" sounds a bit weird to my ears so idk. Zarel's call I suppose.

Zarel:
"boosted by ABILITY" and "weakened by ABILITY" would be the more grammatically correct way to do this, but I rather dislike wordiness.

Maybe "&times;FACTOR ABILITY"?

asgdf:
"×0.75 Rivalry"? Feels a bit too minimalistic to me, maybe "×0.75 due to Rivalry"?

Zarel:
"×0.75 from Rivalry"

Zarel:
Closing because #925 is a better basis.

PR #929, opened 2017-05-26
Don't crash when clicking empty targetting buttons

asgdf:
Even though the buttons were styled like they were disabled, the onclick events still fired. This caused invalid target locations to be added to the choice string, which basically broke the battle.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -524,7 +524,7 @@
 					if (disabled) {
 						targetMenus[0] += '<button disabled="disabled"></button> ';
 					} else if (!pokemon || pokemon.zerohp) {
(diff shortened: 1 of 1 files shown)


Slayer95:
In-game you can still target empty slots.

asgdf:
Oh. What happens if you do? It tries to execute the move but fails due to lack of a target?

Lionyx:
it redirects the move on the remaining enemy pokémon (or fails if you target your fainted partner)

asgdf:
Ok change of plans. Now they 1) actually look clickable and 2) don't cause crashes when you click on them.

PR #931, opened 2017-05-28
Don't format text inside a code block

QuiteQuiet:
Addresses the concerns brought up in #567 

It doesn't add or implement an entire markdown parser, though, just takes a different approach to how the escaping is done if code blocks are present and not suppressed.


Zarel:
I may have closed #567 but this isn't necessary either... Bumba and I are working on a new parser (currently available in `chat.js` on the server) that'll be significantly better at corner cases.

PR #932, opened 2017-05-30
Client: fix default order

sparkychildcharlie:
Change group type name when determining default order


sparkychildcharlie:
(there fixed commit title :x)

PR #934, opened 2017-05-31
Add editorconfig file

FelixRilling:
Having an [editorconfig](http://editorconfig.org/) file allows contributors in many IDEs/editors to automatically check for correct indentation and line breaks.

I configured the file based on the styleguide of the server-repo and the existing files, if I missed something please tell me.

```ini
# editorconfig.org

root = true

[*]
charset = utf-8
end_of_line = lf
indent_size = 4
indent_style = tab
insert_final_newline = true
trim_trailing_whitespace = true

[*.md]
trim_trailing_whitespace = false
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = tab
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false


Zarel:
Which editors support `editorconfig`?

Zarel:
Apparently Sublime, Atom, and VS Code all require plugins. Unfortunate, because I think those are the most popular GUI editors for JavaScript right now.

FelixRilling:
While that is true, I feel like the file could be added regardless.
It doesn't hurt developers without the plugins and helps those who have them.

.editorconfig
```diff
@@ -0,0 +1,14 @@
+# editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
```
Zarel: 2

Zarel:
The other issue is, the entire reason we use tabs is so people can decide their indent depth for themselves. I hope editors let them override editorconfig settings without needing a dirty tree.

Zarel:
?

Zarel:
Was closing this a mistake?

Zarel:
This ended up happening in 9b108361b80cdc58

PR #937, opened 2017-06-09
Fix close button on replay uploaded popup

panpawn:


Zarel:
Close is the only button, it should be the submit button.

Does submit not close the window? That should be fixed.

panpawn:
On regular popups, this is how the "OK" button works

Zarel:
Hm, yeah, this is probably better, then.

PR #938, opened 2017-06-10
Updated Gen 1 Viability

Enigami:
Updates the viability of moves in Gen 1

Edit: Had a couple mistakes, fixed, working on squashing commits.
Edit2: Informed squashing on my end isn't necessary, is that the case?

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -976,19 +976,13 @@
 					if (id === 'disable' || id === 'firepunch' || id === 'icepunch' || id === 'leechseed' || id === 'quickattack' || id === 'roar' || id === 'thunderpunch' || id === 'toxic' || id === 'triattack' || id === 'whirlwind') isViable = false;
 
 					// Viable only when certain moves aren't present
+					if (id === 'bubblebeam') isViable = ((moves.indexOf('surf') < 0 && moves.indexOf('blizzard') < 0));
 					if (id === 'doubleedge') isViable = (moves.indexOf('bodyslam') < 0);
 					if (id === 'doublekick') isViable = (moves.indexOf('submission') < 0);
 					if (id === 'megadrain') isViable = (moves.indexOf('razorleaf') < 0 && moves.indexOf('surf') < 0);
 					if (id === 'megakick') isViable = (moves.indexOf('hyperbeam') < 0);
 					if (id === 'reflect') isViable = (moves.indexOf('barrier') < 0 && moves.indexOf('acidarmor') < 0);
-
-					// Viable for specific types
-					if (id === 'substitute') isViable = (template.types[0] === 'Ground' || template.types[1] === 'Ground');
-					if (id === 'fireblast' || id === 'flamethrower') isViable = (template.types[0] === 'Fire' || template.types[1] === 'Fire');
-
-					// Complicated viability
-					if (id === 'bubblebeam') isViable = ((moves.indexOf('surf') < 0 && moves.indexOf('blizzard') < 0));
-					if (id === 'submission') isViable = ((template.types[0] === 'Fighting' || template.types[1] === 'Fighting') && moves.indexOf('highjumpkick') < 0);
+					if (id === 'submission') isViable = (moves.indexOf('highjumpkick') < 0);
 				}
 				if (isViable) {
 					if (!usableMoves.length) usableMoves.push(['header', "Moves"]);
--- a/js/search.js
+++ b/js/search.js
@@ -969,10 +969,10 @@
 				if (this.gen === 1) {
 					//Usually viable for Gen 1
 					if (id === 'acidarmor' || id === 'amnesia' || id === 'barrier' || id === 'bind' || id === 'clamp' || id === 'confuseray' || id === 'counter' || id === 'firespin' || id === 'hyperbeam' || id === 'mirrormove' || id === 'pinmissile' || id === 'razorleaf' || id === 'sing' || id === 'slash' || id === 'sludge' || id === 'twineedle' || id === 'wrap') isViable = true;
-					
+
 					//Usually not viable for Gen 1
 					if (id === 'disable' || id === 'firepunch' || id === 'icepunch' || id === 'jumpkick' || id === 'leechseed' || id === 'quickattack' || id === 'roar' || id === 'thunder' || id === 'thunderpunch' || id === 'toxic' || id === 'triattack' || id === 'whirlwind') isViable = false;
-					
+
 					//Viable only when certain moves aren't present
 					if (id === 'bubblebeam') isViable = (moves.indexOf('surf') < 0 && moves.indexOf('blizzard') < 0);
 					if (id === 'doubleedge') isViable = (moves.indexOf('bodyslam'));
@@ -980,15 +980,15 @@
 					if (id === 'megadrain') isViable = (moves.indexOf('razorleaf') < 0 && moves.indexOf('surf') < 0);
 					if (id === 'megakick') isViable = (moves.indexOf('hyperbeam') < 0);
 					if (id === 'reflect') isViable = (moves.indexOf('barrier') < 0 && moves.indexOf('acidarmor') < 0);
-					
+
 					//Viable for specific types
 					if (id === 'substitute') isViable = (template.types[0] === 'Ground' || template.types[1] === 'Ground');
-					
+
 					//Viable for specific Pokemon
 					if (id === 'doublekick') isViable = (toId(set.species) === 'jolteon');
 					if (id === 'lightscreen') isViable = (toId(set.species) === 'zapdos');
 					if (id === 'meditate') isViable = (toId(set.species) === 'hitmonlee');
-					
+
 					//Complicated viability
 					if (id === 'fireblast' || id === 'flamethrower') isViable = (template.types[0] === 'Fire' || template.types[1] === 'Fire' || toId(set.species) === 'aerodactyl');
 					if (id === 'submission') isViable = ((template.types[0] === 'Fighting' || template.types[1] === 'Fighting') && moves.indexOf('highjumpkick') < 0);
--- a/js/search.js
+++ b/js/search.js
@@ -960,7 +960,9 @@
 				if (id === 'hiddenpowerfire') isViable = (moves.indexOf('flamethrower') < 0);
 				if (id === 'hiddenpowergrass') isViable = (moves.indexOf('energyball') < 0 && moves.indexOf('gigadrain') < 0);
 				if (id === 'hiddenpowerice') isViable = (moves.indexOf('icebeam') < 0 && template.id !== 'xerneas');
+				if (id === 'hypnosis') isViable = (this.gen < 4 && moves.indexOf('sleeppowder') < 0);
 				if (id === 'icywind') isViable = (toId(set.species).substr(0, 6) === 'keldeo');
+				if (id === 'jumpkick') isViable = (moves.indexOf('highjumpkick') < 0);
 				if (id === 'leechlife') isViable = (this.gen > 6);
 				if (id === 'reflecttype') isViable = (toId(set.species) in {latias:1, starmie:1});
 				if (id === 'selfdestruct') isViable = (this.gen < 5 && moves.indexOf('explosion') < 0);
@@ -971,28 +973,22 @@
 					if (id === 'acidarmor' || id === 'amnesia' || id === 'barrier' || id === 'bind' || id === 'clamp' || id === 'confuseray' || id === 'counter' || id === 'firespin' || id === 'hyperbeam' || id === 'mirrormove' || id === 'pinmissile' || id === 'razorleaf' || id === 'sing' || id === 'slash' || id === 'sludge' || id === 'twineedle' || id === 'wrap') isViable = true;
 
 					// Usually not viable for Gen 1
-					if (id === 'disable' || id === 'firepunch' || id === 'icepunch' || id === 'jumpkick' || id === 'leechseed' || id === 'quickattack' || id === 'roar' || id === 'thunderpunch' || id === 'toxic' || id === 'triattack' || id === 'whirlwind') isViable = false;
+					if (id === 'disable' || id === 'firepunch' || id === 'icepunch' || id === 'leechseed' || id === 'quickattack' || id === 'roar' || id === 'thunderpunch' || id === 'toxic' || id === 'triattack' || id === 'whirlwind') isViable = false;
 
 					// Viable only when certain moves aren't present
 					if (id === 'doubleedge') isViable = (moves.indexOf('bodyslam') < 0);
-					if (id === 'hypnosis') isViable = (moves.indexOf('sleeppowder') < 0);
+					if (id === 'doublekick') isViable = (moves.indexOf('submission') < 0);
 					if (id === 'megadrain') isViable = (moves.indexOf('razorleaf') < 0 && moves.indexOf('surf') < 0);
 					if (id === 'megakick') isViable = (moves.indexOf('hyperbeam') < 0);
 					if (id === 'reflect') isViable = (moves.indexOf('barrier') < 0 && moves.indexOf('acidarmor') < 0);
 
 					// Viable for specific types
 					if (id === 'substitute') isViable = (template.types[0] === 'Ground' || template.types[1] === 'Ground');
-
-					// Viable for specific Pokemon
-					if (id === 'doublekick') isViable = (toId(set.species) === 'jolteon');
-					if (id === 'lightscreen') isViable = (toId(set.species) === 'zapdos');
-					if (id === 'meditate') isViable = (toId(set.species) === 'hitmonlee');
-					if (id === 'thunder') isViable = (toId(set.species) === 'zapdos');
+					if (id === 'fireblast' || id === 'flamethrower') isViable = (template.types[0] === 'Fire' || template.types[1] === 'Fire');
 
 					// Complicated viability
-					if (id === 'bubblebeam') isViable = ((moves.indexOf('surf') < 0 && moves.indexOf('blizzard') < 0) || toId(set.species) === 'raticate');
-					if (id === 'fireblast' || id === 'flamethrower') isViable = (template.types[0] === 'Fire' || template.types[1] === 'Fire' || toId(set.species) === 'aerodactyl');
-					if (id === 'submission') isViable = ((template.types[0] === 'Fighting' || template.types[1] === 'Fighting' || toId(set.species) === 'pinsir') && moves.indexOf('highjumpkick') < 0);
+					if (id === 'bubblebeam') isViable = ((moves.indexOf('surf') < 0 && moves.indexOf('blizzard') < 0));
+					if (id === 'submission') isViable = ((template.types[0] === 'Fighting' || template.types[1] === 'Fighting') && moves.indexOf('highjumpkick') < 0);
 				}
 				if (isViable) {
 					if (!usableMoves.length) usableMoves.push(['header', "Moves"]);
--- a/js/search.js
+++ b/js/search.js
@@ -951,7 +951,6 @@
 				var isViable = BattleMovedex[id] && BattleMovedex[id].isViable;
 				if (id === 'aerialace') isViable = (toId(set.ability) === 'technician');
 				if (id === 'bellydrum') isViable = (toId(set.species) in {azumarill:1, linoone:1, slurpuff:1});
-				if (id === 'clamp' || id === 'hyperbeam' || id === 'wrap') isViable = (this.gen === 1);
 				if (id === 'counter') isViable = (toId(set.species) in {chansey:1, skarmory:1, clefable:1, wobbuffet:1});
 				if (id === 'dynamicpunch') isViable = (toId(set.ability) === 'noguard');
 				if (id === 'focuspunch') isViable = (toId(set.species) === 'breloom');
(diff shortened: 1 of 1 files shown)


Enigami:
Details:

Acid Armor: Viable, superior Reflect in Gen 1
Amnesia: Viable, effectively x2 Calm Mind in Gen 1
Barrier: Viable, superior Reflect in Gen 1
Bind: Viable, partial trapping is strong in Gen 1
BubbleBeam: Viable when Surf and Blizzard aren't available
Clamp: Viable, partial trapping is strong in Gen 1
Confuse Ray: Viable, users can usually afford the slot and it can help break through walls
Counter: Viable, even Rock- and Ghost-types can make niche use of this because of RBY's mechanics.
Disable: Unviable, super inaccurate and there are much better moves its users can run
Double Edge: Viable when Body Slam isn't available, otherwise not enough of an increase of damage to make up for recoil/lack of paraylsis
Double Kick: Viable for Jolteon, usually unviable for everything else
Fire Blast: Viable for Fire-types and Aerodactyl, usually unviable or niche for everything else
Flamethrower: Viable for Fire-types, usually unviable for everything else
Fire Spin: Viable, partial trapping is strong in Gen 1
Fire Punch: Unviable, users are Magmar who has Flamethrower and Hitmonchan who has an unusable Special stat
Growth: Viable, effectively Calm Mind in Gen 1
Hypnosis: Viable, Gen 1 Sleep is very good (unviable with Sleep Powder so it doesn't show up as viable on Exeggutor or Exeggcute)
Ice Punch: Unviable, users are Jynx who has Ice Beam and Hitmonchan who has an unusable Special stat
Jump Kick: Unviable, only user is Hitmonlee who has High Jump Kick
Leech Seed: Unviable
Light Screen: Viable for Zapdos, usually unviable for everything else
Meditate: Viable for Hitmonlee, usually unviable for everything else
Mega Drain: Viable when Razor Leaf and Surf aren't available
Mega Kick: Viable(ish) for being the strongest Normal-type attack available to Pokemon lacking Hyper Beam, such as Hitmonlee
Mirror Move: Viable, users are Fearow and Pidgeot, and both can afford the slot for it
Pin Missile: Viable, users are Jolteon and Beedrill which both have use for it
Quick Attack: Unviable, niche at best even on strong Normal-types.
Razor Leaf: Viable, crits almost always occur greatly boosting utility and damage output
Reflect: Viable when Acid Armor/Barrier aren't available
Roar: Unviable, doesn't work in Gen 1
Sing: Viable, Gen 1 Sleep is very good
Slash: Viable, crits almost always occur greatly boosting utility and damage output
Sludge: Viable, users are Weezing and Muk who can make use of it
Submission: Viable(ish) for Fighting-types, usually unviable for everything else
Substitute: Viable for Ground-types, usually unviable or niche for everything else
Thunder: Unviable, the lack of rain, inaccuracy and only 10% chance of paralysis makes Thunderbolt preferable in nearly all cases.
Thunder Punch: Unviable, users are Electabuzz who has Thunderbolt and Hitmonchan who has an unusable Special stat
Tri Attack: Unviable, bad move on virtually everything and niche at best on Porygon
Twineedle: Viable, best STAB for Beedrill
Whirlwind: Unviable, doesn't work in Gen 1

js/search.js
```diff
@@ -962,11 +961,38 @@
...
+					//Viable for specific types
+					if (id === 'substitute') isViable = (template.types[0] === 'Ground' || template.types[1] === 'Ground');
+
+					//Viable for specific Pokemon
+					if (id === 'doublekick') isViable = (toId(set.species) === 'jolteon');
+					if (id === 'lightscreen') isViable = (toId(set.species) === 'zapdos');
+					if (id === 'meditate') isViable = (toId(set.species) === 'hitmonlee');
+
+					//Complicated viability
+					if (id === 'fireblast' || id === 'flamethrower') isViable = (template.types[0] === 'Fire' || template.types[1] === 'Fire' || toId(set.species) === 'aerodactyl');
```
Zarel: Fire Blast and Flamethrower should be always viable. We reserve complicated viability for very special situations, not 95/100% or 120/85% moves.

js/search.js
```diff
@@ -962,11 +961,38 @@
...
+				if (this.gen === 1) {
+					//Usually viable for Gen 1
+					if (id === 'acidarmor' || id === 'amnesia' || id === 'barrier' || id === 'bind' || id === 'clamp' || id === 'confuseray' || id === 'counter' || id === 'firespin' || id === 'hyperbeam' || id === 'mirrormove' || id === 'pinmissile' || id === 'razorleaf' || id === 'sing' || id === 'slash' || id === 'sludge' || id === 'twineedle' || id === 'wrap') isViable = true;
+
+					//Usually not viable for Gen 1
+					if (id === 'disable' || id === 'firepunch' || id === 'icepunch' || id === 'jumpkick' || id === 'leechseed' || id === 'quickattack' || id === 'roar' || id === 'thunder' || id === 'thunderpunch' || id === 'toxic' || id === 'triattack' || id === 'whirlwind') isViable = false;
+
+					//Viable only when certain moves aren't present
+					if (id === 'bubblebeam') isViable = (moves.indexOf('surf') < 0 && moves.indexOf('blizzard') < 0);
+					if (id === 'doubleedge') isViable = (moves.indexOf('bodyslam'));
```
Zarel: This appears to be a typo.

js/search.js
```diff
@@ -962,11 +961,38 @@
...
 				if (id === 'hiddenpowerice') isViable = (moves.indexOf('icebeam') < 0 && template.id !== 'xerneas');
 				if (id === 'icywind') isViable = (toId(set.species).substr(0, 6) === 'keldeo');
+				if (id === 'leechlife') isViable = (this.gen > 6);
 				if (id === 'reflecttype') isViable = (toId(set.species) in {latias:1, starmie:1});
 				if (id === 'selfdestruct') isViable = (this.gen < 5 && moves.indexOf('explosion') < 0);
 				if (id === 'skyattack') isViable = (toId(set.species) === 'hawlucha');
 				if (id === 'stunspore') isViable = (moves.indexOf('thunderwave') < 0);
-				if (id === 'toxic') isViable = (this.gen > 1);
+				if (this.gen === 1) {
+					//Usually viable for Gen 1
```
Zarel: Our convention is to leave a space after `//`

Enigami: Already fixed.

js/search.js
Enigami: Where's the typo? I can't find it.

js/search.js
Enigami: If you're adamant about this, I can remove the check, but generally Fire Blast and Flamethrower are undesirable in RBY due to inflicting burn and only rarely is useful for coverage. For the most part only Fire-types, Aerodactyl, and maybe Machamp want this.

DisasterAreaPS: Yeah. It would be better for Fire Blast to be listed as never viable than as always viable. On most Pokemon it is actively detrimental, but on Fire-types, on Aerodactyl, and in one or two very niche cases (usually in lower tiers) it is a viable option.

Zarel: Viability is always subjective, and these lists especially are used not only for OU but also for other tiers and OMs. That's why we don't evaluate strong attacks on a case-by-case basis. We display special attacks as viable on physical attackers and vice versa. Even if these moves are only good in obscure situations, maybe you have a Parasect problem in an OM or something. I would still rather show them than not; players can know not to choose them.

js/search.js
```diff
@@ -962,11 +961,39 @@
...
 				if (id === 'selfdestruct') isViable = (this.gen < 5 && moves.indexOf('explosion') < 0);
 				if (id === 'skyattack') isViable = (toId(set.species) === 'hawlucha');
 				if (id === 'stunspore') isViable = (moves.indexOf('thunderwave') < 0);
-				if (id === 'toxic') isViable = (this.gen > 1);
+				if (this.gen === 1) {
+					// Usually viable for Gen 1
+					if (id === 'acidarmor' || id === 'amnesia' || id === 'barrier' || id === 'bind' || id === 'clamp' || id === 'confuseray' || id === 'counter' || id === 'firespin' || id === 'hyperbeam' || id === 'mirrormove' || id === 'pinmissile' || id === 'razorleaf' || id === 'sing' || id === 'slash' || id === 'sludge' || id === 'twineedle' || id === 'wrap') isViable = true;
+
+					// Usually not viable for Gen 1
+					if (id === 'disable' || id === 'firepunch' || id === 'icepunch' || id === 'jumpkick' || id === 'leechseed' || id === 'quickattack' || id === 'roar' || id === 'thunderpunch' || id === 'toxic' || id === 'triattack' || id === 'whirlwind') isViable = false;
```
Zarel: Instead of listing Jump Kick viability per-gen, maybe it should just be "viable only if High Jump Kick not in set", and moved outside of the Gen 1 section.

js/search.js
```diff
@@ -962,11 +961,39 @@
...
+
+					// Viable only when certain moves aren't present
+					if (id === 'doubleedge') isViable = (moves.indexOf('bodyslam') < 0);
+					if (id === 'hypnosis') isViable = (moves.indexOf('sleeppowder') < 0);
```
Zarel: This also seems like a check that should be moved out of Gen 1.

js/search.js
```diff
@@ -962,11 +961,39 @@
...
+					if (id === 'megadrain') isViable = (moves.indexOf('razorleaf') < 0 && moves.indexOf('surf') < 0);
+					if (id === 'megakick') isViable = (moves.indexOf('hyperbeam') < 0);
+					if (id === 'reflect') isViable = (moves.indexOf('barrier') < 0 && moves.indexOf('acidarmor') < 0);
+
+					// Viable for specific types
+					if (id === 'substitute') isViable = (template.types[0] === 'Ground' || template.types[1] === 'Ground');
```
Zarel: Can you tell me more about how Substitute works?

js/search.js
```diff
@@ -962,11 +961,39 @@
...
+
+					// Viable for specific Pokemon
+					if (id === 'doublekick') isViable = (toId(set.species) === 'jolteon');
+					if (id === 'lightscreen') isViable = (toId(set.species) === 'zapdos');
+					if (id === 'meditate') isViable = (toId(set.species) === 'hitmonlee');
+					if (id === 'thunder') isViable = (toId(set.species) === 'zapdos');
```
Zarel: Is Thunder needed for specific KOs on Zapdos?

js/search.js
```diff
@@ -962,11 +961,39 @@
...
+					if (id === 'hypnosis') isViable = (moves.indexOf('sleeppowder') < 0);
+					if (id === 'megadrain') isViable = (moves.indexOf('razorleaf') < 0 && moves.indexOf('surf') < 0);
+					if (id === 'megakick') isViable = (moves.indexOf('hyperbeam') < 0);
+					if (id === 'reflect') isViable = (moves.indexOf('barrier') < 0 && moves.indexOf('acidarmor') < 0);
+
+					// Viable for specific types
+					if (id === 'substitute') isViable = (template.types[0] === 'Ground' || template.types[1] === 'Ground');
+
+					// Viable for specific Pokemon
+					if (id === 'doublekick') isViable = (toId(set.species) === 'jolteon');
```
Zarel: I assume it's outclassed on Fighting types. Can you just check for the moves it's outclassed by, instead of checking specific Pokémon?

js/search.js
Enigami: Thunder has a chance to OHKO Starmie, Slowbro, always KOs Cloyster, 2HKO Tauros, and probably some other things but I think that's sufficient enough. I personally do not prefer it, but others believe it to be a viable 4th move along with moves like Light Screen or Agility and that it should be listed as viable.

js/search.js
Enigami: Doable, I can have it check for Submission. The only users of Double Kick are Jolteon, the Nido family, and Hitmonlee. Hitmonlee and Nidoking/queen have Submission, so the check should reduce it to Jolteon. It also would leave it listed as viable for the NFE Nidos, but they have a small use for it (strongest attack they have against Normal-types), so it should be fine.

Zarel:
So, a general rule of thumb for special-casing viability:

1. If a move is outclassed by another move, check for the presence of the other move
2. If a move has poor coverage but is otherwise the best move of its type, leave it viable: we use these lists in OMs and weird tiers and it is really not hard to just ignore them
3. If a move changed mechanics between gens in a way that changes viability, check the gen

Checking specific Pokémon should nearly never be done; mostly reserved for stuff like Gen 7 Dark Void.

Zarel:
Also, "niche at best" means "viable" in the current version of the teambuilder, so I'd weakly suggest putting Quick Attack back (although I won't insist on that).

js/search.js
DisasterAreaPS: It blocks Confuse Ray and Toxic and the secondary status from moves such as Flamethrower and Ice Beam, but doesn't block thunder wave, hypnosis, or any other sleep move. As such it's only really seen as a filler on Pokemon which can force switches that don't have much else to be doing (usually ground-types: Rhydon is the only common user in OU, whilst Dugtrio / Nidoking / Nidoqueen can use it in (PP's) lower tiers). The other cases where it's used, if you don't mind me talking in the context of very low tiers on PP, in RBY 5U [PU] on PP, Seadra [a top Pokemon, seen on every team] uses it frequently because of its ability to block the secondary effect of Ice Beam, and Pidgeot uses Substitute in combination with Toxic & Rest in RBY 6U [ZU] to wear down Kabutops.

DisasterAreaPS:
Quick Attack isn't _really_ viable in RBY in any case. There are some very very fringe cases, but no Pokemon runs it as its first choice on its standard set in any tier. Probably the most viable use for it is as a filler on Swords Dance Scyther in RBY 4U [NU].

js/search.js
Enigami: Works like later gen Substitutes, except doesn't block status (except in Gen 1 Stadium, which has differing mechanics to normal RBY). Ground-types have a much better time using Substitute because:
1) All Ground-type generally have all the coverage and power they need in Rock Slide + Earthquake leaving them with an easy time fitting Substitute on a set.
2) Pokemon in RBY are usually crippled by status, and most Pokemon that inflict status do so with Thunder Wave which Ground-types are immune to. This reduces the problem with status moves going through Substitute.
3) Ground-types are often very weak to common Water/Ice/Grass attacks that can 1-2HKO, and so successfully finding an opening to use Substitute (such as fishing for Full Paralysis, scaring out an opponent, setting up on a sleeping/frozen Pokemon) can give them an extra turn they wouldn't have had otherwise to hammer the opponent with powerful STAB Earthquakes.

js/search.js
DisasterAreaPS: Well, with Hypnosis vs Sleep Powder, Hypnosis's poor accuracy [60%, except in DP (modified mid-gen so it's 75% in diamond and pearl but 60% in all other gen 4 games)] means that outside of gen 1 / 3, where sleep isn't as powerful, it's usually not listed as viable (idk about gen 7 though), whereas Sleep Powder is considered viable in any generation.

In later generations it might be more viable to run both Body Slam and Double Edge on the same set (maybe Tauros in GSC or ADV or something in lower tiers?) or equally it's possible that either both are viable separate options (normal-type CBers, Snorlax, etc.) in later generations, or that neither is viable (is venusaur going to run either of those in gen 4 onwards? absolutely not). In ADV, Body Slam is viable on some Pokemon (like Jirachi in OU) but Double Edge would be entirely unviable even if Jirachi didn't get Body Slam.

Enigami: Both Double Edge<Body Slam and Hypnosis<Sleep Powder are highlighted, I presume you just meant Hypnosis<Sleep Powder. Will be done, though if you did mean Double Edge<Body Slam, let me know, though that should probably just be a Gen 1 check because Double Edge is significantly stronger from Gen 2 onward.

DisasterAreaPS:
I found an alternative with Submission... if we made it check to see if it was Bug type then that would also work as a means of effectively getting Pinsir into the equation

DisasterAreaPS:
Think is Pinsir really is one of the few Pokemon that really needs Submission's coverage a lot of the time. It's its only way to hit Rock-types (relevant in any tier where it is viable). Thing is there's a huge number of Pokemon with access to submission, and Pinsir is the only thing without STAB on it that would want it. For example, special attackers that have much better things to be doing (most psychic-types, a bunch of water-types, magmar, omanyte, gengar) as well as physical attackers with normal- or ground-type stabs that would never need it because they have coverage of some kind, and then also kabutops, charizard and its line (although charmeleon/charmander maybe could use it...), and raichu (maybe could use it in OU but it's ultra gimmicky) also get it. In short, there's a good logical explanation why Pinsir is a viable user of Submission and everything else without STAB on it isn't, but that logical explanation is so long that if you decided to code it it would be a huge waste of time because you're better off just coding an exception for Pinsir.

js/search.js
Enigami: I've played more RBY tiers and RBY OMs on Pokemon Perfect than will probably exist on Smogon, and I can tell you, non-STAB Fire Blast and Flamethrower are objectively bad and should not be listed as viable. Newer players should not be using these moves if they are not on Fire-types (or Aerodactyl, which only wants it due to its complete lack of coverage). Any possible exceptions require a strong understanding of RBY and the meta they play in, such that it wouldn't need to be listed as 'viable'.

js/search.js
Zarel: Okay, that's fair.

js/search.js
Zarel: Maybe add Seadra and Pidgeot to hardcoding, then?

Enigami: Uh, well, I deleted all the hardcoded Pokemon checks, should I restore them (or just some?)

DisasterAreaPS:
Here's a breakdown of Fire Blast.
- Fire Blast has a 30% burn rate.
- Fire Blast is always run on Fire-types, when they are viable.
- Fire Blast is strongly discouraged in OU because of the inability to land a separate status on burnt Pokemon with recovery, making Chansey, Alakazam, and Starmie, much more difficult to beat. Because of this, the move itself is very rare, and Fire-types are very rare.
- Fire Blast is also discouraged a little in lower tiers due to the possibility of thawing, but in practice this doesn't usually dissuade users from choosing to use the move over other concerns, even in metagames where freezing is very significant (most importantly 5U).
- Whether Fire Blast is a good choice on a Pokemon depends a lot on the specific Pokemon and metagame. For example, a few Pokemon with low special can sometimes make use of it (Golem in OU, Nidos in 3U, Machamp in 6U) because it provides useful coverage versus a common switch-in, in spite of frozen Pokemon sometimes being possible switch-ins. Some Pokemon can make use of it for a lack of any other coverage options (Aerodactyl in 3U for example) but these are rarer cases.
- Because of the big downsides of Fire Blast, combined with how widely it is distributed, means that if the focus is significantly more on OU more than PP's lower tiers, and more targeted at newer/weaker players than experienced/stronger players, that Fire Blast probably shouldn't be listed as viable. Fire Blast on Pokemon like Chansey, Slowbro, and Snorlax is a pretty awful idea, but is often seen on the ladder used by newer players, in particular Fire Blast Chansey. RBY players who are most familiar with OU would absolutely shirk at the thought of Fire Blast being listed as a "viable option"; most experienced players won't use the more niche Fire-type Pokemon in OU or use Fire Blast in the niche capacity that it has on a few stronger attackers in OU (such as Golem or maybe Tauros) because of the numerous downsides the move presents.

Zarel:
There is a reason we removed the entire viable moves section and now just have a usable moves section. "Usable moves" is supposed to just be any move that isn't mostly-useless and isn't outclassed. It is very much _not_ supposed to be a "recommended moves" section, but more "list of moves, minus the ones we're pretty sure you'd never use"

Enigami:
Fair enough. Alright, moving Fire Blast back to viable.

Zarel:
All right, looks great; thanks a lot!

DisasterAreaPS:
I don't think Flamethrower is viable on anything.

Submission is usable on Fighting-types without HJK & Pinsir, and viable for a subset of that. Not sure if that's the right call.

Enigami:
That's where the issue is, though. Zarel's right, Showdown's qualifier is 'usage' not 'viability'. Flamethrower is usable because it is more accurate than Fire Blast despite reduced power, and Submission is "usable" on non Fighting-types due to SE hits on Normal-types, and should be listed as such despite neither being viable.

DisasterAreaPS:
Literally only on Pinsir though and not on the other more than 40 Pokemon with access to it.

DisasterAreaPS:
I guess on usable vs viable, maybe flamethrower is "usable" but I have never used it or seen it considered higher than Other Options on any analysis ever. I guess if that's the definition we're going with though...

DisasterAreaPS:
Also if we're going with Usable over Viable then there's an argument to move Thunder to usable too. Although I would never ever use Thunder on anything not called Zapdos (in which case I'd use it alongside) in any tier currently formed, I also wouldn't ever use Flamethrower. If the trade-off between power / accuracy isn't seen as very relevant between flamethrower and fire blast (where fire blast has better power, worse accuracy, and a 3x higher burn chance) then thunder's worse accuracy (albeit 70% is much worse than fire blast's 85%) probably isn't enough to stop it being considered "usable".

Zarel:
Like I said, the reason we got rid of the "viable moves" section in the first place is because it leads to arguments like this one.

And yes, I also did kind of want Thunder as a usable move.

DisasterAreaPS:
On that note Enigami is intending to change Substitute to being unconditional as it's "usable" on most things.

DisasterAreaPS:
Still, with regards to submission: if we are going to do it on the notion of usability rather than viability, it is still frankly only usable on Fighting-types and Pinsir. On everything else, they have some other means of hitting rock-types that doesn't have poor accuracy + recoil + barely outdamages hyper beam.

Zarel:
That's fine by me re: Submission and Substitute.

PR #940, opened 2017-06-11
Change tenary operators to or statements

panpawn:
"Oh, btw, this isn't standard PS notation." 

(this is the (late) update from that comment)


Zarel:
...this is getting replaced pretty soon, but sure.

PR #941, opened 2017-06-11
Link Dex articles in various data commands output

panpawn:
And let's open them in a new tab, shall we


Zarel:
...this is going to break PSdex.

panpawn:
oh

Zarel:
This is also no longer necessary; `/data` is rendered server-side now (iirc)

PR #942, opened 2017-06-13
Include bumbadadabum in major contributors credits

xfix:
Congratulations, bumbadadabum!


Zarel:
One day we'll just get rid of this credits screen, since I really don't want to constantly sync it. :S

bumbadadabum:
t-thanks

PR #943, opened 2017-06-14
CSS Update — fix some darkmode elements

WallyTheBully:
Teambuilder files manager (darkmode): https://prnt.sc/fjtkgm
Menugroup (darkmode): https://prnt.sc/fjtt1g
Pm Box (darkmode / darkmode notification): https://prnt.sc/fjued6

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -3343,7 +3343,7 @@ a.ilink.yours {
 }
 
 .dark .folderpane {
-    border-left-color: #484848;
+    	border-left-color: #484848;
 }
 .dark .folderlist .foldersep:before,
 .dark .folderlistafter:before,
@@ -3356,36 +3356,36 @@ a.ilink.yours {
 }
 
 .dark .folderpane h3 {
-    margin: 0;
-    padding: 13px 0 0 0;
-    height: 17px;
-    font-size: 9pt;
-    background: #d7e3ec;
-    color: #f9f9f9;
-    border-right: 1px solid #888888;
+    	margin: 0;
+    	padding: 13px 0 0 0;
+    	height: 17px;
+    	font-size: 9pt;
+    	background: #d7e3ec;
+    	color: #f9f9f9;
+    	border-right: 1px solid #888888;
 }
 .dark .folder .selectFolder {
-    display: block;
-    padding: 0 0 0 7px;
-    border: 0;
-    border-right: 1px solid #888888;
-    background: #d7e3ec;
-    color: #f9f9f9;
-    width: auto;
-    max-width: 146px;
-    padding-top: 7px;
-    height: 23px;
-    font-size: 9pt;
-    font-family: Verdana,sans-serif;
-    white-space: nowrap;
-    overflow: hidden;
-    text-align: left;
-    cursor: pointer;
-    user-select: none;
-    -webkit-user-select: none;
+   	display: block;
+        padding: 0 0 0 7px;
+  	border: 0;
+	border-right: 1px solid #888888;
+  	background: #d7e3ec;
+  	color: #f9f9f9;
+  	width: auto;
+ 	max-width: 146px;
+  	padding-top: 7px;
+  	height: 23px;
+   	font-size: 9pt;
+  	font-family: Verdana,sans-serif;
+   	white-space: nowrap;
+  	overflow: hidden;
+ 	text-align: left;
+   	cursor: pointer;
+  	user-select: none;
+   	-webkit-user-select: none;
 }
 .dark .folder .selectFolder:hover {
-    background: #383838;
+    	background: #383838;
 }
 .dark .folder .selectFolder.active,
 .dark .folder .selectFolder:active {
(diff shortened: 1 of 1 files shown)


xfix:
Looks great.

Zarel:
The 3D was somewhat important for the dropdowns though :s

WallyTheBully:
True @Zarel , updating it asap (box shadow and avoiding transparent background considering the 3d)

Zarel:
I mean, you could have a translucent 3D effect, it would look glassy which would be kinda cool.

WallyTheBully:
updating it rn, I'll send sceenshots, as you said drag & drop 3d effect is really important on the teambuilder and I was about to modify it when you commented, anyways, I agree with you

WallyTheBully:
https://prnt.sc/fjvi3c

would it be better like this? (glassy effect; 3d looks alright in my opinion)

Zarel:
It's okay, it looks a bit similar to other buttons but I guess that's not a huge problem.

style/client.css
```diff
@@ -117,11 +117,10 @@ button:disabled {
...
 .dark .textbox {
-	border-color: #888888;
-
-	box-shadow: none;
-	background: #282B2D;
-	color: #DDD;
+    border-color: #888888;
+    box-shadow: none;
+    background: rgb(60, 59, 58);
+    color: #DDD;
```
Zarel: Tabs.........

WallyTheBully:
Done @Zarel waiting for approval, and yeah it looks a bit similar but I might be able to improve it, if so, I'll make another pull request

Zarel:
Ummm.... this is not ok

image

No line should start with a space.

Zarel:
It's still not fixed.

WallyTheBully:
sorry, fixed it, it should be fine now (removed the inappropriate space and added a tab)

panpawn:
Nope, those are still spaces

style/client.css
```diff
@@ -117,26 +117,29 @@ button:disabled {
 }
 
 .dark .textbox {
-	border-color: #888888;
-
-	box-shadow: none;
-	background: #282B2D;
-	color: #DDD;
+    	border-color: #888888;
```
Zarel: This has four extraneous spaces.

style/client.css
```diff
@@ -117,26 +117,29 @@ button:disabled {
...
+    	box-shadow: none;
```
Zarel: This too

style/client.css
```diff
@@ -117,26 +117,29 @@ button:disabled {
...
+    	background: rgb(60, 59, 58);
```
Zarel: This too

style/client.css
```diff
@@ -117,26 +117,29 @@ button:disabled {
...
+    	color: #DDD;
```
Zarel: This too

style/client.css
```diff
@@ -3340,22 +3345,56 @@ a.ilink.yours {
 /* teambuilder */
 
 .dark .folderpane {
-	border-left-color: #77838c;
+	border-left-color: #484848;
 }
 
+.dark .folderpane {
+    	border-left-color: #484848;
```
Zarel: This too

panpawn:
@Zarel, I think they all do

style/client.css
```diff
@@ -3340,22 +3345,56 @@ a.ilink.yours {
...
 .dark .folderpane h3,
 .dark .folder .selectFolder,
 .dark .folderhack1,
 .dark .folderhack2 {
-	background: #77838c;
+	background: #484848;
+}
+
+.dark .folderpane h3 {
+    	margin: 0;
```
Zarel: This too, etc etc for a lot of the following lines

Zarel:
No, one of them doesn't!

Zarel:
(The one I skipped)

WallyTheBully:
could it be fixed, by the way, tell me and I'll make a request asap


fixed;

panpawn:
That's a lot of contrast

WallyTheBully:
I agree with you @panpawn, the white variant looks better, @Zarel what's your opinion on that, do you think that the left and right bar should be black when using darkmode?

Zarel:
The battle contents should not change at all between dark and light modes.

WallyTheBully:
11css

a'ight, what about it?

WallyTheBully:
11css
preview with another background

Zarel:
Are those broken in all themes? Just in dark mode because something interferes with it?

WallyTheBully:
those (reflect & light screen, not weathers in general)are broken in dark mode only beacause I assume, something interferes with it

Zarel:
Can you investigate?

WallyTheBully:
yeah apparently, the "weather" class isn't a thing in darkmode 

`.weather {
    position: absolute;
    display: block;
    top: 0;
    left: 0;
    right: 0;
    bottom: 0;
    text-shadow: #FFFFFF 1px 1px 0, #FFFFFF 1px -1px 0, #FFFFFF -1px 1px 0, #FFFFFF -1px -1px 0;
}`

I'd suggest to add the following code to the client CSS, allowing it to be readable 

`.dark .weather {
color: #000;
}`

this is how the white mode displays those screens moves ;



only "light screen" and "reflect" are affected by this "bug" and it should be like this in darkmode since you said ;

> The battle contents should not change at all between dark and light modes.

Zarel:
You're going to need to be clearer than "isn't a thing in darkmode"

WallyTheBully:
yes, the weather CSS class wasn't coded, as a consequence it isn't on the client CSS file, I'd suggest to add it, sorry for the lack of precision in my last comment

Zarel:
It doesn't need to be, the `.weather` CSS class should cover it.

WallyTheBully:
this is true, then we should just add `color: #000;` to the `.weather` class

Zarel:
yes

Zarel:
I believe this has mostly been superseded by other changes, and can be closed now?

Zarel:
Tested, this is fixed in the current version of PS.

PR #944, opened 2017-06-15
Add <pokemonicon> and <itemicon>

bumbadadabum:
I talked most of this through with Zarel, just leaving this here first as a sanity check. Usage of this feature will be added once this is merged.

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -791,9 +791,9 @@ var Tools = {
 			// Implement <pokemonicon> and <itemicon>
 			str = str.replace(/<(pokemon|item)icon>([^<]*)<\/\1icon>/g, function (match, p1, p2) {
 				if (p1 === 'pokemon') {
-					return '<span class="picon" style="display:inline-block;' + Tools.getPokemonIcon(toId(p2)) + '"></span>';
+					return '<span class="picon" style="display:inline-block;' + Tools.getPokemonIcon(p2) + '"></span>';
 				} else if (p1 === 'item') {
-					return '<span class="itemicon" style="display:inline-block;' + Tools.getItemIcon(toId(p2)) + '"></span>';
+					return '<span class="itemicon" style="display:inline-block;' + Tools.getItemIcon(p2) + '"></span>';
 				}
 			});
 
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -786,10 +786,8 @@ var Tools = {
 			return {attribs: attribs};
 		};
 		return function (input) {
-			var str = getString(input);
-
 			// Implement <pokemonicon> and <itemicon>
-			str = str.replace(/<(pokemon|item)icon>([^<]*)<\/\1icon>/g, function (match, p1, p2) {
+			input = getString(input).replace(/<(pokemon|item)icon>([^<]*)<\/\1icon>/g, function (match, p1, p2) {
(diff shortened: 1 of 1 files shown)


js/battledata.js
```diff
@@ -786,7 +786,18 @@ var Tools = {
 			return {attribs: attribs};
 		};
 		return function (input) {
-			return html.sanitizeWithPolicy(getString(input), tagPolicy);
+			var str = getString(input);
```
Zarel: I'd just reuse `input` instead of creating a new variable for it...

bumbadadabum:
Updated.

Slayer95:
This is a job that can be perfectly done with the Caja API :|

Slayer95:
Just whitelist these tag names and return ``{tag, attribs}`` with the rewritten tag from the ``tagPolicy``.

Slayer95:
Actually, that's not the case if we read the inner HTML of the ``<pokemonicon>`` tag rather than its attributes. Is there a reason for doing it that way rather than with attributes?

Zarel:
I told him it was probably possible with the Caja API! He decided to do it this way so he didn't need to learn the Caja API.

Slayer95:
Heh. I can refactor this now. Would you want any attribute name in particular?
I'd suggest ``<psicon pokemon="heracross"></psicon>`` and ``<psicon item="leppaberry"></psicon>``

Zarel:
There's no other reason not to do it with attributes, but it _is_ shorter and more readable currently.

Slayer95:
My first suggestion is longer just for 1 character for Pokémon and shorter by 4 for items.
If you care enough for length so to skip the attribute quotes, then it's shorter for both cases.

Using attributes also improves extensibility as I already said, prevents doing multiple passes over the HTML code, and of course the correcteness sense of just using the HTML sanitizer module for the HTML jobs.

Zarel:
I think you miscounted, it's longer by 1 for Pokémon and longer by 8 for items. But more importantly I think it's more readable/intuitive this way.

And technically, transforming custom tags _is_ a different job from sanitizing. Putting them in different parts of the codebase may be good for readability.

These are all really minor reasons, though. I don't have a strong opinion either way. @bumbadadabum?

Slayer95:
Playing with HTML sanitizing is never intuitive enough.

Before
```
> Tools.sanitizeHTML('<div id="<pokemonicon>heracross</pokemonicon>"</div>')
<"<div id="&lt;pokemonicon&gt;heracross&lt;/pokemonicon&gt;"></div>"
```

After

```
> Tools.sanitizeHTML('<div id="<pokemonicon>heracross</pokemonicon>"</div>')
< "<div id="&lt;span class=" style="display: inline-block ; background: transparent url(&#34;//play.pokemonshowdown.com/sprites/smicons-sheet.png?a1&#34;) no-repeat scroll -400px -510px">"</div>"
```

This commit definitely makes it harder to reason about how we handle HTML.

Slayer95:
cc @kotarou3 @xfix

Zarel:
That's a fair criticism.

bumbadadabum:
@Slayer95 I'm fine with a refactor obviously. I did mention to @Zarel I did it in the easiest, "dirtiest" way I could think of. Your PR does seem more "proper" so yeah that's great.

PR #945, opened 2017-06-16
Refactor custom icons to a single element <psicon> and use Caja

Slayer95:
Since real custom elements still have very limited availability, this is the most proper way to do it,
and results in a more predictable HTML parsing.
Adds support for custom styles and classes.

Example:
``<psicon pokemon="meloetta" class="pixelated" style="opacity:0.4" />``

Alternative to #944

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -712,14 +712,14 @@ var Tools = {
 				throw new Error('sanitizeHTML requires caja');
 			};
 		}
-		// Add <marquee> and <blink> to the whitelist.
-		// See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/marquee
-		// for the list of attributes.
+		// Add <marquee> <blink> <psicon> to the whitelist.
 		$.extend(html4.ELEMENTS, {
 			'marquee': 0,
-			'blink': 0
+			'blink': 0,
+			'psicon': html4.eflags['OPTIONAL_ENDTAG'] | html4.eflags['EMPTY']
 		});
 		$.extend(html4.ATTRIBS, {
+			// See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/marquee
 			'marquee::behavior': 0,
 			'marquee::bgcolor': 0,
 			'marquee::direction': 0,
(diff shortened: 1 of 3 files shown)


Zarel:
Should we make "psicon" a self-closing tag and save ourselves some trouble?

Zarel:
Specifically, instead of `<psicon pokemon="heracross"></psicon>`

How about just `<psicon pokemon="heracross">`?

Like how `` tags work.

Zarel:
Or maybe we could even just use `` tags.

``

Slayer95:
Ok, made it like ``img``.
As for using ``img`` itself, the server would then send non-conforming HTML due to the lack of required ``src`` and ``alt`` attributes. The semantics of ``img`` are also very special due to its being a replaced element in the CSS meaning.
Although in the end the HTML would be rewritten and end up as valid, applications other than the PS client might meet trouble validating user content (room intros, etc), such as xfix 's room intro tool or any other.

bumbadadabum:
Yeah I'd prefer this to be a completely custom element as well. I like the change though.

Zarel:
Also, it should be pixelated by default, right? That should be fixed in CSS if not.

Zarel:
I don't see any code to make it inline-block?

Zarel:
That should probably be the default for picon/itemicon though.

Slayer95:
> Also, it should be pixelated by default, right? That should be fixed in CSS if not.

Yes, it is already. That was just an example.

> I don't see any code to make it inline-block?
> That should probably be the default for picon/itemicon though.

Oops. The default is currently set to ``display: block`` for some reason.
I'll have a look and report back ~

Zarel:
picon/itemicon are currently `display: block` because most of their uses work best with that. But we should probably change that, and just override it for all the uses that aren't inline-block

Slayer95:
Ok.

The only place where the default update affected the layout was in the Pokémon selection top bar in the teambuilder (``.teambar``). All other icons in the teambuilder already had ``display:block`` set due to more complex selectors.

As for Pokémon icons in the battle screen, battle controls and team preview, as well as the Meloettas in the room list the switch to ``inline-block`` didn't seem to affect their layout.

bumbadadabum:
<3

PR #947, opened 2017-06-20
Support custom colors

supersoniko:
Allows side servers to send a dictionary of userids and a color value. Right now we have to handle custom colors through css, which isn't a pretty solution. This would also improve the speed of the css sanitizer!
_Nii Sama_

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -70,6 +70,7 @@ var colorCache = {};
 
 function hashColor(name) {
 	if (colorCache[name]) return colorCache[name];
+	if (Config.serverCustomColors && Config.serverCustomColors[name]) return Config.serverCustomColors[name];
 	var hash;
 	if (window.Config && Config.customcolors && Config.customcolors[name]) {
 		if (Config.customcolors[name].color) {
--- a/js/client.js
+++ b/js/client.js
@@ -1069,13 +1069,13 @@
 
 			var colors = {};
 
-			for (let user in data) {
+			for (var user in data) {
 				var entry = data[user];
 				if (!entry) continue;
 
 				if (isNaN(entry[0]) || isNaN(entry[1]) || isNaN(entry[2])) continue; // should receive an array of 3 numbers
 
-				colors[user] = 'color:hsl(' + Math.round(entry[0]) + ',' + parseInt(entry[1]) + '%,' + parseInt(entry[2]) + '%);';
+				colors[user] = 'color:hsl(' + Math.round(entry[0]) + ',' + parseInt(entry[1], 10) + '%,' + parseInt(entry[2], 10) + '%);';
 			}
 
 			Config.serverCustomColors = colors;
--- a/js/client.js
+++ b/js/client.js
@@ -879,6 +879,16 @@
 				this.parseGroups(tarRow);
 				break;
 
+			case 'customcolors':
+				var nlIndex = data.indexOf('\n');
+				if (nlIndex > 0) {
+					this.receive(data.substr(nlIndex + 1));
+				}
+
+				var tarRow = data.slice(14, nlIndex);
+				this.parseColors(tarRow);
+				break;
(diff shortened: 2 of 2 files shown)


Slayer95:
This is an XSS vector

supersoniko:
I'll add some validation

supersoniko:
is this better @Slayer95?

Slayer95:
Works in that regard (see Travis)

Slayer95:
Since we did custom groups and now custom colors are brought on the table,
I'll take a chance to ref #441

AustinXII:
@Slayer95 that'd be really cool to get implemented :+1

Zarel:
I still don't think #443 is necessary.

Zarel:
I also still don't really want to support custom colors......

supersoniko:
Would improve the css sanitizer's speed and page load on side servers. These bonker big custom css's  on side servers make initial load on side servers rather slow since they're all loading custom colors through css.

Zarel:
I feel like that's a good reason not to have an n^2 color system in the first place, not to refactor it to even further increase server lag.

supersoniko:
I think there's a difference between lag and lag visual by the user, right now users have to wait ~15-20 seconds for the css to load if not cached. I haven't really tested this code with a larger data set, I can try it out tomorrow morning.

sparkychildcharlie:
What if this was done using some JSON file that the client would get instead, similar to how avatars would work.  The server would host "another" nodestatic server file with the data for the custom colours [similar to how the css file is hosted](https://github.com/Zarel/Pokemon-Showdown/blob/b66e395bf67b08750dbc2194cb2ab7cc1007226e/sockets.js#L234), which wouldn't add another line that could potentially have lengths long enough to crash the client

Zarel:
I'm saying you shouldn't have custom colors at all. It's entirely the wrong approach. You should not be sending a big data blob on login at all. Not from the server, not from the CSS file.

Zarel:
If it's 10MB+, all the more reason to _stop doing it_

Zarel:
Maybe the correct solution is to ban servers with custom colors from registration.

panpawn:
Main client probably doesn't have to support this.

If a server wants custom colors, then they can host their own client or whatever.

DeathlyPlays:
@Zarel Custom Colors are very common on side servers, could we not be this hasty to fully remove them?

AustinXII:
@DeathlyPlays the "bucks" system is also very popular and side servers, popularity doesn't mean main should support it

Zarel:
Well, supporting something on the official client is one thing... I can't deny that there's demand. It's just, unboundedly large data blobs are very blah to force down the throat of every user connecting to your server.

Zarel:
Like, seriously? You're going to send 30MB to every user connecting to you? Did you know that Firefox 1.0 was 5MB and that was an entire browser????

Zarel:
I'd be fine if it was something more like, just the people who've talked in the past few hours, and then stream the rest if someone new has a custom color.

Slayer95:
I've had a quick look through Azure's CSS and I have verified that a single CSS3 update can bring its size down from 2 MB to 300 KB.

This is just telling of how little optimization work CSS-based custom colors have (well, to be fair CSS3 support is new). Maybe we should just revert #946, let people optimize their servers and call it a day?

supersoniko:
I've been able to recude my css to from 2 MB to 126 KB with css optimizations, thanks for not encouraring us to continue these bad practises

sparkychildcharlie:
(Oh oops it was 2MB and not 10.... hm i got my facts wrong)

supersoniko:
It was 16MB at the start of June in Azure.. (I did a custom color reset a week ago) Anyway this won't happen again, with these css optimizations we should be able to keep it under 400 KB at all times.

PR #948, opened 2017-06-20
Fix teambuilder showing current gen base stats for gen 6 teams

asgdf:
Would it be a good idea to have some global "current gen" constant so we don't have to hunt down all the outdated gen numbers every time a new one comes out? Where would it go? battledata.js?


urkerab:
`Tools.gen` (or `Dex.gen` if it gets renamed following the server rename) would seem to be a suitable name...

PR #949, opened 2017-06-21
Display correct type of transformed Pokemon in old gens

xfix:
This makes for instance Ditto transformed into Clefairy appear as Normal type instead of Fairy type. The battle with a bug can be seen on http://replay.pokemonshowdown.com/gen1ou-593692946

Pretty sure there is a cleaner way of doing this, I'm not really used to contributing to client, feel free to point out improvements.

BattleTeambuilderTable is checked before use because there is no requirement it's loaded (it's not loaded in replays for instance). The old version didn't work in replays either, so it's probably not a big deal?


xfix:
ignore this, BattleTeambuilderTable doesn't exist in replays, so it would have to check if it even exists to begin with

PR #950, opened 2017-06-22
Parse HTML time

xfix:
This is supposed to allow for locale aware implementation of /roomevents. Timezones are confusing, and can be misread causing some people coming for an event in wrong time. This is client-side part of locale aware /roomevents.

`<time>` element is part of HTML5 and is already parsed by Caja. Reusing already existing element doesn't appear to be a practical issue, as it's pretty much meant to do stuff like that.

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -706,7 +706,7 @@ var Tools = {
 		return str;
 	},
 
-	htmlToSafeDOM: (function () {
+	sanitizeHTML: (function () {
 		if (!('html4' in window)) {
 			return function () {
 				throw new Error('sanitizeHTML requires caja');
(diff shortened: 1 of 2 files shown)


Zarel:
I'll wait for someone else to review. @Slayer95?

Slayer95:
Could this be possibly done with Caja+CSS?

Say, parse datetime attribute and set a data attribute with the equivalent locale-aware text. Then, add a time:empty::before rule to display the the text from the data slot.

xfix:
CSS can be used, but then it wouldn't be copyable, would it be?

Slayer95:
Fair enough. Apparently pseudoelements can't be selected even after tinkering with ``user-select``.
I am not thrilled by this implementation, but this is a significant nice thing to have, so :+1:

js/battledata.js
```diff
@@ -826,11 +826,66 @@ var Tools = {
...
+				var parsedTime = new Date(parts[1] + 'T' + parts[2] + (parts[3] || 'Z').toUpperCase());
+				// Very old (pre-ES5) web browsers may be incapable of parsing ISO 8601
+				// dates. In such a case, gracefully continue without replacing the date
+				// format.
+				if (!parsedTime.getTime()) continue;
+
+				var formattedTime;
+				// Try using Intl API if it exists
+				if (window.Intl) {
+					formattedTime = new Intl.DateTimeFormat(undefined, {month: 'long', day: 'numeric', hour: 'numeric', minute: 'numeric'}).format(parsedTime);
```
Slayer95: Apparently some mobile browsers such as the default Android browser support some ``Intl`` APIs but not ``DateTimeFormat``.

js/battledata.js
```diff
@@ -706,7 +706,7 @@ var Tools = {
 		return str;
 	},
 
-	sanitizeHTML: (function () {
+	sanitizeHTMLReturningTree: (function () {
```
Slayer95: Very verbose. Does ``.htmlToSafeDOM()`` sound good to you?

xfix:
> I am not thrilled by this implementation

I'm not either. I just cannot think of a better way.

Slayer95:
Ideally this would be handled by browsers automatically if the element ``textContent`` matches its ``datetime`` attribute. At this point, it might require a new attribute such as ``autolocalize="true"`` for compat...

xfix:
What is the state of this pull request? Slayer95 approved it some time ago.

Zarel:
I think DOM manipulation is horrible for performance. Can we use a regex?

Zarel:
(And text search/replace)

xfix:
I agree, DOM construction cost is terrible [after I decided to benchmark it](https://jsperf.com/html-string-manipulation-performance). Could be managed by using htmlToSafeDOM directly, instead of innerHTML assignment.

js/battledata.js
```diff
@@ -826,11 +826,66 @@ var Tools = {
...
+					formattedTime = parsedTime.toLocaleString();
+				}
+				timeElement.textContent = formattedTime;
+			}
+		};
 		return function (input) {
-			return html.sanitizeWithPolicy(getString(input), tagPolicy);
+			var sanitized = html.sanitizeWithPolicy(getString(input), tagPolicy);
+			// Unfortunately Caja doesn't support modifying tags in case of <time> tag,
+			// so Caja's output is converted into DOM just to modify it. Element name
```
Zarel: What I'm saying is, instead of converting to DOM, why not do a text find/replace of the `<time>` element before passing to Caja?

Slayer95: In any case, it should be *after* to prevent the sort of trouble I mentioned in the ``pokemonicon`` implementation (and even worse I'd say). Caja automatically escapes entities which aren't parsed with their HTML special meaning.

Slayer95: Of course, utmost care must be taken not to end up with unsafe HTML...

xfix: I would prefer to avoid accidentally introducing security issues by doing transformations after Caja processes. Doing regex replace before Caja will cause issues, in fact security ones allowing for stuff like room intros tracking user's locale and timezone. Doing regex replace after Caja is possible, albeit on risky side, considering it involves changing safe HTML in weird ways. Probably safe, but I'm not willing to risk it myself.

The reason why I directly work on DOM is that this is very unlikely to cause security issues, as I don't use `innerHTML` in DOM code, instead operate directly on trees.

Zarel: I don't really think doing a regex replace before Caja can cause issues?

Slayer95: > room intros tracking user's locale and timezone

:thinking: This can already be done by just inserting an image hosted in an owned third-party server, albeit only as far as HTTP request headers provide. There is other info that the headers would provide, however, which makes this the least of our concerns as well as our users'.

As long as there is no PS-especific PII we are inserting with regex replace, we probably don't have to worry about privacy vulnerabilities.

## Now, comparing the rewrite before vs rewrite after approaches
*Rewrite after:*
As long as the rewrite is very especific so as not to change the grammar, there should not be issues, for there is not context where a substring ``<time datetime="2017-02-14">2017-02-14 08:00</time>`` is sintactically or semantically different from ``<time datetime="2017-02-14">2017-02-14 09:30</time>``, so none of them could be safer than the other.

**Limitations**: We could only cover a limited set of attributes for ``time``, and with a rigid grammar. Specifically, stay away from ``style``.

*Rewrite before*
The oddities resulting from regex-replacing "tricky" HTML, such as ``<div title="<time>">`` are already documented in the [pokemonicon discussion](https://github.com/Zarel/Pokemon-Showdown-Client/pull/944#issuecomment-308948392).

In particular, a regex replace without backing from a SAX parser *will* change the grammar and semantics of the final HTML output in unpredictable ways. Of course, we are still left with the guarantees provided by Caja, since Caja can process this wild unknown HTML into a sanitized output.

**Limitations**: A best-effort approach can be followed to cover all sorts of global and ``time``-specific attributes. However, bugs arising from tricky HTML would be unavoidable and their prevalence will correlate with the grammar we want to cover. Likely, stay away from ``style``.

Slayer95: ### Comparing the rewrite before vs rewrite after approaches
*Rewrite after:*
As long as the rewrite is very especific so as not to change the grammar, there should not be issues, for there is not context where a substring ``<time datetime="2017-02-14">2017-02-14 08:00</time>`` is sintactically or semantically different from ``<time datetime="2017-02-14">2017-02-14 09:30</time>``, so none of them could be safer than the other.

**Limitations**: We could only cover a limited set of attributes for ``time``, and with a rigid grammar. Specifically, stay away from ``style``.

*Rewrite before*
The oddities resulting from regex-replacing "tricky" HTML, such as ``<div title="<time>">`` are already documented in the [pokemonicon discussion](https://github.com/Zarel/Pokemon-Showdown-Client/pull/944#issuecomment-308948392).

In particular, a regex replace without backing from a SAX parser *will* change the grammar and semantics of the final HTML output in unpredictable ways. Of course, we are still left with the guarantees provided by Caja, since Caja can process this wild unknown HTML into a sanitized output.

**Limitations**: A best-effort approach can be followed to cover all sorts of global and ``time``-specific attributes. However, bugs arising from tricky HTML would be unavoidable and their prevalence will correlate with the grammar we want to cover. Likely, stay away from ``style``.

xfix:
I don't like the regex solution myself considering quirks like `2015-05-05T05:05:05</time>">`, but I would rather see this in the client, so whatever.

PR #952, opened 2017-06-23
fix reflect and light screen battle text in darkmode

WallyTheBully:
screenshots
before: http://i.imgur.com/OAcyAW2.png
after: http://i.imgur.com/SNNr8oS.png

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -3210,7 +3210,7 @@ a.ilink.yours {
 .dark .battle-log-add,
 .dark .battle-controls {
 	background: #444444;
-	color: #DDD;
+	color: #000;
 }
 .dark .userlist-minimized {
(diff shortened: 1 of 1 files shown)


style/client.css
```diff
@@ -3210,7 +3210,7 @@ a.ilink.yours {
 .dark .battle-log-add,
 .dark .battle-controls {
 	background: #444444;
-	color: #DDD;
+	color: #000;
```
Zarel: Why should this affect that?

WallyTheBully:
wrong class mb, the "weather" class will cover it, allowing this text to be black as it should be in both light and dark modes

Zarel:
Um, we already have a weather class...

PR #953, opened 2017-06-23
CSS – update teambuilder folderlist in darkmode

WallyTheBully:
some screenshots
https://prnt.sc/fnbfn3
hover:
http://prntscr.com/fnbhfy
http://prntscr.com/fnbjam

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -3282,26 +3282,24 @@ a.ilink.yours {
 /* semi-generic */
 
 .dark .select, .dark .team {
-	border-color: #888888;
-	box-shadow: inset 0 1px 0 #FFFFFF;
-	background: linear-gradient(to bottom,  #ccd4dc,  #96a3b0);
+	border-color: #000000;
+	box-shadow: none;
+	background: linear-gradient(to bottom,rgba(227, 239, 251, 0.7),rgba(118, 128, 140, 0.7));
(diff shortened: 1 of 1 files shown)


Zarel:
The team hover is a bit subtle. I like the rest of this, though.

style/client.css
```diff
@@ -3282,26 +3282,24 @@ a.ilink.yours {
...
+	box-shadow: none;
+	background: linear-gradient(to bottom,rgba(227, 239, 251, 0.7),rgba(118, 128, 140, 0.7));
 	color: black;
 }
 .dark .select:hover, .dark .team.team-hover {
-	border-color: #555555;
-	background: #FFFFFF;
-	background: linear-gradient(to bottom,  #edf2f8,  #d7e3ec);
+	border-color: #000000;
+	background: linear-gradient(to bottom,rgba(255, 255, 255, 0.85),rgba(161, 175, 189, 0.85));
```
Zarel: Space after all commas.

Also, don't remove the `background: #FFFFFF;` - we still want to support backgrounds on older browsers that don't have background stuff.

Also, you should probably set a default color here. Rule of thumb is to set text color wherever setting background color, just to be safe.

Zarel:
Looks good, thanks!

PR #955, opened 2017-06-24
Update Font Awesome to 4.7.0

HoeenCoder:
I'm not sure if I should remove the `style/font/` directory that was created as a work around for a cloudflare cache issue. (See: https://github.com/Zarel/Pokemon-Showdown-Client/commit/76bbf1d46985acf781f5eaa2e0e98b4bbc8f8301 ) It seems to work while pointed to `style/fonts/` currently but that could just be because i'm using testclient. I currently have it updated but not used at all. I want to confirm its ok to delete that directory.


Zarel:
Yes, you should remove the old `style/font/` directory.

Zarel:
On second thought... caches are tough. Maybe we should keep them for compatibility.........

HoeenCoder:
I'm not sure what having it would do atm @Zarel since nothing is reading the `/style/font/` directory atm.

Zarel:
@HoeenCoder, that's where you're wrong: cached copies of `style/font-awesome.css` are reading from there.

HoeenCoder:
Good point, I didn't think of that @Zarel

Issue #954, opened 2017-06-24
Improve accessibility for screen readers


Neurrone:
Showdown is pretty popular among [screen reader](https://www.nomensa.com/blog/2005/what-screen-reader) users. However, some parts of the UI are problematic, which motivated someone to create a [guide on how to work around some of them](http://forum.audiogames.net/viewtopic.php?id=18547). I plan to try fixing the issues identified in that guide.

* [x] Use role="log" to enable automatic reding of incoming battle log text.
* [] Do the same for chat room screens.
* [] Add labels to unlabelled controls (e.g alt attribute or aria-label).
* [] There are 2 edit boxes in the DOM when on the battle and chat room screens. One of them looks like its not meant for use, so should be hidden.
* [] Add WAI-ARIA landmarks to divide the UI into logical regions for faster navigation. E.g, to quickly check the health of the opponent.
* [] Make custom slider controls used for volume adjustment keyboard operable.

Zarel:
The weird part is that the buttons _are_ labeled in that they have `title` attributes "Sound" and "Options". But I guess `title` attributes are not screen-reader compatible. I'll look into `aria-label`.

Zarel:
The second edit box is designed for edit box scaling, and _is_ hidden. How should it be hidden from screen-readers?

Zarel:
I haven't been able to find a slider library that actually works for PS's use-case and is accessible, unfortunately. Best I can do is a text box, which I should probably be doing anyway...

Zarel:
Okay, I discovered `aria-hidden`. Do you use a screen-reader? Can you help with this?

Neurrone:
@Zarel yup aria-hidden does the trick. Originally thought about fixing the entire issue in a single PR, but I think the improvements I've made to the battle interface improves the experience so much that it should be done first. Lets continue discussion in a PR.

Yup, I use a screen reader.

Zarel:
It looks like PS's menus are weird for screen-readers – the article you linked to mentions "opens a menu at the bottom". How do I make that menu easier to access?

Neurrone:
Which menus do you mean? The main menu (with home etc), or the menus that appear (e.g to choose a format)?

Neurrone:
@Zarel is there an easy way for me to be signed in in 2 places simultaneously so I can leave the battle window open without havinig to actually conduct a battle?

Zarel:
[...]

PR #956, opened 2017-06-24
Accessibility improvements to the battle UI

Neurrone:
Fixes battle UI-related issues in #954 

* [x] Use role="log" to enable automatic reding of incoming battle log text.
* [ ] There are 2 edit boxes in the DOM when on the battle and chat room screens. One of them is not meant for use, so should be marked aria-hidden.
* [x] Add WAI-ARIA landmarks to divide the UI into logical regions for faster navigation. E.g, to quickly check the health of the opponent.
* [ ] Make information conveyed by tooltips read out automatically, as well as being triggered when gaining keyboard focus.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2479,7 +2479,7 @@ var Battle = (function () {
 		} else {
 			this.logFrameElem.html('<div class="battle-options"></div>');
 			this.optionsElem = this.logFrameElem.children().last();
-			this.logFrameElem.append('<div class="inner" role="log" aria-label="Battle Log"></div>');
+			this.logFrameElem.append('<div class="inner" role="log"></div>');
 			this.logElem = this.logFrameElem.children().last();
 			this.logFrameElem.append('<div class="inner-preempt"></div>');
 			this.logPreemptElem = this.logFrameElem.children().last();
--- a/js/battle.js
+++ b/js/battle.js
@@ -2479,7 +2479,7 @@ var Battle = (function () {
 		} else {
 			this.logFrameElem.html('<div class="battle-options"></div>');
 			this.optionsElem = this.logFrameElem.children().last();
-			this.logFrameElem.append('<div class="inner"></div>');
+			this.logFrameElem.append('<div class="inner" role="log"></div>');
 			this.logElem = this.logFrameElem.children().last();
 			this.logFrameElem.append('<div class="inner-preempt"></div>');
 			this.logPreemptElem = this.logFrameElem.children().last();
--- a/js/battle.js
+++ b/js/battle.js
@@ -2479,7 +2479,7 @@ var Battle = (function () {
 		} else {
 			this.logFrameElem.html('<div class="battle-options"></div>');
 			this.optionsElem = this.logFrameElem.children().last();
-			this.logFrameElem.append('<div class="inner" role="log"></div>');
+			this.logFrameElem.append('<div class="inner" role="log" title="Battle Log"></div>');
 			this.logElem = this.logFrameElem.children().last();
 			this.logFrameElem.append('<div class="inner-preempt"></div>');
 			this.logPreemptElem = this.logFrameElem.children().last();
@@ -2511,16 +2511,16 @@ var Battle = (function () {
 		this.spriteElem.append('<div></div>');
 		this.spriteElems[0] = this.spriteElem.children().last();
 
-		this.elem.append('<div></div>');
+		this.elem.append('<div role="complementary" title="Active Pokemon"></div>');
 		this.statElem = this.elem.children().last();
 
 		this.elem.append('<div></div>');
 		this.fxElem = this.elem.children().last();
 
-		this.elem.append('<div class="leftbar"></div>');
+		this.elem.append('<div class="leftbar" role="complementary" title="Your Team"></div>');
 		this.leftbarElem = this.elem.children().last();
 
-		this.elem.append('<div class="rightbar"></div>');
+		this.elem.append('<div class="rightbar" role="complementary" title="Opponent\'s Team"></div>');
 		this.rightbarElem = this.elem.children().last();
 
 		this.elem.append('<div></div>');
--- a/js/battle.js
+++ b/js/battle.js
@@ -2479,7 +2479,7 @@ var Battle = (function () {
 		} else {
 			this.logFrameElem.html('<div class="battle-options"></div>');
 			this.optionsElem = this.logFrameElem.children().last();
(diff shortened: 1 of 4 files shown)


js/battle.js
```diff
@@ -2479,7 +2479,7 @@ var Battle = (function () {
 		} else {
 			this.logFrameElem.html('<div class="battle-options"></div>');
 			this.optionsElem = this.logFrameElem.children().last();
-			this.logFrameElem.append('<div class="inner"></div>');
+			this.logFrameElem.append('<div class="inner" role="log" title="Battle Log"></div>');
```
Zarel: `title` is a bad attribute; I don't want that to appear as a tooltip for sighted readers. Does there exist an attribute like `aria-title`?

Neurrone: Yup, aria-label.

Zarel: Okay, can you change all these `title`s to `aria-label` for me?

js/client-battle.js
```diff
@@ -11,7 +11,7 @@
 
 			this.isSideRoom = Tools.prefs('rightpanelbattles');
 
-			this.$el.addClass('ps-room-opaque').html('<div class="battle">Battle is here</div><div class="foehint"></div><div class="battle-log"></div><div class="battle-log-add">Connecting...</div><div class="battle-controls"></div><button class="battle-chat-toggle button" name="showChat"><i class="fa fa-caret-left"></i> Chat</button>');
+			this.$el.addClass('ps-room-opaque').attr('role', 'main').html('<div class="battle">Battle is here</div><div class="foehint"></div><div class="battle-log" role="complementary" title="Battle Log"></div><div class="battle-log-add">Connecting...</div><div class="battle-controls" role="complementary" title="Battle Controls"></div><button class="battle-chat-toggle button" name="showChat"><i class="fa fa-caret-left"></i> Chat</button>');
```
Zarel: Is it fine to have multiple elements with `role="main"`? It seems like every room should have that role.

Zarel: Also, `role="complementary" title="Battle Log"` is probably unnecessary here. It might conflict with `role="log" title="Battle Log"`. At least one of those should have a different label title?

Neurrone:
The `role="complementary" title="Battle Log"` puts a "bookmark" called battle log that can be navigated to very quickly. The `role="log"` is different - it allows automatic reading of new text as it appears in that element's children, but does not add a landmark in itself.

js/battle.js
Neurrone: On it.

Zarel:
I still feel weird about two different things having the exact same ARIA label. Even something like "Battle log contents" might be better for the `role="log"` element?

js/client-battle.js
```diff
@@ -11,7 +11,7 @@
 
 			this.isSideRoom = Tools.prefs('rightpanelbattles');
 
-			this.$el.addClass('ps-room-opaque').html('<div class="battle">Battle is here</div><div class="foehint"></div><div class="battle-log"></div><div class="battle-log-add">Connecting...</div><div class="battle-controls"></div><button class="battle-chat-toggle button" name="showChat"><i class="fa fa-caret-left"></i> Chat</button>');
+			this.$el.addClass('ps-room-opaque').attr('role', 'main').html('<div class="battle">Battle is here</div><div class="foehint"></div><div class="battle-log" role="complementary" aria-label="Battle Log"></div><div class="battle-log-add">Connecting...</div><div class="battle-controls" role="complementary" aria-label="Battle Controls"></div><button class="battle-chat-toggle button" name="showChat"><i class="fa fa-caret-left"></i> Chat</button>');
```
Zarel: `role="main"` is unfortunately a bad fit for battles, because you can have more than one battle open at a time.

The spec says:

> Within any document or application, the author SHOULD mark no more than one element with the main role.

Zarel: Perhaps `role="application"`?

Neurrone:
@Zarel my bad i didnt realize I had duplicate labels.

js/client-battle.js
Neurrone: I'll remove the main role. `role="application"` should only be used sparingly, and only if the webpage needs to intercept keypresses.

Zarel: Maybe `role="document"`, then?

Neurrone: Will test `role="document"` in 10 hours after some sleep, my hunch is that some other markup is a better fit.

Neurrone: Think the best solution here is to edit the container so that its label is constructed from the participants of the battle or chat. For e.g, `aria-label="Battle with foo"`, `aria-label="Chat with foo", and add an appropriate role.

What'd be the easiest way to do this?

Neurrone:
Is there a reason why the buttons have `tabindex="-1"`? It makes them not participate in sequential navigation, but it looks like they should.

Neurrone:
@Zarel why're there 6 children under the `#foehint` div? The only thing that differs among them is that they use different `thing` parameters when calling functions, but couldn't figure out `your0`, ... `your2` does.

Neurrone:
Just tried setting `tabindex="0"` so they participate in sequential keyboard navigation so that onfocus event would trigger the showing of the tooltip. It fires onfocus, but not read. there are also 6 divs in the tab order now that does the same thing.

Zarel:
Which buttons? Buttons with `tabindex="-1"` are generally buttons that don't do anything important and aren't very button-like.

Zarel:
The `#foehint` div has six children to correspond to the maximum of 6 Pokémon on the field at a time (in a Triples battle). `your0` is the opponent's first (rightmost) Pokémon, `your1` is the opponent's second Pokémon (second from the right), etc.

Zarel:
Oh, I see, the buttons to close the News box. The ability to close the News box isn't very important, is the main reason why the buttons aren't selectable. I'd rather enable some way to focus the News box so it can be closed with `Esc` the way PM conversations can be closed.

Zarel:
I'm merging this. Future changes should go in a new pull request. Thanks for the contribution!

Neurrone:
Does it immediately go live on the main server?

Neurrone:
Would setting `tabindex=0` interfere with the scape thing somehow? Using tab navigation is fairly common for screen readers.

Zarel:
It goes live following a sync, which I've just done.

Zarel:
`tabindex="0"` interferes in the sense that it makes tab navigation harder.

Instead of going from PM to PM, you go from PM to minimize button to close button to PM, so it takes 3x as much time to go between PMs.

PR #958, opened 2017-06-25
Don't require boundary before http in links

xfix:
It seems like forgetting spaces before links like `examplehttp://example.com` is common enough, and considering it's rather clear what this is supposed to do, no reason to force typing the link again for it to work. This only affects links having http, https or www at the start, no other links.

### DIFF
--- a/js/battledata.js
+++ b/js/battledata.js
@@ -322,12 +322,11 @@ var baseSpeciesChart = [
 var domainRegex = '[a-z0-9\\-]+(?:[.][a-z0-9\\-]+)*';
 var parenthesisRegex = '[(](?:[^\\s()<>&]|&amp;)*[)]';
 var linkRegex = new RegExp(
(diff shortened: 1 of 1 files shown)


asgdf:
Would this correctly handle domains that happen to contain `www.` elsewhere than at the beginning of their name?

xfix:
Sounds like a weird edge case. That said, regex matches try to match from the left, so hello.www.example.com will work as expected.

Zarel:
Well, in Japanese, "wwwwwww" means roughly what "lolololol" means in English. I don't think leaving off the space is common when a URL starts with "www".

xfix:
i don't think wwwwwwwwwwwwwww is likely to be followed by a period, but point accepted

Zarel:
Can you also apply this to the server parser in `chat.js` (which will replace the client parser soon)?

PR #959, opened 2017-06-27
Don't boost PP of Trump Card

xfix:


xfix:
Will add unit test after testing for client gets accepted.

PR #965, opened 2017-07-06
Test if Trump Card has 5 PP

xfix:
This I believe was a regression that was fixed by 2fac7642f63.

### DIFF
--- a/test/test.ts
+++ b/test/test.ts
@@ -50,6 +50,7 @@ test('Trump Card has 5 PP', async t => {
         .click('[value="teambuilder"]')
         .click('[name="newTop"]')
         .click('[name="addPokemon"]')
+        // Eevee (e e v), Trump Card (t c)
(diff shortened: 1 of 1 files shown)


test/test.ts
```diff
@@ -44,3 +44,12 @@ test('Teambuilder shows level 50 for VGC', async t => {
         .click('.setdetails')
         .expect(Selector('[name="level"]').value).eql('50');
 });
+
+test('Trump Card has 5 PP', async t => {
+    await t.click('[name="close"]')
+        .click('[value="teambuilder"]')
+        .click('[name="newTop"]')
+        .click('[name="addPokemon"]')
+        .pressKey('e e v enter enter enter t c')
```
Slayer95: Sorry, what's this doing? Adding a Trump Card Eeevee?
Should probably explain it in a comment.

PR #966, opened 2017-07-06
Add tie requests

HoeenCoder:
In collaboration with @panpawn https://github.com/Zarel/Pokemon-Showdown/pull/3665

Adds a tie request button next to the timer button in battle. The button will show a countdown when the foe has requested a tie, and the button will be disabled when you are requesting a tie, or if you were the last user to request a tie. I've sent panpawn code changes for the server side PR that will handle these changes.  Currently the tie button has the same coloration as the timer button, i'm interested in changing it from red to a different color, any suggestions?

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -250,7 +250,7 @@
 					this.$controls.html('<p><button class="button" name="instantReplay"><i class="fa fa-undo"></i><br />First turn</button> <button class="button" name="rewindTurn"><i class="fa fa-step-backward"></i><br />Last turn</button><button class="button" name="skipTurn"><i class="fa fa-step-forward"></i><br />Skip turn</button> <button class="button" name="goToEnd"><i class="fa fa-fast-forward"></i><br />Skip to end</button></p>');
 				} else {
 					// is a player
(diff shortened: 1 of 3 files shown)


HoeenCoder:
Panpawn mentioned that I didn't do this correctly. (Ties should expire on the next turn, not after X seconds). I'll redo this and update when done.

js/client-battle.js
```diff
@@ -446,6 +447,44 @@
...
+				var secondsLeft = this.battle.requestingTie.timer;
+				if (secondsLeft <= 10 && ticking) {
+					ticking = ' tiebutton-critical';
+				}
+				var minutesLeft = Math.floor(secondsLeft / 60);
+				secondsLeft -= minutesLeft * 60;
+				time = '' + minutesLeft + ':' + (secondsLeft < 10 ? '0' : '') + secondsLeft;
+			}
+			console.log(ticking);
+			console.log(time);
```
panpawn: Did you mean to leave these in?

HoeenCoder: Nope, ill remove those too.

panpawn:
Also, a question both Hoeen and I have is: how do we make it so that the server sends something per-format to the client to know if a tier supports ties or not?

I assume it would be kind of like how some tiers show a person's team before the battle starts?

HoeenCoder:
**Do Not Merge**, Server side code is not ready yet so this is untested.

PR #968, opened 2017-07-10
Add protocols for RoomGames in battle rooms

panpawn:
WIP.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -6571,11 +6571,11 @@ var Battle = (function () {
 			break;
 		case 'fieldhtml':
 			var $element = $('.battle');
-			$element.html(args.slice(1, args.length));
+			$elements.html(Tools.sanitizeHTML(args.slice(1).join('|')));
 			break;
 		case 'controlshtml':
 			var $element = $('.battle-controls');
-			$element.html(args.slice(1, args.length));
+			$elements.html(Tools.sanitizeHTML(args.slice(1).join('|')));
 			break;
 		default:
 			this.logConsole('unknown command: ' + args[0]);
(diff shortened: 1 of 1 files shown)


Slayer95:
XSS alert ಠ_ಠ

panpawn:
``Tools.sanitizeHTML`` fixes that right

Zarel:
We should probably actually parse args differently depending on the first thingy.

panpawn:
?

Zarel:
I'll show you what I mean in the code

Zarel:
See 76690a5001f8b2a7092

panpawn:
... so, should I just use ``args[1]`` here instead of ``args.slice(1).join('|')``?

Wouldn't that make it so that if a game contains ``|`` in it's HTML, that it would just stop there?

Zarel:
...see https://github.com/Zarel/Pokemon-Showdown-Client/commit/76690a5001f8b2a7092938af7899c672f5a3c6d1#diff-b502a7533835db1d325f945bc2b7fc08R6588

panpawn:
After thinking about this more, this kind of seems more like a hack than anything else.

Then again, doing it how I don't think it'd be a hack would be repeating a lot of code that's used for battles, so I'm really not sure what to do.

Is this approach to this a good idea, or should it probably be a separate thing similar to battle rooms?

Zarel:
Yes, it's a hack, but do you have a better idea?

Zarel:
@panpawn are you just going to let this stagnate? This is important!

panpawn:
This was incomplete - iirc, it didn't handle clearing the weather background

panpawn:
Yeah, there's also the ``Join Battle`` button that has to be taken care of, too.

There should probably be a ``|initializegame|`` protocol or something that the server sends that initially clears all of this stuff.

PR #970, opened 2017-07-14
Fix remaining timer not being displayed properly

asgdf:
https://github.com/Zarel/Pokemon-Showdown-Client/commit/76690a5001f8b2a7092938af7899c672f5a3c6d1 changed argument parsing behavior so that `args` will no longer get shifted. Due to this, `args[0]` in timer messages is no longer "You have x seconds left", but "inactive".


Zarel:
Good catch!

PR #973, opened 2017-07-21
Add CAP LC as a tier

DeathlyPlays:


DeathlyPlays:
In tandem with https://github.com/Zarel/Pokemon-Showdown/pull/3813

PR #976, opened 2017-07-28
Allow export of the 'uncategorized' format folder

asgdf:
The "uncategorized" folder is currently named "gen7", but uncategorized teams don't have any format, so the current approach wasn't finding the teams correctly.

### DIFF
--- a/test/test.ts
+++ b/test/test.ts
@@ -15,7 +15,7 @@ fixture('Pokemon Showdown')
     })
     .beforeEach(t => t.maximizeWindow());
 
-const expectedTeam = `=== Untitled 1 ===
+const expectedTeam = `=== [gen7] Untitled 1 ===
 
 Abomasnow-Mega @ Abomasite\x20\x20
 Ability: Snow Warning`;
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -706,11 +706,11 @@
 			this.edit(0);
 		},
 		createTeam: function () {
-			var format = this.curFolder;
+			var format = this.curFolder || 'gen7';
 			var folder = '';
 			if (format && format.charAt(format.length - 1) === '/') {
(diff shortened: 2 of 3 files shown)


Zarel:
Looks like a weird special-case?

I actually think I mainly want to add "(uncategorized)" folders for each gen, with gen 7's uncategorized being the default for "no folder".

asgdf:
Okay, but then how should the uncategorized teams of each gen be told apart? By giving them a format like "gen6"? Or having special "gen6uncategorized" (etc.) formats?

Zarel:
Yes, like `gen6`.

asgdf:
Okay, different approach now: all newly created teams without a specified format will be `gen7`. This also fixes the original issue, because now the format name will match the format folder name.
The "isGenericFormat" part is so that [it doesn't literally say 'gen7' in the format selection](https://prnt.sc/g2jh43) instead of "please select a format".

asgdf:
Hm, old uncategoized teams would need to have their format updated though.

Zarel:
Old uncategorized teams should also be gen7.

asgdf:
Okay, now teams without a specified format will be set to `gen7` both when importing and when unpacking, which should update all existing teams.

Zarel:
Hm, I'm trying to remember if this was my original plan. I think so.

asgdf:
Does that mean this is good to go?

test/test.ts
```diff
@@ -15,7 +15,7 @@ fixture('Pokemon Showdown')
     })
     .beforeEach(t => t.maximizeWindow());
 
-const expectedTeam = `=== Untitled 1 ===
+const expectedTeam = `=== [gen7] Untitled 1 ===
```
Zarel: I'm probably most undecided about this part, the part where the export lists the gen. I think this is probably good.

PR #977, opened 2017-07-30
Add refresh button to Ladders tab

Ridaz:


js/client-ladder.js
```diff
@@ -79,6 +79,9 @@
 		selectFormat: function (format) {
 			this.curFormat = format;
 			this.update();
+		},
+		refresh: function () {
+			this.update();
```
Zarel: Fade the refresh button! Prevent double-clicks!

js/client-ladder.js
```diff
@@ -79,6 +79,11 @@
 		selectFormat: function (format) {
 			this.curFormat = format;
 			this.update();
+		},
+		refresh: function () {
+			this.$refreshButton = this.$('button[name=refresh]');
+			this.$refreshButton.fadeOut();
```
Zarel: There is no reason for this to be two lines.

Also `fadeOut`? Ugh no that's horrible. Disable it.

PR #978, opened 2017-08-04
Set the item when a Z move is used

urkerab:
Status moves are a bit of a kludge but I don't have any better ideas.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -3374,6 +3374,14 @@ var Battle = (function () {
 					this.message(pokemon.getName() + ' used <strong>' + move.name + '</strong>!');
 				}
 				if (!fromeffect.id || fromeffect.id === 'pursuit') {
+					if (move.isZ) {
+						pokemon.item = move.isZ;
+					} else if (move.name.slice(0, 2) === 'Z-') {
+						move = Tools.getMove(move.name.slice(2));
+						for (var item in window.BattleItems) {
+							if (BattleItems[item].zMoveType === move.type) pokemon.item = item;
+						}
+					}
 					var pp = (target && target.side !== pokemon.side && toId(target.ability) === 'pressure' ? 2 : 1);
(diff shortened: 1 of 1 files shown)


js/battle.js
```diff
@@ -3374,6 +3374,15 @@ var Battle = (function () {
 					this.message(pokemon.getName() + ' used <strong>' + move.name + '</strong>!');
 				}
 				if (!fromeffect.id || fromeffect.id === 'pursuit') {
+					if (move.name.slice(0, 2) === 'Z-') {
+						move = Tools.getMove(move.name.slice(2));
+						for (item in window.BattleItems) {
```
ascriptmaster: `for (let item in window.BattleItems)` should fix the build error

Slayer95: ``var``!

js/battle.js
```diff
@@ -3374,6 +3374,15 @@ var Battle = (function () {
...
+							if (BattleItems[item].zMoveType === move.type) pokemon.item = item;
+						}
+					}
+					if (move.isZ) {
```
Zarel: This should probably be first

js/battle.js
```diff
@@ -3374,6 +3374,15 @@ var Battle = (function () {
 					this.message(pokemon.getName() + ' used <strong>' + move.name + '</strong>!');
 				}
 				if (!fromeffect.id || fromeffect.id === 'pursuit') {
+					if (move.name.slice(0, 2) === 'Z-') {
```
Zarel: And then this would be an `else if`

js/battle.js
urkerab: (An earlier iteration set `isZ` to the appropriate item for the status move.)

urkerab:
> Status moves are a bit of a kludge but I don't have any better ideas.

`kwargs.anim` has the true move in that case; would that be less of a kludge?

PR #980, opened 2017-08-07
Tooltips: fix wrong base power being displayed for Extreme Evoboost

asgdf:
This way, it's made sure that `basePower` actually reflects the power of the Z move (if applicable), not just always the base move.
Calling `this.getMoveBasePower(move, pokemon, activeTarget) || basePower` is usually enough for all moves, except for Extreme Evoboost, because its BP is both 0 and different from its base move's BP, so Last Resort's 140 were incorrectly shown.

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -203,7 +203,6 @@ var BattleTooltips = (function () {
 
 	BattleTooltips.prototype.showMoveTooltip = function (move, isZ) {
 		var text = '';
-		var basePower = move.basePower;
 		var basePowerText = '';
 		var additionalInfo = '';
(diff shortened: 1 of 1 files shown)


Zarel:
Do you know why the test is failing? It doesn't seem like you'd change anything that'd affect tests.

asgdf:
Oh wow that's a new one
The test is failing because Aboma-Mega dropped to UU, and because the pokemon selection buttons are loaded as needed when scrolling down, the button to select it doesn't exist
I suppose we could just pick a different pokemon for the test, like setting the format to Ubers and then picking Arceus, that way there's a very low chance of it breaking again

xfix:
Restarted the build, now it passes

PR #981, opened 2017-08-14
Fix weird behavior of client commands in multiline inputs

asgdf:
Almsot every client command is parsed incorrectly in a multiline context currently, by having it go through to the server or by having all the remaining lines being interpreted as the argument of a single command.
The biggest issue with this was probably that a `/nick [name]` followed by a newline could lead to the login token being posted in the chat because the newline is happily passed along to the /trn command, splitting it in the process. At least it was usually blocked due to being too long of a message, unless you can bypassall.

Both Array#map and Array#filter are used elsewhere in the client so I assumed they're fine to use. The other half of the client uses the lodash equivalents though.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -93,7 +93,8 @@
 			}
 			this.tabComplete.reset();
 			this.chatHistory.push(text);
-			text = this.parseCommand(text);
+			// make sure commands in multiline inputs are handled correctly
+			text = text.split('\n').map(this.parseCommand, this).filter(function (line) {return (line !== false);}).join('\n');
 			if (this.battle && this.battle.ignoreSpects && app.user.get('userid') !== this.battle.p1.id && app.user.get('userid') !== this.battle.p2.id) {
 				this.add("You can't chat in this battle as you're currently ignoring spectators");
 			} else if (text.length > 80000) {


Slayer95:
Looks like this breaks template strings in multiline eval.

Maybe consider sanitizing the target name passed to /nick instead? No reason to pass newlines around in the rename process, since that's a forbidden name character.

asgdf:
That would fix that particular issue with /nick, but won't make client commands work correctly. `/ignore test <newline> /unignore test2` would still be intrepreted as `/ignore testunignoretest2`.
In what way would it break template strings in multiline eval? Does PS require putting a slash in front of new lines, making them possibly be interpreted as commands?

Zarel:
This should also possibly be fixed server-side.

Like, right now, we support mixing commands and chat messages, and honestly, maybe we just... shouldn't.

asgdf:
>This should also possibly be fixed server-side.

I'm not aware of this problem or a similar one occuring serverside? Doesn't the server already strictly handle things on a per-line basis? Is there a reason to disallow mixing commands and chat in multiline inputs?

Zarel:
Yeah, it's usually a mistake?

Zarel:
We could easily support something like "commands can be mixed, but text must be sent at the top"

Zarel:
e.g. `text, text, command, command` is fine, but not in any other order

Zarel:
At which point, newlines in commands can always be unambiguous, unless the next line starts with `/`. Which is... still un-ideal, but better, I think.

asgdf:
I've never had a issue with that nor really saw it happen, but I suppose it'd be acceptable to do it that way (especially for the longer commands, like trivia stuff, anything that passes in HTML, etc) as long as it doesn't include broadcasted informational commands, because I do occasionally mix those with text in one input.

But on the topic of *this* PR, I'm still not sure what or how it's breaking anything, all it's doing is running parseCommand once per line instead of once per input.

Slayer95:
Regarding multiline eval, this impl is filtering empty lines out, which are significant in the template string grammar.

asgdf:
Ah I see what you mean now. I've changed it so it doesn't filter out blank lines, only `false`, which is what's returned if a command was parsed.

asgdf:
Bump - shouldn't be breaking any other functionality now

Slayer95:
Unfortunately, this still messes with template strings in eval, as lines starting with slashes will be transformed into a different thing altogether.

asgdf:
Hmm, ok, that issue is a bit harder to resolve. The options I see are
1) Maintain a list of commands that can have multiline arguments and don't try to parse client commands following them
\- requires the client to maintain a copy of the list of multiline commands, which takes work and would probably not be remembered to be updated. 
2) Disallow use of client commands following another command
\- would lead to an inexplicable (to users) difference in behavior between commands.
3) Disallow using multiple commands in one input in general
\- would probably be needlessly limiting
4) Keep things as they are
\- like 2 but in addition to not all intended commands being executed, it leads to unintended commands being executed.
5) Introduce a unified way of marking a command as having multiline arguments without having to track them all separately.
\- no idea what that might look like in practise

js/client-chat.js
```diff
@@ -93,7 +93,8 @@
 			}
 			this.tabComplete.reset();
 			this.chatHistory.push(text);
-			text = this.parseCommand(text);
+			// make sure commands in multiline inputs are handled correctly
+			text = text.split('\n').map(this.parseCommand, this).filter(function (line) {return (line !== false);}).join('\n');
```
Zarel: `.split('\n/')`?

asgdf: Not sure if I understand. Is it to split before each command rather than just by newline? Why?

Zarel: Because having the rule be "everything not starting with `/` is part of the previous command" makes the code simpler.

asgdf: Which part would be made easier by that change? It'd require re-adding the slash in order to parse the command, and adding a comment to explain why it's done like that. In terms of user experience, it makes client commands + text in a new line work in unexpected ways (like the current situation)

Zarel: client commands + text is always at least a bit confusing of a situation.

My proposed change would make it slightly less confusing, overall.

Zarel:
For `eval` specifically, `>>` and `>>>` should probably be hardcoded. `/eval` should probably not be used for arbitrary input.

PR #982, opened 2017-08-16
Tooltips: Strong Winds boost Weather Ball's power

MacChaeger:


urkerab:
At this point that covers every type of weather, so wouldn't just testing `this.battle.weather` directly suffice?

Zarel:
urkerab makes a strong point.

MacChaeger:
One wonders why that wasn't done in the first place...

MacChaeger:
Well, probably because whoever did it didn't realize that strong winds boosts weather ball; otherwise they would have included it, probably.

js/client-battle-tooltips.js
```diff
@@ -965,7 +965,7 @@ var BattleTooltips = (function () {
 			if (move.id in table.overrideBP) basePower = table.overrideBP[move.id];
 		}
 		var basePowerComment = '';
-		var thereIsWeather = (this.battle.weather in {'sunnyday': 1, 'desolateland': 1, 'raindance': 1, 'primordialsea': 1, 'sandstorm': 1, 'hail':1});
+		var thereIsWeather = this.battle.weather;
```
Zarel: `!!`

Zarel: (Cast to boolean)

MacChaeger: Oh! I thought you were just expressing alarm of some kind.

Zarel: lol, yeah, in hindsight I realized that was unclear.

PR #985, opened 2017-08-20
Don't be so confused when an Illusion faints

urkerab:
If Zoroark is KO'd while its Illusion is up, the following happens:

- The server updates the Illusion's condition to be `0 fnt`
- The server breaks the Illusion
- The server faints Zoroark

Unfortunately once the client has seen the Illusory Pokémon's condition as `0 fnt` it treats it as fainted and zero HP for the rest of the battle, so even when it switches in for real the client either says that it's fainted or a duplicate.

The `zerohp` flag seems to be a complete waste of time here as everything it does is already covered by the `fainted` flag except in the case of KO'ing Zoroark under Illusion, so removing that is the first step to a solution.

Note that it then remains to decide what to do with the `fainted` flag. There are two approaches here:

 - Don't set the `fainted` flag when the condition is `0 fnt`. This might cause confusion with requests, which can only send the fainted state as `0 fnt`, although presumably the relevant `faint` major would already have been sent.
- Don't send `0 fnt` until after the `faint` major (server-side change).

Note that even after the latter is complete, the client will still show the Illusory Pokémon as having 0% health in the sidebar until it switches in for real, at which point it will return to normal health. (But presumably Illusory Pokémon can show unusual amounts of health anyway.)

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -192,7 +192,6 @@ var Pokemon = (function () {
 		this.prevItemEffect = '';
 		this.species = species;
 		this.fainted = false;
-		this.zerohp = false;
 
 		this.status = '';
 		this.statusStage = 0;
@@ -322,7 +321,7 @@ var Pokemon = (function () {
 			hpstring = hpstring.substr(parenIndex + 1, hpstring.length - parenIndex - 2);
 		}
 
-		var oldhp = (this.zerohp || this.fainted) ? 0 : (this.hp || 1);
+		var oldhp = this.fainted ? 0 : (this.hp || 1);
 		var oldmaxhp = this.maxhp;
 		var oldwidth = this.hpWidth(100);
 		var oldcolor = this.hpcolor;
@@ -705,7 +704,6 @@ var Pokemon = (function () {
 	Pokemon.prototype.reset = function () {
 		this.clearVolatile();
 		this.hp = this.maxhp;
-		this.zerohp = false;
 		this.fainted = false;
 		this.needsReplace = (this.details.indexOf('-*') >= 0);
 		this.status = '';
@@ -719,7 +717,7 @@ var Pokemon = (function () {
 	// This function is NOT used in the calculation of any other displayed
 	// percentages or ranges, which have their own, more complex, formulae.
 	Pokemon.prototype.hpWidth = function (maxWidth) {
-		if (this.fainted || this.zerohp) return 0;
+		if (this.fainted) return 0;
 
 		// special case for low health...
 		if (this.hp == 1 && this.maxhp > 10) return 1;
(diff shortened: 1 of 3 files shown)


Zarel:
I forget what the `zerohp` flag was intended to do, but I think it did have a use at some point.

The best solution for the Illusion thing, I think, is to handle it when the Illusion breaks. If a 0hp Illusion breaks, the pre-illusion Pokémon should probably be set back to full HP, no statuses.

Zarel:
How do fainted Illusion Pokémon work in the games? Does someone have a YouTube video?

Marty-D:
>How do fainted Illusion Pokémon work in the games?

I don't understand the question. The games have never given any indication about what is fainted other than graying out Poké Balls, which isn't going to tell you anything about whether or not you were facing an Illusion because the games don't use minisprites or health indicators or anything like that.

Zarel:
That's what I wanted to know. If the games don't transmit the information that the fainted Pokémon was an Illusion, the server shouldn't, either.

urkerab:
If the Pokémon takes direct damage then the illusion always breaks. See [this video](https://youtu.be/lbaosxvVJbE?t=150). Note that it shows the Zapdos as losing all of its HP first, then the illusion breaks, then the Zoroark faints.

Marty-D:
Oh, I was under the impression we _were_ talking about indirect damage since Zarel's first reply implied he knew Illusion breaks on direct damage all the time. My reply above was about a case like fainting to Spikes or something, since Illusion breaking on direct damage even if you lost all your HP has been implemented properly forever and has nothing to do with the client.

Zarel:
I think you're using different terminology than I'm used to. I usually call the damage types:

- move damage: Tackle, etc
- residual damage: burn, Sandstorm, etc
- true damage: Substitute, Curse (to self), Struggle (to self), etc

I'm already aware that move damage will always reveal Zoroark before it faints. The question was about residual damage and true damage. But now I understand the miscommunication: This pullreq fixes fainting from move damage, but I thought it was about fainting from residual damage.

Zarel:
I REMEMBER NOW! `.zerohp` is backwards compatibility with buggy PO replays where HP was sometimes simultaneously zero and nonzero. It's probably safe to remove now.

Zarel:
So, this pullreq removes the `.zerohp` flag, but does it do anything else? I'd merge this either way, but I'm just wondering if the commit summary is inaccurate. It's certainly under-detailed either way.

urkerab:
This PR started as an attempt to fix illusion-breaking faints, but in the end I stopped after removing the `zerohp` flag because I couldn't decide what the rest of the fix was (see initial comment). So yes, if you want to take this as-is then it wouldn't mind being resummarised.

Zarel:
The other changes I _am_ planning to do. I'll make an issue, I guess.

Zarel:
The rest of the Illusion stuff goes in #987

PR #988, opened 2017-08-26
Auto-hide format menu into sections

panpawn:


panpawn:
Screen caps:

panpawn:
Can someone help me out with the tests?

TheImmortal:
That doesn't look nice at all.

Zarel:
This is also not at all what I was suggesting in the thread...

panpawn:
>This is also not at all what I was suggesting in the thread...

The formats could be moved server-side to for it to look like what you were suggesting in the thread.

panpawn:
>That doesn't look nice at all.

What about it looks particularly bad? Is it the actual organization of the formats, or the way it's still split into columns, or...?

Zarel:
I'm not sure what you're saying? What does moving it server-side accomplish?

panpawn:
>What does moving it server-side accomplish?

Format reogranization

panpawn:
Although, this does lack single-format support currently.  I'll add that soon, but in the meantime, what else is different about this than what you guys had in mind?

Zarel:
Oh. Yes. You'll need server-side changes for this to take effect.

AustinXII:
I agree with TI, this is ugly, the options should no be going vertical

panpawn:
... as opposed to horizontal? They definitely would look better vertical than they would horizontal. Anyways, I think keeping support for different columns is a good idea, so I don't plan on removing that.

Zarel:
I personally think it's too many columns and several other things, but I proposed a pretty decent design imo. :<

AustinXII:
@panpawn it overlapping into the rooms section is ugly though

panpawn:
>I personally think it's too many columns and several other things, but I proposed a pretty decent design imo. :<

I mean, I plan on keeping it all as 1 format with the design you had proposed Zarel, but I would very much like to keep support for multiple columns.

>@panpawn it overlapping into the rooms section is ugly though

... It currently does that already. When I'm done, it won't on main. >_>...

TheImmortal:
This doesn't really reduce clutter; it's practically the same size as the list now. All it does is make it more difficult to select what you want to play. And these buttons are not appealing at all. My idea or Zarel's idea, or like practically any idea posted in the thread is good.

PR #989, opened 2017-08-27
Distinguish between 0HP and 1HP when HP Percentage Mod is off

urkerab:
When HP Percentage Mod is off, 0HP and 1HP look too similar. This makes 0HP take 0 CSS pixels and 1HP take 3 CSS pixels.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -717,10 +717,7 @@ var Pokemon = (function () {
 	// This function is NOT used in the calculation of any other displayed
 	// percentages or ranges, which have their own, more complex, formulae.
 	Pokemon.prototype.hpWidth = function (maxWidth) {
-		if (this.fainted) return 0;
-
-		// special case for low health...
-		if (this.hp == 1 && this.maxhp > 10) return 1;
+		if (this.fainted || !this.hp) return 0;
 
 		if (this.maxhp === 48) {
 			// Draw the health bar to the middle of the range.


Zarel:
How does this make 1HP take 3 CSS pixels?

js/battle.js
```diff
@@ -719,16 +719,13 @@ var Pokemon = (function () {
 	Pokemon.prototype.hpWidth = function (maxWidth) {
 		if (this.fainted) return 0;
```
Zarel: I think it makes sense to add `|| !this.hp` here, and then add back the `|| 1` down there.

urkerab:
Sorry for the delay, I'd mislaid the branch.

> How does this make 1HP take 3 CSS pixels?

Because it scales 48 game pixels to the 150 client pixels.

Zarel:
Do we still want the special case for low health? I still think it makes sense...

urkerab:
Well, I checked a couple of YouTube videos and the cart doesn't special-case low health - if you have ¹⁄₁₂ HP then you see ¹⁄₁₂ of the HP bar, whether that's 1 HP of 12, 10 HP of 120 etc.

Zarel:
The problem is, the cart shows 1 pixel for low health.

We don't actually give as much information as we have pixels for, but I think it's important information to the user to see that "1 pixel = lowest amount of health displayable".

This is also relevant for, Sturdy activating and people complaining about why PS shows their opponent as having more than one pixel of health.

urkerab:
> The problem is, the cart shows 1 pixel for low health.

[Various](https://youtu.be/zKcApfCkTT0?t=123) [sources](http://imgur.com/W9biKWI) disagree.

Zarel:
Yes, we also have an exception for super-low max-HP. Let me rephrase:

The problem is, for Pokémon with more than 40ish max-HP, the cart shows 1 pixel for low health.

Zarel:
I should note that PS nearly always deals with Pokémon with more than 40ish max-HP. The most common exception is Shedinja, which PS also displays correctly.

urkerab:
> Yes, we also have an exception for super-low max-HP.

But the cart doesn't, so I want to remove that exception.

> The problem is, for Pokémon with more than 40ish max-HP, the cart shows 1 pixel for low health.

I thought that had already been covered by adding back the `|| 1` as you previously requested.

Zarel:
> But the cart doesn't, so I want to remove that exception.

The cart doesn't need one, because it already only displays 1 pixel when a Pokémon with more than 30ish max HP is at 1 HP.

> I thought that had already been covered by adding back the `|| 1` as you previously requested.

No, the special case for low health is what's intended to cover it.

Zarel:
Like, the main difference between us and the cart is that the cart has perfect-resolution information about the enemy's HP, while we don't because it would be too much of an information leak.

But I think probably one of the most important signals to send, is that 1 pixel of HP nearly always represents 1 HP left, and vice versa.

Zarel:
Sooooo...

PR #991, opened 2017-08-30
Tooltips: Flare Boost boosts BP, not SpA

MacChaeger:


DeathlyPlays:
Isn't there a place where it modifies the BP for a calc when you hover over the move (iirc guts does that) you could possibly do that for this

Marty-D:
Yes, it's in this same file and he already did that.

DeathlyPlays:
Oh ok :)

PR #992, opened 2017-08-31
Tooltips: Add more BP and stat modifiers

MacChaeger:
Things that I wasn't sure about adding:
- STAB and Adaptability
- Weather and Terrain

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -1173,32 +1173,39 @@ var BattleTooltips = (function () {
 			for (var i = 0; i < allyActive.length; i++) {
 				var ally = allyActive[i];
 				if (!ally || ally.fainted) continue;
-				if (!auraBoosted && ((ally.ability === 'Fairy Aura' && move.type === 'Fairy') || (ally.ability === 'Dark Aura' && move.type === 'Dark'))) {
-					auraBoosted = ally.ability;
-				}
-				if (ally.ability === 'Aura Break') auraBroken = true;
-				if (ally === pokemon) continue;
-				if (ally.ability === 'Battery' && move.category === 'Special') {
-					basePower = Math.floor(basePower * 1.3);
-					basePowerComment = this.makePercentageChangeText(1.3, 'Battery');
+				if (ally.ability === 'Fairy Aura') {
+					if (move.type === 'Fairy') auraBoosted = 'Fairy Aura';
+				} else if (ally.ability === 'Dark Aura') {
+					if (move.type === 'Dark') auraBoosted = 'Dark Aura';
+				} else if (ally.ability === 'Aura Break') {
+					auraBroken = true;
+				} else if (ally.ability === 'Battery') {
+					if (ally !== pokemon && move.category === 'Special') {
+						basePower = Math.floor(basePower * 1.3);
+						basePowerComment = this.makePercentageChangeText(1.3, 'Battery');
+					}
 				}
 			}
 		}
 		var foeActive = this.battle.yourSide.active;
-		if (!auraBoosted && move.category !== 'Status' && (move.type === 'Fairy' || move.type === 'Dark')) {
+		var doneLooking = auraBoosted && auraBroken;
+		if (!doneLooking && move.category !== 'Status' && (move.type === 'Fairy' || move.type === 'Dark')) {
 			for (var i = 0; i < foeActive.length; i++) {
 				var foe = foeActive[i];
 				if (!foe || foe.fainted) continue;
-				if (!auraBoosted && ((foe.ability === 'Fairy Aura' && move.type === 'Fairy') || (foe.ability === 'Dark Aura' && move.type === 'Dark'))) {
-					auraBoosted = foe.ability;
+				if (ally.ability === 'Fairy Aura') {
+					if (move.type === 'Fairy') auraBoosted = 'Fairy Aura';
+				} else if (ally.ability === 'Dark Aura') {
+					if (move.type === 'Dark') auraBoosted = 'Dark Aura';
+				} else if (ally.ability === 'Aura Break') {
+					auraBroken = true;
 				}
-				if (foe.ability === 'Aura Break') auraBroken = true;
 			}
 		}
 		if (auraBoosted) {
 			if (auraBroken) {
 				basePower = Math.floor(basePower * 0.75);
-				basePowerComment = this.makePercentageChangeText(0.75, 'Aura Break');
+				basePowerComment = this.makePercentageChangeText(0.75, auraBoosted + ' + Aura Break');
 			} else {
 				basePower = Math.floor(basePower * 1.33);
 				basePowerComment = this.makePercentageChangeText(1.33, auraBoosted);
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -1240,6 +1240,13 @@ var BattleTooltips = (function () {
 		'Spell Tag': 'Ghost',
 		'TwistedSpoon': 'Psychic'
 	};
+	var orbUsers = {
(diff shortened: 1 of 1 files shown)


Zarel:
Still failing test

Zarel:
(You have mixed tabs and spaces)

js/client-battle-tooltips.js
```diff
@@ -1152,6 +1165,47 @@ var BattleTooltips = (function () {
...
+
+		var allyActive = pokemon.side.active;
+		var auraBoosted = '';
+		var auraBroken = false;
+		if (move.category !== 'Status' && allyActive.length > 1) {
+			for (var i = 0; i < allyActive.length; i++) {
+				var ally = allyActive[i];
+				if (!ally || ally.fainted) continue;
+				if (!auraBoosted && (ally.ability === 'Fairy Aura' && move.type === 'Fairy') ||
+				    		    (ally.ability === 'Dark Aura' && move.type === 'Dark')) {
```
Zarel: We definitely don't support this indent+space thing.

js/client-battle-tooltips.js
```diff
@@ -1152,6 +1165,47 @@ var BattleTooltips = (function () {
...
+				}
+			}
+		}
+		var foeActive = this.battle.yourSide.active;
+		if (!auraBoosted && move.category !== 'Status' && (move.type === 'Fairy' || move.type === 'Dark')) {
+			for (var i = 0; i < foeActive.length; i++) {
+				var foe = foeActive[i];
+				if (!foe || foe.fainted) continue;
+				if (!auraBoosted && (foe.ability === 'Fairy Aura' && move.type === 'Fairy') ||
+				    		    (foe.ability === 'Dark Aura' && move.type === 'Dark')) {
```
Zarel: ditto

js/client-battle-tooltips.js
Slayer95: Also note that && has higher operator precedence than ||.

js/client-battle-tooltips.js
```diff
@@ -1152,6 +1165,47 @@ var BattleTooltips = (function () {
...
+				}
+			}
+		}
+		var foeActive = this.battle.yourSide.active;
+		if (!auraBoosted && move.category !== 'Status' && (move.type === 'Fairy' || move.type === 'Dark')) {
+			for (var i = 0; i < foeActive.length; i++) {
+				var foe = foeActive[i];
+				if (!foe || foe.fainted) continue;
+				if (!auraBoosted && (foe.ability === 'Fairy Aura' && move.type === 'Fairy') ||
+						    (foe.ability === 'Dark Aura' && move.type === 'Dark')) {
```
Zarel: This indentation is still dumb even if `eslint` doesn't complain about it. This is literally only designed for 8-space tabs, which we don't actually use.

js/client-battle-tooltips.js
```diff
@@ -1152,6 +1165,47 @@ var BattleTooltips = (function () {
...
+
+		var allyActive = pokemon.side.active;
+		var auraBoosted = '';
+		var auraBroken = false;
+		if (move.category !== 'Status' && allyActive.length > 1) {
+			for (var i = 0; i < allyActive.length; i++) {
+				var ally = allyActive[i];
+				if (!ally || ally.fainted) continue;
+				if (!auraBoosted && (ally.ability === 'Fairy Aura' && move.type === 'Fairy') ||
+							(ally.ability === 'Dark Aura' && move.type === 'Dark')) {
```
Zarel: I actually don't know what the convention is here. I think Cathy wanted to do +2 tabs, but I think consensus is on +1 tab?

Zarel: +3 tabs is right out, though.

Slayer95: Might as well delete the ``!auraBoosted`` check and there is no longer a debate to have.
That would also fix the ``&&`` vs ``||`` precedence issue.

Slayer95:
@MacChaeger , my comments re: precedence refer to:

The following code
```js
if (!auraBoosted && (ally.ability === 'Fairy Aura' && move.type === 'Fairy') || (ally.ability === 'Dark Aura' && move.type === 'Dark')) {
// Do Stuff
}
```

is parsed as

```js
if (!auraBoosted && (ally.ability === 'Fairy Aura' && move.type === 'Fairy')) {
// Do Stuff 
} else if (ally.ability === 'Dark Aura' && move.type === 'Dark') {
// Do Stuff
}
```

rather than

```js
if (!auraBoosted) {
if ((ally.ability === 'Fairy Aura' && move.type === 'Fairy') || (ally.ability === 'Dark Aura' && move.type === 'Dark')) {
// Do Stuff
}
}
```

MacChaeger:
OH! I get it now. I forgot a set of parentheses.

js/client-battle-tooltips.js
```diff
@@ -1152,6 +1165,45 @@ var BattleTooltips = (function () {
...
 		}
+
+		var allyActive = pokemon.side.active;
+		var auraBoosted = '';
+		var auraBroken = false;
+		if (move.category !== 'Status' && allyActive.length > 1) {
+			for (var i = 0; i < allyActive.length; i++) {
+				var ally = allyActive[i];
+				if (!ally || ally.fainted) continue;
+				if (!auraBoosted && ((ally.ability === 'Fairy Aura' && move.type === 'Fairy') || (ally.ability === 'Dark Aura' && move.type === 'Dark'))) {
```
Zarel: This is way too complicated of a condition and would probably benefit from being simplified in some way.

My first thought is that `!auraBoosted &&` isn't really a necessary condition. It's a micro-optimization that makes the code less readable so I'd probably just leave it off entirely.

Zarel: My second thought is just to split up the `||` into a chain of `switch`es...

```
if (ally.ability === 'Fairy Aura') {
  if (move.type === 'Fairy') auraBoosted = 'Fairy Aura';
} else if (ally.ability === 'Dark Aura') {
  if (move.type === 'Dark') auraBoosted = 'Dark Aura';
} else if (ally.ability === 'Aura Break') {
  auraBroken = true;
} else if (ally.ability === 'Battery') {
  if (ally !== pokemon && move.category === 'Special') {
    // ...
```

js/client-battle-tooltips.js
```diff
@@ -1206,6 +1271,10 @@ var BattleTooltips = (function () {
 		// Type-enhancing items
 		if (itemTypes[item.name] === moveType) return this.battle.gen < 4 ? 1.1 : 1.2;
 
+		// Pokemon-specific items
+		if (item.name === 'Soul Dew' && this.battle.gen < 7) return 0;
+		if (orbTypes[item.name] === moveType || moveType === 'Dragon') return 1.2;
```
Zarel: If I'm reading this correctly, this will make all Dragon-type moves 1.2x damage?

MacChaeger:
OK, with that, what do you think of showing STAB, weather and terrain? Also, I just found out Victory Star's accuracy boost isn't showing up in the tooltips right now and I can't figure out why.

js/client-battle-tooltips.js
```diff
@@ -1206,6 +1278,10 @@ var BattleTooltips = (function () {
 		// Type-enhancing items
 		if (itemTypes[item.name] === moveType) return this.battle.gen < 4 ? 1.1 : 1.2;
 
+		// Pokemon-specific items
+		if (item.name === 'Soul Dew' && this.battle.gen < 7) return 0;
+		if (orbUsers[pokemonData.baseSpecies] === item.name && orbTypes[item.name] === moveType || moveType === 'Dragon') return 1.2;
```
Zarel: This still shows 1.2x damage for a Flygon using Outrage.

You need parens.

PR #993, opened 2017-09-01
Room list: Add "PSPL Winner" section

panpawn:


panpawn:
This has a server counterpart: https://github.com/Zarel/Pokemon-Showdown/pull/3937

js/client-rooms.js
```diff
@@ -75,7 +75,8 @@
 				var rightSide = '<button class="button" name="roomlist" title="Watch an active battle"><strong>' + battleCount + '</strong> active ' + (battleCount == 1 ? 'battle' : 'battles') + '</button> <span style="' + Tools.getPokemonIcon('meloetta-pirouette') + '" class="picon icon-right" title="Meloetta is PS\'s mascot! The Pirouette forme is Fighting-type, and represents our battles."></span>';
 				this.$('.roomlisttop').html('<table class="roomcounters" border="0" cellspacing="0" cellpadding="0" width="100%"><tr><td>' + leftSide + '</td><td>' + rightSide + '</td></tr></table>');
 			}
-			this.$('.roomlist').first().html('<h2 class="rooms-officialchatrooms">Official chat rooms</h2>' + _.map(rooms.official, this.renderRoomBtn).join(""));
+			this.$('.roomlist').first().html('<h2 class="rooms-officialchatrooms">Official chat rooms</h2>' + _.map(rooms.official, this.renderRoomBtn).join("") +
+				(rooms.pspl && rooms.pspl.length ? '<h2 class="rooms-psplchatrooms">PSPL Winner</h2>' + _.map(rooms.pspl, this.renderRoomBtn).join("") : ''));
```
HoeenCoder: I dont think older browsers support ternary operators (ive been told to use `if` instead)

panpawn: ... there's a ternary operator a few lines above?

Slayer95: IE8 isn't in our compatibility matrix for the application bits of the client, so that's not an issue here.

The ternary operator IE8 quirk is probably already an issue for the replay player, however..... /cc @Zarel

Zarel: I'm really not going to fix IE8 bugs unless someone asks...

panpawn:
Thanks! Now the server-end of this just has to be taken care of :p

PR #994, opened 2017-09-01
Locked symbol: Use unicode code point

panpawn:


js/client.js
```diff
@@ -2313,7 +2313,7 @@
 			order: 10012,
 		},
 		'‽': {
```
Slayer95: Here too?

panpawn: Good catch, thanks!

panpawn:
Updated

PR #995, opened 2017-09-02
Handle level changes in detailschange

xfix:
Will be used for a secret project (staff link: http://www.smogon.com/forums/threads/3613633/).

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -6467,11 +6467,13 @@ var Battle = (function () {
 			poke.removeVolatile('typeadd');
 			poke.removeVolatile('typechange');
 
-			var newSpecies;
-			var commaIndex = args[2].indexOf(',');
-			if (commaIndex === -1) {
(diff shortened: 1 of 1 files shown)


js/battle.js
```diff
@@ -6467,11 +6467,13 @@ var Battle = (function () {
...
-			var commaIndex = args[2].indexOf(',');
-			if (commaIndex === -1) {
-				newSpecies = args[2];
-			} else {
+			var newSpecies = args[2];
+			var commaIndex = newSpecies.indexOf(',');
+			if (commaIndex !== -1) {
+				var level = $.trim(newSpecies.substr(commaIndex + 1));
+				if (level.charAt(0) === 'L') {
+					poke.level = +level.substr(1);
```
Zarel: ...this is `NaN` if the Pokémon has a gender.

Zarel: image

PR #996, opened 2017-09-04
Match a shiny Pokemon as their non-shiny Team Preview counterpart

QuiteQuiet:
This relies on a pretty trivial fix server-side that I'll (or anyone else) will do if/when this gets approved (https://github.com/Zarel/Pokemon-Showdown/issues/2367#issuecomment-170441910). 

I believe this would cover most of the bases for correctly matching the Pokemon as they were from Team Preview with an updated shiny version if the first switch-in reveal it to be shiny. Nothing else I've tried so far would cover Species Clause-less metagames like AG, so this was the best I've been able to do.

[Example](https://puu.sh/xrnOl.gif)

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -339,7 +339,6 @@ var Pokemon = (function () {
 	Pokemon.prototype.checkDetails = function (details) {
 		if (!details) return false;
 		if (details === this.details) return true;
-		if (!this.needsReplace) return false;
 		// the actual forme was hidden on Team Preview
 		details = details.replace(/-[A-Za-z0-9]+(, |$)/, '$1');
 		return (details === this.details.replace(/-\*(, |$)/, '$1'));
@@ -705,7 +704,6 @@ var Pokemon = (function () {
 		this.clearVolatile();
 		this.hp = this.maxhp;
 		this.fainted = false;
-		this.needsReplace = (this.details.indexOf('-*') >= 0);
 		this.status = '';
 		this.moveTrack = [];
 		this.name = this.name || this.species;
@@ -6104,16 +6102,13 @@ var Battle = (function () {
 				}
 				if (isSwitch && pokemon == this.p1.lastPokemon && !this.p1.active[slot]) continue;
 				if ((searchid && pokemon.searchid === searchid) || // exact match
-					(!pokemon.searchid && pokemon.checkDetails(details)) || // switch-in matches Team Preview entry
 					(!searchid && pokemon.ident === pokemonid)) { // name matched, good enough
-					if (!pokemon.searchid && createIfNotFound) {
-						pokemon.name = name;
-						pokemon.searchid = searchid;
-						pokemon.ident = pokemonid;
-						if (pokemon.needsReplace) {
-							pokemon = this.p1.newPokemon(this.parseDetails(name, pokemonid, details), i);
-						}
-					}
+					if (slot >= 0) pokemon.slot = slot;
+					return pokemon;
+				}
+				if (!pokemon.searchid && (pokemon.checkDetails(details) || // switch-in matches Team Preview entry
+					pokemon.checkDetails(details.replace(', shiny', '')))) { // switch-in matches non-shiny Team Preview entry
+					pokemon = this.p1.newPokemon(this.parseDetails(name, pokemonid, details), i);
 					if (slot >= 0) pokemon.slot = slot;
 					return pokemon;
 				}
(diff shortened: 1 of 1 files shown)


QuiteQuiet:
Updated because I forgot Gourgeist and Pumpkaboo could be shiny and hide their actual form with `-*`. This should work fine now, unless I'm forgetting more edge cases.

js/battle.js
```diff
@@ -6147,6 +6155,14 @@ var Battle = (function () {
 					if (slot >= 0) pokemon.slot = slot;
 					return pokemon;
 				}
+				if (!pokemon.searchid && !pokemon.shiny && pokemon.checkDetails(details.replace(', shiny', ''))) { // switch-in matches non-shiny Team Preview entry
```
Zarel: You should possibly drag the entire `switch-in matches Team Preview entry` case down here, instead of leaving the cases messy.

QuiteQuiet: From what I've been able to tell, `pokemon.needsReplace` is used for the Arceus/Silvally/etc. that shouldn't reveal their typing on Team Preview. If I move that down, is there any remaining case where

    if (pokemon.needsReplace) {
        pokemon = this.p2.newPokemon(this.parseDetails(name, pokemonid, details), i);
    }

happens?

Zarel: Probably not.

Zarel: I would recommend removing it, yes.

QuiteQuiet:
Updated.

js/battle.js
```diff
@@ -6104,19 +6104,24 @@ var Battle = (function () {
...
 						pokemon.ident = pokemonid;
-						if (pokemon.needsReplace) {
-							pokemon = this.p1.newPokemon(this.parseDetails(name, pokemonid, details), i);
-						}
 					}
 					if (slot >= 0) pokemon.slot = slot;
 					return pokemon;
 				}
+				if (!pokemon.searchid && (pokemon.checkDetails(details) || // switch-in matches Team Preview entry
+					(!pokemon.shiny && pokemon.checkDetails(details.replace(', shiny', ''))))) { // switch-in matches non-shiny Team Preview entry
```
Zarel: You don't need `!pokemon.shiny`; that's sufficiently implicated by `!pokemon.searchid`

js/battle.js
```diff
@@ -6104,19 +6104,24 @@ var Battle = (function () {
 				}
 				if (isSwitch && pokemon == this.p1.lastPokemon && !this.p1.active[slot]) continue;
 				if ((searchid && pokemon.searchid === searchid) || // exact match
-					(!pokemon.searchid && pokemon.checkDetails(details)) || // switch-in matches Team Preview entry
 					(!searchid && pokemon.ident === pokemonid)) { // name matched, good enough
 					if (!pokemon.searchid && createIfNotFound) {
```
Zarel: This entire block is unnecessary in this branch, actually.

Zarel: For the "exact match" case, `pokemon.searchid` will already exist, and for "name match only", not enough information is passed to actually create a Pokémon that isn't already found.

QuiteQuiet:
Did.

js/battle.js
```diff
@@ -6104,16 +6104,16 @@ var Battle = (function () {
...
-					}
+					if (slot >= 0) pokemon.slot = slot;
+					return pokemon;
+				}
+				if (!pokemon.searchid && (pokemon.checkDetails(details) || // switch-in matches Team Preview entry
+					pokemon.checkDetails(details.replace(', shiny', '')))) { // switch-in matches non-shiny Team Preview entry
+					pokemon.name = name;
+					pokemon.searchid = searchid;
+					pokemon.ident = pokemonid;
+					pokemon = this.p1.newPokemon(this.parseDetails(name, pokemonid, details), i);
```
Zarel: `.needsReplace` isn't always true here - and if it were, the three lines above would be useless...

Zarel: In other words, you need to add back `if (pokemon.needsReplace) {` here.

QuiteQuiet: Do I also not have to mark every single Pokemon with `needsReplace` then, since we don't know from Team Preview which of them are shiny or not?

QuiteQuiet: Or something like this?

    if (pokemon.needsReplace) {
        ...
    else {
        ...
    }

Zarel: Good point, we can just always replace the Pokemon object. In which case just remove the above three `pokemon.... = ...` lines.

Zarel: And then remove the `needsReplace` flag, since we're just replacing them all anyway.

QuiteQuiet:
Done

Zarel:
Thanks!

PR #998, opened 2017-09-06
Tooltips: Show BP modifiers from STAB and field effects

MacChaeger:
This is the last of the BP modifiers until I think of something else that I inevitably forgot.

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -1244,7 +1244,7 @@ var BattleTooltips = (function () {
 				basePowerComment += this.makePercentageChangeText(1.5, 'weather');
 			}
 		}
-		var isGrounded = false;
+		var isGrounded = true;
 		var noItem = !pokemonData.item || this.battle.hasPseudoWeather('Magic Room') || pokemonData.volatiles && pokemonData.volatiles['embargo'];
 		if (this.Battle.hasPseudoWeather('Gravity')) {
 			isGrounded = true;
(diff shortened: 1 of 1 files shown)


js/client-battle-tooltips.js
```diff
@@ -1205,12 +1229,56 @@ var BattleTooltips = (function () {
...
+		} else if (ability === 'levitate') {
+			isGrounded = false;
+		} else if (pokemonData.volatiles && (pokemonData.volatiles['magnetrise'] || pokemonData.volatiles['telekinesis'])) {
+			isGrounded = false;
+		} else if (!noItem && pokemonData.item !== 'airballoon') {
+			isGrounded = false;
+		} else if (!(pokemonData.volatiles && pokemonData.volatiles['roost'])) {
+			// If a Fire/Flying type uses Burn Up and Roost, it becomes ???/Flying-type, but it's still grounded.
+			for (var i = 0; i < types.length; i++) {
+				if (types[i] === 'Flying') isGrounded = false;
```
Zarel: > still grounded

> `isGrounded = false;`

MacChaeger: That just explains why it checks for roost before it checks the typing; I literally copied this straight out of `isGrounded` in the server's sim code, complete with the comment.

MacChaeger: That said, I don't actually think it's been tested what happens if a flying/??? type that has roosted also has levitate or something else that lifts it up...

Zarel:
Looks good to me.

MacChaeger:
OK well somebody remind me to change this when Marty confirms that that particular mechanic is [actually wrong](http://www.smogon.com/forums/posts/7503256/)

PR #999, opened 2017-09-06
Support Google login (two-factor auth)

Zarel:
This doesn't support setting accounts up for Google login: that still has to be manually done via the database by setting the email field to `username@gmail.com@`, where the second `@` denotes that it's using Gmail login.

If the email field does end in `@`, `getassertion` will note this by sending `;;@gmail`, to convey that the server is expecting a Google login token rather than a password.

Upon receiving `;;@gmail`, the client will replace the password box will with a Google login button, and then send the resulting Google login token to the server in the `password` field. The server will validate the "password" using the Google server libraries, and otherwise handle the login as normal.  Note that Google login requires various features that a paranoid person might disable; most notably 3rd-party cookies.

Fixes Zarel/Pokemon-Showdown#3394

### DIFF
--- a/index.template.html
+++ b/index.template.html
@@ -46,6 +46,7 @@ http://play.pokemonshowdown.com/development
 		<!-- meta name="mobile-web-app-capable" content="yes" / -->
 		<!-- meta name="apple-mobile-web-app-capable" content="yes" / -->
 		<meta name="robots" content="noindex" />
+		<meta name="google-signin-client_id" content="912270888098-jjnre816lsuhc5clj3vbcn4o2q7p4qvk.apps.googleusercontent.com">
 		<script>
 			var Config = {};
 		</script>
--- a/lib/ntbb-session.lib.php
+++ b/lib/ntbb-session.lib.php
@@ -179,7 +179,7 @@ class NTBBSession {
 	}
 
 	private function passwordVerifyInner($userid, $pass, $user) {
-		global $psdb;
+		global $psdb, $psconfig;
 
 		// throttle
 		$ip = $this->getIp();
@@ -200,6 +200,18 @@ class NTBBSession {
 					'time' => time(),
 				];
 			}
+		}
+
+		if (substr(@$user['email'], -1) === '@') {
+			require_once dirname(__FILE__).'/../vendor/autoload.php';
+			$client = new Google_Client(['client_id' => $psconfig['gapi_clientid']]);
+			$payload = $client->verifyIdToken($pass);
+			if (!$payload) return false;
+			if (strpos($payload['aud'], $psconfig['gapi_clientid']) === false) return false;
+			if ($payload['email'] === substr($user['email'], 0, -1)) {
+				return true;
+			}
+			return false;
 		}
 
 		$rehash = false;
@@ -489,6 +501,7 @@ class NTBBSession {
 				if ($row['password'] && $row['nonce']) {
 					return ';;Your username is no longer available.';
 				}
+				if (substr($row['email'], -1) === '@') return ';;@gmail';
 				return ';';
 			} else {
 				// Unregistered username.
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ eslint-cache/
 .DS_Store
 Thumbs.db
 npm-debug.log
+/vendor/
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,5 @@
+{
+    "require": {
+        "google/apiclient": "^2.2"
+    }
+}
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,702 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "This file is @generated automatically"
+    ],
+    "content-hash": "14b5b31e2f70798045355d7ca0e8c52a",
+    "packages": [
+        {
+            "name": "firebase/php-jwt",
+            "version": "v5.0.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/firebase/php-jwt.git",
+                "reference": "9984a4d3a32ae7673d6971ea00bae9d0a1abba0e"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/firebase/php-jwt/zipball/9984a4d3a32ae7673d6971ea00bae9d0a1abba0e",
+                "reference": "9984a4d3a32ae7673d6971ea00bae9d0a1abba0e",
+                "shasum": ""
+            },
+            "require": {
+                "php": ">=5.3.0"
+            },
+            "require-dev": {
+                "phpunit/phpunit": " 4.8.35"
+            },
+            "type": "library",
+            "autoload": {
+                "psr-4": {
+                    "Firebase\\JWT\\": "src"
+                }
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "BSD-3-Clause"
+            ],
+            "authors": [
+                {
+                    "name": "Neuman Vong",
+                    "email": "neuman+pear@twilio.com",
+                    "role": "Developer"
+                },
+                {
+                    "name": "Anant Narayanan",
+                    "email": "anant@php.net",
+                    "role": "Developer"
+                }
+            ],
+            "description": "A simple library to encode and decode JSON Web Tokens (JWT) in PHP. Should conform to the current spec.",
+            "homepage": "https://github.com/firebase/php-jwt",
+            "time": "2017-06-27T22:17:23+00:00"
+        },
+        {
+            "name": "google/apiclient",
+            "version": "v2.2.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/google/google-api-php-client.git",
+                "reference": "f3fadd538315d62ebd1191d89ac791468c617260"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/google/google-api-php-client/zipball/f3fadd538315d62ebd1191d89ac791468c617260",
+                "reference": "f3fadd538315d62ebd1191d89ac791468c617260",
+                "shasum": ""
+            },
+            "require": {
+                "firebase/php-jwt": "~2.0|~3.0|~4.0|~5.0",
+                "google/apiclient-services": "~0.13",
+                "google/auth": "^1.0",
+                "guzzlehttp/guzzle": "~5.3.1|~6.0",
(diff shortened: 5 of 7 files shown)


index.template.html
```diff
@@ -46,6 +46,7 @@
 		<!-- meta name="mobile-web-app-capable" content="yes" / -->
 		<!-- meta name="apple-mobile-web-app-capable" content="yes" / -->
 		<meta name="robots" content="noindex" />
+		<meta name="google-signin-client_id" content="912270888098-jjnre816lsuhc5clj3vbcn4o2q7p4qvk.apps.googleusercontent.com">
```
Zarel: Hardcoding this here is kind of lazy, but rewriting the update script to pull from `config/config.inc.php` sounds annoying. :<

Slayer95:
Looks good at first glance. However, it's my impression that the core idea of OAuth is precisely to have a single, unique, login system. How does this fit exactly with the "two-factor authentication" idea in the PR title? I am sure this question has already been answered elsewhere, but still *how do we get from here to there*?

Zarel:
The idea is that if you have two-factor auth on the Google account associated with the account, you can't get Google to sign your login token without passing the two-factor auth check. We're outsourcing the two-factor auth to Google, in other words. That's why it's set up so that your password no longer works if your account is set to Google login.

panpawn:
How will this alter the policy to password resets?

Say someone links their PS account to their Google account, but then they lose access to their Google account. Will the password reset policy stay the same as it is now, or will that change?

Slayer95:
So, where is "sign only if two factor authentication is enabled" configured?

Zarel:
Right now, it's configured by manually editing the database to insert `[email address]@` into the `email` column.

lib/ntbb-session.lib.php
```diff
@@ -202,6 +202,18 @@ private function passwordVerifyInner($userid, $pass, $user) {
...
+			if (!$payload) return false;
+			if (strpos($payload['aud'], $psconfig['gapi_clientid']) === false) return false;
+			if ($payload['email'] === substr($user['email'], 0, -1)) {
+				return true;
+			}
+			return false;
+		}
+
 		$rehash = false;
 		if ($user['passwordhash']) {
```
Slayer95: If Google Login is enforced for an account, we no longer need to store the password hash, and it should probably be deleted, least an attacker who vulnerates the login server retrieves it by targetting these specially marked accounts and building a rainbow table from the hash.

Slayer95: Speaking of login server being taken over...

To make sure we are all in the same page, it seems that 2FA wouldn't improve security after that, since with this implementation as long as they get the login private key, they can fake a Google login without ever speaking to Google servers.

Given that, it would be helpful to be clear of the threat model we are working under, to make sure that our 2FA implementation really does address menaces.

Slayer95:
> Right now, it's configured by manually editing the database to insert [email address]@ into the email column.

That looks like "Login only with Google Login", which is a different thing. Unless I am missing something?

lib/ntbb-session.lib.php
Zarel: We don't check the password hash if Google login is enforced, though...

Zarel:
> That looks like "Login only with Google Login", which is a different thing. Unless I am missing something?

An account will either support logging in with Google or with a password, not both. Is that what you mean?

lib/ntbb-session.lib.php
Slayer95: The lessons from the past (rootinabox and last year's incident) would say that the real-world security threat lies in he login server itself. Therefore,  would it be possible to validate Google logins directly in the sim server?

IMO, Ideally a 2FA implementation would require *independent* login assertions from psweb and gapi. However, it seems you want to rely only on Google's own 2FA implementation.

Slayer95:
Yes. That's what I understand from the code.

Zarel:
Yes, that's the intention. Did you have a question about the intention?

It is possible to set up a Google account without 2FA, but if your Google account does, 2FA is required all the way through.

lib/ntbb-session.lib.php
Zarel: I suppose we _could_? The purpose of this commit is to protect the login server, though, which is way more of an issue than the sim server.

Zarel: I don't see what's wrong with relying on Google's own 2FA implementation? You know how 2FA works, right? Adding something else you know and something else you have doesn't actually make it any more secure.

Zarel: Like, validating logins in the sim server is like going through airport security after the plane has landed: it's not _completely_ useless, but it's not the main thing we're trying to protect against.

Zarel: "The lessons from the past is that people keep robbing all the cash from our bank's safes."

"We should get a huge lock..."

"and?"

"and put the huge lock on our break room, so bank robbers can't get into our break room after taking all the money out of our safes."

Slayer95: I don't see what you are arguing for. Analogies are bad, for certain.
How does this protect the login server? Again, what's our threat model?

Slayer95: > We don't check the password hash if Google login is enforced, though...

Just read this. Indeed, but we store it, and if the hashes get exposed they can be eventually(TM) broken to access other sites with the same credentials.

The risk is minimal, but it's still there, and it affects the users.

Zarel: Our threat model is "someone steals a password and uses it to log into a PS admin account".

Slayer95: Seems niche, but as long as you assert that Google credentials are somehow (without the "know and own" condition?!) inherently safer than PS credentials, I guess this works.

Zarel: The Google credentials come with the "know and own" conditions...

Zarel: Like, the threat models you're talking about aren't really important? I've never seen Google worry about them, and they take security very seriously.

Zarel: Also: There's nothing in this code that means we have to store password hashes in Google login mode.

Slayer95: > The Google credentials come with the "know and own" conditions...

But they are opt-in for the user. Unless I am missing something?

> I've never seen Google worry about them.
Of course they are really important. Otherwise our security efforts are firing without aim.
From a quick Google search, in the abstract of Chromium OS sec overview.
https://sites.google.com/a/chromium.org/dev/chromium-os/chromiumos-design-docs/security-overview

Zarel: > But they are opt-in for the user. Unless I am missing something?

Yes, they're opt-in for the user. Why is that a problem? _Most_ two-factor auth is opt-in for the user.

> Of course they are really important. Otherwise our security efforts are firing without aim.

The page you link to makes no mention of the threat models you've been talking about.

Zarel: Once again:

> The lessons from the past (rootinabox and last year's incident) would say that the real-world security threat lies in the login server itself. Therefore, would it be possible to validate Google logins directly in the sim server?

Our biggest threat model is "attacker takes down the login server".

Protecting the sim server against the login server is **useless** because **our goal is to protect the login server**. There is nothing valuable in the sim server, everything of value is in the login server. The sim server can't do anything without a login server anyway!

Slayer95: The problem is that, as you have stated, we care specifically about admin credentials, and this:
- Is unsafe by default, by which I mean "as safe as it already is", unless -again- Google authentication process is otherwise somehow safer than ours', an idea with which I can live, but should be stated clearly.
- Relies on third party authentication methods.
- Empowers a third party which may act as a opportunistic adversary in the adminship context (Would be fixed by still requiring the PS password)

Slayer95: > There is nothing valuable in the sim server, everything of value is in the login server.

Years worth of modlogs and chatlogs. Quite valuable probably.
We can probably dismiss the trivia questions DB (is there even a command to dump it?), as well as the dumpable data stored sim-server-side for all side servers.

Zarel: We can't _force_ admins to be safe, but we've never been able to. Our policy will be to tell admins to enable 2FA on Google, maybe require screenshots of 2FA being enabled, and that will always be the best we can do. Policies that make it harder for admins to circumvent things like this are of very marginal value.

Relying on third-party authentication methods is a good thing when the third party is Google. They say "don't roll your own crypto" and I tend to generally agree with that.

I also don't consider "Google wants to take over PS" as a particularly worrying threat model. I suppose it's something I'm willing to live with.

AustinXII: Are they valuable though? Chat logs and modlogs probably aren't overly important imo. Regular users can store chat logs and I don't see why a hacker would ever want the modlogs. I suppose the dumpable data from side servers could be an issue though

Zarel: If your recovery email is a Gmail account, you already trust Google with basically everything. And I'd argue this is a good thing – I trust Google to secure an email server much more than I trust myself or any other company to.

Slayer95: > Are they valuable though? Chat logs and modlogs probably aren't overly important imo. Regular users can store chat logs and I don't see why a hacker would ever want the modlogs.

I am sure you could get a few zeros from mining/selling PS logs.

> If your recovery email is a Gmail account, you already trust Google with basically everything. And I'd argue this is a good thing – I trust Google to secure an email server much more than I trust myself or any other company to.

I barely rely on Google/Gmail myself, but I can see where you are coming from.

> They say "don't roll your own crypto" and I tend to generally agree with that.

This isn't "our own crypto". We use bcrypt, don't we?

> We can't force admins to be safe, but we've never been able to. Our policy will be to tell admins to enable 2FA on Google, maybe require screenshots of 2FA being enabled, and that will always be the best we can do. Policies that make it harder for admins to circumvent things like this are of very marginal value.
Relying on third-party authentication methods is a good thing when the third party is Google. They say "don't roll your own crypto" and I tend to generally agree with that.
I also don't consider "Google wants to take over PS" as a particularly worrying threat model. I suppose it's something I'm willing to live with.

With that out of the way, lgtm.

AustinXII: Oh I suppose that's possible, I'm totally forgetting about the recent PSRevealer incident

Zarel: The sim server's logs are valuable, but nowhere near as valuable as the data in the login server.

I'm not opposed to adding more checks in the sim server, but for now I consider this pullreq a strict improvement in security, barring, like, actual vulns.

Slayer95: Please also remember that the login server data available to admins is the same as to leaders (at least that was the case last time I checked /users panel as admin).

Just for awareness sake.

(Of course, password reset access is very important; however, that doesn't imply data access per se, but in conjunction with the logs/sim server)

PR #1000, opened 2017-09-06
Fix grammar nit in Leppa Berry message

urkerab:


TheImmortal:
The message is taken directly from the game.

Zarel:
n.b. Any message that isn't surrounded by parens is taken directly from the game.

PR #1001, opened 2017-09-06
Leppa Berry only restores 10 PP

urkerab:


Zarel:
Oh, that makes the code less hacky, too!

PR #1003, opened 2017-09-06
Reveal own Ability in Gen 7

urkerab:
Server-side is Zarel/Pokemon-Showdown#3969.

In Gen7 battles, `pokemonData.ability` will (during a choice) now contain the true own Ability, even after Imposter/Transform. If that's not set, then we fall back to `pokemon.ability` and `pokemonData.baseAbility` as usual for previous Gens.

(Previous code is somewhat inconsistent as to whether it's using ids or names, so feel free to insist on one.)


Zarel:
:D

Zarel:
This is basically exactly how I would have implemented this.

PR #1006, opened 2017-09-10
Quickly mute by double-clicking the sound button

urkerab:
Just for convenience.

### DIFF
--- a/js/client-topbar.js
+++ b/js/client-topbar.js
@@ -5,6 +5,7 @@
 			'click a': 'click',
 			'click .username': 'clickUsername',
 			'click button': 'dispatchClickButton',
+			'dblclick button[name=openSounds]': 'toggleMute',
 
 			'dragstart .roomtab': 'dragStartRoom',
 			'dragend .roomtab': 'dragEndRoom',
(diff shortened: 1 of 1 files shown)


js/client-topbar.js
```diff
@@ -44,7 +44,12 @@
 			app.addPopup(LoginPopup);
 		},
 		openSounds: function () {
-			app.addPopup(SoundsPopup);
+			if (!app.addPopup(SoundsPopup)) {
```
Zarel: This seems to happen whenever the sound button is closed, rather than only on double-click?

urkerab: I can measure the time interval between the two clicks if you care to tell me what interval you think is appropriate...

Zarel: You can also just sidestep manually time-intervaling by listening to the `doubleclick` event.

urkerab: I found out where I originally went wrong with that approach - it's called `dblclick`.

js/client-topbar.js
```diff
@@ -5,6 +5,7 @@
 			'click a': 'click',
 			'click .username': 'clickUsername',
 			'click button': 'dispatchClickButton',
+			'dblclick .volume': 'toggleMute',
```
Zarel: `dblclick button[name=openSounds]`

You don't need the `.volume` class.

urkerab: Sorry, I'm unfamiliar with Backbone's View system (I had to look it up just to find out what it was doing) so I resorted to blithely cutting and pasting...

PR #1007, opened 2017-09-15
Fix tooltips

MacChaeger:


Zarel:
Why are we removing these? What's wrong with them?

MacChaeger:
They aren't BP boosts, but damage boosts instead. Who knew? (apparently
everyone did except me)

urkerab:
They were also wrong when the user had an -ate or -ize Ability.

PR #1008, opened 2017-09-16
README: Add Travis Widget

DeathlyPlays:


Zarel:
How much do people actually want this? If the build ever fails, it usually gets fixed within minutes, mostly because I get an email about it.

Zarel:
I guess it looks good, but considering that the "build" is only a linter and some basic tests, it might be kind of deceptive to have it there, too...

DeathlyPlays:
Yeah, it was really just a proposal

Zarel:
...does anyone else have opinions?

Slayer95:
Deceptive indeed.

DeathlyPlays:
Deceptive...? How.... :/ Thought we just added things to make the server a better place.....

Zarel:
Not your pull request, but the Travis icon itself would be deceptive.

DeathlyPlays:
Ah gotcha....

PR #1009, opened 2017-09-19
Tooltips: Fix Twisted Spoon

MacChaeger:
It has a space in newer generations.

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -1290,6 +1290,7 @@ var BattleTooltips = (function () {
 		'Soft Sand': 'Ground',
 		'Spell Tag': 'Ghost',
 		'TwistedSpoon': 'Psychic'
+		'Twisted Spoon': 'Psychic'
 	};
 	var orbUsers = {
 		'Latias': 'Soul Dew',
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -1289,7 +1289,6 @@ var BattleTooltips = (function () {
 		'SilverPowder': 'Bug',
 		'Soft Sand': 'Ground',
 		'Spell Tag': 'Ghost',
-		'TwistedSpoon': 'Psychic'
 		'Twisted Spoon': 'Psychic'
 	};
 	var orbUsers = {


Zarel:
You forgot to remove the old entry...

MacChaeger:
I left it in because it didn't have a space in the name until Gen 6, but it
seems like older generation mods don't change it back to one word. Should
they?

Zarel:
They intentionally don't. Also you left it in but also didn't add a `,` which is why the build is failing, if you didn't notice.

Zarel:
Either way, I'd recommend taking out the old line entirely.

urkerab:
> Also you left it in but also didn't add a , which is why the build is failing

We don't enforce trailing `,` on the client like we do on the server?

Zarel:
We can't, Safari 5 (?) and IE7 don't support them.

...I guess in theory we can drop support for those by now. The last time we got a bug report about breaking Safari 5 compatibility was like 4 years ago.

Zarel:
Safari 5 is I think relevant for being the most recent version of Safari supported on the iPad 1.

Zarel:
I wouldn't be surprised if the Wii/3DS browser didn't support them either, though.

AustinXII:
The 3DS doesn't even support JS

Zarel:
Weird, I thought I heard someone talking about playing PS on the 3DS browser.

Anyway, I just researched and can confirm that Safari 5 doesn't support trailing commas, and that Safari 5 is the most recent version of Safari supported on the original iPad.

AustinXII:
I've actually tried, but sadly it's impossible on the 3DS :(

urkerab:
> We can't, Safari 5 and IE7 don't support them.

The code that sets the following variables currently uses trailing commas: `BattleTooltips.zMoveTable`, `BattleTooltips.orbUsers`, `BattleTooltips.orbTypes`, `Config.groups`. Do you want those places fixed?

> ...I guess in theory we can drop support for those by now. The last time we got a bug report about breaking Safari 5 compatibility was like 4 years ago.

... or not?

Zarel:
> The code that sets the following variables currently uses trailing commas:

...I thought our linter checked for that...

Zarel:
...why is Travis not catching this?

PR #1010, opened 2017-09-22
Be less confused when Illusion is involved

### LINKED ISSUE #987: Illusion
Illusion screws with the client two ways:

Say you have a Zoroark, Illusioned as a Venusaur.

- If Zoroark faints from move damage, both Zoroark and Venusaur will be treated by the client as fainted.
- If Zoroark faints from residual damage, Venusaur will be treated by the client as fainted.

General plan to deal with this:

- Venusaur should be brought back to life (full HP, no status) if it's revealed to be Zoroark during the faint process (move damage case)
- Zoroark should be marked as fainted if Venusaur is switched in after it faints (residual damage case)

This can't cover Hackmons. I'm thinking of a `|legality|` message to cover things like Species Clause and whether or not hacked mons are allowed, which will give the client more information to work with in terms of what it can assume does and doesn't have Illusion.

QuiteQuiet:
First draft to close #987

I'm not sure this entirely complete as it currently is, so consider it a first draft.

The thing that confuses me is how to sort out Species Clause-less metagames, or metagames where Illusion can be on anything. For now this completely ignores the second part, and marks the Illusion's in order for the first. Ordered marking for Species Clause-less meta games are most likely not ideal, so I'm open to better ideas.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -1734,6 +1734,7 @@ var Side = (function () {
 			var existingTable = {};
 			for (var i = 0; i < this.pokemon.length; i++) {
 				var poke1 = this.pokemon[i];
+				if (!poke1.searchid) continue;
 				if (poke1.searchid in existingTable) {
 					var j = existingTable[poke1.searchid];
 					var poke2 = this.pokemon[j];
@@ -1744,7 +1745,8 @@ var Side = (function () {
 					} else if (poke1.fainted && !poke2.fainted) {
 						this.pokemon.splice(j, 1);
 					} else {
-						this.pokemon.splice(i, 1);
+						this.pokemon.splice(j, 1);
+						poke.replacedFaintedIllusion = poke2.fainted;
 					}
 					break;
 				}
@@ -1802,6 +1804,28 @@ var Side = (function () {
 		pokemon.statbarElem = this.battle.statElem.children().last();
 		this.updateStatbar(pokemon, true);
 		pokemon.side.updateSidebar();
+
+		if (pokemon.replacedFaintedIllusion && !this.battle.hackedPokemon) {
+			var couldIllusion;
+			for (var i = 0; i < this.pokemon.length; i++) {
+				var poke = this.pokemon[i];
+				for (var i in poke.abilities) {
+					if (!poke.fainted && poke.abilities[i] === 'Illusion') {
+						couldIllusion = poke;
+					}
+				}
+				if (couldIllusion) break;
+			}
+			if (couldIllusion) {
+				// we know `something` had Illusion and fainted before, but not exactly what
+				// so mark the first found Illusion-legal Pokemon as fainted.
+				couldIllusion.hp = 0;
+				couldIllusion.fainted = true;
+
+				pokemon.side.updateSidebar();
+			}
+		}
+
 		if (this.battle.fastForward) {
 			pokemon.statbarElem.css({
 				display: 'block',
@@ -2346,7 +2370,8 @@ var Battle = (function () {
 		this.lastMove = '';
 		this.gen = 7;
 		this.maxPokemon = 6;
-		this.speciesClause = false;
+		this.speciesClause = true;
+		this.hackedPokemon = false;
 
 		this.frameElem = frame;
 		this.logFrameElem = logFrame;
@@ -6303,9 +6328,6 @@ var Battle = (function () {
 			for (var i in kwargs) args[1] += '[' + i + '] ' + kwargs[i];
 			this.log('<div style="padding:5px 0"><small>Format:</small> <br /><strong>' + Tools.escapeHTML(args[1]) + '</strong></div>');
 			this.tier = args[1];
-			if (this.tier === 'Random Battle') {
-				this.speciesClause = true;
-			}
 			break;
 		case 'gametype':
 			this.gameType = args[1];
@@ -6331,6 +6353,13 @@ var Battle = (function () {
 		case 'rule':
 			var ruleArgs = args[1].split(': ');
 			this.log('<div><small><em>' + Tools.escapeHTML(ruleArgs[0]) + (ruleArgs[1] ? ':' : '') + '</em> ' + Tools.escapeHTML(ruleArgs[1] || '') + '</div>');
+			break;
+		case 'legality':
+			// default assumption if not sent is:
+			// - Species Clause is in effect
+			// - No hacked Pokemon
+			this.speciesClause = args.indexOf('nospeciesclause') < 0;
+			this.hackedPokemon = args.indexOf('hackmons') >= 0;
 			break;
 		case 'rated':
 			this.rated = true;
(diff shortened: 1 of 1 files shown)


js/battle.js
```diff
@@ -1729,11 +1729,12 @@ var Side = (function () {
 			this.pokemon.push(poke);
 		}
 		if (replaceSlot === -2) this.battle.maxPokemon = this.pokemon.length;
-		if (this.pokemon.length > this.battle.maxPokemon || this.battle.speciesClause) {
```
Zarel: Why though?

js/battle.js
```diff
@@ -1802,6 +1804,26 @@ var Side = (function () {
...
+			if (pokemon.fainted) pokemon.fainted = false; // not anymore!
+			var couldIllusion = [];
+			for (var i = 0; i < this.pokemon.length; i++) {
+				var poke = this.pokemon[i];
+				for (var i in poke.abilities) {
+					if (!poke.fainted && poke.abilities[i] === 'Illusion') couldIllusion.push(poke);
+				}
+			}
+			if (couldIllusion.length) {
+				// we know `something` had Illusion and fainted before, but not exactly what
```
Zarel: No we don't. We're not inside the `if (pokemon.fainted)` block.

js/battle.js
```diff
@@ -1802,6 +1804,26 @@ var Side = (function () {
...
+				// so mark the first found Illusion-legal Pokemon as fainted.
+				couldIllusion[0].hp = 0;
```
Zarel: If we're only marking the first Illusion as fainted, we don't need to collect an array, we can just collect the first candidate and `break`.

js/battle.js
QuiteQuiet: Keeping `this.battle.speciesClause` broke team preview. I'll see if I can solve that some other way...

Zarel: ...why though?

QuiteQuiet: From what I could tell: because every team preview tier had it as false before, so when Pokemon got added with `|poke|` through team preview and `this.pokemon.length === this.battle.maxPokemon`, it would skip this check because it was false. When default is true (from now), and every searchid still is `''` it get loads of false positives.

QuiteQuiet: `if (!poke1.searchid) continue;` may or may not have already fixed that, though, but it might also have other side effects?

Zarel: It should skip searchids that are `''` though

Zarel: That's not necessarily the right solution...

js/battle.js
```diff
@@ -1729,11 +1729,12 @@ var Side = (function () {
 			this.pokemon.push(poke);
 		}
 		if (replaceSlot === -2) this.battle.maxPokemon = this.pokemon.length;
-		if (this.pokemon.length > this.battle.maxPokemon || this.battle.speciesClause) {
+		if (this.pokemon.length > this.battle.maxPokemon) {
 			// check for Illusion
 			var existingTable = {};
 			for (var i = 0; i < this.pokemon.length; i++) {
 				var poke1 = this.pokemon[i];
+				if (!poke1.searchid) continue;
```
Zarel: Yeah, I think this is good...

js/battle.js
Zarel: I thought it through and I don't think it'd have other side effects.

js/battle.js
```diff
@@ -1802,6 +1804,41 @@ var Side = (function () {
 		pokemon.statbarElem = this.battle.statElem.children().last();
 		this.updateStatbar(pokemon, true);
 		pokemon.side.updateSidebar();
+
+		if (pokemon.replacedIllusion && !this.battle.hackedPokemon) {
+			if (this.pokemon.indexOf(pokemon) < 0) {
+				// if the pokemon that is switching in isn't in the team, match it to one that is
```
QuiteQuiet: I don't think this is the best way to do this, but I don't know how much other things I affect if I would edit `newPokemon` more, and I don't think it's relevant anywhere else.

QuiteQuiet:
Did the other stuff too.

js/battle.js
```diff
@@ -1802,6 +1804,41 @@ var Side = (function () {
 		pokemon.statbarElem = this.battle.statElem.children().last();
 		this.updateStatbar(pokemon, true);
 		pokemon.side.updateSidebar();
+
+		if (pokemon.replacedIllusion && !this.battle.hackedPokemon) {
+			if (this.pokemon.indexOf(pokemon) < 0) {
```
Zarel: As far as I'm aware, this will never be true?

js/battle.js
```diff
@@ -1802,6 +1804,41 @@ var Side = (function () {
...
+				// if the pokemon that is switching in isn't in the team, match it to one that is
+				for (var i = 0; i < this.pokemon.length; i++) {
+					if (pokemon.searchid === this.pokemon[i].searchid) {
+						pokemon.fainted = this.pokemon[i].fainted;
+						this.pokemon[i] = pokemon;
+						break;
+					}
+				}
+			}
+			if (pokemon.fainted) {
```
Zarel: This will also never be true, because this is after `getPokemon` overwrites the old faint status.

js/battle.js
QuiteQuiet: This is true if a Pokemon with Illusion faints without revealing itself to be an Illusion.

From what I can tell something like the following happens:
- `Side#newPokemon` creates the new pokemon that switches in and adds it to the team.
- Finds a duplicate on the team, and removes the new Pokemon that it just added to fix this.
- Returns the new Pokemon it just removed from the team.

js/battle.js
QuiteQuiet: It's related to the above regarding the `newPokemon` call in `getPokemon`, but right now it doesn't override the faint status (though, that might what is intended?)

js/battle.js
Zarel: If it finds a duplicate on the team, it should remove the old Pokemon, not the new Pokemon it just added...

QuiteQuiet: That's how `newPokemon` works right now though with regards to Illusion.

https://github.com/Zarel/Pokemon-Showdown-Client/blob/master/js/battle.js#L1747 In every case I can figure out, this splices out the last duplicate it finds, which for `pokemon.length > maxPokemon` will always be the new Pokemon it just pushed onto the team.

Zarel: Oh, that's entirely unintended. I wonder if that's the cause of a number of our bugs. It should definitely be splicing out the old copy.

QuiteQuiet: Double checking before I make any other changes, but isn't this logic also reversed here https://github.com/Zarel/Pokemon-Showdown-Client/blob/master/js/battle.js#L1745? If `poke1` is fainted but `poke2` isn't, why is `poke2` the one that gets spliced out and the fainted one remains?

Zarel: That's... probably a mistake.

Zarel: Wait, no, that's right.

Zarel: The number of fainted Pokémon should be kept the same. We assume that the fainted Pokémon is the one that's not an Illlusion, and if we're wrong, we mark a Zoroark as fainted or something.

QuiteQuiet:
This might be fine then?

Zarel:
I ended up fixing this myself; see #987

QuiteQuiet:
Do you want to close https://github.com/Zarel/Pokemon-Showdown/pull/4006 as well then?

Zarel:
Yeah, good catch.

PR #1012, opened 2017-09-27
Update tooltips

MacChaeger:
- Show Sheer Cold's real accuracy in gen 7
- Show Brine's BP
- Use the real move type when determining base power (except when checking for -ate boosts)
- Misty Terrain checks whether the target is grounded, not the user
- Use the Pokemon's real type when checking for Sandstorm's spd boost
- Steelworker, Water Bubble, Blaze, Overgrow, Swarm, and Torrent boost attacking stats, not base power
- Aerilate, Galvanize, Pixilate, and Refrigerate do not affect the moves Judgment, Multi-Attack, Techno Blast, or Weather Ball, regardless of item held.
- Gen 6 and 7 randdubs use the same ev spreads as their singles counterparts

I can't help but feel like all those `move.id` checks in the base power check should be an else if chain.

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -881,7 +881,7 @@ var BattleTooltips = (function () {
 		}
 		// Other abilities that change the move type.
 		if ('sound' in move.flags && ability === 'Liquid Voice') moveType = 'Water';
-		if (moveType === 'Normal' && move.category && move.category !== 'Status' && !(move.id in {judgment:1, multiattack:1, naturalgift:1, revelationdance:1, technoblast:1, weatherball:1})) {
+		if (moveType === 'Normal' && move.category && move.category !== 'Status' && !(move.id in {judgment: 1, multiattack: 1, naturalgift: 1, revelationdance: 1, struggle: 1, technoblast: 1, weatherball: 1})) {
 			if (ability === 'Aerilate') moveType = 'Flying';
 			if (ability === 'Galvanize') moveType = 'Electric';
 			if (ability === 'Pixilate') moveType = 'Fairy';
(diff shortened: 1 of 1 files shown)


js/client-battle-tooltips.js
```diff
@@ -1138,20 +1147,17 @@ var BattleTooltips = (function () {
...
 			(ability === 'Technician' && basePower <= 60) ||
 			(ability === 'Toxic Boost' && (pokemon.status === 'psn' || pokemon.status === 'tox') && move.category === 'Physical') ||
-			(ability === 'Overgrow' && move.type === 'Grass' && pokemonData.hp * 3 <= pokemonData.maxhp) ||
-			(ability === 'Blaze' && move.type === 'Fire' && pokemonData.hp * 3 <= pokemonData.maxhp) ||
-			(ability === 'Torrent' && move.type === 'Water' && pokemonData.hp * 3 <= pokemonData.maxhp) ||
-			(ability === 'Swarm' && move.type === 'Bug' && pokemonData.hp * 3 <= pokemonData.maxhp)) {
+			(ability === 'Overgrow' && moveType === 'Grass' && pokemonData.hp * 3 <= pokemonData.maxhp) ||
+			(ability === 'Blaze' && moveType === 'Fire' && pokemonData.hp * 3 <= pokemonData.maxhp) ||
+			(ability === 'Torrent' && moveType === 'Water' && pokemonData.hp * 3 <= pokemonData.maxhp) ||
+			(ability === 'Swarm' && moveType === 'Bug' && pokemonData.hp * 3 <= pokemonData.maxhp)) {
```
Marty-D: These four also behave like Steelworker and Water Bubble. Probably some new way of identifying when they're active should be implemented one day. Then again, I'll take "not giving out false information" over "handholding the games don't even do" any day.

MacChaeger: Coulda sworn I checked those, but I must not have, because this is clearly wrong.

js/client-battle-tooltips.js
```diff
@@ -879,7 +880,7 @@ var BattleTooltips = (function () {
 		}
 		// Other abilities that change the move type.
 		if ('sound' in move.flags && ability === 'Liquid Voice') moveType = 'Water';
-		if (moveType === 'Normal' && move.category && move.category !== 'Status' && !(move.id in {'naturalgift': 1, 'struggle': 1})) {
+		if (moveType === 'Normal' && move.category && move.category !== 'Status' && !(move.id in {judgment:1, multiattack:1, naturalgift:1, revelationdance:1, technoblast:1, weatherball:1})) {
```
Marty-D: Struggle still needs to be in this list and the one below. The server doesn't need it because Struggle is ??? type by the time it gets to that point, so isn't affected anyway.

MacChaeger: Whoops, didn't realize struggle got removed there. Also, are judgment, etc. still affected by -ate in gen 6? Cuz if they are I need to change this

Marty-D: Nah, same behaviour.

MacChaeger:
To what extent should Gyro Ball and Electro Ball's BP range account for speed modifiers to the user and target?

MacChaeger:
Why the radio silence?

Zarel:
Sorry, missed this.

PR #1014, opened 2017-09-29
Fix crash on browser console when typing "/r "

Hidden50:
`e.setSelectionRange` is a typo, `setSelectionRange` is a method of the textbox element.

Results in `Uncaught TypeError: e.setSelectionRange is not a function` when the user types in ``"/r "``.


Hidden50:
Errors like this would be a lot more obvious if browser javascript didn't crash silently :\

Zarel:
We actually do log browser JavaScript errors, we've just started to ignore them since the vast majority of browser JavaScript errors are malware crashing.

PR #1016, opened 2017-10-01
Teambuilder Hack for Dual Wielding

Spandamn:
Do not merge; The code for Dual Wielding is not ready yet
[OMotM October](http://www.smogon.com/forums/threads/om-of-the-month-october-winner-dual-wielding.3616125/#post-7527701)

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -2567,7 +2567,10 @@
 				val = (id in BattlePokedex ? BattlePokedex[id].species : '');
 				break;
 			case 'ability':
-				val = (id in BattleAbilities ? BattleAbilities[id].name : '');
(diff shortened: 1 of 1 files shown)


Zarel:
...why not? This doesn't hurt to merge early, does it?

Spandamn:
it doesn't but the only thing holding it back is the fact the the code for dual wielding is incomplete.
If it isn't completed, then another OMotM would have to be chosen, removing the need for this PR

Spandamn:
Actually merge it @Zarel, it'd be easy for teambuilders while testing

Zarel:
Code style is wrong, we require braces for multiline if/else blocks.

PR #1017, opened 2017-10-02
Tooltips: Track sleep and toxic turns

QuiteQuiet:
Doesn't cover gen 1 Toxic/Leech Seed interaction or Tox -> Psn in earlier gens, but should otherwise cover everything.

Sidenote: What is hardcoreMode?

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -410,7 +410,13 @@ var BattleTooltips = (function () {
 			var exacthp = '';
 			if (pokemonData) exacthp = ' (' + pokemonData.hp + '/' + pokemonData.maxhp + ')';
 			else if (pokemon.maxhp == 48) exacthp = ' <small>(' + pokemon.hp + '/' + pokemon.maxhp + ' pixels)</small>';
-			text += '<p>HP: ' + pokemon.hpDisplay() + exacthp + (pokemon.status ? ' <span class="status ' + pokemon.status + '">' + pokemon.status.toUpperCase() + '</span>' : '') + '</p>';
+			text += '<p>HP: ' + pokemon.hpDisplay() + exacthp + (pokemon.status ? ' <span class="status ' + pokemon.status + '">' + pokemon.status.toUpperCase() + '</span>' : '');
+			if (pokemon.status === 'tox') {
+				text += ' Next damage: ' + Math.floor(100 / 16) * Math.min(pokemon.statusData.toxicTurns, 15) + '%';
+			} else if (pokemon.status === 'slp') {
+				text += ' Turns asleep: ' + pokemon.statusData.sleepTurns;
+			}
+			text += '</p>';
 		}
 		var showOtherSees = !this.battle.hardcoreMode && isActive;
 		if (pokemonData) {
--- a/js/battle.js
+++ b/js/battle.js
@@ -202,6 +202,7 @@ var Pokemon = (function () {
 
 		// [[moveName, ppUsed]]
 		this.moveTrack = [];
+		this.statusData = {sleepTurns: 0, toxicTurns: 0};
 
 		this.name = '';
 		this.species = '';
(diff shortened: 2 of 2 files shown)


Zarel:
Hardcore mode hides things not available in-game, to help people practicing for VGC.

js/client-battle-tooltips.js
```diff
@@ -518,6 +518,12 @@ var BattleTooltips = (function () {
 			}
 			text += '</p>';
 		}
+		if (!this.battle.hardcoreMode && (pokemon.sleepTurns || pokemon.toxicTurns)) {
+			text += '<p class="section">';
+			if (pokemon.sleepTurns) text += 'Turns asleep: ' + pokemon.sleepTurns + '<br />';
+			if (pokemon.toxicTurns) text += 'Next Toxic damage: ' + pokemon.toxicTurns + '/16<br />';
```
Marty-D: I would prefer if we didn't perpetuate the common misconception that Toxic damage is a flat X/16 when it's really floor(1/16) * X, where X caps at 15 turns.

Slayer95: We should probably just compute the value, right?

QuiteQuiet: So `Math.floor(Math.min(pokemon.toxicTurns, 15) / 16 * 100)`?

QuiteQuiet: So `(Math.floor(pokemon.maxhp / 16) * Math.min(pokemon.toxicTurns, 15)) / pokemon.maxhp`?

Zarel: We don't have enemy Pokémon maxhp, so something subtler.

Zarel:
I'm not sure I like status state so high up here...

QuiteQuiet:
Do you have anything else you'd prefer over having it at the bottom of the hover-over tooltip?

js/client-battle-tooltips.js
QuiteQuiet: Using 100 would work instead, for an estimate on both sides.

    (Math.floor(100 / 16) * Math.min(pokemon.toxicTurns, 15)) / 100

or for displaying it as a percentage instead of a decimal, leave out the `/ 100` bit.

Zarel:
Oh sorry, I mean in the pokemon object, not in the UI. Should probably be in `statusData` the way it is in the server.

It's also in a bad place in the UI. That one should probably be next to the actual status.

QuiteQuiet:
Is this better?

js/client-battle-tooltips.js
```diff
@@ -410,7 +410,13 @@ var BattleTooltips = (function () {
 			var exacthp = '';
 			if (pokemonData) exacthp = ' (' + pokemonData.hp + '/' + pokemonData.maxhp + ')';
 			else if (pokemon.maxhp == 48) exacthp = ' <small>(' + pokemon.hp + '/' + pokemon.maxhp + ' pixels)</small>';
-			text += '<p>HP: ' + pokemon.hpDisplay() + exacthp + (pokemon.status ? ' <span class="status ' + pokemon.status + '">' + pokemon.status.toUpperCase() + '</span>' : '') + '</p>';
+			text += '<p>HP: ' + pokemon.hpDisplay() + exacthp + (pokemon.status ? ' <span class="status ' + pokemon.status + '">' + pokemon.status.toUpperCase() + '</span>' : '');
+			if (pokemon.status === 'tox') {
+				text += ' Next damage: ' + Math.floor(100 / 16) * Math.min(pokemon.statusData.toxicTurns, 15) + '%';
+			} else if (pokemon.status === 'slp') {
+				text += ' Turns asleep: ' + pokemon.statusData.sleepTurns;
+			}
```
Zarel: Don't show these in hardcore mode.

QuiteQuiet: It doesn't, there's a check for hardcoreMode above this that makes hardcoreMode never enter this section. https://github.com/Zarel/Pokemon-Showdown-Client/blob/master/js/client-battle-tooltips.js#L408

Zarel: Oh okay.

PR #1018, opened 2017-10-03
 Fix Hidden Power IVs Spread during import for Gen7

Nephtyrix:
js/client-teambuilder.js
        Add this.curTeam in importTeam function
    js/storage.js
        Update importTeam function to get curTeam argument
        Update importTeam function to test gen for IVs spread

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1428,7 +1428,7 @@
 		},
 		savePokemonImport: function (i) {
 			i = +(this.$('li').attr('value'));
-			var curSet = Storage.importTeam(this.$('.pokemonedit').val())[0];
+			var curSet = Storage.importTeam(this.$('.pokemonedit').val(), '', this.curTeam)[0];
 			if (curSet) {
 				this.curSet = curSet;
 				this.curSetList[i] = curSet;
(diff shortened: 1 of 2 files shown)


Zarel:
This is an interesting idea, but it's probably the wrong way to do it.

Zarel:
To clarify, I really appreciate the effort you put in here.

Unfortunately, passing moe and more parameters around just complicates the code. An ideal solution would be to have a concept of default and non-default IVs.

Nephtyrix:
Ok, I clearly understand, the only one problem who attempt on this is the grouped import function between import a single pokemon on TeamBuilder and import all team from backup, the last one get the Gen (===[Gen7ou]===) so this issue don't occur, but not on a single pokemon. 
I told that cause the concept of default and non-default IVs is based on Gen. 
I know, we can't add more parameters. I need to find an other solution for that !

Nephtyrix:
Ok, I clearly understand, the only one problem who attempt on this is the grouped import function between import a single pokemon on TeamBuilder and import all team from backup, the last one get the Gen (===[Gen7ou]===) so this issue don't occur, but not on a single pokemon. 
I told that cause the concept of default and non-default IVs is based on Gen. 
I know, we can't add more parameters. I need to find an other solution for that !

PR #1019, opened 2017-10-03
Don't merge normal move boosts in with the Z boost

urkerab:
If you use e.g. Z-Cosmic Power, then you should get 'Clefable boosted its Special Defense using its Z-Power! Clefable's Defense rose! Clefable's Special Defense rose!'.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -3847,7 +3847,7 @@ var Battle = (function () {
 						break;
 					}
 				} else if (kwargs.zeffect) {
-					if (minors.length) {
+					if (minors.length && minors[0][1].zeffect) {
 						actions += "" + poke.getName() + " boosted its stats" + amountString + " using its Z-Power! ";
 						for (var i = 0; i < minors.length; i++) {
 							minors[i][1].silent = '.';


urkerab:
I've just seen a video of Z-Toxic, and the Defense boost happens _before_ the move is Used anyway, but that's a server-side bug.

Also Z-Toxic is still described as full-force Z-move on the cart, which is cute, since by the time the message appears, the Z-move has lost all of its force!

Zarel:
Um, wait, why is there still `kwargs.zeffect` if it's not a z-effect? Is the server sending bad code?

urkerab:
When you use Z-Cosmic Power, you get three boosts, a Z-Effect Special Defense boost, and then regular Defense and Special Defense boosts.

Zarel:
So the regular Def and SpD boosts shouldn't be sending `kwargs.zeffect`... that sounds like a server bug to me.

urkerab:
> So the regular Def and SpD boosts shouldn't be sending kwargs.zeffect

They weren't, but the client wasn't checking, instead it just assumed that if the first boost was a Z-boost then so were the others. I might not have noticed before because ROM's code is slightly closer to the in-game behaviour where you get the Z-boosts before the move is used and therefore the client doesn't attempt to aggregate the boosts.

PR #1021, opened 2017-10-05
Dispatch fragment after processing autojoins

urkerab:
Currently fragment dispatching happens before autojoins. This means that if you're in a battle but need to reload (e.g. due to an interrupted connection) you rejoin the battle before attempting autojoins and they fail because they're now too late.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -411,6 +411,9 @@
 						}
 					}
 					app.send('/autojoin ' + autojoinIds.join(','));
+
+					// HTML5 history throws exceptions when running on file://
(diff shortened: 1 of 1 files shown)


Zarel:
Should happen simultaneously with autojoins imo, but this works too.

Zarel:
This is probably still non-ideal, but oh well. :/

AustinXII:
Regardless, thank you Urkerab this has been a huge annoyance for me personally

urkerab:
Also happens to work around an annoying bug whereby if you had to reload during a battle then it would not play any background music.

PR #1024, opened 2017-10-14
Tooltips: Add exceptions to Toxic damage calcs

KrisOnline:


js/client-battle-tooltips.js
```diff
@@ -415,7 +415,7 @@ var BattleTooltips = (function () {
 			else if (pokemon.maxhp == 48) exacthp = ' <small>(' + pokemon.hp + '/' + pokemon.maxhp + ' pixels)</small>';
 			text += '<p>HP: ' + pokemon.hpDisplay() + exacthp + (pokemon.status ? ' <span class="status ' + pokemon.status + '">' + pokemon.status.toUpperCase() + '</span>' : '');
 			if (pokemon.statusData) {
-				if (pokemon.status === 'tox') {
+				if (pokemon.status === 'tox' && !(pokemon.ability in {poisonheal:1, magicguard:1})) {
```
Zarel: Is `pokemon.ability` an ability ID? I thought it was an ability name...

KrisOnline: oh shit u right

MacChaeger:
I kinda don't like this because Toxic's counter still increments even if the user isn't taking damage from it, so if the ability is replaced mid-turn a badly poisoned Poison Heal or Magic Guard Pokémon can take serious damage, and it would be helpful for users to know just how much it _would_ take, rather than be indirectly reminded what their ability does.

MacChaeger:
I understand that the same could be said of lots of tooltip related stuff, but this is a particular case where the non-ability number is silently changing every turn in a potentially deadly way.

KrisOnline:
@MacChaeger i can make it so that the counter still increments but doesnt appear as long as the Pokemon's ability is one of the two exceptions

MacChaeger:
I'm saying it should still appear even with those abilities so someone can tell how much they would take if their ability were suddenly removed (or suppressed, which isn't currently tracked by the client).

QuiteQuiet:
Reword it to ``Would take: xx%`` if Magic Guard/Poison Heal? It would be the most accurate and also won't hide potentially relevant information.

js/client-battle-tooltips.js
```diff
@@ -415,7 +415,7 @@ var BattleTooltips = (function () {
 			else if (pokemon.maxhp == 48) exacthp = ' <small>(' + pokemon.hp + '/' + pokemon.maxhp + ' pixels)</small>';
 			text += '<p>HP: ' + pokemon.hpDisplay() + exacthp + (pokemon.status ? ' <span class="status ' + pokemon.status + '">' + pokemon.status.toUpperCase() + '</span>' : '');
 			if (pokemon.statusData) {
-				if (pokemon.status === 'tox') {
+				if (pokemon.status === 'tox' && !['Poison Heal', 'Magic Guard'].includes(pokemon.ability)) {
```
Zarel: Something something `includes` not ES3

PR #1025, opened 2017-10-15
Dark mode CSS – Lighten Teambuilder's Credits

WallyTheBully:
Screenshot before/after : https://prnt.sc/gxkczd


Zarel:
...I would rather remove the credits page entirely, than add even more code to support it.

PR #1026, opened 2017-10-15
Dark mode CSS – Fix Notifications' color

WallyTheBully:
this color should be blue in darkmode 
screenshot : before/after

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -117,7 +117,7 @@ select {
 	color: #F9F9F9;
 }
 .dark a.button.subtle-notifying {
-	color: #AA6600;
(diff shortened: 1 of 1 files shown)


Zarel:
Do you have a screenshot of the change to regular notifications? Right now the only change your screenshot covers is subtle notifications. Also can you paste the screenshot instead of just linking to it? Like this:

image

WallyTheBully:
> Also can you paste the screenshot instead of just linking to it? 


ok will do
This is the notification (the pull request was updated), when I made it I forgot the regular notification and added it right after

PR #1027, opened 2017-10-15
Dark mode CSS – Fix Teambuilder Results (hover)

WallyTheBully:
really hard to read and shouldn't be that white in darkmode, as a consequence it should be updated like this;
screenshot : before/after

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -3384,11 +3384,13 @@ a.ilink.yours {
 .dark .teambuilder-results .result a:hover {
 	border-color: #555555;
 	background: rgba(47, 47, 47, 0.8);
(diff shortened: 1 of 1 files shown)


style/client.css
```diff
@@ -3381,6 +3381,16 @@ a.ilink.yours {
 
 /* teambuilder set */
 
+.dark .teambuilder-results .result a:hover {
+	border-color: #555555;
+	background: rgba(47, 47, 47, 0.8);
```
Zarel: Don't define a background without defining a color...

Zarel:
Cool, thanks!

PR #1028, opened 2017-10-15
Building a Page Loader

Markner:
I want to add a page loader since PS does not have anything new lately lets create more attension i will add the rest off the code

PR #1029, opened 2017-10-15
Create loader.js

Markner:


Zarel:
You appear to have pullreqqed this to the wrong repository? I have no idea what you're trying to do here...

whalemer:
in case you didnt see, it's "part" of #1028

PR #1030, opened 2017-10-15
Dark mode CSS – Fix Select Icons (color)

WallyTheBully:
The teambuilder folderlist update looks cool but I'd like to fix this select icon as it should be dark.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -3318,6 +3318,14 @@ a.ilink.yours {
 	border-color: #CCC;
 }
 
+.dark .select:after {
+	color: #000000;
+}
+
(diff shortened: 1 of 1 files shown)


style/client.css
```diff
@@ -3318,6 +3318,14 @@ a.ilink.yours {
 	border-color: #CCC;
 }
 
+.dark .select:after {
+	color: #000000;
+}
+
+.dark .select:disabled:after {
+	color: #000000;
```
Zarel: ...if it's disabled, it should be grayed out, though...

PR #1031, opened 2017-10-15
Dark mode CSS – Fix Teambuilder Results (color)

WallyTheBully:
this teambuilder result in darkmode is also unreadable and should be fixed (the background should be black with a white text-color);


Zarel:
That's not what the before looks like...

Zarel:
This was already fixed in #1027...

PR #1032, opened 2017-10-15
Dark mode CSS – Teambuilder (Fix Some Colors)

WallyTheBully:
I assume Darkmode CSS would be entirely updated with this pull request. All the elements available in Light mode seems to be modified for Dark mode.
Dark mode could potentially be out of beta and to my mind, it would be better like this. That being said, what are your thoughts?

Screenshot (after update):

Screenshot (before/after):

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -3427,7 +3427,7 @@ a.ilink.yours {
 .dark .utilichart .namecol,
 .dark .utilichart .pokemonnamecol,
 .dark .utilichart .movenamecol {
-	color: #FFF;
+	color: #DDD;
 }
 .dark .utilichart .cur .labelcol em,
 .dark .utilichart .cur .widelabelcol em,
--- a/style/client.css
+++ b/style/client.css
@@ -3381,6 +3381,36 @@ a.ilink.yours {
 
 /* teambuilder set */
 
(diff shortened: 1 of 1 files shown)


style/client.css
```diff
@@ -3381,6 +3381,36 @@ a.ilink.yours {
...
+}
+
+.dark .teambar button:disabled, .dark .teambar button:disabled:hover, .dark .teambar button:disabled:active {
+    color: #ffffff;
+    background: #2d343a;
+    border-color: #6bacc5;
+    opacity: 1;
+}
+
+
```
DeathlyPlays: I'd suggest removing 2 of these blank lines (1 blank line is good for organizational purposes, other than that it becomes a bit of an annoyance to read)

style/client.css
```diff
@@ -3399,7 +3427,7 @@ a.ilink.yours {
 .dark .utilichart .namecol,
 .dark .utilichart .pokemonnamecol,
 .dark .utilichart .movenamecol {
-	color: #DDD;
+	color: #FFF;
```
Zarel: Why?

WallyTheBully: (updated) this color shouldn't change , see below

PR #1035, opened 2017-10-29
Darkmode CSS - fix format and background select button

WallyTheBully:
Darkmode CSS Update
These buttons should be updated for darkmode, I can provide more screenshots if needed

This pull request would update;
FormatSelect Button
AvatarlistSelect Button
BackgroundlistSelect Button

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -3243,13 +3243,16 @@ a.ilink.yours {
 	color: #AAA;
 	box-shadow: none;
 }
+.dark .popupmenu button.sel {
+	border-color: #A9A9A9;
+}
 .dark .popupmenu button:hover,
 .dark .popupmenu button.sel:hover,
 .dark .bglist button:hover,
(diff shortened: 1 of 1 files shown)


Zarel:
Why?

Zarel:
It's intentionally like the current look, I think it looks better that way, and also it's this way "selected" and "hovered" look different.

WallyTheBully:
Basically the current background isn't the exact opposite of darkmode
The border shouldn't be in the same color
In whitemode background has a different color as a consequence the border should also be different in darkmode

Zarel:
Dark mode doesn't need to be identical to a color-flipped light mode. I think it looks better like this.

Zarel:
Closing. I still think the current implementation looks better than what you're changing it to. Especially since with my implementation, it's a lot easier to tell the difference between "selected" and "hovered".

PR #1037, opened 2017-10-30
Add cursed effect animation

Ridaz:
... And a few extras


Zarel:
Do run `npm test` before pullreqqing :p

PR #1038, opened 2017-10-30
Add conditional Useful Moves

TheTiksiBranch:
all species-exclusive move additions should be used/slashed in relevant analyses except Trick Room, which is used in Magearna analysis and in Other Options of others


js/search.js
```diff
@@ -962,12 +962,18 @@
...
 				if (id === 'icywind') isViable = (toId(set.species).substr(0, 6) === 'keldeo');
+				if (id === 'infestation') isViable = (toId(set.species) === 'shuckle');
+				if (id === 'irontail') isViable = (toId(set.species) in {absol:1, flygon:1, haxorus:1, lucario:1, mightyena:1, zygarde:1});
 				if (id === 'jumpkick') isViable = (moves.indexOf('highjumpkick') < 0);
 				if (id === 'leechlife') isViable = (this.gen > 6);
 				if (id === 'reflecttype') isViable = (toId(set.species) in {latias:1, starmie:1});
 				if (id === 'selfdestruct') isViable = (this.gen < 5 && moves.indexOf('explosion') < 0);
 				if (id === 'skyattack') isViable = (toId(set.species) === 'hawlucha');
+				if (id === 'smartstrike') isViable = (toId(set.species) === 'kartana');
+				if (id === 'solarbeam') isViable = (toId(set.ability) === 'drought');
```
Zarel: Chlorophyll?

js/search.js
```diff
@@ -962,12 +962,18 @@
...
+				if (id === 'storedpower') isViable = (toId(set.species) === 'necrozma');
```
Zarel: Espeon? Sigilyph?

js/search.js
```diff
@@ -949,25 +949,38 @@
...
-				if (id === 'aerialace') isViable = (toId(set.ability) === 'technician');
+				if (id === 'aerialace') isViable = (toId(set.species) in {scyther:1, aerodactylmega:1, kricketune:1});
 				if (id === 'bellydrum') isViable = (toId(set.species) in {azumarill:1, linoone:1, slurpuff:1});
-				if (id === 'counter') isViable = (toId(set.species) in {chansey:1, skarmory:1, clefable:1, wobbuffet:1});
+				if (id === 'counter') isViable = (toId(set.species) in {chansey:1, skarmory:1, clefable:1, wobbuffet:1, alakazam:1});
+				if (id === 'curse') isViable = (toId(set.species) === 'snorlax');
+				if (id === 'drainingkiss') isViable = (toId(set.ability) === 'triage');
 				if (id === 'dynamicpunch') isViable = (toId(set.ability) === 'noguard');
+				if (id === 'electroball') isViable = (toId(set.species) === 'raichualola');
+				if (id === 'endeavor') isViable = (toId(set.species) in {mamoswine:1, scolipede:1});
```
Zarel: ...would not manually override this; anything can run a FEAR set...

js/search.js
```diff
@@ -949,25 +949,38 @@
...
 				var isViable = BattleMovedex[id] && BattleMovedex[id].isViable;
-				if (id === 'aerialace') isViable = (toId(set.ability) === 'technician');
+				if (id === 'aerialace') isViable = (toId(set.species) in {scyther:1, aerodactylmega:1, kricketune:1});
 				if (id === 'bellydrum') isViable = (toId(set.species) in {azumarill:1, linoone:1, slurpuff:1});
-				if (id === 'counter') isViable = (toId(set.species) in {chansey:1, skarmory:1, clefable:1, wobbuffet:1});
+				if (id === 'counter') isViable = (toId(set.species) in {chansey:1, skarmory:1, clefable:1, wobbuffet:1, alakazam:1});
+				if (id === 'curse') isViable = (toId(set.species) === 'snorlax');
+				if (id === 'drainingkiss') isViable = (toId(set.ability) === 'triage');
 				if (id === 'dynamicpunch') isViable = (toId(set.ability) === 'noguard');
+				if (id === 'electroball') isViable = (toId(set.species) === 'raichualola');
```
Zarel: ??? Does Raichu-Alola actually use it? It needs to be three times faster than its opponent to actually outclass Thunderbolt. Even with Surge Surfer, it seems too inconsistent to be useful...

js/search.js
```diff
@@ -949,25 +949,38 @@
...
+				if (id === 'endeavor') isViable = (toId(set.species) in {mamoswine:1, scolipede:1});
 				if (id === 'focuspunch') isViable = (toId(set.species) === 'breloom');
 				if (id === 'gyroball') isViable = (template.baseStats.spe <= 60);
+				if (id === 'headbutt') isViable = (toId(set.species) === 'dunsparce');
```
Zarel: Why not Serene Grace?

AustinXII: Is Togekiss good with headbutt, also does it even get it?

js/search.js
```diff
@@ -949,25 +949,38 @@
...
 				if (id === 'hiddenpowerelectric') isViable = (moves.indexOf('thunderbolt') < 0);
 				if (id === 'hiddenpowerfighting') isViable = (moves.indexOf('aurasphere') < 0 && moves.indexOf('focusblast') < 0);
 				if (id === 'hiddenpowerfire') isViable = (moves.indexOf('flamethrower') < 0);
 				if (id === 'hiddenpowergrass') isViable = (moves.indexOf('energyball') < 0 && moves.indexOf('gigadrain') < 0);
 				if (id === 'hiddenpowerice') isViable = (moves.indexOf('icebeam') < 0 && template.id !== 'xerneas');
 				if (id === 'hypnosis') isViable = (this.gen < 4 && moves.indexOf('sleeppowder') < 0);
 				if (id === 'icywind') isViable = (toId(set.species).substr(0, 6) === 'keldeo');
+				if (id === 'infestation') isViable = (toId(set.species) === 'shuckle');
+				if (id === 'irontail') isViable = (toId(set.species) in {absol:1, flygon:1, haxorus:1, lucario:1, mightyena:1, zygarde:1});
```
Zarel: This list looks long enough that you should be able to figure out a better rule for it, I think.

js/search.js
TheTiksiBranch: I see why taking out mamo but Scolipede doesn't need to fear w/ access to speed boost and spikes so i find it having more merit

js/search.js
TheTiksiBranch: I think dunsparce is the only one to viably run it, the others are either special, have better things to do (jirachi), or are at awkward speed tier to try (para/)flinch (sawsbuck)

js/search.js
Zarel: (The problem with lists is that they usually don't count for LC and OMs that use middle evolutions, and you have to update them every new gen)

js/search.js
```diff
@@ -949,25 +949,38 @@
...
 				if (id === 'jumpkick') isViable = (moves.indexOf('highjumpkick') < 0);
 				if (id === 'leechlife') isViable = (this.gen > 6);
 				if (id === 'reflecttype') isViable = (toId(set.species) in {latias:1, starmie:1});
+				if (id === 'rocktomb') isViable = (toId(set.species) === 'groudon');
 				if (id === 'selfdestruct') isViable = (this.gen < 5 && moves.indexOf('explosion') < 0);
 				if (id === 'skyattack') isViable = (toId(set.species) === 'hawlucha');
+				if (id === 'smackdown') isViable = (toId(set.species) === 'landorustherian');
```
Zarel: Why not any Ground type?

Zarel:
Remember, Useful Moves aren't supposed to be Viable Moves. They're supposed to be any move that isn't outclassed or nearly outclassed in any possible metagame (unmodded, but with any banlist).

js/search.js
Zarel: I mean, I think Endeavor should just always be usable.

js/search.js
Zarel: Usable Moves is expressly not intended to replace viable movelists. You can make the argument that Iron Head mostly outclasses Headbutt on Jirachi, and maybe that Air Slash outclasses it on Togekiss, but Deerling/Sawsbuck can do things with it. And Togepi might be able to do something interesting with a Hustle set.

js/search.js
Zarel: I'd at least test Surge Surfer instead of Raichu

js/search.js
```diff
@@ -949,25 +949,38 @@
...
+				if (id === 'aerialace') isViable = (toId(set.species) in {scyther:1, aerodactylmega:1, kricketune:1});
 				if (id === 'bellydrum') isViable = (toId(set.species) in {azumarill:1, linoone:1, slurpuff:1});
-				if (id === 'counter') isViable = (toId(set.species) in {chansey:1, skarmory:1, clefable:1, wobbuffet:1});
+				if (id === 'counter') isViable = (toId(set.species) in {chansey:1, skarmory:1, clefable:1, wobbuffet:1, alakazam:1});
+				if (id === 'curse') isViable = (toId(set.species) === 'snorlax');
+				if (id === 'drainingkiss') isViable = (toId(set.ability) === 'triage');
 				if (id === 'dynamicpunch') isViable = (toId(set.ability) === 'noguard');
+				if (id === 'electroball') isViable = (toId(set.species) === 'raichualola');
+				if (id === 'endeavor') isViable = (toId(set.species) in {mamoswine:1, scolipede:1});
 				if (id === 'focuspunch') isViable = (toId(set.species) === 'breloom');
```
Zarel: No one else runs SubPunch?

js/search.js
TheTiksiBranch: that's fine with me, I'm planning a req for the always/never usabilities and I'll put it in there

js/search.js
TheTiksiBranch: I like that better

js/search.js
```diff
@@ -949,25 +949,38 @@
...
+				if (id === 'infestation') isViable = (toId(set.species) === 'shuckle');
+				if (id === 'irontail') isViable = (toId(set.species) in {absol:1, flygon:1, haxorus:1, lucario:1, mightyena:1, zygarde:1});
 				if (id === 'jumpkick') isViable = (moves.indexOf('highjumpkick') < 0);
 				if (id === 'leechlife') isViable = (this.gen > 6);
 				if (id === 'reflecttype') isViable = (toId(set.species) in {latias:1, starmie:1});
+				if (id === 'rocktomb') isViable = (toId(set.species) === 'groudon');
 				if (id === 'selfdestruct') isViable = (this.gen < 5 && moves.indexOf('explosion') < 0);
 				if (id === 'skyattack') isViable = (toId(set.species) === 'hawlucha');
+				if (id === 'smackdown') isViable = (toId(set.species) === 'landorustherian');
+				if (id === 'smartstrike') isViable = (toId(set.species) === 'kartana');
```
Zarel: Should probably just detect the presence of Iron Head.

js/search.js
TheTiksiBranch: sure

Zarel:
Also remember: Kyu-B has HP Grass listed as a Useful move.

Useful moves are not viable moves. The bar is much lower.

js/search.js
TheTiksiBranch: kartana is the only steel to get it and not iron head, and unstabbed 70bp steel coverage is pretty outclassed everywhere

js/search.js
TheTiksiBranch: is a two-part like steel type + !ironhead OR dark|dragon + !ironhead !gunkshot possible

the latter part would cover most everything on its own...except probably the most important user in luke

js/search.js
TheTiksiBranch: off 20 attack and unstabbed? sg as the param wouldn't help hustle toge anyway. is dunsparce+deers enough? i guess jirachi maaaaybe and that adding it would make this push it but i dont think the broad serenegrace->headbutt is not outclassed link is there

js/search.js
Zarel: Steel-type and no Iron Head, then.

Zarel: (Mostly, I don't want to have to update this next gen)

TheTiksiBranch: sure

js/search.js
Zarel: Yeah, that works.

Zarel: Are you asking "possible" as in, you'd need my help to write the code? Or are you just asking if I'd support that, because I definitely consider that better than a list.

js/search.js
Zarel: Oh, good point. Maybe Serene Grace + STAB?

TheTiksiBranch: great

TheTiksiBranch:
speaking of lists and breaking them, how about changing trick room to spe<100 instead of cress/p2/mag/diancie/uxie/dialga 

and while im here hypnosis should be viable for darkrai

Zarel:
Spe<100 is way too high...

Approve Hypnosis for Darkrai though.

TheTiksiBranch:
Base 100 at level 100 with 0 IVs, 0- EVs: 184.
Base 43 at level 100 with 31 IVs, 252 EVs: 185.
Base 75 at level 100 with 31 IVs, 0 EVs: 186.

Base 100 at level 100 with 31 IVs, 0- EVs: 212.
Base 57 at level 100 with 31 IVs, 252 EVs: 213.

The right investment can make 100 "out-slow" a surprising lot, to set for the team, sweep, or disrupt fast offense. It's hard to outclass TR due to its uniqueness, some mons over 80 that should be caught in a TR rule are cresselia (85) dialga (90) deoxysd (90) uxie (95) and solgaleo (97). Not always viable, but I'd def say useful.

js/search.js
TheTiksiBranch: i meant the latter but the former may well be true... how do you make "if" check the mon's type (a.k.a. what is the Pokemon type counterpart to toId(set.species) === 'magikarp')? 

After that, would it be this?
if (id === 'irontail') isViable = ((steeltypecheck && moves.indexOf('ironhead') < 0) || ((darktypecheck || dragontypecheck) && moves.indexOf('ironhead') < 0 && moves.indexOf('gunkshot') < 0));

js/search.js
TheTiksiBranch: i mean i dont think even breloom 'runs' it anymore (not sure whether it did when somebody put it there) but its probably tolerable on a fair few others (megamawile? buzzwole? toxicroak?)

Zarel:
Hm, fair.

TheTiksiBranch:
Electro Ball, Endeavor, Hypnosis, TR changes in

js/search.js
Zarel: Might as well just leave it generally Usable, it's a funny prediction move that could see niche use basically everywhere.

js/search.js
Zarel: `template.types` contains types. Do an `indexOf` to check them.

And yes, that looks right to me.

PR #1039, opened 2017-11-01
Remove spook from battles

Ridaz:
Halloween is over


panpawn:
If we plan on doing this again next year, maybe comment them out instead of removing them?

Ridaz:
I'm doing something else next year

PR #1040, opened 2017-11-02
Fix Doubles UU showing DNU in teambuilder

n10sit:
It was returning that Pokemon not banned by the duu banlist were still dou, and the same for the dnu banlist, i think.

### DIFF
--- a/githooks/build-indexes
+++ b/githooks/build-indexes
@@ -312,13 +312,13 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 					}
 					banlist = Tools.getFormat(genPrefix + 'doublesuu').banlist;
 					if (!banlist) {
-						defaultTier = "DOU";
+						defaultTier = "DUU";
 					} else if (banlist.indexOf(template.species) >= 0 || banlist.indexOf(template.baseSpecies) >= 0) {
 						return "DOU";
 					}
 					banlist = Tools.getFormat(genPrefix + 'doublesnu').banlist;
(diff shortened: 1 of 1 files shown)


Zarel:
This does not seem to do what you think it does? Have you tested this?

githooks/build-indexes
```diff
@@ -312,13 +312,13 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 					}
 					banlist = Tools.getFormat(genPrefix + 'doublesuu').banlist;
 					if (!banlist) {
-						defaultTier = "DOU";
+						defaultTier = "DUU";
```
Zarel: This says "if there is no Doubles UU format in this gen, all pokemon are DUU" which is clearly wrong...

n10sit: i read this incorrectly then. fsr i read "!banlist" as "not on the banlist" instead of banlist existing, but it would have to like, check the pokemon first for that to be the case... so im dumb ig. When you have a moment can i talk you about this chunk of code to see if we can get it working as intended? see this post: http://www.smogon.com/forums/threads/bug-reports-v2-0-read-op-before-posting.3469932/page-373#post-7567416 if you haven't already

Zarel: Sure, find me on Discord or PS.

KrisXV:
This PR probably can be closed

Zarel:
Yeah, I don't know when we fixed this, but I would be surprised if it was still an issue.

PR #1041, opened 2017-11-02
Keep /raw PMs from showing up in main chat

bumbadadabum:


js/client-chat.js
```diff
@@ -1145,7 +1145,7 @@
 			if (this.$chatFrame.scrollTop() + 60 >= this.$chat.height() - this.$chatFrame.height()) {
 				autoscroll = true;
 			}
-			this.addChat(user, message, pm);
+			if (!message.startsWith('/raw')) this.addChat(user, message, pm);
```
panpawn: This should probably be done in ``this.addChat``

Zarel:
`/html` tho

bumbadadabum:
Is that ever sent over PM?

Zarel:
Yes. The difference is that

> Zarel: `/raw <i>foo</i>`

will become 

> *foo*

while

> Zarel: `/html <i>foo</i>`

will become

> Zarel: *foo*

bumbadadabum:
Guess I'll add that to the PR then

Zarel:
Poke. I hope you didn't forget about this.

bumbadadabum:
No I just haven't had the time, not home all day today. Will update when
I'm home.

PR #1047, opened 2017-11-21
Fix issues with horizontal scroll bar in Chrome

panpawn:
Several users have been complaining that a horizontal scroll bar is getting in the way in chat rooms and PM boxes, this fixes that.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -1343,6 +1343,8 @@ a.ilink.yours {
 	resize: none;
 	font-size: 9pt;
 	font-family: Verdana, Helvetica, Arial, sans-serif;
+	overflow-x: hidden;
+	overflow-wrap: break-word;
 }
 
 /* Tournaments */


Zarel:
Can you make it not overflow horizontally?

Zarel:
Specifically: `overflow-wrap: break-word;`

panpawn:
That won't break vertical scrolling, right?

Zarel:
No, it won't.

Zarel:
image

panpawn:
How is that?

Zarel:
...you're supposed to have both.

panpawn:
Updated

panpawn:
... I have no idea how what this build is complaining about

Zarel:
Heisenbugs make me uncomfortable... :/

PR #1049, opened 2017-11-24
Proxy POST requests in the testclient too

urkerab:
This seems to work although I don't know whether it's something you want the testclient to be able to do.


Zarel:
That's fine, but that Travis error is definitely your fault. :P

urkerab:
Sorry, I was confused because the error wasn't at the end of the log...

PR #1053, opened 2017-11-27
Do not change URL for HTML rooms

sparkychildcharlie:
- atm when you do /modlog the url of the entire page changes to the modlog tab, mainly because it's focused and on the left side of the page.  This causes some inconveniences because when you refresh, the changed URL causes a "blank page" to show.
- (also rather annoying for people using an external pm logging script, if checking logs causes the url to change, but this is just a trivial detail for me)

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -1412,9 +1412,8 @@
 				}
 				this.curRoom = window.room = room;
 				this.updateLayout();
-				if (this.curRoom.id === id) {
+				if (this.curRoom.id === id && room.type !== 'html') {
 					this.fragment = id;
-					this.navigate(id);
 					this.updateTitle(this.curRoom);
 				}
 			}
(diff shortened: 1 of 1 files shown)


Zarel:
Some rooms really should be kept open, though. Really we need something server-side to intercept `/join view-blah-blah` and render it.

sparkychildcharlie:
atm i think only modlog uses html type rooms, and modlog doesnt need to be kept open past a refresh.  html type rooms are only used to display information called from the server (query type display only) instead of a lasting panel.  Plus even after you refresh, the html room is empty anyways.

the join thing doesnt seem like a good idea to me - it could possibly be used to drain the server's resources if it works with /autojoin... unless i'm wrong about that somewhere

Zarel:
Things planned to use HTML-type rooms:

- modlog
- room FAQs
- Help ticket list
- Help ticket form

Other than specific modlog searches, everything needed to recreate them is in the room ID.

Zarel:
I can't see it possibly draining a server's resources any more than joining any other room. Really, it should be a lot better, because of the lack of state.

Zarel:
Closing. I still think that what we really need is an HTML-room dispatcher.

PR #1054, opened 2017-11-27
Make disabled tour validator button not validate

TheMezStrikes:


Zarel:
Why?

TheMezStrikes:
Clicking the "validate" button in a random battle tournament validates the team, but the button class is set as disable so I figured it'd actually need to be disabled

Zarel:
Wouldn't it make more sense to remove the validate button completely?

TheMezStrikes:
My Javascript knowledge is very limited. So if you think that's the preferred course of action then I won't be of much help.

Zarel:
TI's deathly afraid of client, so I guess this falls to me... :/

Zarel:
It says "Your team is valid for this tournament".

I don't see a problem with this?

TheMezStrikes:
My problem with this is that the button looks like it's disabled, is of the disabled class but is in fact fully functional. 
[21:29:05] Light HaZe: its just a little goofy imo

Zarel:
"Does nothing except show a message telling you you didn't need to press it" does not mean "fully functional" to me, and is in fact the ideal reaction of a disabled button.

Zarel:
Like, a good analogy is a locked door. A door that doesn't open when you try to open it is good, but even better is a door that tells you "this door is locked; use another door" when you try to open it.

TheMezStrikes:
Wouldn't a message saying "Random Battle has random teams, so there is nothing to validate." be ideal then?

TheMezStrikes:
The way I see the current situation, using your door analogy, is you walk up to a door with a huge lock on it that is clearly locked, but it still opens.

Zarel:
Yes, that would certainly be more ideal, but the new situation is that random battle just doesn't display a validate button at all. Which is I think the best solution here.

PR #1055, opened 2017-11-29
Don't faint a newly created Pokemon to reduce the team size

urkerab:
DG9 had challenged Quantum Tesseract to a Balanced Hackmons battle. His team was two Rayquaza-Megas, a Gengar, a Palkia, a Mega Tyranitar, and a Registeel. On turn 11, he mega-evolved his Rayquaza-Mega, although it was really his Illusion Gengar. Then when he later switched his (now Mega) Gengar back in, the client started to get confused, because this was a Pokemon that it hadn't seen in the Team Preview. But this is just standard for Hackmons.
The problem arises when one of his Rayquaza-Megas fainted. Later, he wanted to bring in his other Rayquaza-Mega. The client duly started to created a second Rayquaza-Mega because it had seen two in the team preview. But, the client was still confused by Gengar's Illusion, so in order to reduce the team size, it looked for a random unfainted Pokémon (as per the comment on line 1773). What did it choose? None other but the very Rayquaza-Mega that it was trying to create at the time...
I don't know whether the change on line 1764 is correct but I thought it should match 1779.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -1761,6 +1761,7 @@ var Side = (function () {
 					var illusionFound = null;
 					for (var i = 0; i < this.pokemon.length; i++) {
 						var curPoke = this.pokemon[i];
+						if (curPoke === poke) continue;
 						if (curPoke.fainted) continue;
 						if (this.active.indexOf(curPoke) >= 0) continue;
 						if (curPoke.speciesid === 'zoroark' || curPoke.speciesid === 'zorua' || curPoke.ability === 'Illusion') {
@@ -1775,6 +1776,7 @@ var Side = (function () {
 						// and reguessed.
 						for (var i = 0; i < this.pokemon.length; i++) {
 							var curPoke = this.pokemon[i];
+							if (curPoke === poke) continue;
 							if (curPoke.fainted) continue;
 							if (this.active.indexOf(curPoke) >= 0) continue;
 							illusionFound = curPoke;


Zarel:
Can you explain "the client started to get confused, because this was a Pokemon that it hadn't seen in the Team Preview".

If Team Preview has taken effect, the client should not be overriding the team list in the first place.

Zarel:
A replay would be _really_ nice

urkerab:
> Can you explain "the client started to get confused, because this was a Pokemon that it hadn't seen in the Team Preview".

The Team Preview showed two Rayquaza-Megas, a Gengar, a Palkia, a Mega Tyranitar, and a Registeel. However, the Gengar had Illusion, so it mega-evolved invisibly. When it was switched in for real, the client didn't recognise it as being on the team. 

I don't know what you mean by "overriding the team list", but `Side.prototype.newPokemon` contains the following comment:

                        // This is Hackmons; we'll just guess a random unfainted Pokemon.
                        // This will keep the fainted Pokemon count correct, and will
                        // eventually become correct as incorrect guesses are switched in

And when the second Rayquaza-Mega switched in, which the client was expecting but hadn't seen yet, it was still confused as to how big the team was, and decided to faint the Pokémon it was just creating.

Looks like someone saved the replay: https://replay.pokemonshowdown.com/rom-gen6customgame-188036 (it's a custom game because I forgot to update the banlist).

Zarel:
Yes, this should definitely happen, but I just wanted to get the replay because there were other bugs involved, too.

PR #1056, opened 2017-12-01
Properly update Earthquake animation

Ridaz:
eq1

### DIFF
--- a/data/graphics.js
+++ b/data/graphics.js
@@ -13592,6 +13592,29 @@ var BattleMoveAnims = {
 			var attacker = args[0];
 			var defender;
 
+			battle.bgElem.animate({
+				left: -55,
+				right: -5,
+			}, 75).animate({
+				left: -45,
+				right: -15
+			}, 100).animate({
+				left: -55,
+				right: -5
+			}, 100).animate({
+				left: -45,
+				right: -15
+			}, 100).animate({
+				left: -55,
+				right: -5
+			}, 100).animate({
+				left: -45,
+				right: -15
+			}, 100).animate({
+				left: -50,
+				right: -10,
+			}, 100);
+
 			attacker.anim({
 				x: attacker.x - 30,
 				time: 75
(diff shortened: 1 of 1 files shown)


Zarel:
Confusion and EQ should really be two separate pullreqs.

Zarel:
Confusion can probably just be pushed directly tbh

data/graphics.js
```diff
@@ -13592,6 +13618,34 @@ var BattleMoveAnims = {
...
+			}, 100).animate({
+				left: 5,
+				right: -5
+			}, 100).animate({
+				left: 0,
+				right: 0,
+			}, 100, function () {
+				battle.bgElem.animate({
+					position: 'static'
+				});
```
Zarel: Isn't it supposed to be `absolute` to begin with?

I'm actually 80% sure the background has a lot of padding for a reason.

data/graphics.js
```diff
@@ -13592,6 +13592,34 @@ var BattleMoveAnims = {
...
+				right: 5
+			}, 100).animate({
+				left: 5,
+				right: -5
+			}, 100).animate({
+				left: 0,
+				right: 0,
+			}, 100, function () {
+				battle.bgElem.animate({
+					position: 'absolute'
```
Zarel: ...????

First of all, `position` is not a CSS attribute that can be animated. This will actually do exactly nothing (which is good because the `position` attribute is already set to `absolute` so you don't need it to do anything at all.

data/graphics.js
```diff
@@ -13592,6 +13592,34 @@ var BattleMoveAnims = {
 			var attacker = args[0];
 			var defender;
 
+			battle.bgElem.animate({
+				position: 'absolute',
```
Zarel: (Same goes for this instance)

data/graphics.js
```diff
@@ -13592,6 +13592,34 @@ var BattleMoveAnims = {
...
+				left: -5,
```
Zarel: The `bgElem` starts at `-90, -50`. Don't you notice it being positioned weirdly if you do this? Have you tested a gen 1 replay with Earthquake?

Ridaz:
Should look good now. Tested on a Gen 3 game:
eq gen3

Zarel:
Where did your self-hit confusion animation go?

Zarel:
Feel free to just push that one directly.

PR #1057, opened 2017-12-02
Refactor |uhtml|

HoeenCoder:
`|uhtml|id|html`
If a uhtml element exists already, update it.
If not, add one to chat.
if "html" is blank, delete the uhtml element from chat (not just make it blank).

Breaks anything that needs to add multiple uhtml elements to chat (any suggestions for this? or just let them break.)
`|uhtmlchange|` deprecated

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1266,13 +1266,19 @@
 					break;
 
 				case 'uhtml':
-					this.$chat.append('<div class="notice uhtml-' + toId(row[1]) + '">' + Tools.sanitizeHTML(row.slice(2).join('|')) + '</div>');
-					break;
-
 				case 'uhtmlchange':
 					var $elements = this.$chat.find('div.uhtml-' + toId(row[1]));
-					if (!$elements.length) break;
-					$elements.html(Tools.sanitizeHTML(row.slice(2).join('|')));
+					var html = row.slice(2).join('|');
+					if (!html) {
+						$elements.remove();
+					} else if (!$elements.length) {
+						this.$chat.append('<div class="notice uhtml-' + toId(row[1]) + '">' + Tools.sanitizeHTML(html) + '</div>');
+					} else if (row[0] === 'uhtmlchange') {
+						$elements.html(Tools.sanitizeHTML(html));
+					} else {
+						$elements.remove();
+						this.$chat.append('<div class="notice uhtml-' + toId(row[1]) + '">' + Tools.sanitizeHTML(html) + '</div>');
+					}
 					break;
 
 				case 'unlink':


Zarel:
What do you mean by "needs to add multiple uhtml elements to chat"?

js/client-chat.js
```diff
@@ -1266,9 +1266,19 @@
 					break;
 
 				case 'uhtml':
```
Zarel: Just roll in `uhtmlchange` into this.

js/client-chat.js
```diff
@@ -1266,9 +1266,19 @@
...
-					this.$chat.append('<div class="notice uhtml-' + toId(row[1]) + '">' + Tools.sanitizeHTML(row.slice(2).join('|')) + '</div>');
+					var $elements = this.$chat.find('div.uhtml-' + toId(row[1]));
+					if (!$elements.length) {
+						this.$chat.append('<div class="notice uhtml-' + toId(row[1]) + '">' + Tools.sanitizeHTML(row.slice(2).join('|')) + '</div>');
+					} else {
+						if (!toId(row.slice(2).join('|'))) {
```
Zarel: `else if`

Heck, this should probably be the first `if`, the appending should be the `else if`.

HoeenCoder:
Before, `|uhtml|id|html` would add a new uhtml element to chat with that id every time it was received, allowing for multiple uhtml elements with the same id in a room.

Zarel:
Oh, that was never intended to be supported. I don't think anything uses it that way.

js/client-chat.js
```diff
@@ -1266,13 +1266,15 @@
...
 
 				case 'uhtml':
-					this.$chat.append('<div class="notice uhtml-' + toId(row[1]) + '">' + Tools.sanitizeHTML(row.slice(2).join('|')) + '</div>');
-					break;
-
 				case 'uhtmlchange':
 					var $elements = this.$chat.find('div.uhtml-' + toId(row[1]));
-					if (!$elements.length) break;
-					$elements.html(Tools.sanitizeHTML(row.slice(2).join('|')));
+					if (!toId(row.slice(2).join('|'))) {
```
Zarel: You should probably cache `row.slice(2).join('|')` somewhere. Instead of doing it three times, one in each branch.

HoeenCoder: good catch

bumbadadabum:
Uhhh I use multiple uhtml elements with the same id a lot in the Wi-Fi chat
plugin.

bumbadadabum:
It would actually suck really hard to have that functionality removed.

js/client-chat.js
```diff
@@ -1266,13 +1266,16 @@
...
 				case 'uhtml':
-					this.$chat.append('<div class="notice uhtml-' + toId(row[1]) + '">' + Tools.sanitizeHTML(row.slice(2).join('|')) + '</div>');
-					break;
-
 				case 'uhtmlchange':
 					var $elements = this.$chat.find('div.uhtml-' + toId(row[1]));
-					if (!$elements.length) break;
-					$elements.html(Tools.sanitizeHTML(row.slice(2).join('|')));
+					var html = row.slice(2).join('|');
+					if (!toId(html)) {
```
Zarel: Do we need to `toId` here?

HoeenCoder:
That should be handled in one way or another (either change the wifi plugin or support it here)

js/client-chat.js
HoeenCoder: I did it incase of a blank space, though `trim()` would probably be better if anything

Zarel:
@bumbadadabum what exactly do you need that for?

Zarel:
Hm, one thing we could do is keep the existing API, and just make a small adjustment that `uhtml` always creates a new element.

PR #1058, opened 2017-12-02
Allow cloning of teams

acz13:
Just a little helpful thing I edited a while back in my testclient and never PRd... code style can be changed (Not sure how you want it)


acz13:
oops, messed that up

js/client-teambuilder.js
```diff
@@ -398,7 +398,7 @@
 					// support dragging and dropping buttons.
 					buf += '<li><div name="edit" data-value="' + i + '" class="team" draggable="true">' + formatText + '<strong>' + Tools.escapeHTML(team.name) + '</strong><br /><small>';
 					buf += Storage.getTeamIcons(team);
-					buf += '</small></div><button name="edit" value="' + i + '"><i class="fa fa-pencil"></i>Edit</button><button name="delete" value="' + i + '"><i class="fa fa-trash"></i>Delete</button></li>';
+					buf += '</small></div><button name="edit" value="' + i + '"><i class="fa fa-pencil"></i></button><button name="newTop" value="' + i + '"><i class="fa fa-clone"></i></button><button name="delete" value="' + i + '"><i class="fa fa-trash"></i></button></li>';
```
Zarel: I'm not sure if this is the right place to put a button...

Maybe a "More" menu.

acz13: IMHO it looks pretty good like this

screen shot 2017-12-02 at 19 35 41

Zarel: Icons without labels are pretty useless. I wouldn't recommend them.

acz13: hmm, then perhaps shorten the width of the teams?

screen shot 2017-12-02 at 22 49 50

Zarel: I don't know if "Clone" needs top billing among "Edit" and "Delete"

Hence a "More" menu. Maybe just a dropdown arrow to the right of Delete, or to the left of Edit. It would take some space but not nearly as much.

Zarel: Or just have a More menu that replaces "Edit". Clicking on the team does edit it.

acz13: I think having a more menu with only one option is kinda pointless, you might as well have just the option at the top level...

if someone thinks up of other things to put in such a menu that would work

Koushien: A menu item to validate the team?

Zarel: Edit, Clone is at least two.

acz13: I don't see the point in an edit button...

Zarel: Mostly, in case people aren't thinking and forget that clicking on the big button means "Edit".

Zarel:
I've been doing this by copy/pasting the export, but yeah a button would be nice I guess but I still would prefer a More menu here. I suppose I'll have to implement that part myself.

Zarel:
Okay, here's my compromise: at least label Delete.

Zarel:
c6fb5858014f67b8df9e52c

PR #1059, opened 2017-12-03
Revert "Properly update Earthquake animation (#1056)"

KrisOnline:
This reverts commit f444d6b8f437e163771bf25c828e762462feb850.

A ton of global auth, staff, regular users, and other people who don't fall into one of those categories have voiced publicly and privately to me and other users that this new animation gives them headaches and is generally unappealing to watch over and over again.

### DIFF
--- a/data/graphics.js
+++ b/data/graphics.js
@@ -13618,29 +13618,6 @@ var BattleMoveAnims = {
 			var attacker = args[0];
 			var defender;
 
-			battle.bgElem.animate({
-				left: -55,
-				right: -5,
-			}, 75).animate({
-				left: -45,
-				right: -15
-			}, 100).animate({
-				left: -55,
-				right: -5
-			}, 100).animate({
-				left: -45,
-				right: -15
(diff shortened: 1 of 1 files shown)


Dialga204:
I find it unappealing to watch over and over again. Good idea to revert it back to the old animation.

Ridaz:
A ton? I've only heard one complain about 1 person. Who are these people and why are they not speaking up to me?
Also, if people are getting headaches from this, how did they even play pokemon in, for example, gen 3? The animation is much more violent in game. 
We should look for a way to make it less prone to headaches. Not an outright revert

Ridaz:
Did it in 400bfa5b59b5726dd14a6497f2a71e1d99da1c6d

Zarel:
It's "A ton of" for me as well, at least 5 people on Inside Scoop Discord.

Dialga204:
Keep improving. I may continue using the move Earthquake. Improvement could be made, less violent shaking is better.

PR #1060, opened 2017-12-08
Fixed the case where Zoroark switches in/out in Random Battle

yuzeh:
This change should fix tracking in species clause battles where an Illusion user switches in or out for the pokemon that it is disguised as.

Let me know what tests I should have (I have one on a local fork but I'm not sure how/where to port) and what else you need from me!

### DIFF
--- a/data/graphics.js
+++ b/data/graphics.js
@@ -13619,132 +13619,191 @@ var BattleMoveAnims = {
 			var defender;
 
 			battle.bgElem.animate({
-				left: -55,
-				right: -5,
+				top: -90,
+				bottom: 0,
 			}, 75).animate({
-				left: -45,
-				right: -15
+				top: -100,
+				bottom: -10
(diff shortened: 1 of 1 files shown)


js/battle.js
```diff
@@ -6654,10 +6654,16 @@ var Battle = (function () {
...
-					poke.side.switchOut(poke.side.active[slot]);
+				if (poke.side.battle.speciesClause &&
+						poke.side.active[slot] &&
+						poke.side.active[slot].ident === poke.ident) {
+					poke.side.replace(poke);
+				} else {
+					if (poke.side.active[slot]) {
+						poke.side.switchOut(poke.side.active[slot]);
+					}
+					poke.side.switchIn(poke);
```
Zarel: Good idea, but this probably doesn't need this complicated if statement here. You should probably just test Species Clause directly in `getPokemon`.

PR #1063, opened 2017-12-11
FormatText: don't embed links within links

sparkychildcharlie:
same as https://github.com/Zarel/Pokemon-Showdown/pull/4238


Zarel:
...I can compile this myself...

Zarel:
You really aren't supposed to edit this by hand, as evidenced by the "transpiled" comment. You're supposed to use Closure Compiler on it.

sparkychildcharlie:
ah ok, sorry ;-;

PR #1065, opened 2017-12-20
Battle list: Add more elo filters

panpawn:


panpawn:
Screen caps:

panpawn:
Someone was asking me if I could go up to 2000, is this necessary do you think?

Zarel:
I'd recommend not jumping up by 100 each time. Maybe 200.

Zarel:
1300, 1500, 1700

Zarel:
Also I think this makes more sense as radio buttons. Save yourself a click.

panpawn:
How would I make it so you can deselect a radio button?

panpawn:
I guess I could have a radio button for "(none)"?

Zarel:
Yes. I'd probably call it "Off".

Zarel:
I'd also probably make Elo Filter a button that reveals the choices? Actually that sounds like probably the right thing to do here. "Elo Filter" reveals the tabbar.

panpawn:
By "reveals the tabbar" do you mean like, details and summary tags to hide/show the radio buttons?

Zarel:
Do you guys really have nothing better to do than details/summary tags? Use them in bots if you really want, but keep those tags off the repo, pls.

Zarel:
I mean actual JavaScript for revealing/hiding the tabbar.

Zarel:
...in hindsight, details/summary is fine too >_>

panpawn:
... is that a yes?

Zarel:
That's a yes.

panpawn:
... wait, if we're doing detail and summary tags, then it'd still be 2 clicks...

I think the dropdown looks much better, considering its the same number of clicks.

sparkychildcharlie:
Why is 'Elo Filter" still a button if you're using details/summary?  I think non-hidden radio buttons look best

panpawn:
>Why is 'Elo Filter" still a button if you're using details/summary?

I thought that's what Zarel was asking for :|?

Zarel:
It's still 2 clicks to open the menu, but it's no longer 2 clicks to switch between settings. That's the main thing I wanted to improve on.

Zrp200:
what about just having the user able to type in a desired elo? Less buttons and clicks.

AustinXII:
I agree that sounds way better

panpawn:
That sounds more complicated for the end-user to me

Zrp200:
I shouldn't, because they should already have in mind what elo range they want to see @panpawn. I'm willing to take that assumption.

Slayer95:
Ladder inflation exists, and the Elo ranges therefore vary for each format, and even depending on the time since last ladder reset.

Therefore, IMO the filters should actually be qualitative (``High/Mid/All``), and the thresholds should be decided server-side according to the requested battle format.

Zrp200:
actually that's way better than my idea @Slayer95

Zarel:
Typing in a desired Elo is definitely worse. Clicking _and_ typing is like 20x slower than just clicking.

AustinXII:
Sorry I wasn’t talking about speed, in general I thinking typing out what you want to see is better than choosing for a few options we pick for you to see. I’m always for more accessibility and it seems like a good compromise for the “what should make make thec range” issue and also accounts for ladder ranks getting higher and higher with more people coming in the future.

Zarel:
For certain situations, being able to access the power of exact selectors is really useful. Like `/dexsearch` is really powerful.

Filtering a list by a visible number really isn't. Just give rough filters, it's not remotely hard to just use your eyes to do the rest.

Zarel:
The thing is that using rough filters and eyes is much _faster_, and that matters.

panpawn:
Should I be doing something on this before its ready for merge?

Zarel:
This ended up being fixed by b09a58dbe99c07e0

After a lot of debate, we ended up doing a rough approximation of your original approach.

PR #1066, opened 2017-12-25
Fix toxicTurns counter for Magic Guard/Poison Heal

cyap:
Issue: If a Magic Guard/Poison Heal Pokemon is badly poisoned, the
tooltip will always say, "Next damage:/Would take if ability removed:
6%”. If the ability is removed, the forecasted poison damage will be
one turn behind. The reason for this is that currently, the Toxic
counter only increments if the client receives a message indicating
that the Pokemon has taken Poison damage.

Proposed fix: The Toxic counter is now handled under upkeep; at the end
of every turn, all active Pokemon are checked for Toxic status, and
their counters are each incremented accordingly.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -3032,6 +3032,14 @@ var Battle = (function () {
 		this.turnsSinceMoved = 0;
 		this.acceleration = (this.messageFadeTime < 150 ? 2 : 1);
 	};
+	Battle.prototype.updateToxicTurns = function () {
+		for (var i = 0; i < this.sides.length; i++) {
+			for (var slot = 0; slot < this.sides[i].active.length; slot++) {
+				var poke = this.sides[i].active[slot];
+				if (poke.statusData.toxicTurns) poke.statusData.toxicTurns++;
+			}
+		}
+	};
 	Battle.prototype.changeWeather = function (weather, poke, isUpkeep, ability) {
 		weather = toId(weather);
 		var weatherTable = {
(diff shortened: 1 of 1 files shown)


js/battle.js
```diff
@@ -3032,6 +3032,13 @@ var Battle = (function () {
 		this.turnsSinceMoved = 0;
 		this.acceleration = (this.messageFadeTime < 150 ? 2 : 1);
 	};
+	Battle.prototype.updateToxicTurns = function () {
+		for (var side in this.sides) {
```
KrisOnline: shouldn't these be `const...of` loops?

js/battle.js
```diff
@@ -7221,4 +7228,4 @@ var Battle = (function () {
 	};
 
 	return Battle;
-})();
+})();
```
KrisOnline: there should also be a newline at the end of the file

Zarel:
The client should maybe track turn count internally, the way the server does...

js/battle.js
AustinXII: Yes

Zarel: ...no! This is the client! What are you thinking!

Zarel: Also `this.sides` is an array, isn't it? You need to iterate it manually...

AustinXII: Oh lol I didn’t realize this was client yeah that’s mb

AustinXII: This is why you don’t go on github at 4 in the morning

KrisOnline: Oops lol

cyap: Fixed the array iterations and the newline.

cyap:
I assumed upkeep fulfilled this purpose, as it also handles the other case I know of where turn information needs to be tracked beyond server messages (weather + Cloud Nine). Would there be a difference in implementation?

Zarel:
Yeah, this is better than nothing; I don't like this implementation but `upkeep` _was_ added for a reason.

PR #1068, opened 2017-12-28
handle unfound pokemon in -curestatus

### LINKED ISSUE #1062: Zoroark/HealBell
My understanding of heal bell: the server sends a -curestatus message to the client for each pokemon the client knows about. Zoroark, as always, throws a wrench into things -- the server thinks the client has seen a Zoroark, whereas the client has either

1) not seen a zoroark or
2) seen an illusion.

I haven't read the server code, but given that this doesn't crash all the time with healbell, I can only imagine 2) is the reason.

I've attached a replay log that should fail: [zoroark-heal-bell.txt](https://github.com/Zarel/Pokemon-Showdown-Client/files/1541822/zoroark-heal-bell.txt)

I'm not sure if the fix here should be client or server-side because I'm not familiar with how this is handled in the cartridge games.

yuzeh:
Should fix #1062.

Let me know how I should be testing this... I tested locally and everything seems to work well.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -4468,15 +4468,24 @@ var Battle = (function () {
 				var poke = this.getPokemon(args[1]);
 				var effect = Tools.getEffect(kwargs.from);
 				var ofpoke = this.getPokemon(kwargs.of);
-				poke.status = '';
+				var pokeName, pokeSideN;
+				if (poke) {
+					poke.status = '';
+					pokeName = poke.getName();
+					pokeSideN = poke.side.n;
+				} else {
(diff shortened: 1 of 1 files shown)


Zarel:
Are you aware of the failed test?

yuzeh:
@Zarel yes! I think... you mean `npm test`?

I just fixed that issue.

yuzeh:
@Zarel anything else you want me to do for this issue?

Zarel:
I'm not sure I _like_ this implementation, but I can't think of obvious changes.

PR #1070, opened 2018-01-08
Battles: Make sure the tie message is actually a tie message

whalemer:


Zarel:
What else could it be?

panpawn:
>+jumbowhales: '|tier'.substr(0, 4) is '|tie'

PR #1071, opened 2018-01-14
Remove surplus space prefix from battle room users

urkerab:
The server sends the user's rank as the prefix. (I discovered this while trying to implement user lists for battle rooms.)

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -6593,7 +6593,11 @@ var Battle = (function () {
 			break;
 		case 'join':
 		case 'j':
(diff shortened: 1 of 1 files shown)


Zarel:
Not always.

Can you add the space only if the first character is alphanumeric?

urkerab:
Like this?

Zarel:
Yes. Can you separate that out into four lines for readability?

```
if (blah) {
  add group symbol if not there
  update room.users
}
```

urkerab:
Third time lucky?

Zarel:
Yep, thanks!

PR #1073, opened 2018-01-15
Make /challenge work in PMs

urkerab:
As suggested halfway down [this forum post](http://www.smogon.com/forums/posts/7649603).

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -314,21 +314,30 @@
 			var $chat = $pmWindow.find('.inner');
 			// this.tabComplete.reset();
 			this.chatHistories[userid].push(text);
-			if (text.toLowerCase() === '/ignore') {
+			var data = '';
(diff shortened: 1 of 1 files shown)


Zarel:
You should support passing a format name.

urkerab:
There aren't any PM commands that accept arguments yet, so I had to improvise...

js/client-mainmenu.js
```diff
@@ -314,21 +314,30 @@
 			var $chat = $pmWindow.find('.inner');
 			// this.tabComplete.reset();
 			this.chatHistories[userid].push(text);
-			if (text.toLowerCase() === '/ignore') {
+			var cmd = text;
```
Zarel: `= text.toLowerCase()` would save you some repetitive code.

urkerab: Well, it's not code that I had to repeat, but I see what you mean.

PR #1075, opened 2018-01-28
Teambuilder: Support AAA, BH, M&M, and STABmons

KrisXV:


js/search.js
```diff
@@ -892,6 +892,13 @@
 				abilitySet.push(['header', "Special Event Ability"]);
 				abilitySet.push(['ability', toId(template.abilities['S'])]);
 			}
+			if (format === 'almostanyability' || format === 'balancedhackmons' || format === 'bh') {
+				abilitySet.push(['header', "Other Abilities"]);
```
Zarel: This is "All Abilities".

Zarel: It's not like you're intentionally skipping abilities the pokemon already has.

Also you're not going to filter out unviable abilities? Just filter out `score <= 0`.

js/search.js
```diff
@@ -922,6 +929,34 @@
 				}
 				learnsetid = this.nextLearnsetid(learnsetid, template.id);
 			}
+			if (format === 'balancedhackmons' || format === 'bh') {
```
Zarel: Just combine this with the Sketch code, no need to duplicate it.

js/search.js
```diff
@@ -962,37 +997,46 @@
...
 				if (id === 'storedpower') isViable = (toId(set.species) in {necrozma:1, espeon:1, sigilyph:1});
 				if (id === 'stunspore') isViable = (moves.indexOf('thunderwave') < 0);
-				if (id === 'thunder') isViable = (toId(set.ability) === 'drizzle' || (toId(set.ability) === 'primordialsea') || (toId(set.species) === 'xerneas'));
+				if (id === 'thunder') isViable = (toId(set.ability) === 'drizzle' || (toId(set.ability) === 'primordialsea') || (toId(set.species) === 'xerneas') || (format === 'mixandmega' && toId(set.item) === 'pidgeotite' && moves.indexOf('zapcannon') < 0));
 				if (id === 'trickroom') isViable = (template.baseStats.spe <= 100);
+				if (id === 'waterpulse') isViable = ((toId(set.ability) === 'megalauncher' && moves.indexOf('primordialsea') < 0) || (format === 'mixandmega' && toId(set.item) === 'blastoisinite' && moves.indexOf('primordialsea') < 0));
+				if (id === 'weatherball') isViable = (format === 'mixandmega' && toId(set.item) === 'redorb');
+				if (id === 'zapcannon') isViable = (format === 'mixandmega' && toId(set.item) === 'pidgeotite');
```
Zarel: I would put the M&M viabilities in an `if (format === 'mixandmega') {` block, instead of interspersing them.

KrisXV: This causes a crash

KrisXV: I’ll update the commit after school

Zarel: What's the crash message?

js/search.js
```diff
@@ -892,6 +892,38 @@
 				abilitySet.push(['header', "Special Event Ability"]);
 				abilitySet.push(['ability', toId(template.abilities['S'])]);
 			}
+			if (format === 'almostanyability' || format === 'balancedhackmons' || format === 'bh') {
+				template = Tools.getTemplate(set.species);
+				var abilities = [];
+				if (template.isMega) {
+					if (format === 'almostanyability') abilitySet.unshift(['html', '<p>Will be <strong>' + Tools.escapeHTML(template.abilities['0']) + '</strong> after Mega Evolving.</p>']);
```
Zarel: This looks like code duplication. Is there any reason the instance 20 lines up needs to be repeated?

js/search.js
```diff
@@ -892,6 +892,38 @@
...
+				var poorAbilities = [['header', "Situational Abilities"]];
+				var badAbilities = [['header', "Unviable Abilities"]];
+				for (var i = 0; i < abilities.length; i++) {
+					var id = abilities[i];
+					var rating = BattleAbilities[id] && BattleAbilities[id].rating;
+					if (id === 'normalize' && (format === 'balancedhackmons' || format === 'bh')) rating = 3;
+					if (rating >= 3) {
+						goodAbilities.push(['ability', id]);
+					} else if (rating >= 2) {
+						poorAbilities.push(['ability', id]);
```
Zarel: Is there a reason you're excluding ratings 0.5 to 1.5 in this list?

Are they straight outclassed or something?

KrisXV: i looked over them all and everything with a 1.5 rating or lower was just bad

js/search.js
```diff
@@ -914,17 +946,20 @@
 						}
 						if (moves.indexOf(moveid) >= 0) continue;
 						moves.push(moveid);
-						if (moveid === 'sketch') sketch = true;
+						if (moveid === 'sketch' && format !== 'balancedhackmons' && format !== 'bh') sketch = true;
```
Zarel: You know you can set a separate variable instead of repeating `format !== 'balancedhackmons' && format !== 'bh'` in like six places.

I actually strongly recommend it.

js/search.js
KrisXV: especially in AAA and BH

js/search.js
Zarel: `isBH`, put it on line 895.

js/search.js
KrisXV: It will give this message in BH as well if it isn't included, and you're not required to start in your base forme BH

js/search.js
KrisXV: I had this originally and I'm not quite sure why I removed it

js/search.js
Zarel: Do you have examples?

js/search.js
Zarel: It sounds like you should be removing this message in BH if someone selects a Mega forme, not adding it.

Zarel:
Actually looks great at this point.

PR #1077, opened 2018-01-31
Tournaments: validate button should not submit a team when disabled

HoeenCoder:


Zarel:
I swear we went over this. :|

Zarel:
It's okay for a button that looks disabled to give an error message when you click on it.

HoeenCoder:
If you mean doing nothing here, it returns 
```
Your team was rejected for the following reasons:

- You sent invalid team data. If you're not using a custom client, please report this as a bug.
```
so something needs to change, are you maybe suggesting an error reply popup from client? `You need to select a team first` maybe.

Zarel:
Yeah.

HoeenCoder:
This appears to have been forgotten and fixed elsewhere at some point so closing.

PR #1081, opened 2018-02-22
Teambuilder: Filter Wormadam correctly in STABmons

KrisXV:


js/search.js
```diff
@@ -995,6 +995,8 @@
 					var baseTemplate = Tools.getTemplate(template.baseSpecies);
 					for (var j = 0; j < template.types.length; j++) {
 						if (template.battleOnly) continue;
+						// Temporary until all in-battle formes are supported
+						if (template.forme === 'Ultra' || template.forme === 'Pirouette' || template.forme === 'Zen' || template.baseSpecies === 'Castform') continue;
```
Zarel: ...?

Where is `template.battleOnly` set?

AustinXII: Looks like it’s one line above

KrisXV: BattleOnly currently only covers mega evolutions and primals

Zarel: @AustinXII I was asking for where it's set, not where it's read. :|

@KrisXV Sounds like that should be fixed.

KrisXV:
I'll get to this in a bit, I'm having issues with my local file right now so it could take a while

PR #1084, opened 2018-03-02
Power Lens and co should now correctly show Speed halved.

Zrp200:
Right now, when using items such as Power Anklet, the speed is not shown as being halved.


Zrp200:
I know I probably should combine those two, but I don't know how to do it on here, or if it is possible.

Zarel:
Am not a huge fan of using regexes for these kinds of simple matches, but I guess it's concise and sufficiently readable here.

PR #1086, opened 2018-03-03
Classify EV boosting items as "usually useless"

Zrp200:
See title.


Marty-D:
Aren't there 6 of these? You're missing one.

Zrp200:
You're right...

Zrp200:
Thanks @Marty-D

PR #1087, opened 2018-03-04
Don't throw errors when trying to animate Mummy activating but failing

asgdf:
When Mummy activates but fails due to Protective Pads, it will send an -activate message, but not the old/new ability of the attacker. Previously, it would throw errors when trying to animate the ability change. Now it just shows this:
grafik


urkerab:
Ugh, yes of course, I forgot to test with the actual client. Thanks again for cleaning up after me!

PR #1088, opened 2018-03-09
Add property for battle-only formes

KrisXV:


Zarel:
Do you use this anywhere?

KrisXV:
@zarel I’ll use it in my other pull request once I get on a computer

PR #1091, opened 2018-03-13
Suppress crash in trick room

HoeenCoder:
Something I noticed when setting trick room with `/editbattle`

I'm not sure if this is an issue in un-edited battles, so feel free to reject this if you don't care about protecting against this issue from `/editbattle` or `/evalbattle`.

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -5997,9 +5997,6 @@ var Battle = (function () {
 				this.addPseudoWeather(effect.name, 5, maxTimeLeft);
 
 				switch (effect.id) {
-				case 'trickroom':
-					actions += "" + poke.getName() + ' twisted the dimensions!';
-					break;
 				case 'wonderroom':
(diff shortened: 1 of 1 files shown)


Zarel:
Is that a real message from in-game though?

HoeenCoder:
Thats a good question actually, I don't know if there is a message that for trick room ingame that dosen't include the user's name. If not maybe it should just be `undefined twisted the dimensions`.
@Marty-D do you know anything about the above?

Marty-D:
The in-game message has the user's name in it because it's always started by a Pokemon. I would suggest just moving it to the bottom of this `switch`, i.e.
```javascript
switch (effect.id) {
...
case 'trickroom':
	if (poke) {
		actions += "" + poke.getName() + ' twisted the dimensions!';
		break;
	}
	// falls through
default:
	actions += effect.name + " started!";
	break;
}

Slayer95:
@HoeenCoder it seems you forgot to comment after you updated the PR.
Merging given Marty's review.

PR #1092, opened 2018-03-25
Show userlist widget in battles

urkerab:
Always shows in collapsed mode, like for side rooms. Also handles small layouts.


Zarel:
That looks approximately what I'd expect it to.

Zarel:
I'm still planning a client rewrite <_<

PR #1094, opened 2018-03-29
Graphics: Implement missing ballistic retreat for contact moves (ballistic2Back)

### LINKED ISSUE #1093: Transition 'ballistic2Back' not implemented
First referenced [back in 2013](https://github.com/Zarel/Pokemon-Showdown-Client/blob/b16c1527daf8bd23fb378dfdfb7b7b01aa7edb0e/data/graphics.js#L175)

There is a [ballistic2Under](https://github.com/Zarel/Pokemon-Showdown-Client/blob/b16c1527daf8bd23fb378dfdfb7b7b01aa7edb0e/js/battle.js#L661-L665), however.

Slayer95:
Fixes #1093

Before                                                           | After
------------------------------------------------|------
Before | After

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -2772,6 +2772,9 @@ var Battle = (function () {
 		if (transition === 'ballistic2Under') {
 			transitionMap.top = (pos.top < oldpos.top ? 'quadDown' : 'quadUp');
 		}
+		if (transition === 'ballistic2Back') {
+			transitionMap.left = (pos.left < oldpos.left ? 'quadDown' : 'quadUp');
+		}
 		if (transition === 'swing') {
 			transitionMap.left = 'swing';
 			transitionMap.top = 'swing';


Zarel:
I don't understand:

1. Why `ballistic2`? What was wrong with `ballistic`? Specifically, what was wrong that required introducing `ballistic2` instead of just changing `ballistic`?

2. Why `ballistic2Back`? "Ballistic" refers to a ballistic (parabolic) arc, it should be the same forward and back.

3. Why make it go under? `ballisticUnder` is exactly the opposite of how physics works outside of balloons, and it was really only designed as a hack for spinning (for moves like Gyro Ball).

Zarel:
> This is stuff you wrote :| The `2` means quadratic easing I guess?

Okay, I figured it out, `ballistic2` doesn't go outside of the range of the start/end positions, `ballistic` does.

`ballistic2Back` was apparently also wrote by, but I think in the end I just meant to use `ballistic2` in those places.

> It goes "under" only if the attacker is in the opposite side to the spectator. If the attacker is in the own side of the spectator, it goes "up". Overall, the 2D bias can be understood as towards the side of the attacker.

That's even worse! It should look approximately the same in either direction.

> As for aesthetics and practicality, a `ballistic2Back` transition looks more like an assault, so I find the new animation superior.

Can you clarify what "an assault" means? I'm pretty sure I meant to for `ballistic2Back` to look exactly the same as `ballistic2`.

I think having the retreat be a weird slow-dig back looks very weird... Digging through the ground should look like the Dig animation, with a fade-out and a fade-back-in, anyway. Not that Slam should involve digging through the ground in the first place.

Zarel:
My vision for Body Slam et al was always "jump on opponent, then jump back towards start point", I think.

Slayer95:
> Can you clarify what "an assault" means?

The language barrier might have a role here, but I allude to the longer time in which the two sprites are in apparent contact.

> I'm pretty sure I meant to for ballistic2Back to look exactly the same as ballistic2.

Do you want to me to replace them all? Or maybe someone else (e.g. @Ridaz) should look into this in a case by case basis? - There are 43 instances of ``ballistic2Back`` in ``graphics.js`` currently, and several involve animations done by Ridaz or @ascriptmaster

Zarel:
Yeah, I'd prefer if @Ridaz looked at them in a case-by-case basis.

Ridaz:
Yeah I'll have a look. Will make a commit on it in a couple of days

Zarel:
...Ridaz did not make a commit on it.

PR #1096, opened 2018-04-01
Add client command to disable AFD sprites

asgdf:
As described by Zarel.
However, currently the preference would not last beyond a refresh, which would be quite annoying if someone really doesn't like the afd sprites. We could add a setting, and then have the value of Config.server.afd be `!settingToDisable && valueSetByZarelDuringAfd`? That would stop people from enabling them when it's not AFD though. Otherwise we'd need 3 values - "always off", "always on" and "only on during AFD".

Also, currently this requires opening/closing the battle room to have it take effect. There is a `Site.updateSprites()`, but that will have all sprites hidden and not properly positioned. Is that intentional or can the method be modified to properly display the active mon?

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -888,6 +888,17 @@
 				}
 				return text; // Send the /avatar command through to the server.
 
+			case 'afd':
+				var cleanedTarget = toId(target);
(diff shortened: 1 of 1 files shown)


Zarel:
If you want it to take effect instantly, the usual way is to replay the battle.

PR #1097, opened 2018-04-01
Switch sprites immediately when toggling AFD mode

asgdf:


Slayer95:
There is a ``Side#updateSprites()`` method that we should probably use instead of replaying the battle...

asgdf:
Well that's what I was [planning to use](https://github.com/Zarel/Pokemon-Showdown-Client/pull/1096), but Zarel said I should just replay the battle instead.

Zarel:
I didn't realize there was an `updateSprites` function. Does it work? I assume no, because I remember having to refactor various features away from previously broken implementations into the reset+fastForward trick.

Zarel:
If you do use `fastForward`, you shouldn't fastforward to `-1` but rather to the current turn.

asgdf:
>I didn't realize there was an updateSprites function. Does it work?
>>Also, currently this requires opening/closing the battle room to have it take effect. There is a Site.updateSprites(), but that will have all sprites hidden and not properly positioned. Is that intentional or can the method be modified to properly display the active mon?

So basically it works except but doesn't show the currently active pokemon.

>If you do use fastForward, you shouldn't fastforward to -1 but rather to the current turn.

I saw that the side switching did that, but I tested it and it seemed to work fine with -1? What's the difference?

Slayer95:
> So basically it works except but doesn't show the currently active pokemon.

A way to make it show would be to ``.animDragIn()`` each sprite afterwards -passing their respective slots. Maybe a tad too hacky?

Zarel:
The difference is that if you're _not_ at the end of the log, it skips back to the part that's currently being played, rather than the end.

Zarel:
This seems kind of dangerous to do with a lot of battles open, though – huge lag. In theory we probably want it to be able to defer the replay until the next time the battle window is opened.

Zarel:
This is fine for now, though.

PR #1098, opened 2018-04-02
Fix ignoring for (semi)locked users

asgdf:
Previously, messages would still go through because locked/semilocked users were considered as having auth in the room.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1488,7 +1488,7 @@
 			var userid = toUserid(name);
 
 			var speakerHasAuth = " +\u2606".indexOf(name.charAt(0)) < 0;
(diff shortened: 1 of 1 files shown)


Zarel:
In what situation would you not want this to go through, anyway?

asgdf:
I think the idea was that global staff can't block messages in rooms?

Zarel:
Yes but locked users can't send messages in rooms other than Help rooms which obviously shouldn't be blocked...

asgdf:
Well, yes. But this was not about messages sent by locked users, not messages received by blocked users. Locked users were unable to ignore in rooms.

Zarel:
Oh, I see.

PR #1104, opened 2018-04-16
Fix Ignore Nicknames setting not applying to messagebar

asgdf:
Since removing the nicknames is done by adding a CSS class, add the same class to the messagebar.
(is there a better name for that temporary move/switch popup message thingie than "message(bar)"?)


Zarel:
I don't actually know what to call it, just the "messages section" or something.

PR #1105, opened 2018-04-17
Teambuilder: Fix CAP LC list

asgdf:
- make sure CAP uses the singles teambuilder table (since that's where the CAPmons are)
- reorder to CAP -> CAP NFE -> CAP LC (was CAP -> CAP LC -> CAP NFE)


Marty-D:
Hey you beat me to it! 👍

PR #1107, opened 2018-04-25
Change request protocol to fetch protocol

Frensly:
Fixes loading client improperly on new clients.

### DIFF
--- a/crossdomain.php
+++ b/crossdomain.php
@@ -12,7 +12,7 @@ if (preg_match('/^([a-z0-9-_\.]*?)\.psim\.us$/', $host, $m)) {
 	die; // not authorised
 }
 
-$protocol = @$_REQUEST['protocol'] ?? 'http:';
+$protocol = @$_URI['protocol'];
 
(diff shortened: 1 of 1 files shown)


Zarel:
http://php.net/manual/en/reserved.variables.php

I don't see `$_URI` on here...

AustinXII:
Perhaps your confusing ``$_uri`` with ``$request_uri``?

AustinXII:
Yeah I did mean that sorry my php is pretty rusty but also I was talking to Frensly :P

Frensly:
Yes autstin your wrong actually, i was kind of, and $_SERVER['REQUEST_URI'] is a neat way i suppose. i mean i suggest using GET_URI instead of REQUEST_URI but im gonna test it first on myside

Zarel:
You have literally never tested this considering you've sent three pullreqs so far and none of them work.

Zarel:
I'm blocking you, Adam Snowden. :|

PR #1108, opened 2018-04-26
Fix new clients loading improperly

Frensly:
Fixes new clients loading improperly.


Zarel:
Don't make new pullreqs every time you want to edit something, also this is still wrong. :|

PR #1111, opened 2018-05-04
Disable AFD mode by default

asgdf:
Most notably, testclient always had afd enabled. Not sure how it's disabled on main, as an unstaged change? or an unpushed commit?


Zarel:
Yeah, unstaged change.

Zarel:
I can take yours.

PR #1113, opened 2018-05-08
Oldgens: Don't show abilities the Pokemon didn't have at the time

asgdf:
In the teambuilder and in battle tooltips, for example, Pelipper will no longer have Drizzle displayed as a possible ability if it didn't have it in that gen.

By the way, is there a reason we do multi-line conditions by putting the conditional operators at the end of the previous line instead of the start of the next one? I think it's legitimately harder to read.


MacChaeger:
It might be easier to read if it were indented based on paren depth, but that would likely defeat the purpose of breaking it up in the first place. I definitely agree that it's really hard to parse multiline conditionals that are really complex like the ones that plague the randbats generator.

Zarel:
My solution for multiline conditionals has always been to break it apart. This also lets you name the intermediate value.

```
if (tickets[t].ip === user.latestIp && tickets[t].banned &&
  (!Punishments.sharedIps.has(user.latestIp) ||
    Punishments.sharedIps.has(user.latestIp) && !user.autoconfirmed)) {
```

is bad because it's too much information in one statement, and takes a long time to read. Breaking it up into multiple lines doesn't help very much at all.

```
const checkIp = !(Punishments.sharedIps.has(user.latestIp) && user.autoconfirmed);
if (checkIp && tickets[t].ip === user.latestIp && tickets[t].banned) {
```

names the intermediate value `checkIp`, which makes it clearer what it's for, and also prevents there from being too much code on one line.

Zarel:
You might want a helper function `hasAbility` or something.

asgdf:
Ok so this turned into a bit more of a refactor than planned, but now "which abilities does this pokemon have in this gen" and "does this pokemon have this ability in this gen" run off one central implementation rather than having like 5 implementations all over the place.
This is probably slightly less performant than before (especially because it searches for the correct BattleTeambuilderTable on every call), but I didn't notice any perf issues while testing. Should I just have the caller pass in the table instead of the gen?

asgdf:
Oh wait ``Object.assign()`` is ES6. Do we have an established way of doing that in the client?

Zarel:
`$.extend`

Zarel:
https://api.jquery.com/jQuery.extend/

Zarel:
Also feel free to just polyfill `Object.assign`. I started lightly polyfilling other functions because people kept on breaking the client by using ES6 functions

asgdf:
Alright I just replaced it with $.extend.

build-tools/build-indexes
```diff
@@ -619,6 +619,8 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 		BattleTeambuilderTable[gen].overrideType = overrideType;
 		const overrideAbility = {};
 		BattleTeambuilderTable[gen].overrideAbility = overrideAbility;
+		const removeSecondAbility = {};
```
Zarel: This seems overkill? I believe `ability.gen == 4 && gen === 3` is the only situation you'd want to remove the second ability?

asgdf: What motivated me to do this was that from gen 6 to 7, Beartic got Slush Rush, Gigalith got Sand Stream, Pelipper got Drizzle and Vamilluxe got Snow Warning.

However, I just noticed that the gen 3->gen 4 ability gains aren't caught by this, because e.g. ``Dex.mod('gen3').data.Pokedex.hitmontop`` already has both Intimidate and Technician listed despite the latter not existing yet. Is that intentional?

asgdf: Hm that's annoying, so I will have to add the ability gen check after all. I had hoped to avoid that with this whole ordeal.

asgdf:
Okay, rebased, typescripted these two functions, still seems to work.

asgdf:
Bump - unless you still think the separate table is overkill and I should just hardcode exceptions in hasAbility?

build-tools/build-indexes
```diff
@@ -636,6 +638,11 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 			if (pastEntry.abilities['0'] !== nowEntry.abilities['0']) {
 				overrideAbility[id] = pastEntry.abilities['0'];
 			}
+			// in the gen 3 dex, Pokemon already have the abilities that were added in gen 4
+			const hasAbilityFromNewerGen = Tools.getTemplate(id).gen <= genNum && Tools.getAbility(pastEntry.abilities['1']).gen > genNum;
```
Zarel: Looks like you chose to add the ability gen in the table builder. Good choice.

js/client-teambuilder.js
```diff
@@ -2795,10 +2795,7 @@
 			} else {
 				set.item = '';
 			}
-			set.ability = template.abilities['0'];
-			var table = BattleTeambuilderTable['gen' + this.curTeam.gen];
-			var overrideAbility = table && table.overrideAbility[template.id];
-			if (overrideAbility) set.ability = overrideAbility;
+			set.ability = Tools.getAbilitiesFor(template.id, this.curTeam.gen)['0'];
```
Zarel: Does ability 0 actually change between generations? Ick.

src/battle-dex.ts
```diff
@@ -852,6 +852,38 @@ const Tools = {
 		return type;
 	},
 
+	getAbilitiesFor: function (template: any, gen: number) {
```
Zarel: `getAbilitiesFor(`, not `getAbilitiesFor: function (`. Look at the surrounding code.

(Ditto `hasAbility`)

src/battle-dex.ts
```diff
@@ -852,6 +852,38 @@ const Tools = {
...
+		template = this.getTemplate(template);
+		if ((!isNaN(gen) && gen < 3) || !template.abilities) return {};
```
Zarel: `!isNaN`?

I assume the `gen` parameter is intended to be optional? What's the intended behavior when `gen` isn't passed? You probably want to specify a default gen, like:

`getAbilitiesFor: function (template: any, gen = 7) {`

or

`getAbilitiesFor: function (template: any, gen = 0) {`

src/battle-dex.ts
```diff
@@ -852,6 +852,38 @@ const Tools = {
...
+		if (!removeSecondAbility && templAbilities['1']) {
+			abilities['1'] = templAbilities['1'];
+		}
+		if (gen >= 5 && templAbilities['H']) abilities['H'] = templAbilities['H'];
+		if (gen >= 7 && templAbilities['S']) abilities['S'] = templAbilities['S'];
+
+		return abilities;
+	},
+
+	hasAbility: function (template: any, ability: string, gen: number) {
```
Zarel: If `gen` is optional, it should be specified as `gen?: number`

js/client-teambuilder.js
asgdf: Gengar being the main case (and Naviathan).

asgdf:
Fixed the issues I think (also replaced all my `var`s with `const`s in the ts file)

PR #1114, opened 2018-05-09
Support mray- and necrozma-type mega messages for arbitrary Pokemon

MacChaeger:
also see https://github.com/Zarel/Pokemon-Showdown/pull/4654

### DIFF
--- a/js/battle.js
+++ b/js/battle.js
@@ -4994,7 +4994,7 @@ var Battle = (function () {
 			case '-mega':
 				var poke = this.getPokemon(args[1]);
 				var item = Tools.getItem(args[3]);
-				if (args[2] === 'Rayquaza') {
+				if (!item) {
 					actions += "" + Tools.escapeHTML(poke.side.name) + "'s fervent wish has reached " + poke.getLowerName() + "!";
 				} else {
 					poke.item = item.name;
@@ -5008,7 +5008,7 @@ var Battle = (function () {
 				break;
 			case '-burst':
 				var poke = this.getPokemon(args[1]);
-				actions += "Bright light is about to burst out of " + poke.getLowerName() + "!";
+				actions += "Bright light is about to burst out of " + poke.getLowerName() + "! <br />" + poke.getName() + " regained its true power through Ultra Burst!";
 				break;
 
 			case '-start':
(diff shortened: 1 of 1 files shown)


js/battle.js
```diff
@@ -6745,8 +6748,6 @@ var Battle = (function () {
 				this.message('<small>' + poke.getName() + "'s disguise was busted!</small>");
 			} else if (toId(newSpecies) === 'zygardecomplete') {
 				this.message('' + poke.getName() + ' transformed into its Complete Forme!');
-			} else if (toId(newSpecies) === 'necrozmaultra') {
-				this.message('' + poke.getName() + ' regained its true power through Ultra Burst!');
```
Marty-D: You can't just remove this because we still have to support old replays. Also this message happens after the animation, which is why it's separate from the one that happens prior to the animation.

MacChaeger: This isn't a change in protocol, though, just how the client handles it. Old replays work fine. As for the message order, this is exactly how mega evolution is handled too.

MacChaeger: OK in that case I got nothing. Somebody other than me with experience in rewriting client protocol will have to fix this, and while they're at it, get the right custom forme change animation to happen appropriately as well.

js/battle.js
```diff
@@ -6711,6 +6711,9 @@ var Battle = (function () {
 			poke.removeVolatile('typechange');
 
 			var newSpecies = args[2];
+			if (!newSpecies) {
+				poke.sprite.animTransform($.extend(spriteData, Tools.getTemplate(poke)), true);
+			}
```
MacChaeger: If all the rest is wrong, this is actually a new protocol added in the server PR in case of an Pokemon with an active Illusion undergoing a permanent forme change so that it does a `'-detailschange'` without actually changing details. The current protocol supports a details change that doesn't actually change anything but still does an animation that doesn't necessarily reveal it as an Illusion, but this allows the server to effectively tell the client ahead of time that there's nothing to actually change and to just play the animation.

Zarel:
I think my most recent changes have mostly fixed message order? Tell me if something still needs to be done.

MacChaeger:
What changes are you referring to?

Zarel:
while implementing `BattleText`, I fixed most of the relevant message order issues.

PR #1116, opened 2018-05-11
Improve titles of uploaded replays

MacChaeger:
Split off as requested

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1244,6 +1244,19 @@
 			var reader = new FileReader();
 			reader.onload = function (e) {
 				var html = e.target.result;
+				var titleStart = html.indexOf('<title>');
+				var titleEnd = html.indexOf('</title>');
+				var title = 'Uploaded Replay';
+				if (titleStart >= 0 && titleEnd > titleStart) {
+					title = html.slice(titleStart + 7, titleEnd - 1);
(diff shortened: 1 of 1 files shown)


MacChaeger:
Actually this is probably pointless since it would be dumb to have replay parsing code in two places in the client so the other PR will end up conflicting with this.

Zarel:
What do you mean?

MacChaeger:
Oh I guess you wanted me to just pull the log directly from the replay server's database instead of having it build all the embedding just to strip it away, which makes a lot of sense now that I understand replays better. In that case, I get why you wanted this split off, but I'm going to need some help with how to access that log.

js/client-battle.js
```diff
@@ -1255,7 +1268,7 @@
 				var index3 = html.indexOf('</script>');
 				html = html.slice(0, index3);
 				html = html.replace(/\\\//g, '/');
-				app.receive('>battle-uploadedreplay\n|init|battle\n|title|Uploaded replay\n' + html);
+				app.receive('>battle-uploadedreplay\n|init|battle\n|title|' + title + '\n' + html);
```
Zarel: Should `title` maybe be sanitized?

Zarel: I guess it isn't a huge deal.

PR #1117, opened 2018-05-12
Teambuilder: Prevent some instances of duplicate search results

asgdf:
In particular, this is so that when you search "mega" in the teambuilder, you don't get Megazard/Mega Mewtwo twice. These have two aliases each (like megacharizardx/megax) which are not consecutive, so they were not caught by the previous duplicate check.


asgdf:
Hm. Is it worth introducing an alternate data structure to make it possible in O(n)?

Zarel:
You could just hardcode the cases that are affected.

asgdf:
Okay, just hardcoded it to skip megax/megay when searching for any part of "mega".

PR #1118, opened 2018-05-14
Fix build script on windows

asgdf:
~~Also update package-lock while I'm at it~~


Zarel:
lol

Zarel:
I committed a package-lock removal an hour ago (which is why I didn't commit package-lock changes in the first place) and was just lazy about pushing it.

https://github.com/Zarel/Pokemon-Showdown-Client/commit/194e07e6d0d8495018fa95fdc81af6232e74b476

PR #1119, opened 2018-05-17
Build process: Don't crash when no config found

asgdf:
I know you mentioned an init script, but since update already writes to the config file, stopping it from erroring out while reading first seemed more straightforward.
I think all other files can reasonably assumed to be present, unless the TS compliation doesn't work.

### DIFF
--- a/build-tools/update
+++ b/build-tools/update
@@ -28,7 +28,12 @@ process.stdout.write("Updating version... ");
 
 let version = require('../package.json').version;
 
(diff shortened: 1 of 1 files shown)


Zarel:
Why not just write a default config file?

Zarel:
I did this better in 66c83f14cda26a843d

Main difference is that the config file it generates is actually important for making the client not crash, and also installing dependencies needs to happen before any build script is run.

PR #1120, opened 2018-05-18
Support UHTML in PMs

sirDonovan:


Zarel:
Client rewrite will merge how PMs and chatrooms work. We might want to just wait until then?

Zarel:
I guess I can merge this for now, though, if you're in a rush.

PR #1125, opened 2018-06-01
Battles: set BG and BGM properly

whalemer:
and maybe some other things that numericid is for


Zarel:
Maybe I should just push my refactor? I should probably test it more, though; this is probably fine in the interim.

PR #1131, opened 2018-07-10
Fix Transform of female Smeargle into Rotom

urkerab:
See this [replay](https://replay.pokemonshowdown.com/gen7hackmonscup-773575134) although given that this is Hackmons Cup it's actually a (female) Salandit that impostors Rotom, but I assume female Smeargle using Transform on Rotom would exhibit the same bug. Since Salandit is female, the client tries to look for a female Rotom sprite but actually finds the Rotom-Frost sprite instead.

I'm assuming here that Transform copies the visual appearance of any gender the target has, so apologies if that is in fact incorrect mechanics.


Zarel:
@Marty-D

Zarel:
(Just to confirm the mechanics)

PR #1134, opened 2018-07-16
Use either command character for clientside commands

whalemer:


Zarel:
This won't actually broadcast, though.

Zarel:
Better to show an error when someone tries to broadcast a client-only command.

whalemer:
That's just mirroring the behavior for non-broadcastable server commands though.

Zarel:
Non-broadcastable server commands should be fixed, too.

whalemer:
How should server commands expose if they're broadcastable? (in the ``ChatCommand`` structure)

whalemer:
Can you elaborate? Do you mean the ``!command`` entry?

Zarel:
Yes.

whalemer:
What should be used as the flag?

Zarel:
I was saying we might want to just use the same flag for both. They line up pretty well anyway.

Zarel:
We can do `"!command": 'broadcast',` or `"!command": 'pm',` for the ones that are _only_ one or the other, but those should be very rare.

PR #1137, opened 2018-08-04
Refactor Battle constructor

yuzeh:
This change makes it possible to initialize a `Battle` object with a
different implementation of `BattleScene`.

Sites where `new Battle()` was used are now replaced with
`Battle.create`.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -22,7 +22,7 @@
 			this.$foeHint = this.$el.find('.foehint');
 
 			BattleSound.setMute(Tools.prefs('mute'));
-			this.battle = new Battle(this.$battle, this.$chatFrame, this.id);
+			this.battle = Battle.create(this.$battle, this.$chatFrame, this.id);
 			this.tooltips = new BattleTooltips(this.battle, this);
 
 			this.battle.roomid = this.id;
--- a/js/replay-embed.js
+++ b/js/replay-embed.js
@@ -25,7 +25,7 @@ var Replays = {
 			if (action) self[action]();
 		});
 
-		this.battle = new Battle(this.$('.battle'), this.$('.battle-log'));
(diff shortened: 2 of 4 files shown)


yuzeh:
I figured out how to work around this, so this PR is now moot... but this was my usecase (including the workaround):

```js
Shim.createBattle = function createBattle() {
  let battle;
  if (Battle.create) {
    // We're using the forked version of psc
    battle = new Battle(makeScene());
    battle.init();
  } else {
    // We're using master branch on psc, replace BattleScene with a proxy
    BattleScene = window.BattleScene = new Proxy({}, {
      construct: function (target, args) {
        return makeScene();
      },
    });
    battle = new Battle(null, null);
  }
  return battle;
};
```

Zarel:
I would probably make it so that if `new Battle()` is called with no arguments (or null arguments), the `Scene` it constructs is a shim Scene.

yuzeh:
> I would probably make it so that if new Battle() is called with no arguments (or null arguments), the Scene it constructs is a shim Scene.

I think this solution makes sense if you want to support a headless client in this repo.

Doesn't seem like the path that this repo is taking, but if so I'm happy to take a stab at it. I just spent this morning doing a variant of this project (getting headless client to run on v8) and there are a few other things that need to change:

* `PokemonSprite` also needs a separate implementation
* Dependency on jQuery utility functions like `$.trim`, `$.isArray`
* `Tools.resourcePrefix`, `Tools.fxPrefix` use `document.location`

(My current implementation uses ES6 Proxy objects for `BattleScene`, `PokemonSprite`, and `$`, so I'm not sure how fast it is.)

Zarel:
I definitely wouldn't mind moving towards supporting a headless client, yes.

PR #1139, opened 2018-08-21
Stringify PM target

HoeenCoder:
A userid should be a string, `.data` can return values other than strings, for example if the userid is `false` it returns the boolean `false`. This may also happen with names that are all numbers.

This fixes a bug where trying to PM a user with a name that resolves to a falsey value causes you to PM `~` instead.

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -310,7 +310,7 @@
 			var text = $.trim($target.val());
 			if (!text) return;
 			var $pmWindow = $target.closest('.pm-window');
-			var userid = $pmWindow.data('userid');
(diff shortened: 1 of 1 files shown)


Zarel:
This would cause a different crash (`.toString()` can't be called on `null`/`undefined`); why not do the usual `'' + variable` cast?

Zarel:
Fixed

Zarel:
For future reference, you should never call `variable.toString()` (or `variable.valueOf()`), it's a method used internally for casting.

There are two ways to cast to string in JavaScript, they are `String(variable)` and `'' + variable`. PS prefers the latter.

PR #1143, opened 2018-08-30
Adding commands for a friends list

tmagicturtle:
These changes add commands for /fl, /friends, /friendslist, and /getfriends as well as /addfriend and /removefriend. The /friendslist command creates a table of existing friends (with buttons to remove or message the friend), outgoing (sent) pending requests, and incoming (received) pending requests with buttons to accept or reject.

When removing a friend, there is currently a bug where a "Could not remove PLAYER from your friend list." message will appear, though it will successfully remove the friend. I am not sure if my changes fix this. Friends list functionality is currently limited to the above and can be expanded on.

### DIFF
--- a/lib/dispatcher.lib.php
+++ b/lib/dispatcher.lib.php
@@ -491,7 +491,7 @@ class DefaultActionHandler {
 
 		$userid = $psdb->escape($curuser['userid']);
 		$player = $psdb->escape($reqData['player']);
-		$res = $psdb->query(
+		if($res = $psdb->query(
 			"DELETE FROM `ntbb_friendlist` " .
 			"WHERE (" .
 				"`p1`='" . $userid . "' AND `p2`='" . $player . "'" .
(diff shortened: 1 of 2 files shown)


lib/dispatcher.lib.php
```diff
@@ -491,7 +491,7 @@ public function removefriend($dispatcher, &$reqData, &$out) {
 
 		$userid = $psdb->escape($curuser['userid']);
 		$player = $psdb->escape($reqData['player']);
-		$res = $psdb->query(
+		if($res = $psdb->query(
```
Zarel: This `if(` appears unclosed by a matching `)`.

Zarel:
Unfortunately, I think we're going to move away from doing a friends list client-side. A server-side friends list makes a lot more sense, especially since the server needs to know who your friends are to update you on their online status.

lib/dispatcher.lib.php
tmagicturtle: Ah, thank you. Because I couldn't actually test this code, I wasn't sure if fixing the bug would be if($res = $psdb->query [...]) ']' . $reqData['player'] . ' has been removed from your friend list.'; or what I've changed on the following line, and I guess when I was still looking into which was correct, I forgot to undo the unclosed if(. Seeing as the pull request was closed, I'll instead start looking into a server-side friends list system.

PR #1144, opened 2018-09-02
Move Mythical pokemon down in VGC teambuilder

HoeenCoder:
This should in theory work, but it didnt apply when I tested it via testclient.html. PRed anyways in case its something that needs to be applied main's (build script for teambuilder probably).

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -352,7 +352,7 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 
 		const tierOrder = (() => {
 			if (isVGC) {
-				return ["Mythical", "Restricted Legendary", "Regular", "NFE", "LC"];
+				return ["Restricted Legendary", "Regular", "NFE", "LC"];
 			}
 			if (isDoubles) {
 				return ["DUber", "DOU", "DBL", "(DOU)", "DUU", "DNU", "NFE", "LC Uber", "LC"];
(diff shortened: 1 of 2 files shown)


Zarel:
The reason Mythical is first is so it's easy to remove. Putting it last makes the removal a bit more complicated.

Zarel:
And also doesn't actually remove them...

Zarel:
(You need to run `./build indexes` to rebuild the search index, if you're wondering.)

Zarel:
This is not what I mean by remove...

Zarel:
https://github.com/Zarel/Pokemon-Showdown-Client/blob/9bf108dad87fbf77a26578bad3d94c2906a67cd6/js/search.js#L825-L848

Zarel:
This is the slicing; you should slice at `slices["Restricted Legendary"]` I think

HoeenCoder:
Ah that makes sense

Zarel:
@HoeenCoder bump?

HoeenCoder:
Ah shit I forgot about this, can do.

build-tools/build-indexes
```diff
@@ -352,7 +352,7 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 
 		const tierOrder = (() => {
 			if (isVGC) {
-				return ["Mythical", "Restricted Legendary", "Regular", "NFE", "LC"];
+				return ["Restricted Legendary", "Regular", "NFE", "LC"];
```
Zarel: This should be reverted:

```suggestion
				return ["Mythical", "Restricted Legendary", "Regular", "NFE", "LC"];
```

PR #1145, opened 2018-09-09
Tooltips: Show status Z-move effects

asgdf:


Zarel:
Thanks! Is this all of them?

Zarel:
https://github.com/Zarel/Pokemon-Showdown-Dex/blob/master/js/pokedex-moves.js#L154-L176

is PSdex's implementation, if you want to use that as reference.

asgdf:
Yeah I think I got all of these - I copypasted it from the /dt implementation.

PR #1146, opened 2018-09-10
Tooltips: Properly show shortdesc for moves without long desc

asgdf:
All this is basically just removing the `if (move.desc)` around the whole tooltip.
There are moves that don't have a desc, but all moves have a shortdesc. Both branches of the condition already support longdescless moves.

On a related note, is it supposed to be the case that the priority/flag/ability boost hints in the move tooltip are only shown in the current gen?

### DIFF
--- a/js/client-battle-tooltips.js
+++ b/js/client-battle-tooltips.js
@@ -322,80 +322,78 @@ var BattleTooltips = (function () {
 		if (additionalInfo) text += '<p>' + additionalInfo + '</p>';
 		text += '<p>Accuracy: ' + accuracy + '</p>';
 		if (zEffect) text += '<p>Z-Effect: ' + zEffect + '</p>';
-		if (move.desc) {
-			if (this.battle.gen < 7 || this.battle.hardcoreMode) {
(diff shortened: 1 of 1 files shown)


Zarel:
Which moves don't have a shortdesc?

Zarel:
And yes, the reason the flag hints are only shown in the current gen is because past-gen flags are not available in PS.

Zarel:
If you want to do an override table for past-gen descs, I won't stop you.

It might be good just to write long descs for all the moves that don't have them. `shortDesc`s contain information that `desc` doesn't, namely info about priority and flags.

asgdf:
Some shortdescs contain priority information, but far from all of them, so it doesn't really seem like something to rely on? And which ones contain flag info?

Zarel:
I mean, the idea is that a `shortDesc` is intended to stand on its own, while a `desc` is intended to be shown next to autogenerated information about priority, targets, and flags.

The "bypasses Protect" flag is definitely shown, but I guess that's also mentioned in long move descriptions, too. I guess can't think of any other flag important enough to be shown.

PR #1147, opened 2018-09-21
Fix Illusion bug where it shows more than 6 pokes for species clause

Shashwat986:
Fixes Illusion bug for all battles with species clause. That includes randbats, so a major pain point should be fixed.


Zarel:
This is wrong, though. Zoroark and the Pokémon Zoroark is trying to imitate are clearly different Pokémon.

Zarel:
You're not removing a bug, you're adding a bug where there was no bug before.

PR #1151, opened 2018-10-06
Headless client

### LINKED ISSUE #1148: Support headless client
Picking up where we left off on #1137: I'm interested in implementing a headless client in the PSC repo.

This will let other projects use the battle state inference code that's already been implemented in this repo in a modular fashion. More importantly, it allows for *unit testing* of the battle state inference code (there doesn't seem to be any such testing currently).

My vision of how this will work:

* Per https://github.com/Zarel/Pokemon-Showdown-Client/pull/1137#issuecomment-410474024, `new Battle(null, null)` (or `new Battle()`) creates a headless Battle object.
* On constructor call, a stub `BattleScene` object is created that basically no-ops all of its methods. This can be done in at least two ways:
    * Common interface between a handwritten `StubBattleScene` class (along with a handwritten `StubPokemonSprite`) and the existing `BattleScene` class, or
    * ES6 `Proxy`.
* Unit tests are written to demonstrate how this should be used. I'm not quite sure myself how this is supposed to be done, but I've had luck doing the following:
```js
battle.activityQueue.push(message);
battle.play();
battle.fastForwardTo(-1);
```

If this sounds good, I can start working on a PR.

yuzeh:
Adds a way to instantiate a `Battle` object that works in non-browser environments (e.g. Node, V8) with minor modifications. Fixes #1148.

To do:
- [x] unit tests.
- [x] testclient.html

### DIFF
--- a/src/battle-dex.ts
+++ b/src/battle-dex.ts
@@ -193,12 +193,12 @@ const Tools = {
 
 	resourcePrefix: (() => {
 		let prefix = '';
-		if (document.location!.protocol !== 'http:') prefix = 'https:';
+		if (!window.document || !document.location || document.location.protocol !== 'http:') prefix = 'https:';
 		return prefix + '//play.pokemonshowdown.com/';
 	})(),
 
 	fxPrefix: (() => {
-		if (document.location!.protocol === 'file:') {
+		if (window.document && document.location && document.location.protocol === 'file:') {
 			if (window.Replays) return 'https://play.pokemonshowdown.com/fx/';
 			return 'fx/';
 		}
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -245,26 +245,26 @@ class Pokemon {
 		return this.ident.substr(0, 2) + slots[this.slot] + this.ident.substr(2);
 	}
 	removeVolatile(volatile: ID) {
-		this.sprite.removeEffect(volatile);
+		this.side.battle.scene.removeEffect(this, volatile);
 		if (!this.hasVolatile(volatile)) return;
 		delete this.volatiles[volatile];
 	}
 	addVolatile(volatile: ID, ...args: any[]) {
 		if (this.hasVolatile(volatile) && !args.length) return;
 		this.volatiles[volatile] = [volatile, ...args] as EffectState;
-		this.sprite.addEffect(volatile);
+		this.side.battle.scene.addEffect(this, volatile);
 	}
 	hasVolatile(volatile: ID) {
 		return !!this.volatiles[volatile];
 	}
 	removeTurnstatus(volatile: ID) {
-		this.sprite.removeEffect(volatile);
+		this.side.battle.scene.removeEffect(this, volatile);
 		if (!this.hasTurnstatus(volatile)) return;
 		delete this.turnstatuses[volatile];
 	}
 	addTurnstatus(volatile: ID) {
 		volatile = toId(volatile);
-		this.sprite.addEffect(volatile);
+		this.side.battle.scene.addEffect(this, volatile);
 		if (this.hasTurnstatus(volatile)) return;
 		this.turnstatuses[volatile] = [volatile];
 	}
@@ -278,7 +278,7 @@ class Pokemon {
 		this.turnstatuses = {};
 	}
 	removeMovestatus(volatile: ID) {
-		this.sprite.removeEffect(volatile);
+		this.side.battle.scene.removeEffect(this, volatile);
 		if (!this.hasMovestatus(volatile)) return;
 		delete this.movestatuses[volatile];
 	}
@@ -286,7 +286,7 @@ class Pokemon {
 		volatile = toId(volatile);
 		if (this.hasMovestatus(volatile)) return;
 		this.movestatuses[volatile] = [volatile];
-		this.sprite.addEffect(volatile);
+		this.side.battle.scene.addEffect(this, volatile);
 	}
 	hasMovestatus(volatile: ID) {
 		return !!this.movestatuses[volatile];
@@ -301,7 +301,7 @@ class Pokemon {
 		this.volatiles = {};
 		this.clearTurnstatuses();
 		this.clearMovestatuses();
-		this.sprite.clearEffects();
+		this.side.battle.scene.clearEffects(this);
 	}
 	markMove(moveName: string, pp?: number, recursionSource?: string) {
 		if (recursionSource === this.ident) return;
@@ -475,7 +475,7 @@ class Pokemon {
 
 		pokemon.boosts = {};
 		pokemon.volatiles = {};
-		pokemon.sprite.removeTransform();
+		pokemon.side.battle.scene.removeTransform(pokemon);
 		pokemon.statusStage = 0;
 	}
 	copyTypesFrom(pokemon: Pokemon) {
@@ -613,19 +613,7 @@ class Side {
 	}
 	updateSprites() {
 		this.z = (this.n ? 200 : 0);
-		if (this.missedPokemon) {
-			this.missedPokemon.sprite.destroy();
-		}
-		this.missedPokemon = {
-			sprite: new PokemonSprite(null, {
-				x: this.leftof(-100),
-				y: this.y,
-				z: this.z,
-				opacity: 0,
-			},
-			this.battle.scene, this.n)
-		} as Pokemon;
-		this.missedPokemon.sprite.isMissedPokemon = true;
+		this.battle.scene.updateSpritesForSide(this);
 	}
 	setAvatar(spriteid: string) {
 		this.spriteid = spriteid;
@@ -801,7 +789,7 @@ class Side {
 			this.battle.message('' + Tools.escapeHTML(pokemon.side.name) + ' sent out ' + pokemon.getFullName() + '!');
 		}
 
-		pokemon.sprite.animSummon(pokemon, slot);
+		this.battle.scene.animSummon(pokemon, slot);
 
 		if (this.battle.switchCallback) this.battle.switchCallback(this.battle, this);
 	}
@@ -811,7 +799,7 @@ class Side {
 		if (oldpokemon === pokemon) return;
 		this.lastPokemon = oldpokemon;
 		if (oldpokemon) {
-			oldpokemon.sprite.animDragOut(oldpokemon);
+			this.battle.scene.animDragOut(oldpokemon);
 			oldpokemon.clearVolatile();
 		}
 		pokemon.clearVolatile();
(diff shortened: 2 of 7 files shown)


src/battle-scene-interface.ts
```diff
@@ -0,0 +1,176 @@
...
+	$statbar: JQuery<HTMLElement> | null; // cannot remove; accessed by BattleScene
+	$el: JQuery<HTMLElement> | null;
+} 
+
+class PokemonSpriteStub implements IPokemonSprite {
+	top: number = NaN;
+	left: number = NaN;
+	x: number = NaN;
+	y: number = NaN;
+	z: number = NaN;
```
Zarel: It seems unnecessary to implement these at all if they're going to just be NaN.

yuzeh: The core problem is that the `Pokemon` object holds a reference to its `sprite`, which is used in methods like `BattleScene.abilityActivateAnim` access:

https://github.com/Zarel/Pokemon-Showdown-Client/blob/24ed8577bf51057633be35115bb1d750e1f0ef51/src/battle-animations.ts#L1125-L1133

I went with a solution that I think would minimize the diff size, but I'm open to other solutions. One way to achieve this is to get rid of the `Pokemon.sprite` reference, and have all accesses gated by a `Battle.getSpriteFor(pokemon: Pokemon)` method.

Zarel: This could be implemented with a `Map`, but the client needs to be ES3-compatible.

You would be overriding the relevant `BattleScene` methods, such as `BattleScene#abilityActivateAnim`. I would not implement a sub Sprite at all, just leave `pokemon.sprite = null`. The stub Scene shouldn't be accessing it anyway, after all.

This may involve some complexity with TypeScript – I'm open to either `Pokemon#sprite` being typed as `Sprite` and the null assignment being `pokemon.sprite = null!;` or with `Pokemon#sprite` being typed as `Sprite | null` and the usage being `pokemon.sprite!`.

Zarel: In other words, reading from this stub Sprite at all would be an error. In that case, there's no reason to have it fail buggily rather than have it fail hard.

src/battle-scene-interface.ts
```diff
@@ -0,0 +1,176 @@
+interface IBattleScene {
```
Zarel: Explicitly defining an interface at all seems unnecessary. TypeScript supports duck typing, after all. You can just do `scene: BattleScene | BattleSceneStub` instead of `scene: IBattleScene`

src/battle-scene-interface.ts
```diff
@@ -0,0 +1,69 @@
...
+	destroyPokemonSprite(pokemon: Pokemon) { }
+	clearEffects(pokemon: Pokemon) { }
+	removeTransform(pokemon: Pokemon) { }
+	animFaint(pokemon: Pokemon) { }
+	animReset(pokemon: Pokemon) { }
+	anim(pokemon: Pokemon, end: ScenePos, transition?: string) { }
+	beforeMove(pokemon: Pokemon) { }
+	afterMove(pokemon: Pokemon) { }
+	updateSpritesForSide(side: Side) { }
+}
```
Zarel: image

Zarel: (You're missing a trailing linebreak)

Zarel:
I don't know how to feel about trading some readability and performance for better separation.

Zarel:
It's not even _that_ much better separation, you're still just tracking `spriteIndex` instead of `Sprite`, which is basically the same thing.

Zarel:
Yeah, sorry, having thought about it more, I don't think your `.spriteIndex` change adds anything useful. Can you revert that part?

Zarel:
I don't oppose moving functions like `removeEffect`, `addEffect`, `animSummon` etc from `pokemon.sprite.*` to `this.scene.*(pokemon,`. That part can stay. Just `spriteIndex` I consider unnecessary.

yuzeh:
I've added back the `sprite` reference, but I'll add this comment: the `spriteIndex` system makes it harder to break the relationship between model (`Battle`, `Side`, `Pokemon`) and view-controller (`BattleScene`, `PokemonSprite`). In this system, you could accidentally make this mistake, and even VSCode won't catch the error.

On the other hand, unit tests that are written should ultimately catch the issue.

src/battle.ts
```diff
@@ -245,26 +245,26 @@ class Pokemon {
 		return this.ident.substr(0, 2) + slots[this.slot] + this.ident.substr(2);
 	}
 	removeVolatile(volatile: ID) {
-		this.sprite.removeEffect(volatile);
+		this.side.battle.scene.removeEffect(this, volatile);
```
Zarel: `this.battle` exists on every `Pokemon` instance; there's no need to go through `.side`

Zarel:
Anyway, I can merge this as-is, and wait for a different pullreq for tests, or I can wait for tests. It's up to you.

yuzeh:
Let me add some tests.

yuzeh:
I wasn't entirely sure what the best way to add testing was... I ended up adding unit tests that would run in the browser (powered by mocha and chai).

All you need to do to run the tests is to run `./build`, and then open up `./unittest/index.html` in your browser.

Zarel:
I don't really like the idea of in-browser tests. We actually had them earlier but dropped them because they were inconsistent and failing for confusing reasons (it wasn't network issues because we were spinning up a local server – I guess a race condition).

The headless `battle.js` should run in Node, right? Can you just test with Mocha like that?

Zarel:
Is there still a jQuery dependency? I thought we got rid of it. Hold on a moment.

Zarel:
Oh, it's just a few random things. I can move those out of `battle.ts` and `battle-dex.ts` or you can, depending on whichever is easier for you.

yuzeh:
I'm pretty uncomfortable making this change, because I'm not entirely sure what good replacements are for the browser.

* `$.parseJSON` -- AFAIK this is the only thing that works in IE7, `JSON.parse` doesn't exist.
* `$.trim` -- this can be replaced fairly easily.
* `$.isArray` -- `Array.isArray` isn't introduced until IE9.

Zarel:
I moved everything relevant out of `battle-dex.ts` and `battle.ts`.

There are still some references to jQuery I think, but they shouldn't affect anything you're testing.

src/battle-dex.ts
```diff
@@ -193,12 +193,12 @@ const Tools = {
 
 	resourcePrefix: (() => {
 		let prefix = '';
-		if (document.location!.protocol !== 'http:') prefix = 'https:';
+		if (!document.location || document.location.protocol !== 'http:') prefix = 'https:';
```
yuzeh: @Zarel 

Wasn't sure what to do here. An alternate solution is to leave this alone and stub `document.location` in the test file.

Zarel: This is good.

Zarel: I'm surprised `document` exists at all here. Are you using Node or...?

test/battle-test.mocha.js
```diff
@@ -0,0 +1,98 @@
...
+			"|inactive|Time left: 150 sec this turn | 740 sec total",
+		].forEach(run);
+
+		chai.expect(p2gliscor.isActive()).to.be.false;
+		var p2kyurem = p2.pokemon[1];
+		chai.expect(p2kyurem.ident).to.equal('p2: Kyurem');
+		chai.expect(p2kyurem.details).to.equal('Kyurem-White, L73');
+		chai.expect(p2kyurem.hp).to.equal(226);
+		chai.expect(p2kyurem.maxhp).to.equal(303);
+		chai.expect(p2kyurem.isActive()).to.be.true;
```
Zarel: Ew, BDD

Any chance I can get you to write tests normally?

So like:

```
assert(p2kyurem.maxhp === 303);
assert(p2kyurem.isActive());
```

test/battle-test.mocha.js
```diff
@@ -0,0 +1,98 @@
...
+	});
+
+	it('should process a bunch of messages properly', function () {
+		var battle = new Battle();
+		battle.debug = true;
+
+		var run = function (msg) {
+			battle.add(msg);
+			battle.fastForwardTo(-1);
+		};
```
Zarel: Is there any reason you're adding these one by one instead of all at once?

yuzeh: There's no reason... fixed.

test/battle-test.mocha.js
```diff
@@ -0,0 +1,98 @@
...
+			"|rated|",
+			"|seed|",
+			"|rule|Sleep Clause Mod: Limit one foe put to sleep",
+			"|rule|HP Percentage Mod: HP is shown in percentages",
+			"|",
+			"|start",
+			"|switch|p1a: Leafeon|Leafeon, L83, F|100/100",
+			"|switch|p2a: Gliscor|Gliscor, L77, F|242/242",
+			"|turn|1",
+		].forEach(msg => battle.add(msg));
```
Zarel: You're looking for `battle.setQueue`.

test/battle-test.mocha.js
```diff
@@ -0,0 +1,98 @@
...
+		var p2kyurem = p2.pokemon[1];
+		assert(p2kyurem.ident === 'p2: Kyurem');
+		assert(p2kyurem.details === 'Kyurem-White, L73');
+		assert(p2kyurem.hp === 226);
+		assert(p2kyurem.maxhp === 303);
+		assert(p2kyurem.isActive());
+		assert(p2kyurem.item === '');
+		assert(p2kyurem.prevItem === 'Leftovers');
+
+		assertJSONEqual(p1leafeon.moveTrack, [['Knock Off', 1]]);
```
Zarel: You could `assert.deepEqual` or you could assert both stringified, but it seems unnecessary to do both.

By the way, I prefer using strict mode to having to type `Strict` every time I assert something, like this:

```
const assert = require('assert').strict;
assert.strictEqual(p1leafeon.moveTrack, [['Knock Off', 1]]);
```

test/battle-test.mocha.js
yuzeh: Ah, I see.

What would you recommend doing for the second block (where the results depend on the events in the first block)?

Zarel: We still don't have a better API than `fastForwardTo` at the moment.

yuzeh:
Tried adding mocha to `npm test` but failed because the JavaScript is missing.

I'll leave adding mocha to the test suite for someone else to do.

src/battle-dex.ts
```diff
@@ -193,12 +193,12 @@ const Tools = {
...
 	resourcePrefix: (() => {
 		let prefix = '';
-		if (document.location!.protocol !== 'http:') prefix = 'https:';
+		if (!window.document || !document.location || document.location.protocol !== 'http:') prefix = 'https:';
 		return prefix + '//play.pokemonshowdown.com/';
 	})(),
 
 	fxPrefix: (() => {
-		if (document.location!.protocol === 'file:') {
+		if (window.document && document.location && document.location.protocol === 'file:') {
```
Zarel: *shakes fist at TC39*

https://github.com/tc39/proposal-optional-chaining/pull/48#issuecomment-377612223

Zarel:
I added it.

PR #1153, opened 2018-10-08
Fix Forme Change and Autotomize Interaction (Client Side)

Cureja:
From issue: https://github.com/Zarel/Pokemon-Showdown/issues/2367#issuecomment-381371924

Removed Autotomize tag from pokemon who undergo formeChange in battle
Server side fix: https://github.com/Zarel/Pokemon-Showdown/pull/4893

Replay download showing it works on the server side: https://drive.google.com/file/d/1LdIgE0evOpPWvAbbGFrFL-GLa51XXWmQ/view?usp=sharing

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -2874,6 +2874,7 @@ class Battle {
 				let isCustomAnim = false;
 				poke.removeVolatile('typeadd' as ID);
 				poke.removeVolatile('typechange' as ID);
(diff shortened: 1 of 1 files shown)


Marty-D:
See comment in server-side fix: https://github.com/Zarel/Pokemon-Showdown/pull/4893#issuecomment-427834796

Issue #1154, opened 2018-10-08
Toxic Orb causes toolbox to show wrong 'next damage' 


Cureja:
Because toxic orb doesn't cause damage on the turn it inflicts toxic, the shown and actual 'next damage'  not the same.

Proof:
![image](https://user-images.githubusercontent.com/24979820/46632127-777d3b80-cafe-11e8-8355-82f899ed26fc.png)
![image](https://user-images.githubusercontent.com/24979820/46632164-8b28a200-cafe-11e8-92ed-124f8532995e.png)
![image](https://user-images.githubusercontent.com/24979820/46632170-97146400-cafe-11e8-8c2e-46679e5f2655.png)
![image](https://user-images.githubusercontent.com/24979820/46632181-a3002600-cafe-11e8-9ea0-22b4b4f0e4b2.png)
http://replay.pokemonshowdown.com/gen7ou-817401448

PR #1155, opened 2018-10-08
Fix Toxic Orb and 'Next Damage' toolbox interaction

Cureja:
Issue: #1154 

Fixed the toxicTurnCount to accurately represent toxic orb poisoning by making toxic orbs first turn's toxicTurnCount = 0

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -868,7 +868,7 @@ class Side {
 				this.battle.message('' + Tools.escapeHTML(pokemon.side.name) + ' withdrew ' + pokemon.getFullName() + '!');
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/battle.ts
```diff
@@ -2472,7 +2473,7 @@ class Battle {
 						else actions += "" + pokeName + " healed its burn!";
 						break;
 					case 'tox':
-						if (poke) poke.statusData.toxicTurns = 0;
+						if (poke) poke.statusData.toxicTurns = null;
```
Zarel: Man is the underlying code here pretty bad though, I'll need to fix it elsewhere.

Zarel:
This is good, though. Thanks!

PR #1161, opened 2018-11-09
Add teambuilder support for Battle Spot Special 13

LegoFigure11:


Zarel:
Is Battle Spot Special 13 supposed to default to level 100 or something?

Zarel:
Oh, I see your comments in Dev, thanks!

PR #1167, opened 2018-11-20
Hide Abilities in Let's Go Tooltips

tmagicturtle:
Not sure if this is the cleanest way of doing it, but other sections of the file have tier.indexOf('Random Battle') so I figure it should be okay.


Zarel:
This all is slated to be rewritten, so this works for now.

I'd normally check the ID, though.

PR #1169, opened 2018-11-20
Add more teambuilder support for Let's Go

KrisXV:
@Zarel

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -9,7 +9,7 @@ const rootDir = path.resolve(__dirname, '..');
 process.chdir(rootDir);
 
 if (!fs.existsSync('data/Pokemon-Showdown')) {
-	child_process.execSync('git clone https://github.com/Zarel/Pokemon-Showdown.git', {
+	child_process.execSync('git clone https://github.com/KrisXV/Pokemon-Showdown.git', {
 		cwd: 'data',
 	});
 }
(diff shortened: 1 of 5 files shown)


TheImmortal:
Change the tiers to Uber, OU, NFE, LC.

TheImmortal:
And if you can edit the mod on the server as well to change the tiers to those then that'll be cool. Basically, it's a new game so everything starts at OU. Mewtwo is the only Uber for now.

Zarel:
So, none of this is your fault, but the code in this file is really bad and needs to be redone at some point.

js/client-teambuilder.js
```diff
@@ -1069,6 +1069,7 @@
 		},
 		renderSet: function (set, i) {
 			var template = Tools.getTemplate(set.species);
+			var isNotLetsGo = !this.curTeam.format.startsWith('gen7letsgo');
```
Zarel: I don't like the double negative here. Just do `isLetsGo`

KrisXV:
done @Zarel

js/search.js
```diff
@@ -846,6 +849,8 @@
 			else if (format === 'doublesou') tierSet = tierSet.slice(slices.DOU);
 			else if (format === 'doublesuu') tierSet = tierSet.slice(slices.DUU);
 			else if (format === 'doublesnu') tierSet = tierSet.slice(slices.DNU);
+			else if (isLetsGo && format.includes('ou')) tierSet = tierSet.slice(slices.OU);
```
Zarel: Why are you doing `includes`? Shouldn't you just strip off the preamble so you're actually just left with `ou`?

js/search.js
```diff
@@ -793,6 +793,7 @@
 		var requirePentagon = (format === 'battlespotsingles' || format === 'battledoubles' || format.slice(0, 3) === 'vgc');
 		var template;
 		var isBH = (format === 'balancedhackmons' || format === 'bh');
+		var isLetsGo = format.startsWith('letsgo');
```
Zarel:
```suggestion
```

js/search.js
```diff
@@ -793,6 +793,7 @@
 		var requirePentagon = (format === 'battlespotsingles' || format === 'battledoubles' || format.slice(0, 3) === 'vgc');
```
Zarel:
```suggestion
		var isLetsGo = format.startsWith('letsgo');
		if (isLetsGo) format = format.slice(6);
		var requirePentagon = (format === 'battlespotsingles' || format === 'battledoubles' || format.slice(0, 3) === 'vgc');
```

js/search.js
Zarel:
```suggestion
```

build-tools/build-indexes
```diff
@@ -301,6 +302,17 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 				if (template.tier === 'CAP' || template.tier === 'CAP NFE' || template.tier === 'CAP LC') {
 					return template.tier;
 				}
+				if (isLetsGo) {
+					let baseTemplate = Tools.mod(gen).getTemplate(template.baseSpecies);
+					if ((baseTemplate.num > 151 || baseTemplate.num < 1) && baseTemplate.num !== 808 && baseTemplate.num !== 809) {
+						if (!['Alola', 'Mega', 'Mega-X', 'Mega-Y', 'Starter'].includes(template.forme)) return 'Illegal';
```
Zarel: err, this line does nothing

Zarel: I'm pretty sure you're making the same mistake you made server-side.

Zarel: https://github.com/Zarel/Pokemon-Showdown/commit/ed54a6eb4db5d5271174cc42537432bb0bf1e602

You understand how this change made the logic a lot clearer, and makes it a lot harder to make mistakes, right?

Zarel: Can you do the same kind of change here?

KrisXV: done

PR #1170, opened 2018-11-20
Fix escaping HTML in localized time

HoeenCoder:
Something to note: battle-log.ts uses both `this.escapeHTML` and `BattleLog.escapeHTML`. I assume which is used depends on where in the code its being called from.

In this case, `this` refered to `window` when parsing HTML from !events, and subsequently crashed.
Let me know if this isn't the correct fix.

### DIFF
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -674,7 +674,7 @@ class BattleLog {
 			// if it exists.
 			formattedTime = parsedTime.toLocaleString();
 		}
-		return '<time>' + this.escapeHTML(formattedTime) + '</time>';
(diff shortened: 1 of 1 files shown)


Zarel:
Note that `this` is correct but unbound.

Think of it this way:

```
class BattleLog {
  static escapeHTML() {}
  static localizeTime() {
    this.escapeHTML();
  }
}

// will work correctly
BattleLog.localizeTime();
// will work correctly
setTimeout(() => BattleLog.localizeTime());
// will fail
setTimeout(BattleLog.localizeTime);
```

This is the usual binding rules.

PR #1173, opened 2018-11-24
Fix log of ability immunity text

urkerab:
I wanted to test this but it turns out that testclient loads `graphics.js` from main, so it doesn't see the changes. Anyway, this just copies other messages such as `-fail`.

### DIFF
--- a/src/battle-text-parser.ts
+++ b/src/battle-text-parser.ts
@@ -826,8 +826,8 @@ class BattleTextParser {
 		}
 
 		case '-immune': {
-			let [, pokemon, effect] = args;
-			const line1 = this.maybeAbility(effect, kwArgs.of || pokemon);
+			const [, pokemon] = args;
+			const line1 = this.maybeAbility(kwArgs.from, kwArgs.of || pokemon);
 			let template = this.template('block', effect);
 			if (!template) {
 				const templateId = kwArgs.ohko ? 'immuneOHKO' : 'immune';


Zarel:
testclient loads `graphics.js` from main because apparently a lot of Windows computers don't have the resources to build it. If you do, you can build it yourself with `./build full`.

src/battle-text-parser.ts
```diff
@@ -826,8 +826,8 @@ class BattleTextParser {
 		}
 
 		case '-immune': {
-			let [, pokemon, effect] = args;
-			const line1 = this.maybeAbility(effect, kwArgs.of || pokemon);
+			const [, pokemon] = args;
```
Zarel:
```suggestion
			const [, pokemon, effect] = args;
```

This is probably also the Travis failure – `effect` is still used to determine the template.

src/battle-text-parser.ts
```diff
@@ -826,8 +826,8 @@ class BattleTextParser {
...
+			const line1 = this.maybeAbility(kwArgs.from, kwArgs.of || pokemon);
 			let template = this.template('block', effect);
```
Zarel: Although maybe this should be `kwArgs.from` too?

I don't have replays so I can't see this.

Zarel:
I should probably move `text.js` outside of `graphics.js` though; that might be a better solution.

Zarel:
Yeah, `battle.js` might be a better place for it.

src/battle-text-parser.ts
urkerab: You're right of course, I notice that for example Own Tempo needs it. Also sorry but I did this very late at night so I didn't stay around to watch for the test failure myself. And thanks for the tip about `build full`, it only takes half a minute on my machine so that won't be a problem.

Zarel: You don't need to do it anymore (I moved text resource files out of `graphics.js`), and should probably actually delete `data/graphics.js` from your computer to speed up build times.

PR #1176, opened 2018-11-25
Fix Dry Skin healing

urkerab:
This was reported on the [thread](https://www.smogon.com/forums/posts/7965951/) but I think this is the fix. Basically if an Ability is involved then it's that of the Pokémon being healed. The only difference with Dry Skin/Volt/Water Absorb is that there can be an attacker involved. This also avoids having to special case Volt/Water Absorb in the text parser.


Zarel:
This pull request **introduces 1 alert** when merging 10828550770a913a957fe5a35e29ae7fac7a83f1 into 3540d9d1f4b53f10bcc9a3c603614e39b1cfb023 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown-Client/rev/pr-8b41f3a66cee72d475723fc8cb70a78d54d056d2)

**new alerts:**

---

urkerab:
Indeed, we don't care about the source of healing.

urkerab:
The code that reports the foe was (according to the blame I found) originally meant for Leech Seed, presumably to show whose health is being drained, and the Abilities didn't send a foe, because they bypassed healing and Heal Block in particular. The fix for that used the draining code path, which is great from the point of view of Heal Block, but it makes the foe get sent with the absorb effect.

Zarel:
This is also only half the fix, apparently.

urkerab:
Sorry, I'm a bit behind with my notifications as usual... what's missing?

Zarel:
Wow, this was a month ago. I don't remember anymore; I mainly remember that the old-protocol compatibility code in the client needs to be refactored.

PR #1178, opened 2018-11-25
Fix Gen 1 partial trapping message

urkerab:
In Gen 1, partial trapping prevents you from moving. The old code used to check whether a move had been provided and if so display "can't move" instead of "can't use " + move. However I guess the right way to do this is to add a "partiallytrapped" effect template instead.


Zarel:
I would code this as `cantNoMove`.

Zarel:
(Refer to how `mega` vs `megaNoItem` is implemented.)

urkerab:
Speaking of which, it looks as if `megaNoItem` isn't working - it should give the "fervent wish" style message for Rayquaza. I don't have time to look at that tonight though.

PR #1180, opened 2018-11-26
Correct rounding error in Lets Go stat calculation

Ezaphs:
the stat value must be floored before being passed into the friendship boost calculations to avoid rounding inaccuracies.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -3362,13 +3362,13 @@
 			if (natureOverride) {
 				val *= natureOverride;
 			} else if (BattleNatures[set.nature] && BattleNatures[set.nature].plus === stat) {
-				val = Math.floor(val * 1.1);
+				val *= 1.1;
 			} else if (BattleNatures[set.nature] && BattleNatures[set.nature].minus === stat) {
-				val *= Math.floor(val * 0.9);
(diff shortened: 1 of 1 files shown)


KrisXV:
It'd be simpler  to just take the code from line 3371 (`val = Math.floor(val * friendshipValue / 100) + (supportsAVs ? ev : 0);` and then change it to `val = Math.floor(Math.floor(val) * friendshipValue / 100) + (supportsAVs ? ev : 0);`

js/client-teambuilder.js
```diff
@@ -3368,7 +3368,7 @@
 			}
 			if (!supportsEVs) {
 				var friendshipValue = Math.floor(((!set.happiness ? 255 : set.happiness > 255 ? 255 : set.happiness) / 255 / 10 + 1) * 100);
-				val = Math.floor(val * friendshipValue / 100) + (supportsAVs ? ev : 0);
+				val = Math.floor(Math.floor(val) * friendshipValue / 100) + (supportsAVs ? ev : 0);
```
```suggestion
				val = Math.floor(val) * friendshipValue / 100) + (supportsAVs ? ev : 0;
```

The inner floor is necessary because the outer floor is unnecessary.

Zarel:
Cool, thanks!

PR #1181, opened 2018-11-28
Fix missing trace activation

urkerab:
If `kwArgs.from` is an Ability then this means that it's one of Trace, Receiver or Power of Alchemy and it's just replaced itself with another Ability.

Spotted when Gardevoir switches in at the end of [Turn 4](https://replay.pokemonshowdown.com/gen7doublesou-827762114).

### DIFF
--- a/src/battle-text-parser.ts
+++ b/src/battle-text-parser.ts
@@ -393,13 +393,9 @@ class BattleTextParser {
 		}
 
 		case '-ability': {
-			let [, pokemon, ability, oldAbility, arg4] = args;
+			let [, pokemon, ability, foe] = args;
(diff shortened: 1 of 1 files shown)


Zarel:
This is wrong, this code is here to support a new protocol which puts the old ability in arg3. Can you retain that functionality here?

Zarel:
I'd do the revert in suggestions, but you actually need to revert basically everything here.

Zarel:
Come to think of it, it's easier for me to fix myself.

urkerab:
Sorry, I didn't realise that this was supposed to be a new protocol.

PR #1183, opened 2018-11-30
Fix messages for some effects that prevent moves from executing

urkerab:
Followup to #1178.

It turns out that `partiallytrapped` is not the only effect that doesn't send the move that can't be executed; `attract`, `flinch`, `frz`, `par`, `recharge`, `slp`, `throatchop` and `truant` don't either. I've left the `cant` versions in in case they get changed to send the move later, but they can readily be removed if not needed.

FYI the other `cant` effects that do provide a move: `damp`,  `dazzling`,  `disable`,  `focuspunch`,  `gravity`,  `healblock`,  `imprison`,  `nopp`,  `queenlymajesty`,  `shelltrap`,  `taunt`


Zarel:
I would just have `cantNoMove` fall back to `cant` if it doesn't exist.

PR #1186, opened 2018-11-30
No target uses generic fail message in Gen 5 or later

urkerab:
This quirk got lost in the text parser rewrite.


Zarel:
No, the server should be sending `-fail` in this case.

urkerab:
Maybe, but that doesn't fix existing servers and replays.

Zarel:
Generic/non-generic failure messages aren't so big of an issue that I care about existing servers and replays.

PR #1188, opened 2018-12-02
Restore Mind Blown recoil message

urkerab:
The old code actually used "because of [effect.name]" but I'm not sure how that would be implemented.

### DIFF
--- a/data/text.js
+++ b/data/text.js
@@ -943,6 +943,9 @@ exports.BattleText = {
 	magicbounce: {
 		move: '#magiccoat',
 	},
+	mindblown: {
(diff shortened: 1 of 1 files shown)


Marty-D:
Mind Blown doesn't actually have a damage message, so if we want to keep this it should probably be in parentheses. I assume the old one was some default since it's not a game message for any effect.

Zarel:
What Marty said.

urkerab:
Since you lose the HP even if the move misses, hits a substitute or protection move, or is "absorbed" by Flash Fire, I avoided using the word "damage" in the message.

> Mind Blown doesn't actually have a damage message

Bah, I can't seem to get anything right these days...

PR #1190, opened 2018-12-04
Fix teambuilder search by tier

spixi:
When you search for ```caplc``` or ```capnfe```, the teambuilder does not show any result. This is because of a mismatch of the tier tables in 

https://github.com/Zarel/Pokemon-Showdown-Client/blob/1ef228ebcbf64c8df8f011508d9bc21368abc752/js/search.js#L1242

and

https://github.com/Zarel/Pokemon-Showdown/blob/a5a2586c7bfbb3134843d1011272cdb8ec28abff/chat-plugins/datasearch.js#L274

This commit synchronizes both tables that the search works again.


Zarel:
Thanks!

PR #1192, opened 2018-12-05
Fix some messages for abilities blocking effects

urkerab:
I saw this with Flower Veil myself; the Ability activated on the correct Pokémon but the protected Pokémon was not correctly identified, instead the message just said "[TARGET]". I assume the other two Abilities work in the same way.


Zarel:
Don't worry about the build error, I have it fixed locally.

PR #1194, opened 2018-12-13
Improve Damp/Dazzling messages (client)

mathfreak231:
See https://github.com/Zarel/Pokemon-Showdown/pull/5027

### DIFF
--- a/src/battle-text-parser.ts
+++ b/src/battle-text-parser.ts
@@ -320,7 +320,6 @@ class BattleTextParser {
 
 		case 'cant': {
 			let [, pokemon, effect, move] = args;
-			let id = this.effectId(effect);
 			const template = this.template('cant', effect, 'NODEFAULT') ||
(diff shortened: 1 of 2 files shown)


Zarel:
This is where the backwards compatibility issue should be fixed.

Zarel:
The main thing I'd want, is just to display the `block` message when receiving the `cant` message, here.

src/battle-text-parser.ts
```diff
@@ -320,13 +320,6 @@ class BattleTextParser {
 
 		case 'cant': {
 			let [, pokemon, effect, move] = args;
-			let id = this.effectId(effect);
-			switch (id) {
-			case 'damp': case 'dazzling': case 'queenlymajesty':
-				// thanks Marty
-				[pokemon, kwArgs.of] = [kwArgs.of, pokemon];
-				break;
-			}
```
Zarel: This should probably be kept and hardcoded somehow...

PR #1195, opened 2018-12-16
-anim should only animate

urkerab:
As part of commit fc603c0 the `useMove` code so that used to actually log the message got refactored out. As such, the `kwArgs.silent` guard was also removed. However, the `-anim` code was relying on the side effect of setting `kwArgs.silent` to suppress the other bookkeeping that `useMove` does, such as deducting PP. As such, `-anim` now causes a deduction of PP. This is particularly noticeable in Super Staff Bros Brawl but also anywhere `-anim` is used such as for Solar Beam in Sun.

To simplify the code, I've split the bookkeeping away from the animation so that `useMove` now calls both (and also the message logging) while `-anim` now just invokes the animation.

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -1193,12 +1193,6 @@ class Battle {
 			pokemon.removeTurnstatus('focuspunch' as ID);
 		}
 		this.scene.updateStatbar(pokemon);
-		if (!target) {
-			target = pokemon.side.foe.active[0];
-		}
-		if (!target) {
-			target = pokemon.side.foe.missedPokemon;
-		}
 		if (fromeffect.id === 'sleeptalk') {
 			pokemon.rememberMove(move.name, 0);
 		} else if (!fromeffect.id || fromeffect.id === 'pursuit') {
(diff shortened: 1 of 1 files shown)


src/battle.ts
```diff
@@ -1219,6 +1213,19 @@ class Battle {
 			let pp = (target && target.side !== pokemon.side && toId(target.ability) === 'pressure' ? 2 : 1);
 			pokemon.rememberMove(moveName, pp);
 		}
+		pokemon.lastMove = move.id;
+		this.lastMove = move.id;
+		if (move.id === 'wish' || move.id === 'healingwish') {
+			pokemon.side.wisher = pokemon;
+		}
+        }
```
Zarel: >:(

urkerab: Oops, I'd changed my editor settings to work on another project and forgot to check them. Sorry about that...

Zarel:
Looks good, although I would early-return in `animateMove` now that it's possible.

Zarel:
I can implement that myself, though.

PR #1198, opened 2018-12-23
Fix status data not resetting after getting phazed

asgdf:


src/battle.ts
```diff
@@ -745,6 +745,8 @@ class Side {
 		this.lastPokemon = oldpokemon;
 		if (oldpokemon) {
 			this.battle.scene.animDragOut(oldpokemon);
+			oldpokemon.statusData.toxicTurns = 0;
+			if (this.battle.gen === 5) oldpokemon.statusData.sleepTurns = 0;
```
Zarel: Should this stuff all go inside `clearVolatile`?

Zarel:
Thanks!

PR #1199, opened 2018-12-27
Magic Bounce should activate visibly

urkerab:
Reported [here](https://www.smogon.com/forums/posts/7995630). This would have been done using `maybeAbility` but unfortunately the server doesn't send the type of the effect, and similarly `activateAbility` can't be passed the effect object either.

### DIFF
--- a/src/battle-text-parser.ts
+++ b/src/battle-text-parser.ts
@@ -310,9 +310,12 @@ class BattleTextParser {
 
 		case 'move': {
 			const [, pokemon, move] = args;
+			let id = this.effectId(kwArgs.from);
 			let line1 = '';
 			if (kwArgs.zEffect) {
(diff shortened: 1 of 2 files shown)


Zarel:
This is good. But you might want to start on the `lineParse`-based filtering, instead?

Like, set `kwArgs.from` to `"ability: Magic Bounce"` as relevant?

urkerab:
Sorry, I'm not sure what you mean; do you have an example?

Zarel:
I guess I should implement this myself.

Zarel:
Can you paste a replay for ease of testing on my part?

Zarel:
@urkerab Your future fixes should probably use `updateArgs` like in my commit.

PR #1202, opened 2018-12-31
Display actual target for Veil abilities

urkerab:
These Abilities can protect allies too, so their messages need to reflect that.

### DIFF
--- a/data/text.js
+++ b/data/text.js
@@ -932,7 +932,7 @@ exports.BattleText = {
 		start: "  [POKEMON] is drowsing!",
 	},
 	damp: {
-		block: "  [SOURCE] cannot use [MOVE]!",
+		block: "  [POKEMON] cannot use [MOVE]!",
 	},
 	darkaura: {
 		start: "  [POKEMON] is radiating a dark aura!",
--- a/src/battle-text-parser.ts
+++ b/src/battle-text-parser.ts
@@ -73,7 +73,7 @@ class BattleTextParser {
 				'electricterrain', 'mistyterrain', 'psychicterrain', 'telepathy', 'stickyhold', 'suctioncups', 'aromaveil',
 				'flowerveil', 'sweetveil', 'disguise', 'safetygoggles', 'protectivepads',
 			].includes(id)) {
-				return {args: ['-block', target || pokemon, effect, arg3], kwArgs: {}};
+				return {args: ['-block', target || pokemon, effect, arg3], kwArgs: {of: pokemon || target}};
 			}
 
 			if ([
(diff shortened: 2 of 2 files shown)


Zarel:
...we may need to re-spec how the `block` message works.

Zarel:
My original intent was that `[POKEMON]` would be the target.

I would make `[SOURCE]` the thing doing the blocking, but it makes more sense for `[SOURCE]` to be the thing doing the attack that was blocked.

I guess your interpretation might be sensible:

`[POKEMON]` - thing doing the blocking
`[TARGET]` - intended target of the blocked attack
`[SOURCE]` - thing trying to make the blocked attack

Zarel:
Under this interpretation, though, Psychic Terrain should probably use `[TARGET]`, then?

urkerab:
I see what you mean about the inconsistency.

For the `block:` messages used by `-activate` then the only time that `pokemon` and `target` can be different are for the Veil Abilities, so the relevant `block:` messages can readily be changed from `[POKEMON]` to `[TARGET]`.

However there are a couple of other `block:` messages.
- Primal weather blocking other weather is a `-fail` for a move and a `-ability` for an ability, although they don't actually care about the target (except for correct ability activation of course).
- Uproar uses `-fail` which wants arg1 aka `[POKEMON]` to be the source that's getting blocked. Fortunately the `block:` cases happen first, so I think they can be changed too.
- Own Tempo uses `-immune` which assumes arg1 aka `[POKEMON]` is actually the intended target. Changing this would require changing `immune:` too.

I've pushed separate commits to demonstrate so that you can see what I mean.

Zarel:
Do you have any replays? I want to know if my latest change fixed this.

urkerab:
I finally got around to updating this to work with the "latest" change mentioned above.

I assume that you want the `-block` message arguments to be the Pokémon being blocked and the effect doing the blocking, with `kwArgs.of` being the Ability holder if different from the above. Damp/Dazzling/Queenly Majesty are like the Veil abilities in that they have an ability holder which differs from the Pokémon being blocked. For consistency, I changed them to use `[POKEMON]` as the placeholder to match the other `-block` messages. No block messages mention the ability holder, although I didn't remove the `[SOURCE]` placeholder from the code.

The other case where a Veil ability takes affect on another Pokémon is where Flower Veil blocks a stat drop. I've rearranged the `-fail` code that tries to handle that case so that the correct messages appear.

I created a reply for testing those specific messages: https://replay.pokemonshowdown.com/gen7doublesuu-868637836

Zarel:
This pull request **fixes 1 alert** when merging bc3e57b063df55e8b305597e5ce583f166f6b392 into 5a4d4da3a85a875e7d8ccdd2079f7a3edfe57d35 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown-Client/rev/pr-ebcfecda0b93313a97bbbd305812672f2cddef9b)

**fixed alerts:**

---

Zarel:
Fixed by e47d261b46ccb0faefed36

Zarel:
I did end up re-speccing `-block`. In kind of a rush, but hey, it's the morning after and I don't mind the new spec.

PR #1206, opened 2019-01-06
Re-implement unlinking in battle chat

asgdf:
Warning and muting unlink without hidetexting, so this makes the behavior consistent between chat and battle rooms.

### DIFF
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -126,7 +126,11 @@ class BattleLog {
 			return;
 
 		case 'unlink':
-			this.hideChatFrom(toId(args[2] || args[1]));
+			const user = toId(args[2]) || toId(args[1]);
+			this.unlinkChatFrom(user);
+			if (args[2]) {
+				this.hideChatFrom(user);
+			}
 			return;
 
 		case 'debug':
@@ -291,6 +295,33 @@ class BattleLog {
 		lastNode.appendChild(document.createTextNode(' '));
 		lastNode.appendChild(button);
 	}
+	unlinkChatFrom(userid: ID) {
+		const classStart = 'chat chatmessage-' + userid + ' ';
+		const innerNodeList = this.innerElem.childNodes;
+		const preemptNodeList = this.preemptElem.childNodes;
+
+		const unlinkNodeList = (nodeList: ArrayLike<HTMLElement>) => {
+			for (const node of nodeList as HTMLElement[]) {
+				if (node.className && (node.className + ' ').startsWith(classStart)) {
+					const linkList = node.getElementsByTagName('a');
+					// iterate in reverse because linkList will update as links are removed
+					for (let i = linkList.length - 1; i >= 0; i--) {
+						const linkNode = linkList[i];
+						const parent = linkNode.parentElement;
+						if (!parent) continue;
+						for (const childNode of linkNode.childNodes as any) {
+							parent.insertBefore(childNode, linkNode);
+						}
+						parent.removeChild(linkNode);
+					}
+				}
+			}
+		};
+
+		unlinkNodeList(innerNodeList as NodeListOf<HTMLElement>);
+		unlinkNodeList(preemptNodeList as NodeListOf<HTMLElement>);
+	}
+
 	preemptCatchup() {
 		if (!this.preemptElem.firstChild) return;
 		this.innerElem.appendChild(this.preemptElem.firstChild);


src/battle-log.ts
```diff
@@ -126,7 +126,13 @@ class BattleLog {
 			return;
 
 		case 'unlink':
-			this.hideChatFrom(toId(args[2] || args[1]));
+			const user = toId(args[2]) || toId(args[1]);
+			const $messages = $('.chatmessage-' + user);
```
Zarel: jQuery is not a supported dependency in battle logs.

Zarel: Look at the header comment:

https://github.com/Zarel/Pokemon-Showdown-Client/blob/bf09524187f2f7d997ed0ef6f3d5dd7e604c7349/src/battle-log.ts#L4-L10

Zarel: I'd suggest manually iterating child nodes.

Zarel: Like, look at the source code for `hideChatFrom` and adapt that.

Zarel: Also, you can't use `getElementsByClassName` because that's not supported by all supported browsers.

asgdf:
Okay, I *think* I did this in a IE7-compliant way now.

asgdf:
Ok I'm getting some TS issues here, let me see if I can figure these out

Zarel:
`document.getElementsByTagName` returns an HTMLCollection in IE, Edge, and Firefox, and a NodeList in Chrome and Safari.

`element.getElementsByTagName` returns an HTMLCollection in most modern browsers, and a NodeList in some older versions of Chrome, Safari, and Firefox.

Treat it as an array-like object (only use `.length`, `[index]`, and `for...of`), don't try to cache it anywhere, and you'll be fine.

Zarel:
Would it be easier to replace the link with a span than to unwrap it?

src/battle-log.ts
```diff
@@ -126,7 +126,11 @@ class BattleLog {
 			return;
 
 		case 'unlink':
-			this.hideChatFrom(toId(args[2] || args[1]));
+			const user = toId(args[2]) || toId(args[1]);
+			this.unlinkChatFrom(user);
```
Zarel: This shouldn't be done if `args[2]` exists.

src/battle-log.ts
```diff
@@ -291,6 +295,37 @@ class BattleLog {
 		lastNode.appendChild(document.createTextNode(' '));
 		lastNode.appendChild(button);
 	}
+	unlinkChatFrom(userid: ID) {
+		const classStart = 'chat chatmessage-' + userid + ' ';
+		const innerNodeList = this.innerElem.getElementsByTagName('a');
+		const preemptNodeList = this.preemptElem.getElementsByTagName('a');
+
+		const unlinkNodeList = (nodeList: NodeList) => {
```
```suggestion
		const unlinkNodeList = (nodeList: NodeList | HTMLCollection) => {
```

In theory, this should be all that's needed here.

Zarel:
```suggestion
		const unlinkNodeList = (nodeList: ArrayLike<HTMLAnchorElement>) => {
```

Actually, this might be better. Feel free to just use `ArrayLike` if VS Code doesn't like `<HTMLAnchorElement>`.

src/battle-log.ts
```diff
@@ -291,6 +295,37 @@ class BattleLog {
 		lastNode.appendChild(document.createTextNode(' '));
 		lastNode.appendChild(button);
 	}
+	unlinkChatFrom(userid: ID) {
+		const classStart = 'chat chatmessage-' + userid + ' ';
+		const innerNodeList = this.innerElem.getElementsByTagName('a');
+		const preemptNodeList = this.preemptElem.getElementsByTagName('a');
```
Zarel: Also: can you benchmark this, vs iterating child elements matching `classStart`, and using `getElementsByTagName` on each matching element?

The latter is much more readable to me, so unless the former is noticeably faster, I'd prefer that method.

src/battle-log.ts
asgdf: Since the whole thing is |unlink|hide|, I assumed it wasn't a bad idea to actually unlink _and_ hide, since I see no reason to leave links intact after they've been hidden like that (in case someone else unhides the messages)

src/battle-log.ts
asgdf: I've switched it around

asgdf:
Okay I think this works now.

src/battle-log.ts
```diff
@@ -291,6 +295,33 @@ class BattleLog {
 		lastNode.appendChild(document.createTextNode(' '));
 		lastNode.appendChild(button);
 	}
+	unlinkChatFrom(userid: ID) {
+		const classStart = 'chat chatmessage-' + userid + ' ';
+		const innerNodeList = this.innerElem.childNodes;
+		const preemptNodeList = this.preemptElem.childNodes;
+
+		const unlinkNodeList = (nodeList: ArrayLike<Node>) => {
+			for (const node of nodeList as any) {
```
```suggestion
			for (const node of nodeList as Node[]) {
```

Zarel:
I like the rest, so it's just that one last change.

asgdf:
Okay, no more ``any``, no more mixing of ``Node`` and ``HTMLElement``, no errors thrown

PR #1208, opened 2019-01-08
Add a getTier function for the teambuilder

KrisXV:
@Zarel 

I combined the getTier and getDoublesTier functions into one because I felt that it should be easy to just use the one function interchangeably.

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -295,6 +295,7 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 		const pokemon = Object.keys(Dex.data.Pokedex);
 		pokemon.sort();
 		const tierTable = {};
+		const tierStorage = {};
 		for (const id of pokemon) {
 			const template = Dex.mod(gen).getTemplate(id);
 			if (template.gen > genNum) continue;
@@ -326,6 +327,7 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 				}
 				return template.tier;
 			})();
+			tierStorage[template.speciesid] = tier;
 			if (template.forme) {
 				if (template.baseSpecies in {Aegislash:1, Castform:1, Cherrim:1, Darmanitan:1, Meloetta:1, Mimikyu:1, Minior:1, Wishiwashi:1} || template.forme.includes('Totem')) {
 					continue;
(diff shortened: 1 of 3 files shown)


Zarel:
Do you think you could do this without a new teambuilder table?

If you really need teambuilder tables, it should be the same format as the others.

Zarel:
In hindsight, a new one probably isn't a bad idea.

build-tools/build-indexes
```diff
@@ -362,13 +364,16 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 		} else if (isDoubles) {
 			BattleTeambuilderTable[gen + 'doubles'] = {};
 			BattleTeambuilderTable[gen + 'doubles'].tiers = tiers;
+			BattleTeambuilderTable[gen + 'doubles'].overrideTier = overrideTier;
 			BattleTeambuilderTable[gen + 'doubles'].formatSlices = formatSlices;
 		} else if (gen === 'gen7') {
 			BattleTeambuilderTable.tiers = tiers;
 			BattleTeambuilderTable.items = items;
+			BattleTeambuilderTable.overrideTier = overrideTier;
```
Slayer95: @Zarel , We don't need this one, right?

Zarel: Technically no.

js/search.js
```diff
@@ -1307,14 +1309,12 @@
...
-			tier = pokemon.tier;
+		if (pokemon.tier) {
+			tier = Dex.getTier(pokemon, this.gen, this.isDoubles);
 		} else if (pokemon.forme && pokemon.forme.endsWith('Totem')) {
-			tier = Dex.getTemplate(pokemon.species.slice(0, (pokemon.forme.startsWith('Alola') ? -6 : pokemon.baseSpecies.length + 1))).tier;
+			tier = Dex.getTier(BattlePokedex[toId(pokemon.species.slice(0, (pokemon.forme.startsWith('Alola') ? -6 : pokemon.baseSpecies.length + 1)))], this.gen, this.isDoubles);
 		} else {
-			tier = Dex.getTemplate(pokemon.baseSpecies).tier;
+			tier = Dex.getTier(BattlePokedex[toId(pokemon.baseSpecies)], this.gen, this.isDoubles);
 		}
```
Zarel: I seem to have missed this, but the point of `getTier` was to be able to replace these 7 lines with a single call to `getTier`, which seems to not be what's going on here.

PR #1209, opened 2019-01-11
Introduce serverinfo action

Kaiepi:


Zarel:
This is being deployed, without the URL parser, to:

https://pokemonshowdown.com/servers/smogtours.json

https://pokemonshowdown.com/servers/textnonexistentserver.json

The URL parser seems interesting, but I don't think I want to provide an HTTP service for it.

PR #1210, opened 2019-01-12
Fix crash damage message after server PR 5042

urkerab:
Crash damage's effect name was changed from `highjumpkick` and `jumpkick` to `crash`.

... unless you think it should be changed back on the server?


Zarel:
......ugh

PR #1211, opened 2019-01-13
Teambuilder: Properly label healing berry usefulness in gen 7

asgdf:
As of gen 7, Aguav and friends are actually useful.

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -458,8 +458,16 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 				break;
 			case 'lumberry':
 			case 'sitrusberry':
-				if (genNum === 2) goodItems.push(id);
+				if (genNum === 2 || genNum >= 7) goodItems.push(id);
 				else greatItems.push(id);
+				break;
+			case 'aguavberry':
+			case 'figyberry':
+			case 'iapapaberry':
+			case 'magoberry':
+			case 'wikiberry':
+				if (genNum >= 7) greatItems.push(id);
+				else poorItems.push(id);
 				break;
 			case 'berryjuice':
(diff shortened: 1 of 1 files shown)


xfix:
Sure.

build-tools/build-indexes
```diff
@@ -461,6 +461,14 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 				if (genNum === 2) goodItems.push(id);
 				else greatItems.push(id);
 				break;
+			case 'aguavberry':
+			case 'figyberry':
+			case 'iapapaberry':
+			case 'magoberry':
+			case 'wikiberry':
+				if (genNum >= 7) goodItems.push(id);
```
xfix: Why is it merely a good item? I think it would be better to make it a great item, and move Sitrus Berry to outclassed items in Gen7 only.

Zarel: Sitrus Berry activates at 50% and the others at 25%, right? So it would still have uses.

DaWoblefet: Iirc teambuilder usefulness is not based on Doubles/VGC play, but the super Sitrus (Aguav, etc.) Berries are pivotal parts of both tiers, with regular Sitrus Berry having more niche use, but still being a viable choice.

asgdf: I suppose I just didn't want to clutter the Great Items list with 5 near-identical berries. So do we put Aguav etc at Great, Sitrus at good?

xfix: Oh right, forgot the difference of activation level.

asgdf:
Okay now Sitrus is in "good items" in gen7 while Aguav et al are in "great items"

PR #1216, opened 2019-01-17
Add more doubles tier support (client side)

KrisXV:
This is necessary for Mega Sableye (Gen 6 and Gen 7) and Mega Metagross (Gen 6)


Zarel:
This could use a cross-link to https://github.com/Zarel/Pokemon-Showdown/pull/5112

PR #1219, opened 2019-01-22
Teambuilder: Make Trick Room first result for 'TR'

asgdf:
"TR" is an alias of Trick Room, but it was being skipped because Trick Room itself starts with these letters and it wasn't included in the exceptions.
The second condition is apparently needed to prevent duplicate results, but move results don't get duplicated and that's why we can do this? I'm not sure I understand how it works exactly but this seems to work.


Zarel:
This code still confuses me, but whatever, I guess I'll figure it out later.

PR #1223, opened 2019-02-01
Fix PP in tooltips (Trump Card, Pressure)

mathfreak231:
- Trump Card now shows as having 5 PP in tooltips as it does on the simulator
- Pressure won't cause tooltips to show -1 PP anymore


Zarel:
The negative PP is useful in obscure situations, though.

PR #1224, opened 2019-02-03
Make named avatars persistent

HoeenCoder:


js/client-chat.js
```diff
@@ -881,13 +881,8 @@
...
 			case 'avatar':
 				var parts = target.split(',');
-				var avatarString = toId(parts[0]);
-				var avatar = parseInt(avatarString, 10);
-				if (avatar) {
-					Dex.prefs('avatar', avatar);
-				} else if (['bw2elesa', 'teamrocket', 'yellow', 'zinnia', 'clemont'].indexOf(avatarString) > -1) { // custom avatar exceptions
-					Dex.prefs('avatar', '#' + avatarString);
-				}
+				var avatar = toId(parts[0]);
```
Zarel: Close, but avatars can have `-` in them.

Zarel: The server is currently using a hardcoded regex, you can just copy/paste it.

HoeenCoder: Whoops, slipped my mind when doing this.

js/client-chat.js
```diff
@@ -881,13 +881,8 @@
 
 			case 'avatar':
 				var parts = target.split(',');
-				var avatarString = toId(parts[0]);
-				var avatar = parseInt(avatarString, 10);
-				if (avatar) {
-					Dex.prefs('avatar', avatar);
-				} else if (['bw2elesa', 'teamrocket', 'yellow', 'zinnia', 'clemont'].indexOf(avatarString) > -1) { // custom avatar exceptions
-					Dex.prefs('avatar', '#' + avatarString);
```
Slayer95: What about #avatars?

HoeenCoder: No longer "custom" (#), they are now where all the other avatars are. eg: https://play.pokemonshowdown.com/sprites/trainers-byname/yellow.png

True custom avatars (#hoeenhero) aren't set through this command anyways.

Zarel: They probably will be, in the future. But it's not really a problem for now.

PR #1225, opened 2019-02-07
Fix PP counting for Trick Room/Pressure interaction

xfix:
Not using `Array#some`, as I believe this part of the code has to work in old web browsers.

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -1220,18 +1220,16 @@ class Battle {
 					}
 				}
 			}
-			let isAffectedByPressure;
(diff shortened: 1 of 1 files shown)


xfix:
This is different to #1223 by the way, it's just a coincidence there were two Pressure-related changes recently.

Zarel:
You can just polyfill `Array#some`, btw.

Zarel:
Also isn't Pressure +1 per Pressure user, not just +1 overall?

xfix:
Oh right.

PR #1226, opened 2019-02-15
Change (PU) to Untiered

KrisXV:
This would be so much easier if we could just allow ZU teambuilder support directly via ZUBL and ZU

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -296,6 +296,7 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 		pokemon.sort();
 		const tierTable = {};
 		const overrideTier = {};
+		const zuBans = {};
 		for (const id of pokemon) {
 			const template = Dex.mod(gen).getTemplate(id);
 			if (template.gen > genNum) continue;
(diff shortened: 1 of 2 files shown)


bumbadadabum:
Why are we changing the name of a tier to something that gets cut off in the teambuilder? It looks really bad

KrisXV:
I was asked to make it Untiered, I’m fine with any name. I was just doing what I was asked to do

bumbadadabum:
In that case I'm asking the person who asked you why he chose this name.

KrisXV:
@theimmortal

TheImmortal:
This what Smogon uses to define Pokemon that do not make it into the lowest usage-based tier. (OU) (PU) and such should be reserved for Megas imo since those are accurate. But classifying Untiered Pokemon as "PU by technicality" is wrong and doesn't match Smogon.

One solution could be abbreviating it to "U", much like AG or any other tier.

bumbadadabum:
Yeah UT or something or "no tier"

KrisXV:
If everyone is fine with UT, I'll update it in a little bit

Zarel:
I don't like "UT". "U" is a bit more acceptable, except that could be confused with Ubers. "None"?

TheImmortal:
Or you could make it display a dash. That seems nicer.

Zarel:
I'm fine with the dash.

PR #1228, opened 2019-02-17
Add https://randbatscalc.github.io/ to URL whitelist

uzelbot:
Smoochyena: /rcalc brings up a link to the randbats damage calc
Smoochyena: but when you click on the link it puts you through the interstice
Smoochyena: any way that could be removed to go directly to the calc?


Zarel:
That's just the example config. I think it's already out of sync with the real config file. I'll add it myself.

PR #1231, opened 2019-02-19
Tooltips: Fix Rivalry and Power Trip/Stored Power

asgdf:
Rivalry had the boosts switched, PT/SP used the target's boosts instead of the source's.


MacChaeger:
I looked right at the rivalry part and missed that the boosts were backwards because I was so enamored by how creative the gender check was.

PR #1232, opened 2019-02-19
Tooltips: Only show 2 digits max after decimal point

asgdf:
Sucker Punch under Dark Aura has the not-very-beautiful base power of 93.10000000000001. With this, it only shows 2 digits past the decimal point. Casting back to Number gets rid of trailing zeroes from `toFixed()`.

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -116,13 +116,15 @@ class ModifiableValue {
 	}
 	toString() {
 		let valueString;
+		const roundedValue = this.value ? Number(this.value.toFixed(2)) : 0;
+		const roundedMaxValue = this.maxValue ? Number(this.maxValue.toFixed(2)) : 0;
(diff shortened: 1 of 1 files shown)


Zarel:
`Math.round` is actually the relevant function here.

Zarel:
I'll take this, though.

Zarel:
(And fix it myself later)

asgdf:
I didn't want to throw away everything past the decimal point entirely. I mean I guess you could multiply and divide by 100 before and after rounding, but that would be even uglier imo.

Zarel:
Oh, yeah, I'd do the multiply/divide. I guess you're right that they're both about the same ugliness.

Zarel:
Conceptually I don't like the idea of a number -> string -> number -> string conversion, but I guess it's not any messier than a multiply/round/divide in practice.

PR #1234, opened 2019-02-20
Open replay links in PS

MacChaeger:


js/client.js
```diff
@@ -1228,12 +1228,15 @@
...
 				if (this.className.indexOf('minilogo') >= 0) return; // handled elsewhere
 				if (!this.href) return; // should never happen
-				if (this.host === 'play.pokemonshowdown.com' || this.host === 'psim.us' || this.host === location.host) {
+				if (this.host === 'play.pokemonshowdown.com' || this.host === 'replay.pokemonshowdown.com' || this.host === 'psim.us' || this.host === location.host) {
 					if (!e.cmdKey && !e.metaKey && !e.ctrlKey) {
 						var target = this.pathname.substr(1);
 						var shortLinks = /^(rooms?suggestions?|suggestions?|adminrequests?|bugs?|bugreports?|rules?|faq|credits?|news|privacy|contact|dex|insecure)$/;
 						if (target === 'appeal' || target === 'appeals') target = 'view-help-request--appeal';
 						if (target === 'report') target = 'view-help-request--report';
+						if (this.host === 'replay.pokemonshowdown.com' && target !== 'search') {
```
```suggestion
						if (this.host === 'replay.pokemonshowdown.com' && target && target !== 'search') {
```

It's funny that you thought of `/search` but forgot `/` which is probably way more common.

Zarel: You're not actually handling it correctly, anyway; you should probably set `target` to `.` so it doesn't try anything weird.

MacChaeger:
(1234 hype)

js/client.js
```diff
@@ -1228,12 +1228,16 @@
...
 				if (!this.href) return; // should never happen
-				if (this.host === 'play.pokemonshowdown.com' || this.host === 'psim.us' || this.host === location.host) {
+				if (this.host === 'play.pokemonshowdown.com' || this.host === 'replay.pokemonshowdown.com' || this.host === 'psim.us' || this.host === location.host) {
 					if (!e.cmdKey && !e.metaKey && !e.ctrlKey) {
 						var target = this.pathname.substr(1);
 						var shortLinks = /^(rooms?suggestions?|suggestions?|adminrequests?|bugs?|bugreports?|rules?|faq|credits?|news|privacy|contact|dex|insecure)$/;
 						if (target === 'appeal' || target === 'appeals') target = 'view-help-request--appeal';
 						if (target === 'report') target = 'view-help-request--report';
+						if (this.host === 'replay.pokemonshowdown.com') {
+							if (!target || target === 'search') return;
```
Zarel: This breaks the desktop client! Do `target = '.';` instead (and put the next part in an `else`)

MacChaeger: I had forgotten that even existed. Why does this break it?

MacChaeger:
If this fails because of white space, I cannot do anything to fix it on mobile.

MacChaeger:
COol it only failed because of that other commit

MacChaeger:
Oh, this one actually didn't get broken by the force-push, so it should be good to go.

PR #1236, opened 2019-02-22
Trim clientside commands

whalemer:


Zarel:
We might want to slice off the first line, rather than simply trimming.

PR #1237, opened 2019-02-22
Allow pm with no target to open a window

whalemer:


MacChaeger:
I’d prefer this to require a comma but no message, as I think that users accidentally omitting the comma between the username and their message is more common (or at least more pressing) than users trying to open a PM window without actually sending a message.

whalemer:
so no comma should error out as normal?

Zarel:
Works for me. Why did you close this, though?

whalemer:
I'm bad at git, apparently

PR #1242, opened 2019-03-01
Tooltips: Fix Normalize

MacChaeger:


src/battle-tooltips.ts
```diff
@@ -1101,13 +1101,15 @@ class BattleTooltips {
...
-			if (value.abilityModify(0, 'Normalize')) moveType = 'Normal';
+		if (allowTypeOverride && category !== 'Status') {
+			if (moveType === 'Normal') {
+				if (value.abilityModify(0, 'Aerilate')) moveType = 'Flying';
+				if (value.abilityModify(0, 'Galvanize')) moveType = 'Electric';
+				if (value.abilityModify(0, 'Pixilate')) moveType = 'Fairy';
+				if (value.abilityModify(0, 'Refrigerate')) moveType = 'Ice';
 			}
+			if (value.abilityModify(0, 'Normalize')) moveType = 'Normal';
+		}
```
Zarel: Does this change actually do anything.

Zarel: Oh, I understand now.

PR #1245, opened 2019-03-06
Don't intercept links in replay-uploaded popups

MacChaeger:


Zarel:
Oh I didn't realize it was a client-side popup; that makes things _much_ easier.

PR #1247, opened 2019-03-09
Tooltips: Fix Flower Gift

asgdf:

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -858,10 +858,10 @@ class BattleTooltips {
 					stats.spa = Math.floor(stats.spa * 1.5);
 				}
 				let allyActive = clientPokemon && clientPokemon.side.active;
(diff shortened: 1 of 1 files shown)


src/battle-tooltips.ts
```diff
@@ -858,10 +858,10 @@ class BattleTooltips {
 					stats.spa = Math.floor(stats.spa * 1.5);
 				}
 				let allyActive = clientPokemon && clientPokemon.side.active;
-				if (allyActive && allyActive.length > 1) {
+				if (allyActive && allyActive.length > 0) {
```
xfix: Checking whether `allyActive` is non-empty seems unnecessary when merely iterating over it (which for empty arrays does nothing).

asgdf: Yeah, fair point. The code in the condition used to look differently halfway through this and I never changed that back.

PR #1250, opened 2019-03-13
Add Ctrl+` as a keyboard shortcut for ``code``

scheibo:
One misleading thing is the 'Usable Formatting' information says to use `Cmd + <X>` because I'm on a Mac, but both `Cmd + <X>` and `Ctrl + <X>` (for `X = B or I`) work on my machine, but the 'grave accent' (I don't know how to escape it in GH markdown! :P) only works with `Ctrl`

### DIFF
--- a/js/client-topbar.js
+++ b/js/client-topbar.js
@@ -612,7 +612,7 @@
 			var ctrlPlus = '<kbd>' + (navigator.platform === 'MacIntel' ? 'Cmd' : 'Ctrl') + '</kbd> + ';
 			buf += '<p class="optlabel">**<strong>bold</strong>** (' + ctrlPlus + '<kbd>B</kbd>)</p>';
 			buf += '<p class="optlabel">__<em>italics</em>__ (' + ctrlPlus + '<bkd>I</kbd>)</p>';
-			buf += '<p class="optlabel">``<code>code formatting</code>`` (' + ctrlPlus + '<bkd>`</kbd>)</p>';
+			buf += '<p class="optlabel">``<code>code formatting</code>`` (<kbd>Ctrl</kbd> + <kbd>`</kbd>)</p>';
 			buf += '<p class="optlabel">~~<s>strikethrough</s>~~</p>';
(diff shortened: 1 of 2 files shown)


js/client-topbar.js
```diff
@@ -612,7 +612,7 @@
 			var ctrlPlus = '<kbd>' + (navigator.platform === 'MacIntel' ? 'Cmd' : 'Ctrl') + '</kbd> + ';
 			buf += '<p class="optlabel">**<strong>bold</strong>** (' + ctrlPlus + '<kbd>B</kbd>)</p>';
 			buf += '<p class="optlabel">__<em>italics</em>__ (' + ctrlPlus + '<bkd>I</kbd>)</p>';
-			buf += '<p class="optlabel">``<code>code formatting</code>``</p>';
+			buf += '<p class="optlabel">``<code>code formatting</code>`` (' + ctrlPlus + '<bkd>`</kbd>)</p>';
```
Zarel:
```suggestion
			buf += '<p class="optlabel">``<code>code formatting</code>`` (<kbd>Ctrl</kbd> + <kbd>`</kbd>)</p>';
```

You spelled `kbd` wrong, but also Cmd+` is specifically "switch window" on Mac and won't work.

Zarel:
We probably want to ignore Ctrl+B on Mac, actually, because it conflicts with the emacs keybinding for "left".

Zarel:
``` `` ` `` ``` is how you escape it in Markdown: `` ` ``. Works in PS, too, actually.

Zarel:
Oh, also, if you wanted to escape it without putting it in a code block, it's `` \` ``, like this: \`

Zarel:
Amusingly, "Quote link" doesn't quote my comment correctly, but "Reference in new issue" does.

js/client-topbar.js
scheibo: wtf. i literally just did a copy and paste? how did it get transposed? thanks.

PR #1253, opened 2019-03-15
Don't display illegal HMs as legal in Teambuilder

asgdf:
The learnset table build script did not take into account that HMs cannot be transferred from gen 3 to 4 and 4 to 5. This led to, for example, Exploud having Rock Climb displayed as legal in gen 4-7 (it's only legal in gen 4) and Pikachu having Surf displayed as legal in gen 5.

(Yeah the condition could probably be made a bit prettier)

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -628,6 +628,9 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 	// Learnset table
 	//
 
+	const gen3HMs = new Set(['cut', 'fly', 'surf', 'strength', 'flash', 'rocksmash', 'waterfall', 'dive']);
+	const gen4HMs = new Set(['cut', 'fly', 'surf', 'strength', 'rocksmash', 'waterfall', 'rockclimb']);
+
 	const learnsets = {};
 	BattleTeambuilderTable.learnsets = learnsets;
 	for (const id in Dex.data.Learnsets) {
@@ -637,11 +640,32 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 			const gens = learnset[moveid].map(x => Number(x[0]));
 			const minGen = Math.min(...gens);
 			const vcOnly = minGen === 7 && learnset[moveid].every(x => x[0] !== '7' || x === '7V');
-			if (minGen === 7) learnsets[id][moveid] = '7';
-			if (minGen === 6) learnsets[id][moveid] = '67';
-			if (minGen === 5) learnsets[id][moveid] = '567';
-			if (minGen === 4) learnsets[id][moveid] = '4567';
-			if (minGen === 3) learnsets[id][moveid] = '34567';
+
+			if (minGen <= 4 && (gen3HMs.has(moveid) || gen4HMs.has(moveid))) {
+				let legalGens = '';
+				let available = false;
+
+				if (minGen === 3) {
+					legalGens += '3';
+					available = true;
+				}
+				if (available) available = !gen3HMs.has(moveid);
+
+				if (available || gens.includes(4)) {
+					legalGens += '4';
+					available = true;
+				}
+				if (available) available = !gen4HMs.has(moveid);
+
+				let minUpperGen = available ? 5 : Math.min(
+					...gens.filter(gen => gen > 4)
+				);
+				legalGens += '01234567'.slice(minUpperGen);
+				learnsets[id][moveid] = legalGens;
+			} else {
+				learnsets[id][moveid] = '01234567'.slice(minGen);
+			}
+
 			if (gens.indexOf(6) >= 0) learnsets[id][moveid] += 'p';
 			if (gens.indexOf(7) >= 0 && !vcOnly) learnsets[id][moveid] += 'q';
 		}


build-tools/build-indexes
```diff
@@ -628,6 +628,9 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 	// Learnset table
 	//
 
+	const gen3HMs = {'cut':1, 'fly':1, 'surf':1, 'strength':1, 'flash':1, 'rocksmash':1, 'waterfall':1, 'dive':1};
```
Zarel: We haven't done this in a long time; benchmarking shows arrays and sets are faster and more readable. I'd probably use a `Set` here.

scheibo: +1 for `Set` for readability, even though somewhat counterintutively an array would likely be faster.

asgdf: Ah shoot, yeah. Got too used to that idiom. Since this is a build script, readability is probably more important than speed, so ``Set`` it is.
Wouldn't an Array#includes be O(n) while an object or Set is closer to O(1)? Does the abstraction offset the complexity for these relatively small amounts of data?

Zarel: I have suspicions that there are enough elements here for `Set` to be faster than `Array`, but yes, for small arrays, arrays are faster than sets because of the relative amounts of overhead.

build-tools/build-indexes
```diff
@@ -637,11 +640,34 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 			const gens = learnset[moveid].map(x => Number(x[0]));
 			const minGen = Math.min(...gens);
 			const vcOnly = minGen === 7 && learnset[moveid].every(x => x[0] !== '7' || x === '7V');
-			if (minGen === 7) learnsets[id][moveid] = '7';
-			if (minGen === 6) learnsets[id][moveid] = '67';
-			if (minGen === 5) learnsets[id][moveid] = '567';
-			if (minGen === 4) learnsets[id][moveid] = '4567';
-			if (minGen === 3) learnsets[id][moveid] = '34567';
+
+			if (gen3HMs.has(moveid) || gen4HMs.has(moveid)) {
```
Zarel:
```suggestion
			if (minGen <= 4 && (gen3HMs.has(moveid) || gen4HMs.has(moveid))) {
```

build-tools/build-indexes
```diff
@@ -637,11 +640,34 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
+					} else {
+						if (isHMInPrevGen || !availableInPrevGen) {
+							availableInPrevGen = false;
+						} else {
+							legalGens += gen;
+							availableInPrevGen = true;
+						}
+					}
+					isHMInPrevGen = (gen === 3 && gen3HMs.has(moveid)) || (gen === 4 &&  gen4HMs.has(moveid));
+				}
```
Zarel: `availableInPrevGen` and `isHMInPrevGen` are pretty redundant. Here's my refactor of this loop.

```js
				let available = false;
				for (let gen = minGen; gen <= 7; gen++) {
					if (available || gens.includes(gen)) {
						legalGens += gen;
						available = true;
					}
					if (gen === 3 && gen3HMs.has(moveid) || gen === 4 && gen4HMs.has(moveid)) {
						available = false;
					}
				}
```

Which actually can be further unrolled:

```js
				let legalGens = '';
				let available = false;
				if (minGen === 3) {
					legalGens += '3';
					available = !gen3HMs.has(moveid);
				}
				if (available || gens.includes(4)) {
					legalGens += 4;
					available = !gen4HMs.has(moveid);
				}
				let minUpperGen = available ? 5 : Math.min(
					...gens.filter(gen => gen > 4)
				);
				legalGens += '01234567'.slice(minUpperGen);
```

build-tools/build-indexes
```diff
@@ -637,11 +640,34 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
+				learnsets[id][moveid] = legalGens;
+			} else {
+				if (minGen === 7) learnsets[id][moveid] = '7';
+				if (minGen === 6) learnsets[id][moveid] = '67';
+				if (minGen === 5) learnsets[id][moveid] = '567';
+				if (minGen === 4) learnsets[id][moveid] = '4567';
+				if (minGen === 3) learnsets[id][moveid] = '34567';
```
Zarel: Meanwhile, these can all just be

```js
				learnsets[id][moveid] = '01234567'.slice(minGen);
```

build-tools/build-indexes
Zarel: n.b. `minUpperGen` can be `Infinity`, but this code will still work correctly if it is.

build-tools/build-indexes
scheibo: Yeah, my intuitions were wrong about how big the `Array` would need to be before `Set` won out - I figured around 5 elements would be the break even point.
https://jsperf.com/pokemon-showdown-includes-vs-has/ (on my monster Linux box, `has` wins, and on my Macbook `includes` wins, and they're pretty close and thats for 10 elements) . Yes. algorithmic complexity is a thing, but in the real world constant factors tend to really matter. I'd never want to discourage using `Set` for something like this though - semantically it makes sense, even if linear search of small arrays beats it.

Zarel: In my experience, in the real world, constant factors stop mattering at around n=15. n=8 is a toss-up, so it makes sense that it'd be different between powerful/unpowerful machine.

asgdf:
Did all these, seems to work fine.
(At this point I basically just made you do this commit for me, haven't I? Sorry)

Zarel:
If it helps, I think writing the original code is much easier than optimizing. Optimizing is fun for me, anyway.

Issue #1252, opened 2019-03-14
Improve the build process to make iterative development easier


scheibo:
> You can make and test client changes simply by building after each change

I think there is room to improve the client `build` process to make it a little faster/more ergonomic.
I don't think we can/should switch the client from `babel` to `sucrase` like the server because the client depends on a number of plugins which `sucrase` by design will not support, but I think restructuring our build process to support some sort of `'--watch'` mode (to at least support interactive edits to `src/`) and highlighting it in our `README.md` would make things easier for new developers.

scheibo:
Typescript incremental builds in 3.4 might help with this.

Also, `node build full` seems broken, as reported in chat:

```
C:\Users\feeds\Desktop\Pokemon-Showdown-Client>node build full
Cloning into 'Pokemon-Showdown'...
Checking out files: 100% (391/391), done.
Syncing data from Git repository... DONE
internal/modules/cjs/loader.js:582
    throw err;
    ^

Error: Cannot find module '../data/Pokemon-Showdown/.sim-dist/dex'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:580:15)
    at Function.Module._load (internal/modules/cjs/loader.js:506:25)
    at Module.require (internal/modules/cjs/loader.js:636:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (C:\Users\feeds\Desktop\Pokemon-Showdown-Client\build-tools\build-indexes:21:13)
    at Module._compile (internal/modules/cjs/loader.js:688:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:699:10)
    at Module.load (internal/modules/cjs/loader.js:598:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:537:12)
    at Function.Module._load (internal/modules/cjs/loader.js:529:3)
child_process.js:651
    throw err;
    ^

Error: Command failed: node ./build-tools/build-indexes
    at checkExecSyncError (child_process.js:611:11)
    at execSync (child_process.js:648:13)
    at Object.<anonymous> (C:\Users\feeds\Desktop\Pokemon-Showdown-Client\build:36:2)
    at Module._compile (internal/modules/cjs/loader.js:688:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:699:10)
    at Module.load (internal/modules/cjs/loader.js:598:32)
[...]

PR #1229, opened 2019-02-19
Preact rewrite

Zarel:
Apparently I can't convert a pull request into a draft pull request, but I can re-create the pull request. Since the Preact rewrite is probably our longest-term project, this is probably the correct thing to do here.

PR #1255, opened 2019-03-21
Add keyboard shortcuts to the Preact client

scheibo:
Challenges encountered:

- [ ] logging in doesn't work, the instructions from [here](https://github.com/Zarel/Pokemon-Showdown-Client/pull/1229#issuecomment-467430493) are helpful as a workaround and would be useful to spread more broadly (section in the `README`?) if we want others to work on this Preact client. I also didn't know which file to open `preactalpha.template.html` looked like a better choice than `testclient-beta.html` (it has `'preact'` in the name!). Also, `alpha` vs. `beta` adds confusion. Maybe rename to `testclient-preact.html`?
- [ ] running into issues with building and caching. Given that more code needs to be built, I think #1252 becomes higher priority. I imagine we should be able to split off a `--watch`able part of the build out which would be much more convenient. Regarding caching - I ended up just using `Ctrl + Shift + R` a lot - is this something thats also a problem on the non-Preact client (I don't remember it from implementing #1250)?
- [ ] Preact seems to be doing fishy things with `keyCode`. `'a'` spits out `keyCode` 1. Either way, `keyCode` seems [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode), so I switched to `key` which seems to support IE9+. Enter still worked as `keyCode` 13 on my computer, but for consistency I switched it over.
- [ ] `Ctrl + GRAVE` (seriously, just spent another few minutes trying to get Github to escape it and lost patience) seems to cause the text element to lose focus and doesn't register as a keypress, so doesn't work.

### DIFF
--- a/src/panel-chat.tsx
+++ b/src/panel-chat.tsx
@@ -112,16 +112,77 @@ class ChatTextEntry extends preact.Component<{room: PSRoom, onMessage: (msg: str
 		elem.focus();
 	};
 	keyPress = (e: KeyboardEvent) => {
+		const cmdKey = (((e.cmdKey || e.metaKey) ? 1 : 0) + (e.ctrlKey ? 1 : 0) === 1) && !e.altKey && !e.shiftKey;
 		let elem = e.currentTarget as HTMLTextAreaElement;
-		if (e.keyCode === 13 && !e.shiftKey) {
+		if (e.key === 'Enter' && !e.shiftKey) {
 			this.props.onMessage(elem.value);
 			elem.value = '';
 			this.update();
 			e.preventDefault();
 			e.stopImmediatePropagation();
-			return;
+		} else if (e.key === 'i' && cmdKey) {
+			if (ChatTextEntry.toggleFormatChar(elem, '_')) {
+				e.preventDefault();
+				e.stopImmediatePropagation();
+			}
+		} else if (e.key === 'b' && cmdKey) {
+			if (ChatTextEntry.toggleFormatChar(elem, '*')) {
+				e.preventDefault();
+				e.stopImmediatePropagation();
+			}
+		} else if (e.key === '`' && cmdKey) {
+			// FIXME: doesn't work :(
+			if (ChatTextEntry.toggleFormatChar(elem, '`')) {
+				e.preventDefault();
+				e.stopImmediatePropagation();
+			}
 		}
 	};
+	static toggleFormatChar(elem, formatChar) {
+		if (!elem.setSelectionRange) return false;
+
+		let value = elem.value;
+		let start = elem.selectionStart;
+		let end = elem.selectionEnd;
+
+		// make sure start and end aren't midway through the syntax
+		if (value.charAt(start) === formatChar && value.charAt(start - 1) === formatChar &&
+			value.charAt(start - 2) !== formatChar) {
+			start++;
+		}
+		if (value.charAt(end) === formatChar && value.charAt(end - 1) === formatChar &&
+			value.charAt(end - 2) !== formatChar) {
+			end--;
+		}
+
+		// wrap in doubled format char
+		const wrap = formatChar + formatChar;
+		value = value.substr(0, start) + wrap + value.substr(start, end - start) + wrap + value.substr(end);
+		start += 2;
+		end += 2;
+
+		// prevent nesting
+		var nesting = wrap + wrap;
+		if (value.substr(start - 4, 4) === nesting) {
+			value = value.substr(0, start - 4) + value.substr(start);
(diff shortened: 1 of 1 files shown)


src/panel-chat.tsx
```diff
@@ -112,16 +112,77 @@ class ChatTextEntry extends preact.Component<{room: PSRoom, onMessage: (msg: str
 		elem.focus();
 	};
 	keyPress = (e: KeyboardEvent) => {
+		const cmdKey = (((e.cmdKey || e.metaKey) ? 1 : 0) + (e.ctrlKey ? 1 : 0) === 1) && !e.altKey && !e.shiftKey;
 		let elem = e.currentTarget as HTMLTextAreaElement;
-		if (e.keyCode === 13 && !e.shiftKey) {
+		if (e.key === 'Enter' && !e.shiftKey) {
```
Zarel: `e.key` has pretty bad compatibility; I'd rather not.

src/panel-chat.tsx
```diff
@@ -112,16 +112,77 @@ class ChatTextEntry extends preact.Component<{room: PSRoom, onMessage: (msg: str
...
+			}
+		} else if (e.key === '`' && cmdKey) {
+			// FIXME: doesn't work :(
+			if (ChatTextEntry.toggleFormatChar(elem, '`')) {
+				e.preventDefault();
+				e.stopImmediatePropagation();
+			}
 		}
 	};
+	static toggleFormatChar(elem, formatChar) {
```
Zarel: I guess you haven't run TypeScript yet.

```suggestion
	static toggleFormatChar(elem: HTMLTextareaElement, formatChar: string) {
```

src/panel-chat.tsx
xfix: What does that mean? `e.key` is supported in IE9, and I don't think we support IE8 for non-replay related parts of client (which this isn't).

Zarel: See the notes "Partial support refers to these versions of Firefox returning `'MozPrintableKey'` for all character keys". And "Has non-standard key identifiers and incorrect behaviour with AltGraph."

This table doesn't even cover _all_ the problems with `event.key` – it says Firefox 29+ is standards compliant, and doesn't even mention that it took until Firefox 37+ to return `" "` instead of `"Spacebar"` for the spacebar (IE9-IE11, supported browsers, also have this problem). Other keys with incompatibilities in supported browsers include the arrow keys (`"Left"` vs `"ArrowLeft"`, all the way up to Edge 16), and Esc (`"Escape"` vs `"Esc"`).

None of these incompatibilities exist in `event.keyCode`.

And even if I don't care about compatibility with IE11 (I do), there's still the question of, I don't want to have to learn multiple key APIs just to detect a keypress. The one I know is consistent and works in all browsers; I don't care if it's deprecated; I write other software that does need to support older browsers, and I don't want myself or other contributors to have to learn multiple APIs just because some people want to use the new hot API.

Like, sure, if the new hot API was appreciably better or something, but it clearly isn't.

Zarel: This is actually one of the things that I really am annoyed at MDN for; their advice is "`keyCode` is deprecated; you should use an alternative. No true alternatives exist."

They don't even bother to explain the tradeoffs alternatives have. It's just a vague "you might have compatibility issues sometimes, but we won't tell you which times."

Zarel: Googling for advice on how to handle key presses give [Medium articles](https://medium.com/@uistephen/keyboardevent-key-for-cross-browser-key-press-check-61dbad0a067a) by "experts" who don't know the difference between `preventDefault` and `stopPropagation`:

> I’m curious if anyone knows for sure what the behavior is and how necessary it is to include this. Since it doesn’t hurt anything, I’m including it anyway.

With such glorious recommended code as:

```
var key = event.key || event.keyCode;
if (key === 'Escape' || key === 'Esc' || key === 27) { }
```

Oh yes, this is _so_ much better than `if (event.keyCode === 27) {}`. I wonder why webdevs aren't switching in _droves_ to this _new better API_...

xfix:
```suggestion
		if (e.keyCode === 13 && !e.shiftKey) {
```

xfix:
```suggestion
		if (e.keyCode === 13 && !e.shiftKey) {
```

Zarel:
> * [ ]  `Ctrl + GRAVE` (seriously, just spent another few minutes trying to get Github to escape it and lost patience)

`` Ctrl + ` `` ;)

(``` `` Ctrl + ` `` ```)

PR #1256, opened 2019-03-22
fix toxic damage display

arandompieceofcheese:
There's a mistake (?) in the toxic display that causes next damage to display as `6 * toxicTurns` instead of `1/16 * toxicTurns`.

This doesn't really affect gameplay at all, and actually keeps the display more in line with what the mon's toxic damage message, but it _is_ still a mistake and it annoyed me enough to spend 5 minutes fixing it

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -651,9 +651,9 @@ class BattleTooltips {
 			if (clientPokemon) {
 				if (pokemon.status === 'tox') {
 					if (pokemon.ability === 'Poison Heal' || pokemon.ability === 'Magic Guard') {
-						text += ' <small>Would take if ability removed: ' + Math.floor(100 / 16) * Math.min(clientPokemon.statusData.toxicTurns + 1, 15) + '%</small>';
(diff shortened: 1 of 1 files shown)


src/battle-tooltips.ts
```diff
@@ -651,9 +651,9 @@ class BattleTooltips {
 			if (clientPokemon) {
 				if (pokemon.status === 'tox') {
 					if (pokemon.ability === 'Poison Heal' || pokemon.ability === 'Magic Guard') {
-						text += ' <small>Would take if ability removed: ' + Math.floor(100 / 16) * Math.min(clientPokemon.statusData.toxicTurns + 1, 15) + '%</small>';
+						text += ' <small>Would take if ability removed: ' + Math.floor(100 / 16 * Math.min(clientPokemon.statusData.toxicTurns + 1, 15)) + '%</small>';
```
scheibo: I think you can add `const damage = Math.floor(100 / 16 * Math.min(clientPokemon.statusData.toxicTurns + 1, 15))` after line 652 and use the variable in both places.

Zarel:
My mistake, I seem to have missed this pull request. Thanks for the contribution!

PR #1258, opened 2019-03-29
Tooltips: Fix Plus/Minus/Victory Star

asgdf:
These ability checks were using the client pokemon, which won't contain ability information if the ability does not announce itself. This meant that Plus, Minus and Victory Star just didn't work in tooltips. I changed it to use the server pokemon instead.

I also applied that change to the Cherrim check, even though it currently works, because the boost is only active after the forme change which announces the ability. I don't know if this is worth it, but it feels so brittle to rely on the other behavior.

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -866,7 +866,8 @@ class BattleTooltips {
 				if (allyActive) {
 					for (const ally of allyActive) {
 						if (!ally || ally.fainted) continue;
-						if (ally.ability === 'Flower Gift' && (ally.getTemplate().baseSpecies === 'Cherrim' || this.battle.gen <= 4)) {
+						let allyAbility = this.getAllyAbility(ally);
+						if (allyAbility === 'Flower Gift' && (ally.getTemplate().baseSpecies === 'Cherrim' || this.battle.gen <= 4)) {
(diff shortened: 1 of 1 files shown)


Zarel:
In theory they should all be using the client ability and falling back to the server ability.

This is important for Ditto's Transform, for which the server ability will be Imposter but the client ability will be the real ability, if revealed.

asgdf:
Ah. Okay, how would that look in practise? Every time the tooltips are being affected by an ally's ability, go through both server and client allies and coordinate them so we can use the server data as a fallback? Do we put both into one data structure first so it's easier to navigate?

Zarel:
In theory we shouldn't need much coordination. Active pokemon should have the same `serverPokemon` index as `active` index, which should both be `pokemon.slot`.

asgdf:
Alright, I did that.
I also rephrased the conditions for Plus/Minus a bit, because I found that rather difficult to parse with the barely-visible negation (it's not very distinct from the parentheses that surround it) that applies to the whole condition which also introduces double negatives.

asgdf:
That test failure seems to come from the AFD changes?

Zarel:
Yeah, I had zero free time and a bad internet connection on April Fool's Day so I pushed a partial change to the repo. Restarting the build should fix it.

PR #1259, opened 2019-03-31
Battle UI: Don't use `switchposition` controls in Gen 7

Slayer95:
Forced slot switching

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -350,7 +350,7 @@
 			// this.choice.choices = array of choice strings
 			// this.choice.switchFlags = dict of pokemon indexes that have a switch pending
 			// this.choice.switchOutFlags = ???
-			// this.choice.freedomDegrees = in a switch request: number of empty slots that can't be replaced
+			// this.choice.forcedSwitchLocations = in a switch request: whether Pokémon must be sent on the left
 			// this.choice.type = determines what the current choice screen to be displayed is
 			// this.choice.waiting = true if the choice has been sent and we're just waiting for the next turn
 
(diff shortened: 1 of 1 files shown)


js/client-battle.js
```diff
@@ -379,8 +379,8 @@
 					if (this.request.forceSwitch !== true) {
 						var faintedLength = _.filter(this.request.forceSwitch, function (fainted) {return fainted;}).length;
 						var freedomDegrees = faintedLength - _.filter(switchables.slice(this.battle.mySide.active.length), function (mon) {return !mon.fainted;}).length;
-						this.choice.freedomDegrees = Math.max(freedomDegrees, 0);
-						this.choice.canSwitch = faintedLength - this.choice.freedomDegrees;
+						this.choice.canSwitch = faintedLength - Math.max(freedomDegrees, 0);
+						this.choice.freedomDegrees = this.battle.gen >= 7 ? 0 : Math.max(freedomDegrees, 0);
```
Zarel: I'm confused. Shouldn't `freedomDegrees` be sent from the server?

Zarel:
I think I'd rather have this as a part of `request` than to have it be inferred from the gen.

In general I think data about what sort of choices are legal should be sent explicitly through the request, rather than needing to be inferred from anything else.

js/client-battle.js
Slayer95: That makes more sense. Though are you fine with that key name?

> In statistics, the number of degrees of freedom is the number of values in the final calculation of a statistic that are free to vary

Zarel: I really dislike the key name.

Zarel: Maybe the flag should be `canChooseSwitchLocation`.

Slayer95:
@Zarel , actually this should be merged as is before implementing the flag in the server for compatibility.

Zarel:
I don't see a check for the flag. Should I wait for that?

Slayer95:
There is an issue with ``canChooseSwitchLocation``.

Strictly speaking, the cartridge regularly requests the player to choose switch locations when 2 Pokémon faint. Now, because the issued commands are commutative, we could avoid requesting for locations unless the player could opt for leaving an amount of slots empty (equal to ``freedomDegrees``).

In Gen 7, the player can no longer opt for leaving slots empty. However, the cartridge still requests for switch locations if 2 Pokémon faint at the same time and can both be replaced. Following that, it would be a shame to have a ``canChooseSwitchLocation`` flag in the protocol which isn't accurate to the in-game behavior.

Slayer95:
Now, that issue would be non-existent if we followed the in-game UI to the letter. However, I believe that would add undeserved clunkyness.

Zarel:
Our UI already doesn't match cartridge though, right? Feel free to either fix or not depending on your preference.

Slayer95:
> There is an issue with canChooseSwitchLocation.

I'd rather have a different name, and punt the issue of matching cartridge regarding requests for switch locations for the future.

Slayer95:
Alright. I made up my mind on an action path.
1. Use ``canChooseSwitchLocation``
2. Keep our divergences in UI from cartridge.
3. If we ever decide to change point 2, it can be easily toggled server-side.
4. Hardcore mode can be made to override ``canChooseSwitchLocation``, but user demand should be probed.

With that stated and if there are no objections, this is ready !

Slayer95:
Server PR: https://github.com/Zarel/Pokemon-Showdown/pull/5390

Zarel:
Maybe `forcedSwitchLocations`?

Zarel:
Specifically, if switch locations are chosen in double-faints normally, the flag isn't so much about "you can choose the switch location" so much as "you're forced to switch into the leftmost available slot".

Slayer95:
> If we ever decide to change point 2, it can be easily toggled server-side.

I believe there is value in that. Even if we don't change the UI, a VGC-dedicated server could opt for doing so.

On the other hand, the semantics of ``forcedSwitchLocations`` imply that it's equivalent to ``battle.gen === 7```, so the UI would be determined by the client. Or it could just be the negation of ``canChooseSwitchLocations`` ? .....

Sigh, this bikeshedding is getting tiresome.

Zarel:
Hey, protocol changes are important enough that I personally don't consider them bikeshedding. They're like the one thing we have to support long-term.

Slayer95:
Do let me know if you reach a final plan though.

Zarel:
Anyway, it doesn't make sense for accurate simulation of something UI-only to be done server-side. It should be client-side, and the client-side option (hardcore mode) already exists.

Zarel:
So yes, `forcedSwitchLocations` is supposed to be equal to `battle.gen >= 7`, and that's still my preferred approach. Feel free to suggest other names, but you've yet to present a convincing argument otherwise for mechanics.

And if at any point you want to stop arguing and decide it all yourself, you have my approval for that, too.

Slayer95:
Well, that's fine too. I'll check the implementation details to account for that naming later.

PR #1263, opened 2019-04-07
Add a persistent /blockchallenges & /blockpms toggle

scheibo:
> [17:43:31] %Flare ✇: I dream of the day where the blockchallenges function can be something that can be integrated to the settings menu ;_;

Seems like a reasonable enough request. The two concerns I can see being raised:

- adds clutter to the `Settings` dropdown, sets up a slippery slope for what settings we put in there
- Blocking challenges is server side state, so the checkbox can be out of sync with the client (if the user issues `/bch`, the checkbox won't update and will not reflect reality). UI wise this isn't ideal, but the server doesn't really care - it will just send an `|error|` which gets ignored. This can possibly be alleviated by adding a mechanism to filter all chat messages and observing `/bch`, `/blockchallenges`, `/unblockchallenges` commands to be able to update the checkbox correctly, but I don't want to go down that road without getting some feedback. TBH, maybe just renaming the checkbox 'Block challenges on login' or something (and still allowing the user to manually change the state during the session, allowing the checkbox to get out of sync until they refresh) might be sufficient?

Also worth noting that the important part of this change is to make the block challenges state persistent, so that users don't need to issue `/bch` every time they log in.


bumbadadabum:
Yeah the second concern is exactly what kept me from implementing it this way years ago. I don't know if that is per se a blocker though. It just seems weird to make a serverside option permanent by just sending the message every time even when it's redundant.

scheibo:
Yup, 100% agree. But given how its a common pain point, it seems like we should figure out a solution instead of continuing to punt because its not 'pure'. Thinking about it some more, my suggestion of simply snooping on the client's input to see if a `blockchallenge` related command has been issued is not sufficient (and seemed hacky anyway), as a user could have multiple clients open. We'd need a way of not just *querying* the server for the current state of `blockchallenge`, but *subscribing* to changes of the `blockchallenge` bit to properly be able to reflect the state in the UI. 

Though once again, maybe this is overkill and we can deal with a little inconsistencies in the UI in order to provide 99% of the feature users want? (who am I kidding, @Zarel will shoot this down... :P)

Zarel:
I think ideally we should probably have user-settings like `blockPMs` and `blockchallenges` sent by `|updateuser|`, so they can be kept in sync.

js/client.js
```diff
@@ -165,6 +166,9 @@
 					self.nameRegExp = new RegExp('(?:\\b|(?!\\w))' + escaped + '(?:\\b|\\B(?!\\w))', 'i');
 				}
 			});
+			this.on('change:settings', function () {
+				Dex.prefs('usersettings', self.get('settings'));
```
Zarel: This works? I thought `Storage.prefs` was the setter, and `Dex.prefs` was a compatibility-getter written for battles.

Zarel: Hm, apparently `Dex.prefs` forwards everything. It probably shouldn't.

scheibo: I just copied the pattern from the other code I saw. Should this be `Storage.prefs` instead?

js/client.js
```diff
@@ -931,7 +953,8 @@
 					name: name,
 					userid: userid,
 					named: named,
-					avatar: parts[3]
+					avatar: parts[3],
+					settings: JSON.parse(parts[4]),
```
Zarel: A trailing comma? I'm surprised ESLint isn't yelling at you. This is the client! Safari 5 doesn't support trailing commas.

js/client.js
```diff
@@ -138,7 +138,8 @@
 			userid: '',
 			registered: false,
 			named: false,
-			avatar: 0
+			avatar: 0,
+			settings: {},
```
Zarel: Ditto my other comment on trailing commas.

js/client.js
Zarel: The old client is rather messy. For now, either probably works.

scheibo: Alright, well I went with `Storage.prefs`, but I dont think thats the more typical way given the code I was looking at.

js/client.js
```diff
@@ -913,8 +935,8 @@
 				var nlIndex = data.indexOf('\n');
 				if (nlIndex > 0) {
 					this.receive(data.substr(nlIndex + 1));
-					nlIndex = parts[3].indexOf('\n');
-					parts[3] = parts[3].substr(0, nlIndex);
+					nlIndex = parts[4].indexOf('\n');
```
Zarel: Wait, this isn't backward compatible!

Zarel: Wait, the server change that's already in will break the client!

Zarel: Okay, I fixed it but it's a merge conflict. Your code will also need to be updated for backward compatibility.

scheibo: Alright, I'm just hiding the new toggles if the server doesn't have the new protocol.

js/client.js
scheibo: Actually, I went back on this. `Dex.prefs` for settings appears to be the overwhelming favorite, so Im going to opt for consistency.

js/client-topbar.js
```diff
@@ -485,6 +488,10 @@
 			buf += '<hr />';
 			buf += '<p><strong>Chat</strong></p>';
 			buf += '<p><label class="optlabel"><input type="checkbox" name="notournaments"' + (Dex.prefs('notournaments') ? ' checked' : '') + ' /> Ignore tournaments</label></p>';
+			if (Object.keys(settings).length) {
```
Zarel: This is above and beyond, I just wanted it not to _crash_ on older servers. :p

js/client.js
Zarel: I also don't like `usersettings` as the pref key here. All of `Storage.prefs` are user settings. These are just the settings that need to be sent to the server.

Zarel: I know I know bikeshedding, but client localStorage is literally the most permanent support burden we have. Even old server protocols we tend to stop supporting after a few years.

Zarel: `serversettings` possibly. It also might be an idea to just have them at the top level with everything else.

js/client-topbar.js
scheibo: Why *not* be more robust? :)

scheibo: I've even gone further - I now should be robust to *future* server protocol changes (Zarel/Pokemon-Showdown#5424) as well!

js/client.js
scheibo: I went with `serversettings`. I'd be OK with pulling them out (I assume not just at the `localStorage layer`, but also in `User`?), but this results in the simplest code, so I'd only do it if you wanted it?

> I know I know bikeshedding, but client localStorage is literally the most permanent support burden we have.

Its still not set in stone? We could write `blockchallenges` and `blockpms` settings in the future, and add migration logic which, if it detects a `serversettings` setting will perform the translation and delete the old storage. At a certain point (after a sunset period) you would want to remove the migration logic though, and users who hadn't used the app during the entire migration period would have their settings lost. Which in these edge cases, seems like an acceptable risk (oh no, they'll need to toggle the checkboxes again!)

That being said, I totally agree that protocol and naming is important. localStorage is effectively a database on the user's computer, so it takes schema migrations which are already painful and makes them 10x worse.

scheibo: I think I actually prefer `serversettings` to individually setting them. I think because we get them all from the same place (`|updateuser|`) editing them together makes sense.

That being said, 6fb7989 pulls them out of localStorage if you'd like (I left them together as `settings` in `User` because I think its nice that the'yre namespaced there instead of cluttering the top level properties) . You can revert 4b737f2 and merge that instead if you want them to be stored separately in users' browsers (though as mentioned above, whatever you decide isn't exactly set in stone).

Zarel:
Fine by me.

PR #1265, opened 2019-04-11
Handle AFK '@-in-names' protocol

scheibo:
Client changes for Zarel/Pokemon-Showdown#5002.

- make user names which are away show up in grey
- sort users who are away to the bottom
- add an 'Away' pill where approriate (not 100% on the styling)

This should be as backwards compatible as possible (it might interpret names ending in `@` as away when really they're just legacy, but thats an artifact of the protocol).

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1414,7 +1414,7 @@
 				var users = userList.substr(commaIndex + 1).split(',');
 				for (var i = 0, len = users.length; i < len; i++) {
 					if (users[i]) {
-						var user = parseNameParts(users[i]);
+						var user = BattleTextParser.parseNameParts(users[i]);
 						this.users[toUserid(user.name)] = user;
 					}
 				}
@@ -1425,7 +1425,7 @@
 			this.userList.construct();
 		},
 		addJoinLeave: function (action, name, oldid, silent) {
-			var user = parseNameParts(name);
+			var user = BattleTextParser.parseNameParts(name);
 			var userid = toUserid(user.name);
 			if (!action) {
 				this.$joinLeave = null;
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1757,11 +1757,11 @@
 			if (aUser.away !== bUser.away) return aUser.away - bUser.away;
 
 			var aRank = (
-				Config.groups[(aUser ? aUser.name.charAt(0) : Config.defaultGroup || ' ')] ||
+				Config.groups[aUser ? aUser.name.charAt(0) : Config.defaultGroup || ' '] ||
 				{order: (Config.defaultOrder || 10006.5)}
 			).order;
 			var bRank = (
-				Config.groups[(bUser ? bUser.name.charAt(0) : Config.defaultGroup || ' ')] ||
+				Config.groups[bUser ? bUser.name.charAt(0) : Config.defaultGroup || ' '] ||
 				{order: (Config.defaultOrder || 10006.5)}
 			).order;
 
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -26,7 +26,6 @@
 		},
 		updateUser: function () {
 			var name = app.user.get('name');
-			var userid = app.user.get('userid');
 			if (this.expired) {
 				this.$chatAdd.html(this.expired === true ? 'This room is expired' : BattleLog.sanitizeHTML(this.expired));
 				this.$chatbox = null;
@@ -37,7 +36,8 @@
 				this.$chatAdd.html('<form><button name="login">Join chat</button></form>');
 				this.$chatbox = null;
 			} else {
-				this.$chatAdd.html('<form class="chatbox"><label style="' + BattleLog.hashColor(userid) + '">' + BattleLog.escapeHTML(name) + ':</label> <textarea class="textbox" type="text" size="70" autocomplete="off"></textarea></form>');
+				var color = app.user.get('away') ? 'color:#AAA;' : BattleLog.hashColor(app.user.get('userid'));
+				this.$chatAdd.html('<form class="chatbox"><label style="' + color + '">' + BattleLog.escapeHTML(name) + ':</label> <textarea class="textbox" type="text" size="70" autocomplete="off"></textarea></form>');
 				this.$chatbox = this.$chatAdd.find('textarea');
 				this.$chatbox.autoResize({
 					animate: false,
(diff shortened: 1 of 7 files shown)


Zarel:
I would probably want `@` to be banned from usernames first, before doing this.

Zarel:
Also, to be clear, I wanted `@` to be a status separator: everything after `@` would be a user's status text.

scheibo:
@bumbadadabum

scheibo:
Would it be possible for you to spec out more completely what you have in mind? I feel it would be more productive than doing several rounds of 'guess and check' implementations?

Zarel:
Step 1: ban `@` from usernames, everywhere.

Step 2: `@` in "usernames" is now a status separator. `|n|`, `|j|`, `|l|`, `|users|`, and maybe a few other messages that previously took `USERNAME` can now also optionally take `USERNAME@STATUS`. There'll be no need to update `toId` or `toUserid` because `@STATUS` should be stripped off before it gets passed to `toId`; it should happen at the stage where the username itself is extracted.

Step 3: Figure out which statuses are considered "away"; maybe just have a symbol, like any status beginning with `!`.

Step 4: The rest is implementation detail.

bumbadadabum:
I did step 1, will try to implement step 2 into the protocol.

scheibo:
What is '`STATUS'`? Is this like 'idle' vs. 'away' vs. 'busy' vs. 'brb' etc? Is it an arbitrary string set by the user (open to abuse?) How is it displayed? How do statuses sort relative to each other? etc

> I did step 1

I think you want to pull out that part on its own and push it now, independent of the other changes, @bumbadadabum .

bumbadadabum:
It's mostly a future thing. Right now the only status we need is away, but
we could have different statuses in the future. We can think of how to
display them if we ever want to add new statuses. Just is nice to have the
option there in the protocol.

Zarel:
It's an arbitrary string set by the user. I don't see it as open to abuse; users can already set arbitrary names and send messages. I would restrict the permission with namelocks.

scheibo:
> There'll be no need to update `toId` or `toUserid` because `@STATUS` should be stripped off before it gets passed to `toId`; it should happen at the stage where the username itself is extracted.

Done (I hope?). `parseNameParts` could use a better name perhaps, but seems to solve the issue? I'm happier with the latest code, definitely feels less hacky.

- styling the status message is an open ended question - right now its just italic and sits under the name in the popup. 'Offline' is treated the same as a status now, only it shows up in red (can `text-transform: uppercase` and `font-weight: bold` the `.offline` class if you don't want to 'break' the previous style)
- the user options popup deliberately doesn't do anything with `status` or the `away` bit because it didn't look go there. I think when we support a `/status` command there should be a text box to change your status (or `contenteditable`?) with a checkbox for whether the status toggles the `away` bit. The `/status` command should be pretty trivial to add in the future, but bumba and I have decided its out of scope for the current implementation (though users can still set a custom `/away MSG` manually...)

src/battle-dex.ts
```diff
@@ -128,6 +128,22 @@ function toUserid(text: any) {
 	return toId(text);
 }
 
+function parseNameParts(text: string) {
```
Zarel: I definitely don't think we need more globals.

I guess this could be stuck next to `parseLine`.

js/client-chat.js
```diff
@@ -1736,12 +1751,17 @@
...
 		comparator: function (a, b) {
 			if (a === b) return 0;
+
+			var aUser = this.room.users[a];
+			var bUser = this.room.users[b];
+			if (aUser.away !== bUser.away) return aUser.away - bUser.away;
+
 			var aRank = (
-				Config.groups[(this.room.users[a] ? this.room.users[a].charAt(0) : Config.defaultGroup || ' ')] ||
+				Config.groups[(aUser ? aUser.name.charAt(0) : Config.defaultGroup || ' ')] ||
```
```suggestion
				Config.groups[aUser ? aUser.name.charAt(0) : Config.defaultGroup || ' '] ||
```

(parens are useless here)

js/client-chat.js
```diff
@@ -1736,12 +1751,17 @@
...
 				{order: (Config.defaultOrder || 10006.5)}
 			).order;
 			var bRank = (
-				Config.groups[(this.room.users[b] ? this.room.users[b].charAt(0) : Config.defaultGroup || ' ')] ||
+				Config.groups[(bUser ? bUser.name.charAt(0) : Config.defaultGroup || ' ')] ||
```
```suggestion
				Config.groups[bUser ? bUser.name.charAt(0) : Config.defaultGroup || ' '] ||
```

src/battle-dex.ts
scheibo: I'd want this to be a static method on `User`, but `User` isn't accessible by `UserList`. Globals are the result of the current design in the client from what I can tell, and there's no suitable `Util` class to stick this under. As for `lineParse` - isn't that in the Preact code?

Zarel: There should be a line parser in `BattleLog`. Let me check.

Zarel: https://github.com/Zarel/Pokemon-Showdown-Client/blob/4d2f86be7dbf79f75513d2180c897cf8f0d8d09f/src/battle-text-parser.ts#L18

Aha, yeah, put it next to this.

scheibo: This still requires adding a global, and I don't feel `BattleTextParser` is the appropriate place for this because the logic there seems more related to battle protocol parsing, not generic crap like this, but OK.

Zarel: I mean, it's adding it to an existing global. Which I prefer to making a new global. Namespacing is useful!

And `@` in names is part of the battle protocol, because `|j|` etc are part of the battle protocol.

Zarel:
bdov is so hot: u know how u can type a letter and tab and it fills in someones username
bdov is so hot: thats not working

Zarel:
image

Zarel:
Heh, honestly, changing the data type of `room.users` is definitely the kind of change I wouldn't dare do without TypeScript backing me up.

scheibo:
> Heh, honestly, changing the data type of `room.users` is definitely the kind of change I wouldn't dare do without TypeScript backing me up.

I mean, this is why I only work in statically typed languages these days. If you dont, you have to instead basically write a ton of tests to provide the same kind of guarantees, and oh whoops the client doesn't have those either...

TBH I didn't expect this to be merged at this point, I think I had told you I wanted to cover the Preact client in the same change after you were OK with it, and maybe the fact that the Preact client is TS might have caught this? (not sure though)

PR #1266, opened 2019-04-14
Teambuilder: Fix changed hidden abilities in older gens

asgdf:
This is relevant for the Scolipede line (used to have Quick Feet instead of Speed Boost) and Aurumoth (used to have Illusion instead of Light Metal).

I don't know if just slapping new properties into the teambuilder tables like this is the best solution, but it works at least.


asgdf:
(bump)

PR #1267, opened 2019-04-18
Fix possible error in spread move animation

asgdf:
I don't know what that `': ?'` was supposed to be for, it's been there since this feature was implemented 3 years ago, but I found no use for it in the current code.
Adding it to the end of the target breaks things when there is an identity where the slot is not specific enough to identify the target on its own, so it will have to resort to comparing idents. The added suffix will make it not find a match.
Currently, this causes a big ugly client error message when using Mind Blown in Doubles.

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -1356,7 +1356,7 @@ class Battle {
 			targets.push(target.side.missedPokemon);
 		} else {
 			for (const hitTarget of kwArgs.spread.split(',')) {
-				targets.push(this.getPokemon(hitTarget + ': ?')!);
+				targets.push(this.getPokemon(hitTarget)!);
 			}
 		}
 


Zarel:
Do you know why replays are fine?

`: ?` is supposed to match any ident, hence the `?`.

asgdf:
It's weird, the replay (https://replay.pokemonshowdown.com/gen7randomdoublesbattle-894137477 turn 4) was fine the first time I watched it, but now I always get the error there too. Maybe I had some old JS cached where the error didn't happen.

So if the `: ?` is supposed to match all idents, shouldn't it be used _instead_ of an invalid ident instead of being appended to the end of all of them?

Zarel:
```
|move|p2b: Blacephalon|Mind Blown|p1a: Cradily|[spread] p2a: Marowak,p1a: Cradily,p1: Thundurus
|-damage|p2a: Marowak|107\/243
|-damage|p1a: Cradily|40\/269
|-damage|p1b: Thundurus|0 fnt
|-damage|p2b: Blacephalon|0 fnt|[from] Mind Blown
```

Well, there's your first problem.

Zarel:
Apparently the problem is the `[spread]` tag's output was changed at some point.

`[spread] p2a: Marowak,p1a: Cradily,p1: Thundurus`

should have been

`[spread] p2a,p1a,p1b`

Hence the search for `p2a: ?`, which matches any Pokémon in the `p2a` slot.

Zarel:
@MacChaeger do you know anything abou this?

asgdf:
Ohhh if it's just supposed to be the slot number there, then I can see what the :? is for. So this would purely be a serverside issue, then.

Zarel:
https://replay.pokemonshowdown.com/smogondoubles-75588440

See how this goes:

```
|move|p2b: BUT FOR YOOOUUU|Earthquake|p1b: Gliscor|[spread] p2a,p1a,p1b
```

Zarel:
(Found the cause in https://github.com/Zarel/Pokemon-Showdown/pull/5216)

PR #1268, opened 2019-04-19
Refactor ability display in tooltips

asgdf:
I got reports about tooltips becoming kind of wonky when abilities were being suppressed (during my tests, it didn't work at all when viewing the player's own mons). In the process of trying to figure out why, I extracted the "which ability does this have?" logic and the "how do we display this abilty?" logic into separate functions, and changed it to fall back to server pokemon data more readily in case we're missing some information. This fixes some of the issues with suppression not being displayed properly. 
Is declaring an interface like this the cleanest option? I just wanted some ad-hoc data structure so I could separate the logic.
`abilityObj` is also kind of a lame name, but I wanted something that isn't too long while making it clear that this isn't just "an ability" but some kind of container. I'm open to suggestions there.

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -133,12 +133,6 @@ class ModifiableValue {
 	}
 }
 
-interface AbilityData {
-	ability: string;
-	baseAbility: string;
-	possibilities: string[];
-}
-
 class BattleTooltips {
 	battle: Battle;
 
@@ -668,14 +662,15 @@ class BattleTooltips {
 		}
 
 		const supportsAbilities = this.battle.gen > 2 && !this.battle.tier.includes("Let's Go");
+
+		let abilityText = '';
 		if (supportsAbilities) {
-			text += this.getPokemonAbilityText(clientPokemon, serverPokemon);
+			abilityText = this.getPokemonAbilityText(clientPokemon, serverPokemon, isActive);
 		}
-		if (serverPokemon) {
-			if (serverPokemon.item) {
-				let itemName = Dex.getItem(serverPokemon.item).name;
-				text += '<p><small>Item:</small> ' + itemName + '</p>';
-			}
+
+		let itemText = '';
+		if (serverPokemon && serverPokemon.item) {
+			itemText = '<small>Item:</small> ' + Dex.getItem(serverPokemon.item).name;
 		} else if (clientPokemon) {
 			let item = '';
 			let itemEffect = clientPokemon.itemEffect || '';
@@ -687,8 +682,17 @@ class BattleTooltips {
 			}
 			if (pokemon.item) item = Dex.getItem(pokemon.item).name;
 			if (itemEffect) itemEffect = ' (' + itemEffect + ')';
-			if (item) text += '<p><small>Item:</small> ' + item + itemEffect + '</p>';
+			if (item) itemText = '<small>Item:</small> ' + item + itemEffect;
 		}
+
+		text += '<p>';
+		text += abilityText;
+		if (itemText) {
+			// ability/item on one line for your own switch tooltips, two lines everywhere else
+			text += (!isActive && serverPokemon ? ' / ' : '</p><p>');
+			text += itemText;
+		}
+		text += '</p>';
 
 		text += this.renderStats(clientPokemon, serverPokemon, !isActive);
 
@@ -1555,8 +1559,10 @@ class BattleTooltips {
 		}
 		return allyAbility;
 	}
-	getPokemonAbilityData(clientPokemon: Pokemon | null, serverPokemon: ServerPokemon | null | undefined): AbilityData {
-		const abilityData: AbilityData = {ability: '', baseAbility: '', possibilities: []};
+	getPokemonAbilityData(clientPokemon: Pokemon | null, serverPokemon: ServerPokemon | null | undefined) {
+		const abilityData: {ability: string, baseAbility: string, possibilities: string[]} = {
+			ability: '', baseAbility: '', possibilities: [],
+		};
 		if (clientPokemon) {
 			if (clientPokemon.ability) {
 				abilityData.ability = clientPokemon.ability || clientPokemon.baseAbility;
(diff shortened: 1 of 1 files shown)


Zarel:
This pull request **introduces 1 alert** when merging 3e44ab25613ca083c8a5f2c71be58bda97f865a2 into 4d2f86be7dbf79f75513d2180c897cf8f0d8d09f - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown-Client/rev/pr-d2434b96ba0e9dc1a774825f87a7e2cdd6e04932)

**new alerts:**

---

scheibo:
> `abilityObj` is also kind of a lame name, but I wanted something that isn't too long while making it clear that this isn't just "an ability" but some kind of container. I'm open to suggestions there.

Given `abilityObj: TooltipAbility`, why is `abilityObj` not just named `tooltipAbility`?

asgdf:
It's partly "I only added that template at the very end and forgot about re-using that name".
But it's also a little bit that I feel that name works better for a class/interface than for a variable name, because variable names often have context nouns at the front, so it still sounds more like "a singular ability which is used for tooltips" than "some tooltip-specific ability data structure" to me. Not sure if that makes sense.

scheibo:
`abilityData` is probably the preferable name to `abilityObj`, as it would mirror the pattern in the server code (`effect`/`effectData`, `status`/`statusData`, `weather`/`weatherData` etc - the first variable is just the ID, the second is the actual data).

That being said, I think `tooltipAbility` is still the obvious name for a variable implementing the `TooltipAbility` interface. I'd also consider renaming `getPokemonAbility` to `getPokemonTooltipAbility`, because once again, it's returning a `TooltipAbility`. (though you could also choose to name `TooltipAbility` -> `AbilityData` and rename things to match that instead)

asgdf:
Oooh, `AbilityData`. I like that name. I'll use that.

Zarel:
At least in the server, `abilityData` etc should probably be renamde `abilityState` etc.

src/battle-tooltips.ts
```diff
@@ -663,41 +668,15 @@ class BattleTooltips {
...
-				if (ability && (ability !== abilityText)) {
-					abilityText = ability + ' (base: ' + abilityText + ')';
-				}
-				text += '<p><small>Ability:</small> ' + abilityText;
-				if (serverPokemon.item) {
-					text += ' / <small>Item:</small> ' + Dex.getItem(serverPokemon.item).name;
-				}
-				text += '</p>';
-			} else if (serverPokemon.item) {
+			if (serverPokemon.item) {
```
Zarel: but but `Ability / Item` on one line...

Zarel: (Which I mainly want for switch tooltips...)

src/battle-tooltips.ts
```diff
@@ -133,6 +133,12 @@ class ModifiableValue {
 	}
 }
 
+interface AbilityData {
```
Zarel: I honestly wouldn't bother naming an interface that's used in exactly one function's return value, though.

scheibo: Good point. I dont even think you need to write the interface at all? TS should be able to infer it based on the `abilityData` variable shape, so you dont even need to write out the type anywhere.

asgdf:
Ok I removed that interface, and also made it so your own mons are back to Ability / Item. The latter makes it a little bit more ugly, but I guess there's no way around it unless I include the item as well in the extracted function.

src/battle-tooltips.ts
```diff
@@ -663,41 +662,17 @@ class BattleTooltips {
...
-				if (serverPokemon.item) {
-					text += ' / <small>Item:</small> ' + Dex.getItem(serverPokemon.item).name;
-				}
-				text += '</p>';
-			} else if (serverPokemon.item) {
+			if (serverPokemon.item) {
 				let itemName = Dex.getItem(serverPokemon.item).name;
-				text += '<p><small>Item:</small> ' + itemName + '</p>';
+				// ability/item on one line for your own tooltips
+				text += '/ <small>Item:</small> ' + itemName;
```
Zarel: Non-base abilities shouldn't be shown on the switch list, probably. And this shouldn't be a one-liner if you have an active Pokémon... I would detect `isActive`.

Like, if `isActive` is false, don't call `getPokemonAbilityText` at all, and put the ability/item on the same line. Otherwise, do call `getPokemonAbilityText` and put the item on a separate line.

asgdf: So the behavior should be changed from what it currently does? Because currently it shows ability/item in one line even for the active mon.

Non-base abilities already don't show up in the switch list, not because of anything this function is doing but because the temporary ability changes are getting properly cleaned up on switchout I imagine.
I'd prefer to avoid a situation where we have a dedicated function for something and then we bypass it sometimes.

Zarel: Yeah, I wouldn't mind changing the behavior slightly.

Also: I was referring to non-base abilities for the active Pokémon in the switch list.

asgdf: Is there a way to distinguish between "switch tooltip for active pokemon" and "battle mouseover tooltip of active pokemon"? I was under the impression that there wasn't right now. Because the latter definitely needs to show the non-base ability.
But I also don't really understand what the point of this is in the first place. The base ability is being displayed in parens already, what's to gain from hiding the non-base one? Do people get confused over this?

Zarel: > Is there a way to distinguish between "switch tooltip for active pokemon"

Yes, `isActive` will be `false` in this case.

> and "battle mouseover tooltip of active pokemon"?

And `isActive` will be `true` in this case.

> But I also don't really understand what the point of this is in the first place. The base ability is being displayed in parens already, what's to gain from hiding the non-base one? Do people get confused over this?

Showing only the base ability makes it consistent with other Pokémon in the switch menu. It's mostly to keep switch menu tooltips consistent, which should make them easier to read.

asgdf:
Okay I believe I got it now. Without bypassing the function I wrote specifically for that purpose!

Zarel:
I'm not sure this is ideal, but I'm also not sure it's not, and in any case the main part of this PR is a huge improvement, so I'll merge this.

PR #1269, opened 2019-04-22
Fix guests not showing in options

TheJetOU:
[Reported here](https://www.smogon.com/forums/posts/8107508). I'm not sure why updateuser has it's first  element empty when it's a guest? Maybe figuring that out is the best solution.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -937,6 +937,7 @@
 					this.receive(data.substr(nlIndex + 1));
 					parts = data.substr(0, nlIndex).split('|');
 				}
+				if (!parts[0]) parts.shift();
 				var name = parts[1];
 				var named = !!+parts[2];
 
--- a/js/client.js
+++ b/js/client.js
@@ -935,9 +935,8 @@
 				var nlIndex = data.indexOf('\n');
 				if (nlIndex > 0) {
 					this.receive(data.substr(nlIndex + 1));
-					parts = data.substr(0, nlIndex).split('|');
+					parts = data.substr(1, nlIndex).split('|');
 				}
-				if (!parts[0]) parts.shift();
 				var name = parts[1];
 				var named = !!+parts[2];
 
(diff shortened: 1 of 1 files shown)


scheibo:
Maybe related to https://github.com/Zarel/Pokemon-Showdown-Client/pull/1263/files#diff-096f4b2e17f7b47c7ed33370746751b5R951?

TheJetOU:
@scheibo, the problem was that you weren't removing the first letter which is '|', thus causing an empty element.

js/client.js
```diff
@@ -935,7 +935,7 @@
 				var nlIndex = data.indexOf('\n');
 				if (nlIndex > 0) {
 					this.receive(data.substr(nlIndex + 1));
-					parts = data.substr(0, nlIndex).split('|');
+					parts = data.substr(1, nlIndex).split('|');
```
```suggestion
					parts = data.slice(1, nlIndex).split('|');
```

Remember, `substr(start, length)`, `slice(start, end)`

TheJetOU:
Why does the client use `substr` instead of `substring`? Isn't `substr` a legacy function? `substring` also seems to [support every browser](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#Browser_compatibility).

js/client.js
TheJetOU: They technically mean the same thing here, but yeah `slice` is clearer`.

Zarel:
When I started programming PS, it was a different time, and no one had strong opinions of which of `substr`, `substring`, and `slice` were best. I used `substr` because I come from a C++ background so that one was most familiar to me.

These days, I've mostly switched to `slice`, but there's a lot of old code in PS.

js/client.js
Zarel: No, they don't mean the same thing, that's why I said **length** vs **end**.

Zarel:
Also, you might not realize this, but "`substr` is a legacy function" is entirely a subjective sentence. It's not a _deprecated_ function, it's just a function some people think you shouldn't use. No one ever told me I shouldn't use it when I started programming PS, so I didn't know.

Zarel:
(Also MDN was a pretty bad programming resource back when I first started working on PS – it's always been the wrong resource if you care about browser support. It's better but still not great about browser support information, but there aren't really any alternatives.)

Issue #1264, opened 2019-04-09
Pokemon Showdown should support access through arbitrary third party clients


scheibo:
> Something like OAuth might also work.

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown-Client/pull/1218_

Currently, if someone wants to play on PS! with a modded client they can't - they can either attempt to get the code submitted into master (eg. #1218), create an elaborate browser extension, or run their own login server which doesn't interop with the existing PS! login system (so they're arguably not playing on PS! at that point, they're playing in their own parallel system). I believe the login flow could be modified so that users of third party clients could be redirected to PS! to authorize (to avoid leaking the credentials), but then continue to play with the token received on the custom client.

I think OAuth is probably the current best practice here, though its non trivial to implement correctly and has flaws.

Arguably not a client bug or a server bug, because the login server is its own thing, but this is relevant to clients so I'm filing this here to hopefully start a discussion. :)

Slayer95:
Isn't the current implementation enough so long as all third-party clients are enabled for CORS?

scheibo:
> Isn't the current implementation enough so long as all third-party clients are enabled for CORS?

What does this entail, some sort of whitelisting? What would I need to do to be able to serve my own client at `play.pkmn.cc` that can still auth against the official login server, for example (I don't actually have plans of doing anything special client yet, I'm just offering that as a strawman to see if its possible)? Or, more concretely, how can we let Digimon Showdown host its own client? If it is in fact possible today, can we add information to the `README.md` about it?

Slayer95:
The current implementation is a [https://github.com/Zarel/Pokemon-Showdown-Client/blob/5bb7326b4320cba0104868756fa91f6d3440fd8b/config/config-example.inc.php#L26-L33](whitelist), but it could be changed to ``Access-Control-Allow-Origin: *``. With that, you only need to serve ``index.html`` for https://github.com/Zarel/Pokemon-Showdown-Client/, and update the domain settings in ``js/storage.js``.

- Each third party client would get an exclusive ``HttpOnly`` ``sid`` cookie which is a hash of a string including the origin domain.
- The ``sid`` cookie is then sent to the login server in the ``?act=upkeep`` request, and the response is the string that needs to be sent in ``/trn`` for the PS Verifier service to validate it.
- That's all!

[...]

PR #1270, opened 2019-04-22
Handle AFK '@-in-names' protocol (take 2)

scheibo:
Rollforward of #1265. **Currently this is unchanged from what was committed and then rolled back** (thus why this is a draft), but I will be going through everything in [`git grep users`](https://gist.github.com/scheibo/9b74f47f1299778b68b7075f9308b011) to avoid breaking things.

### DIFF
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -90,20 +90,23 @@ class BattleLog {
 			}
 			break;
 
-		case 'join': case 'j':
-			divHTML = '<small>' + BattleLog.escapeHTML(args[1]) + ' joined.</small>';
+		case 'join': case 'j': {
+			const user = BattleTextParser.parseNameParts(args[1]);
+			divHTML = '<small>' + BattleLog.escapeHTML(user.name) + ' joined.</small>';
 			break;
-
-		case 'leave': case 'l':
-			divHTML = '<small>' + BattleLog.escapeHTML(args[1]) + ' left.</small>';
+		}
+		case 'leave': case 'l': {
+			const user = BattleTextParser.parseNameParts(args[1]);
+			divHTML = '<small>' + BattleLog.escapeHTML(user.name) + ' left.</small>';
 			break;
-
-		case 'name': case 'n':
-			if (toId(args[2]) !== toId(args[1])) {
-				divHTML = '<small>' + BattleLog.escapeHTML(args[1]) + ' renamed from ' + BattleLog.escapeHTML(args[2]) + '.</small>';
+		}
+		case 'name': case 'n': {
+			const user = BattleTextParser.parseNameParts(args[1]);
+			if (toId(args[2]) !== toId(user.name)) {
+				divHTML = '<small>' + BattleLog.escapeHTML(user.name) + ' renamed from ' + BattleLog.escapeHTML(args[2]) + '.</small>';
 			}
 			break;
-
+		}
 		case 'chatmsg': case '':
 			divHTML = BattleLog.escapeHTML(args[1]);
 			break;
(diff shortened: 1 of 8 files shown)


scheibo:
Went through everything in `git grep users` and I think 7ed7e9f covers what was missed (but I'm not a compiler and I can't really guarantee that). Little surprised to see code duplicated in `battle.ts` which is present in `client-chat.js` as well?

I've confirmed in the test client that tab completion of usernames now works. Are the additional things I should be testing? Does someone else want to patch this in to help provide some extra assurance it won't break things this time around?

scheibo:
@Zarel

Zarel:
Is the new `|j|` etc handled correctly in `BattleLog`?

scheibo:
> Is the new `|j|` etc handled correctly in `BattleLog` and `BattleTextParser`?

I don't see any `|j|` etc in `BattleTextParser`? I had missed `BattleLog`, but I've now updated it and `Battle` which were also missing in the original PR.

Zarel:
Have you played through an entire battle on testclient, with and without `@` in your name, and confirmed that it works?

Zarel:
(If so, feel free to merge.)

scheibo:
> (People in the Staff room are pretty willing to help devs test this sort of thing.)

Doesn't this require me to set up a server + client to allow other people to access it? And they still have to go through a hacky login flow because #1264 has not been resolved?

Easier to just play out battles myself TBH... (I don't have any Pokémon code or even Node for that matter on any of my servers yet...). It'd be great to have a staging environment where we could test out client + server features without pushing to main, but I think a proper set up for that is once again blocked on #1264.

> (Once you have, feel free to merge.)

I'm not a collaborator on this repro, but if I were **I would be adding a 'Do Not Merge' label currently**. I found one bug after playing through a battle (fixed with d701d93), and there's currently another issue where if you are /away and refresh you end up in a mixed state where parts of the UI think your /away and other parts don't (I think due to which UI elements listen to which things to know when to update). Once I've fixed that and gone through another round of poking around looking for edge cases I'll let you know.

scheibo:
> and there's currently another issue where if you are /away and refresh you end up in a mixed state where parts of the UI think your /away and other parts don't

I'm mistaken about this. The 'bad UI state' is really a combination of two things, neither which are caused by this PR.

1. After refreshing chat in the test client, the user list shows '0 users; Only Guests' even though the account youre testing with is not a guest. I've confirmed that on `master` this is also a thing.
2. I was worried about the 'UI being in an inconsistent state' because if you have 2 accounts in different tabs open and refresh one of the tabs (you need 2, because if you refresh with just one account open the userlist gets bugged and shows noone as described in 1), the user list shows the away account as not being away (ie. not greyed out), even though clicking on the account shows the user as away. When one account refreshes, the other window just gets a 
```
|l| pre
|j| pre
```
not any `|n|` info etc. I believe this is a bug in the server change - leaving/joining seems like it should clear your away state (as should actively making moves in a Pokemon battle - during my testing I noticed that doesn't affect away state either...). I will work with @bumbadadabum to fix the server code, but the client code seems ready to merge as is.

Zarel:
> Doesn't this require me to set up a server + client to allow other people to access it? And they still have to go through a hacky login flow because #1264 has not been resolved?

I mean the test would be: Battle on Main, staff member on Main, you on testclient. Bugs that only crop up with the test server aren't important; critical bugs are bugs that affect Main like the ones we hit last time.

Zarel:
Once you've tested that, comment and I'll merge immediately.

scheibo:
Oh yeah, I'd tested both against the main server and against a local server already, but I just did it again to be sure:

Screen Shot 2019-05-03 at 19 08 43

Poked around the Lobby etc afterwards on the test client and no crashes. I'm significantly more confident in this change than last time, but still find it hard to promise much in an untyped codebase without tests.

src/battle-log.ts
```diff
@@ -90,20 +90,23 @@ class BattleLog {
...
 			break;
-
-		case 'name': case 'n':
-			if (toId(args[2]) !== toId(args[1])) {
-				divHTML = '<small>' + BattleLog.escapeHTML(args[1]) + ' renamed from ' + BattleLog.escapeHTML(args[2]) + '.</small>';
+		}
+		case 'name': case 'n': {
+			const user = BattleTextParser.parseNameParts(args[1]);
+			if (toId(args[2]) !== toId(user.name)) {
+				divHTML = '<small>' + BattleLog.escapeHTML(user.name) + ' renamed from ' + BattleLog.escapeHTML(args[2]) + '.</small>';
```
scheibo: I should note - this is going to look kind of weird once the server code lands, as `args[2]` is the ID, which shouldn't have actually changed, so its going to be `'Foo renamed from foo'` when `Foo@!Away -> Foo`.

Zarel: This should probably be a different message in that case; something like `Foo changed status to ...`

PR #1272, opened 2019-05-07
Support |groups| protocol

Kaiepi:


Zarel:
Is this the entire thing? The only changes I see are `|groups|` instead of `|customgroups|`, and `"Regular user"` instead of `""`.

Kaiepi:
Yes. I didn't remove the default assignment to `Config.groups` so compatibility with servers that don't use `|groups|` yet would be kept. I don't know of any other changes needed, but I'm not very familiar with the client

Zarel:
I don't understand the need for either of those changes.

Kaiepi:
`|customgroups|` as a name doesn't make sense anymore since the message is always sent on connect regardless of whether there are any custom groups. `"Regular User"` is there alongside `null` because that's what the name of the default group is when using the newer `Config.grouplist` on the server

Zarel:
What happens if someone renames the default group?

Kaiepi:
Fair point. I'll add `default` as another group type, since nothing in the client depends on regular users being normal type

Zarel:
I guess I don't have a problem with this, but calling this a new protocol is kind of an overstatement; it seems like minor tweaks to an existing protocol.

monsanto:
Old PR, I'm pretty sure the developer has moved on since.

Issue #939, opened 2017-06-10
Supporting custom sprites


panpawn:
I think this has been mentioned that it's planned, but I guess making an issue for it in case someone is looking for something to do might find this interesting.

Right now, if a server wanted to use custom sprites, they would have to host their own custom client.  Allowing servers to send something that has custom sprites for the client to display could be beneficial to some, and I know this has been requested on more than a few occasions.

Zarel:
Yeah, I've had ideas about doing this...

PR #1273, opened 2019-05-08
Digimon Showdown Crys for CSS Bypass

AlphaWind:
This was the only feature that was not able to be bypassed

### DIFF
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -37,6 +37,7 @@ class BattleScene {
 
 	/** Note: Not the actual generation of the battle, but the gen of the sprites/background */
 	gen = 7;
+	digi = false;
 	/** 1 = singles, 2 = doubles, 3 = triples */
 	activeCount = 1;
 
@@ -106,6 +107,7 @@ class BattleScene {
 		let numericId = 0;
 		if (battle.id) {
 			numericId = parseInt(battle.id.slice(battle.id.lastIndexOf('-') + 1), 10);
+			if (this.battle.id.includes('digimon')) this.digi = true;
 		}
 		if (!numericId) {
 			numericId = Math.floor(Math.random() * 1000000);
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -37,7 +37,7 @@ class BattleScene {
 
 	/** Note: Not the actual generation of the battle, but the gen of the sprites/background */
 	gen = 7;
-	mod: string = '';
+	mod = '';
 	/** 1 = singles, 2 = doubles, 3 = triples */
 	activeCount = 1;
 
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -37,7 +37,7 @@ class BattleScene {
 
 	/** Note: Not the actual generation of the battle, but the gen of the sprites/background */
 	gen = 7;
-	digi = false;
+	mod: string;
 	/** 1 = singles, 2 = doubles, 3 = triples */
 	activeCount = 1;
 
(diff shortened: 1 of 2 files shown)


src/battle-dex.ts
```diff
@@ -528,6 +528,12 @@ const Dex = new class implements ModdedDex {
 			return spriteData;
 		}
 
+		// Digimon Crys
```
scheibo: Cries

src/battle-dex.ts
```diff
@@ -528,6 +528,12 @@ const Dex = new class implements ModdedDex {
...
+		if (options.digi) {
+		spriteData.cryurl = 'sprites/digimon/audio/' + toId(template.baseSpecies);
+		spriteData.cryurl += (window.nodewebkit ? '.ogg' : '.mp3');
+		}
```
TheJetOU: This should be indented (tab)

src/battle-dex.ts
```diff
@@ -528,6 +528,12 @@ const Dex = new class implements ModdedDex {
 			return spriteData;
 		}
 
+		// Digimon Crys
+		if (options.digi) {
```
scheibo: Why not instead make this generic, like `options.mod` and make it a `mod?: string`, and set it to `'digimon'` for digi. Then below, do `sprites/${options.mod}/audio` instead, so another mod might be able to use the same code.

src/battle-animations.ts
```diff
@@ -106,6 +107,7 @@ class BattleScene {
 		let numericId = 0;
 		if (battle.id) {
 			numericId = parseInt(battle.id.slice(battle.id.lastIndexOf('-') + 1), 10);
+			if (this.battle.id.includes('digimon')) this.digi = true;
```
scheibo: See below - I think you can make this more generic.

AshleyPikachu: I see, Slayer or Fender set up that code, it's very stable so we do not wish to change it, for example when our test client fell apart that still worked perfectly. If you would like to change it you have my permission to do so.

scheibo: This change is all of 9 lines, it really should be trivial at this stage to change and I would be very concerned if you were unable to at this point.

I wouldn't merge this as is, but maybe Zarel won't care because he just wants to be done with this.

scheibo:
Seems like this PR could help with #939 if done more thoughtfully.

src/battle-animations.ts
AshleyPikachu: your more then welcome to change it but we are not doing it, we want to be done just as much as zarel.

Zarel:
Yeah, if this could be just slightly more generic, that'd be nice. Like I said elsewhere, you can put whatever you want in your own namespaced folder, but if you're going to modify my files, it should at least be generic.

src/battle-animations.ts
```diff
@@ -37,6 +37,7 @@ class BattleScene {
 
 	/** Note: Not the actual generation of the battle, but the gen of the sprites/background */
 	gen = 7;
+	mod: string = '';
```
scheibo: The compiler should be able to infer `string` from the initializer?

src/battle-dex.ts
```diff
@@ -438,7 +438,7 @@ const Dex = new class implements ModdedDex {
 		document.getElementsByTagName('body')[0].appendChild(el);
 	}
 	getSpriteData(pokemon: Pokemon | Template | string, siden: number, options: {
-		gen?: number, shiny?: boolean, gender?: GenderName, afd?: boolean, noScale?: boolean,
+		gen?: number, shiny?: boolean, gender?: GenderName, afd?: boolean, noScale?: boolean, mod?: boolean,
```
scheibo:
```suggestion
		gen?: number, shiny?: boolean, gender?: GenderName, afd?: boolean, noScale?: boolean, mod?: string,
```

src/battle-dex.ts
```diff
@@ -528,6 +528,12 @@ const Dex = new class implements ModdedDex {
 			return spriteData;
 		}
 
+		// Mod Cries
+		if (options.mod) {
+			spriteData.cryurl = 'sprites/' + options.mod + '/audio/' + toId(template.baseSpecies);
```
scheibo:
```suggestion
			spriteData.cryurl = `sprites/${options.mod}/audio/${toId(template.baseSpecies)}`;
```

PR #1277, opened 2019-05-10
Fix modified stat display for Spc

arandompieceofcheese:
This fixes a big where tooltips don't show stat changes to Special in gen 1 games.

`serverPokemon.stats` uses `spa` to store the special stat, but `clientPokemon.boosts` uses `spc`.
`Dex.statNamesExceptHP` also does not include `spc`, so the original iteration would have skipped it anyway.


Zarel:
The reason we use `statNamesExceptHP` is because TypeScript, which you're running afoul of here.

Zarel:
If I could get a replay, I could probably test this...

PR #1281, opened 2019-05-13
Don't disable 'Chat' button for self

scheibo:
> chat button not being grayed out when you click on your own name, or make it more obvious people can pm themselves

_'More common Suggestions' from [Suggestions for Pokémon Showdown](https://www.smogon.com/forums/threads/suggestions-for-pok%C3%A9mon-showdown.3534365/)_

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -2520,10 +2520,13 @@
 
 			buf += '<p class="buttonbar">';
 			if (userid === app.user.get('userid') || !app.user.get('named')) {
-				buf += '<button disabled>Challenge</button> <button disabled>Chat</button>';
+				buf += '<button disabled>Challenge</button> ';
 				if (userid === app.user.get('userid')) {
+					buf += '<button name="pm">Chat</button>';
 					buf += '</p><hr /><p class="buttonbar" style="text-align: right">';
 					buf += '<button name="login"><i class="fa fa-pencil"></i> Change name</button> <button name="logout"><i class="fa fa-power-off"></i> Log out</button>';
+				} else {
+					buf += '<button disabled>Chat</button>';
 				}
 			} else {
 				buf += '<button name="challenge">Challenge</button> <button name="pm">Chat</button> <button name="userOptions">\u2026</button>';
--- a/js/client.js
+++ b/js/client.js
@@ -2522,7 +2522,7 @@
 			if (userid === app.user.get('userid') || !app.user.get('named')) {
 				buf += '<button disabled>Challenge</button> ';
 				if (userid === app.user.get('userid')) {
-					buf += '<button name="pm">Chat</button>';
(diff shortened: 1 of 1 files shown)


Zarel:
I think I'd prefer it with a different label, like "Chat with self" or "Test chat"

Zarel:
Maybe even add it to the More menu.

scheibo:
> I think I'd prefer it with a different label, like "Chat with self" or "Test chat"

This seems fine to me.

> Maybe even add it to the More menu.

This seems to defeat the request of 'making it more obvious people can PM themselves'.

Slayer95:
Any of these options empowers users to *easily* figure out whether they are shadowbanned (and explore|through) implementation loopholes.

On the other hand, the benefit is not clear, and arguably just a "feature request for the sake of the feature request".

scheibo:
**I'm 100% fine with dropping this.** I wasn't aware of the logic for not having it, and I don't particularly care about the feature, I just saw it as low hanging fruit on the 'More common suggestions list' that was negligible work to implement. However, I'm not sure I buy that it opens up loopholes - any user clever enough to be using it to detect whether theyre shadowbanned/which filters they can avoid is surely clever enough to figure out they can get the chat window through `/msg <self>,<foo>` as well? Or am I missing something?

I'd rather we either remove self messaging entirely or surface it properly. [Security through obscurity](https://en.wikipedia.org/wiki/Security_through_obscurity) is kind of a lame middleground, IMO. I do appreciate you explaining your reasoning though - a thumbs down is a lot less productive than a discussion :)

Slayer95:
My reasoning is not really a "security" issue though, and I 100% agree that security through obscurity is bad.

Rather, this is about making it harder for users to engage in socially unacceptable behavior. Influencing people's psychology is not a matter of absolute safety, but a scale in which incentives and obstacles do matter.

Now, you claim that the difficulty is the same. However, I disagree. To start with, one of the methods is reliant only on the GUI. If you even put a "Test chat" label there, that's even more of a hint for its fitness towards an evil purpose. Furthermore, it's not obvious how the command ``/msg`` interacts with embedded commands, or whether ``/msg name, /announce x`` actually parses the "announce" command in fact.

Slayer95:
Overall, I see self-chat as a development feature. Debug features shouldn't be visible to end users, am I right?

scheibo:
I don't even know why I'm arguing with you :D

Zarel:
I'm confused why this would expose implementation loopholes in shadowbanning? In theory we can just not have shadowbanning affect self-messages at all?

Zarel:
The reason I suggested having it in the "More" menu was because it _is_ a development feature and I don't want it to be confused for a feature having an actual effect.

PR #1282, opened 2019-05-14
Make <<replay>> link to the replay.pokemonshowdown.com

scheibo:
> make <<replay>> link to the replays.pokemonshowdown.com

_'More common Suggestions' from [Suggestions for Pokémon Showdown](https://www.smogon.com/forums/threads/suggestions-for-pok%C3%A9mon-showdown.3534365/)_


Zarel:
You also need to update the list in `client.js`.

Honestly, those should probably both be autogenerated.

scheibo:
> You also need to update the list in `client.js`.

Done.

> Honestly, those should probably both be autogenerated.

Not disagreeing, just not willing to take that on :)

PR #1284, opened 2019-05-17
Sort rooms in preact client

TheJetOU:
The Preact client code looks really clean and readable, I like it :)

### DIFF
--- a/src/panel-rooms.tsx
+++ b/src/panel-rooms.tsx
@@ -166,7 +166,8 @@ class RoomsPanel extends PSRoomPanel {
 	}
 	renderRoomList(title: string, rooms?: RoomInfo[]) {
 		if (!rooms || !rooms.length) return null;
-		const sortedRooms = rooms.sort((a, b) => /* Ascending order */ (b.userCount || 0) - (a.userCount || 0));
+		// Ascending order
+		const sortedRooms = rooms.sort((a, b) => (b.userCount || 0) - (a.userCount || 0));
 		return <div class="roomlist">
 			<h2>{title}</h2>
 			{sortedRooms.map(roomInfo => <div>
--- a/src/panel-rooms.tsx
+++ b/src/panel-rooms.tsx
@@ -166,10 +166,7 @@ class RoomsPanel extends PSRoomPanel {
 	}
 	renderRoomList(title: string, rooms?: RoomInfo[]) {
 		if (!rooms || !rooms.length) return null;
-		const sortedRooms = rooms.sort((a, b) => {
-			// ascending order
-			return (b.userCount || 0) - (a.userCount || 0);
-		});
+		const sortedRooms = rooms.sort((a, b) => /* Ascending order */ (b.userCount || 0) - (a.userCount || 0));
 		return <div class="roomlist">
 			<h2>{title}</h2>
 			{sortedRooms.map(roomInfo => <div>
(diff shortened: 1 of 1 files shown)


src/panel-rooms.tsx
```diff
@@ -166,9 +166,14 @@ class RoomsPanel extends PSRoomPanel {
 	}
 	renderRoomList(title: string, rooms?: RoomInfo[]) {
 		if (!rooms || !rooms.length) return null;
+		const sortedRooms = rooms.sort((a, b) => {
+			if (!b.userCount || !a.userCount) return 0;
```
Zarel: ?

Zarel: Defaulting to `0` seems a better plan than making a comparison function that doesn't uphold global comparison guarantees.

src/panel-rooms.tsx
```diff
@@ -166,9 +166,14 @@ class RoomsPanel extends PSRoomPanel {
...
+			// ascending order
+			return b.userCount - a.userCount;
```
Zarel:
```suggestion
			return (b.userCount || 0) - (a.userCount || 0);
```

src/panel-rooms.tsx
Zarel:
```suggestion
```

TheJetOU: Because `userCount` can be a false-y value as per the definition https://github.com/Zarel/Pokemon-Showdown-Client/blob/master/src/panel-mainmenu.tsx#L8

You also can't return undefined, so I returned zero because that's the number that most makes sense in this case.

src/panel-rooms.tsx
TheJetOU: If we're going this direction, isn't it better if we one-liner it?

scheibo: Yes. But the suggested changes API is limited. Go ahead and remove the `return`

TheJetOU: Did.

scheibo: I would move the comment above the variable declaration instead of doing it inline.

src/panel-rooms.tsx
```diff
@@ -166,9 +166,10 @@ class RoomsPanel extends PSRoomPanel {
 	}
 	renderRoomList(title: string, rooms?: RoomInfo[]) {
 		if (!rooms || !rooms.length) return null;
+		const sortedRooms = rooms.sort((a, b) => /* Ascending order */ (b.userCount || 0) - (a.userCount || 0));
```
Zarel: Hm, I think it looks better on multiple lines.

```suggestion
		const sortedRooms = rooms.sort((a, b) =>
			// Ascending order
			(b.userCount || 0) - (a.userCount || 0)
		);
```

Zarel: Possible alternative?

```suggestion
		// Ascending order
		const sortedRooms = rooms.sort((a, b) => (b.userCount || 0) - (a.userCount || 0));
```

Zarel: Also wait you mean "Descending order", right? Most to least users?

TheJetOU: I prefer the second one. I don't think it's long/complex enough to warrant more than one line.

> Also wait you mean "Descending order", right? Most to least users?

 I forgot how English works I guess, lol.

src/panel-rooms.tsx
TheJetOU: That is done.

PR #1285, opened 2019-05-21
Separate battle and sprite animation prefs

scheibo:
Users can still select BW sprites if they desire separately from whether or not they want battle animations.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -108,7 +108,8 @@
 			this.tooltips.hideTooltip();
 			if (this.battle.playbackState === 3 && !this.battlePaused) {
 				this.battle.play();
-				if (Dex.prefs('noanim')) this.battle.fastForwardTo(-1);
+				var animations = Dex.prefs('animations');
+				if (animations === 'none' || animations === 'sprites') this.battle.fastForwardTo(-1);
 			}
 			ConsoleRoom.prototype.focus.call(this);
 		},
@@ -215,7 +216,8 @@
 					this.battle.activityQueue.push(logLine);
 				}
 			}
-			this.battle.add('', Dex.prefs('noanim'));
+			var animations = Dex.prefs('animations');
+			this.battle.add('', animations === 'none' || animations === 'sprites');
 			this.updateControls();
 		},
 		toggleMessages: function (user) {
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -108,8 +108,7 @@
 			this.tooltips.hideTooltip();
 			if (this.battle.playbackState === 3 && !this.battlePaused) {
 				this.battle.play();
-				var animations = Dex.prefs('animations');
-				if (animations === 'none' || animations === 'sprites') this.battle.fastForwardTo(-1);
+				if (Dex.prefs('animations') === 'none') this.battle.fastForwardTo(-1);
 			}
 			ConsoleRoom.prototype.focus.call(this);
 		},
@@ -216,8 +215,7 @@
 					this.battle.activityQueue.push(logLine);
 				}
 			}
-			var animations = Dex.prefs('animations');
-			this.battle.add('', animations === 'none' || animations === 'sprites');
+			this.battle.add('', Dex.prefs('animations') === 'none');
 			this.updateControls();
 		},
 		toggleMessages: function (user) {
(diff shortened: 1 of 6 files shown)


Zarel:
BW sprites are different from still sprites – the default BW sprites are animated.

I don't think I approve of this. Maybe better to autogenerate some stills from the gen 6+ sprites, and use those.

Zarel:
But in lieu of that, "stop animations when someone turns off animations" is more important than "match art style". A user turning off animations doesn't want an animated GIF, either. GIF animation isn't appreciably different from other animation, from an end-user perspective.

scheibo:
> A user turning off animations doesn't want an animated GIF, either. GIF animation isn't appreciably different from other animation, from an end-user perspective.

Yeah, I disagree with this. People (I'm assuming its not just me) turn off animations to speed up play, having the GIF sprite bouncing around doesn't do anything to slow things down (arguably it uses up client resources, but theyre so negligible - its not a fixed delay like the battle animations add).

Furthermore, if I turn off the 'Battle Scene' setting on the cartridge, its only the moves that do reduced animations, the sprites are still animating (check out a [speedrun](https://www.youtube.com/watch?v=VUS3z5XyH8o) for an example, as all are going to use that setting)

Zarel:
> Yeah, I disagree with this. People (I'm assuming its not just me) turn off animations to speed up play, having the GIF sprite bouncing around doesn't do anything to slow things down (arguably it uses up client resources, but theyre so negligible - its not a fixed delay like the battle animations add).

GIFs are non-negligible resource-wise; they're the biggest bottleneck to client perf. If you use an old cell phone or Chrome 64, it's really noticeable – attack animations run fine but GIFs are consistently stuttery.

> Furthermore, if I turn off the 'Battle Scene' setting on the cartridge, its only the moves that do reduced animations, the sprites are still animating (check out a [speedrun](https://www.youtube.com/watch?v=VUS3z5XyH8o) for an example, as all are going to use that setting)

I've said this repeatedly; I never intended "no animation" to be a "speed up play" setting. If you want a "speed up play" setting, it should probably hook into the acceleration setting, instead of overloading the "no animation" setting.

Zarel:
I'm open to a dropdown:

- Regular animation
- Fast animation (no move animation)
- Absolutely no animation, not even GIFs

But the way you have it currently set up, I think it's unnecessarily confusing to users.

scheibo:
Thanks for the explanation, all very valid. I will add a dropdown for `Animation: None | Sprites | All`, where `None` and `All` are  what is provided by the current toggle (open to different descriptions of the various animation states, perhaps eg. `None | Just Sprites | Sprites and Moves` or whatever).

Zarel:
I think the DO NOT MERGE label is intentionally scary so people know not to accidentally merge it. Draft PRs don't need scariness because they already have an interlock – the merge button is greyed out and unclickable. The big red allcaps label is an important reminder that GitHub's bright green inviting merge button is a lie.

Zarel:
I also think the middle ground should have more than just sprite animations; it could have other non-blocking animations. I'm thinking result animations, too.

Zarel:
Or, seriously, the acceleration setting. Set acceleration to 3, it should be about as fast as you want, right?

scheibo:
> I also think the middle ground should have more than just sprite animations; it could have other non-blocking animations. I'm thinking result animations, too.

> Or, seriously, the acceleration setting. Set acceleration to 3, it should be about as fast as you want, right?

Will play around with these now to see what the effect is

scheibo:
OK, I'm not completely in love with `BattleScene.acceleration = 3` - things kind of feel rushed? And some animations seem disproportionately slower or faster than they should be (is `acceleration` applied correctly to all animations?) I think youre right though that *some* animation in addition to the sprite movement might be nice, but im not entirely sure what exactly, and just turning on eg `resultAnim` looks somewhat involved?

Zarel:
Acceleration is definitely custom-coded to apply differently to different animations, yes.

Zarel:
Acceleration is currently used to automatically accelerate repetitive double-switching. It might need tweaking to accommodate other use cases.

Zarel:
Since this, I've improved acceleration 3 for Blitz games. I'm still confused that you want no animation, but you're also complaining that fast animation is too rushed?

But either way, that would be a new PR, not this one.

scheibo:
>  I'm still confused that you want no animation, but you're also complaining that fast animation is too rushed?

Yes.

I agree with closing this PR, I'll open up a new one when I tackle this again.

PR #1287, opened 2019-05-29
Don't include the ! in away status display

scheibo:
https://github.com/Zarel/Pokemon-Showdown/pull/5002/commits/e373c1d026ef4f70a3ee25c133dacfbf0333ef93 will cause the rank to sometimes be erroneously displayed in the UI: 

Screen Shot 2019-05-27 at 15 40 42
Screen Shot 2019-05-27 at 15 41 15

Zarel/Pokemon-Showdown#5002 will not _break_ the client when it lands, but to preserve the current display semantics **this PR should be committed and pushed in coordination with the server restart that  deploys the protocol change**. I've marked this change as 'DO NOT MERGE' because if deployed without the server change it will result in lopping off the first character of usernames, which is clearly undesirable.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -939,7 +939,7 @@
 					this.receive(data.substr(nlIndex + 1));
 					parts = data.slice(1, nlIndex).split('|');
 				}
-				var parsed = BattleTextParser.parseNameParts(parts[1]);
+				var parsed = BattleTextParser.parseNameParts(parts[1].slice(1));
 				var named = !!+parts[2];
 
 				var userid = toUserid(parsed.name);
--- a/js/client.js
+++ b/js/client.js
@@ -939,7 +939,7 @@
 					this.receive(data.substr(nlIndex + 1));
 					parts = data.slice(1, nlIndex).split('|');
 				}
-				var parsed = BattleTextParser.parseNameParts(parts[1].slice(1));
+				var parsed = BattleTextParser.parseNameParts(parts[1]);
 				var named = !!+parts[2];
 
 				var userid = toUserid(parsed.name);
--- a/src/battle-text-parser.ts
+++ b/src/battle-text-parser.ts
@@ -58,6 +58,12 @@ class BattleTextParser {
 	}
 
 	static parseNameParts(text: string) {
+		let rank: string | undefined;
+		// Names cannot start with symbols, so if we do see a symbol then we know it is the user's rank.
+		if (text.match(/^A-Za-z0-9/)) {
+			rank = text.charAt(0);
+			text = text.slice(1);
+		}
 		const atIndex = text.indexOf('@');
 		let name = text;
 		let status = '';
@@ -70,7 +76,7 @@ class BattleTextParser {
 				status = status.substr(1);
 			}
 		}
-		return {name, status, away};
+		return {rank, name, status, away};
 	}
 
 	static upgradeArgs({args, kwArgs}: {args: Args, kwArgs: KWArgs}): {args: Args, kwArgs: KWArgs} {
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -383,9 +383,6 @@ class BattleLog {
 		if (this.colorCache[name]) return this.colorCache[name];
 		let hash;
 		if (window.Config && Config.customcolors && Config.customcolors[name]) {
-			if (Config.customcolors[name].color) {
-				return (this.colorCache[name] = Config.customcolors[name].color);
-			}
 			hash = MD5(Config.customcolors[name]);
 		} else {
 			hash = MD5(name);


Zarel:
We have more than one server, you know.

Zarel:
i.e. I have zero intention of merging in any backwards-incompatible client change.

Zarel:
Honestly, the rank should be a name part. Every other usage of `parseNameParts` also includes the rank symbol, after all. Just test to see if the first letter is alphanumeric.

scheibo:
> Just test to see if the first letter is alphanumeric.

Did not realize we already filtered those out.

scheibo:
[HAHAHA DISREGARD THAT...](http://bash.org/?5775)

yeah so this is a mess. i didn't realize that group/rank was being implicitly included in name sometimes and i need to go figure out which places to and do not want rank/basically unify everything. also, 82e7f1a is just terrible in so many ways

I'll look at this on my flight tomorrow, but man I'm probably just going to retreat into stats processing again because figuring out how to rip through GB of data >>>> figuring out how to unbreak the UI

Zarel:
> Since when? Rank is always prefixed seperately AFAICT.

Every other message from the server is `getIdentityWIthStatus` or whatever. Ideally, it should all be parsed the same way.

src/battle-text-parser.ts
```diff
@@ -70,7 +76,9 @@ class BattleTextParser {
 				status = status.substr(1);
 			}
 		}
-		return {name, status, away};
+		// TODO: stop prefixing name with group and update callers
```
scheibo: It is very challenging to trace through not just the immediate callers of `parseNameParts` but the things they pass 'name' to etc, such that actually updating `name` to be just the name part and not group + name is out of scope for this PR (we can followup with a cleanup change to make people use `group` and  instead of all the places we use `test`/`chartAt(0)`/`substr(1)` and what not, but I'd really like to not block landing statuses on this).

Zarel: I'm fine with that.

scheibo:
I think we want this to land before the server changes

src/battle-text-parser.ts
Zarel: Wait, why are you manually adding a group to this function at all?

scheibo: This function needs to understand that names can be prefixed with a group to be able to properly parse text like `'@pre@!(Away) coding'`. Because `@` can be the moderator rank and the symbol used for separating name and status, the parsing will break if we don't handle the possibility. 

We could handle this by changing line 67 to just be `const atIndex = text.indexOf('@', 1);` instead (ie/ make use of the `fromIndex` to avoid getting choked up by the leading `@`), which would be totally valid. However, I left the actual group parsing in because `parseNameParts` seems like it really should, you know, parse the parts of the name. I'm fine to just use from index and not mess around with group at all in this method, but I believe in the future we'd probably want this method to be able to handle all parsing as mentioned in the first of these comments, its just annoying to do at this time.

scheibo:
There are other parts to this PR which need to land outside of `parseNameParts` (and I'm not sure 8e9bc3fe1e739a92c8b25075a6a0479ffc56e0aa parses all names correctly?)

scheibo:
OK, I think this is what I'd still like to merge. I still worried 8e9bc3f will have broken things, but thats somewhat orthogonal to away status display

js/client.js
```diff
@@ -2472,7 +2472,8 @@
 			buf += '<strong><a href="//pokemonshowdown.com/users/' + userid + '" target="_blank">' + BattleLog.escapeHTML(name) + '</a></strong><br />';
 			var offline = data.rooms === false;
 			if (data.status || offline) {
-				buf += '<span class="userstatus' + (offline ? ' offline' : '') + '">' + (offline ? '(Offline)' : BattleLog.escapeHTML(data.status)) + '</span><br />';
+				var status = offline ? '(Offline)' : data.status.startsWith('!') ? data.status.slice(1) : data.status;
+				buf += '<span class="userstatus' + (offline ? ' offline' : '') + '">' + BattleLog.escapeHTML(status) + '</span><br />';
```
Zarel: I guess this is the main thing you'd like to merge? Sure, you should probably give your PR a new name, though.

Zarel: I don't really have much idea what this does, though.

Zarel:
(Also you have another conflict)

js/client.js
scheibo: I mean, once you pull out the other parts of my PR and commit them separately, yeah :). I dont know if the change in `client-topbar.js` is strictly required, but the Chrome inspect seems to like it better.  The material change in this file is effectively

`data.status` -> `data.status.startsWith('!') ? data.status.slice(1) : data.status`

ie/ 'Don't include the `!` in away status display'.

scheibo:
> (Also you have another conflict)

Hmmm... I wonder where that came from?

:D

PR #1288, opened 2019-06-03
Display base power of Hidden Power/Frustration/Return in Pokemon tooltip if not max

scheibo:
Won't do anything without https://github.com/Zarel/Pokemon-Showdown/pull/5519 which plumbs through happiness as part of the `|request|`.

### DIFF
--- a/src/battle-dex.ts
+++ b/src/battle-dex.ts
@@ -259,7 +259,7 @@ const Dex = new class implements ModdedDex {
 		return new PureEffect(id, name);
 	}
 
-	getMove(nameOrMove: string | Move | null | undefined): Move {
+	getMove(nameOrMove: string | Move | null | undefined, gen = 7): Move {
 		if (nameOrMove && typeof nameOrMove !== 'string') {
 			// TODO: don't accept Moves here
 			return nameOrMove;
@@ -278,7 +278,7 @@ const Dex = new class implements ModdedDex {
 			let [, hpWithType, hpPower] = /([a-z]*)([0-9]*)/.exec(id)!;
 			data = {
 				...(window.BattleMovedex[hpWithType] || {}),
-				basePower: Number(hpPower) || 60,
+				basePower: Number(hpPower) || (gen < 6 ? 70 : 60),
 			};
 		}
 		if (!data && id.substr(0, 6) === 'return' && id.length > 6) {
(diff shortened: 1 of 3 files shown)


src/battle-tooltips.ts
```diff
@@ -733,6 +734,14 @@ class BattleTooltips {
 		return text;
 	}
 
+	getMoveDisplay(move: Move, pokemon: ServerPokemon) {
+		if (((move.id === 'frustration' || move.id === 'return') && move.basePower !== 102) ||
+			(move.id.startsWith('hiddenpower') && this.battle.gen < 7 && move.basePower !== 70)) {
+			return `${move.name} - ${move.basePower}BP`;
```
Zarel:
```suggestion
			return `${move.name} ${move.basePower}`;
```

src/battle-tooltips.ts
```diff
@@ -733,6 +734,14 @@ class BattleTooltips {
 		return text;
 	}
 
+	getMoveDisplay(move: Move, pokemon: ServerPokemon) {
+		if (((move.id === 'frustration' || move.id === 'return') && move.basePower !== 102) ||
+			(move.id.startsWith('hiddenpower') && this.battle.gen < 7 && move.basePower !== 70)) {
```
Zarel: Does this correct display the type?

Zarel:
```
> Dex.mod('gen5').getMove('hiddenpowerflying59').name
"Hidden Power Flying"
> Dex.mod('gen5').getMove('hiddenpowerflying59').id
"hiddenpowerflying59"
```

Huh, apparently yes.

src/battle-tooltips.ts
scheibo: You like `Frustration 98 (29/32)` vs `Frustration - 98BP (29/32)` (this was one thing I brought up on the [thread](https://www.smogon.com/forums/threads/return-frustration-base-power-tooltip.3650730/))?

Zarel: Yeah `Frustration 98 (29/32)` was always how I would have implemented it, and it matches how variable base power moves are usually talked about.

https://www.google.com/search?q="hidden+power+70"

See how people just say "Hidden Power 70" when talking about it.

scheibo: Ack. Just wanted to make sure

scheibo:
So did some testing (btw, the teambuilder is super obnoxious about not wanting me to have non 0 or 255 happiness, but thats besides the point), needs more work

src/battle-dex.ts
```diff
@@ -278,7 +278,7 @@ const Dex = new class implements ModdedDex {
 			let [, hpWithType, hpPower] = /([a-z]*)([0-9]*)/.exec(id)!;
 			data = {
 				...(window.BattleMovedex[hpWithType] || {}),
-				basePower: Number(hpPower) || 60,
+				basePower: Number(hpPower) || (gen < 6 ? 70 : 60),
```
Zarel: In theory, the server should send the HP power if it's 70.

Zarel: Also there's no need to ask for `gen`, you can just test `this.gen`.

PR #1291, opened 2019-06-05
[WIP] Begin work to build on top of nwjs v0.13/v0.14

NotTsunami:
Begin to migrate as per http://docs.nwjs.io/en/latest/For%20Users/Migration/From%200.12%20to%200.13/

If anything, anyone more knowledgeable on nwjs can and should definitely take over. If we continue to stick with the NSIS install system, the script would need to be updated to match the file structure of nwjs v0.13 and beyond. This should also allow us to check in a clean SockJS 1.1.5 versus our modded 1.1.1.


whalemer:
Does this need to be done? The current client works fine for me on nwjs v0.38.4 stable (windows) and I'm pretty sure this breaks backwards compatibility.

NotTsunami:
Wait, really? I followed the instructions and the exe kept crashing.

whalemer:
What are you doing?

Clone -> unzip nwjs into /desktop -> run nw.exe works on my machine

NotTsunami:
Apparently, too many extra steps. Will close as that worked fine.

PR #1292, opened 2019-06-05
Upgrade ancient jquery-cookie to js-cookie 1.5.1

NotTsunami:
Let's modernize a tiny bit and upgrade jquery-cookie to js-cookie. Untested.


Zarel:
...why though?

NotTsunami:
Just trying to modernize a bit, feel free to close if you believe it's inappropriate!

Zarel:
jQuery Cookie is 40ish lines of easily readable and clear JavaScript which still works on modern browsers and has no known vulnerabilities.

js-cookie is 150ish lines of greater complexity and greater abstraction (which will result in worse performance) to add features we'll never use.

I just don't see a reason.

Zarel:
The new client naturally won't depend on jQuery Cookie, but I honestly don't think I'll be using a cookie library at all for that one.

PR #1293, opened 2019-06-07
Fix groundedness check

asgdf:
`Pokemon#isGrounded` was erroneously returning `false` in a lot of cases because of a flipped operator, which mostly showed itself in moves lacking their terrain boosts.

I suppose the choice between using `else if` and just chaining `if`s with early returns here is a matter of taste, I switched to the latter because that makes it consistent with the conditions immediately above, and the previous inconsistency was more confusing to read at a glance.


Slayer95:
Amazed this didn't have more visible effects. This is huge! Good catch @asgdf !

PR #1296, opened 2019-06-12
Teambuilder: Fix pokemon display for Doubles.

jmclemo6:
This fixes a regression introduced by 9114cb98945a12b2de01319294a47335dda0fa66. This was a quick and dirty fix, so there is a possibility it breaks again.

Licensing is covered by: https://github.com/Zarel/Pokemon-Showdown/pull/2853#issue-90513765

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -1321,11 +1321,11 @@
 		// buf += '<span class="col numcol">' + (pokemon.num >= 0 ? pokemon.num : 'CAP') + '</span> ';
 		var tier;
 		if (pokemon.tier) {
-			tier = Dex.getTier(pokemon, this.gen, this.isDoubles);
+			tier = Dex.getTier(Dex.forGen(this.gen).getTemplate(id), this.gen, this.isDoubles);
 		} else if (pokemon.forme && pokemon.forme.endsWith('Totem')) {
-			tier = Dex.getTier(BattlePokedex[toID(pokemon.species.slice(0, (pokemon.forme.startsWith('Alola') ? -6 : pokemon.baseSpecies.length + 1)))], this.gen, this.isDoubles);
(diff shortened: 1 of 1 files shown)


Zarel:
Can you not? We can't accept contributions that aren't MIT licensed.

Zarel:
Please retract your license and issue a new one for the MIT license, or I won't be able to accept your pull requests in the future.

src/battle-dex.ts
```diff
@@ -424,7 +424,7 @@ const Dex = new class implements ModdedDex {
 		}
 		if (!table) return pokemon.tier;
 
-		let id = pokemon.id;
+		let id = toID(pokemon.species);
```
Zarel: Can you clarify which pokemon for which `pokemon.id` doesn't match `toID(pokemon.species)`?

Slayer95:
> Please retract your license

No need to "retract it". The contributions can be dual-licensed AGPLv3 and MIT.
Still, the core idea is... please grant us a MIT License :)

src/battle-dex.ts
jmclemo6: `id` didn't exist on the pokemon template that was being passed to getTier. That was what was causing this problem. This is a quick fix to get the Double's teambuilder working again until I can investigate more and find out why.

jmclemo6:
Licensing is done.

jmclemo6:
Yeah, the problem here is that `BattlePokedex` doesn't include `id` in its entries. So, possible fixes that I can think of are:

1. Call toID on the species like I do in this change.
2. Add `id` to `BattlePokedex` entries.
3. Put the `id` onto the `pokemon` object within `Search.prototype.renderRow`

src/battle-dex.ts
scheibo: Revert?

Zarel: Oh, I see. Yeah, this works.

PR #1298, opened 2019-06-12
Fix bug in hiding tournaments

TheJetOU:


js/client-chat-tournament.js
```diff
@@ -245,6 +245,8 @@
 		};
 
 		TournamentBox.prototype.parseMessage = function (data, isBroadcast) {
+			var hide = Dex.prefs('tournaments') === 'hide';
```
scheibo: this can probably just be all on one line?

PR #1299, opened 2019-06-13
Make `getTemplate` able to override tiers for Doubles

jmclemo6:
This lets `getTemplate` handle overriding tiers for Doubles. Doing it this way means we can remove `getTier` entirely.

### DIFF
--- a/src/battle-dex.ts
+++ b/src/battle-dex.ts
@@ -367,7 +367,7 @@ const Dex = new class implements ModdedDex {
 		return ability;
 	}
 
-	getTemplate(nameOrTemplate: string | Template | null | undefined): Template {
+	getTemplate(nameOrTemplate: string | Template | null | undefined, isDoubles: boolean = false): Template {
 		if (nameOrTemplate && typeof nameOrTemplate !== 'string') {
 			// TODO: don't accept Templates here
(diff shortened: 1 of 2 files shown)


Zarel:
no :[

src/battle-dex.ts
```diff
@@ -367,7 +367,7 @@ const Dex = new class implements ModdedDex {
 		return ability;
 	}
 
-	getTemplate(nameOrTemplate: string | Template | null | undefined): Template {
+	getTemplate(nameOrTemplate: string | Template | null | undefined, isDoubles: boolean = false): Template {
```
Zarel: This signature doesn't match the signature in the server, and it's a digression I wouldn't want to maintain.

It would make more sense to support `gen7doubles` etc as a mod name internally. Another option would be to have separate `tier` and `doublesTier` properties on the Template object.

Zarel: Both are probably good ideas: `doublesTier` as a teambuilder-table property, and `gen7doubles` as a modid.

jmclemo6: Alright yeah. I had mentioned doing it those ways in the dev room as well. Will do when I get the chance! :+1:

PR #1304, opened 2019-06-17
Fix roomgroup in user popup

whalemer:
I don't //think// that making passed data override the cache should be an issue.

I don't really like the amount of cache invalidation that goes on, but short of hardcoding roomGroup to not cache I'm not sure what else to do.


Zarel:
Hardcoding roomGroup to not cache seems like the right option.

Alternatively, just accept that roomGroup is a part of name. :/ I'm still not super happy with breaking it out like it was.

PR #1305, opened 2019-06-17
Tooltips: Fix possible abilities display in old gens

asgdf:
1. Use the battle's modded dex instead of the default gen 7 one
2. Change the conditions in `getPokemonAbilityData()` so it's no longer possible for ability and baseAbility to both be set to the same value (which was happening in oldgens randbats for some reason).

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -1575,11 +1575,11 @@ class BattleTooltips {
 		if (clientPokemon) {
 			if (clientPokemon.ability) {
 				abilityData.ability = clientPokemon.ability || clientPokemon.baseAbility;
-				if (clientPokemon.baseAbility && clientPokemon.baseAbility !== clientPokemon.ability) {
+				if (clientPokemon.baseAbility && clientPokemon.baseAbility !== abilityData.ability) {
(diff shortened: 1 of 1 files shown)


Zarel:
Why shouldn't `ability` and `baseAbility` be set to the same value? That seems like what should happen if you both know the base ability and the current ability, and it hasn't been changed by Simple Beam et al?

asgdf:
The impression I had was that baseAbility was only supposed to be used if the ability was changed to something different, rather than always being set. 
I was aiming to avoid "Ability: Levitate (Base ability: Levitate)", which was how every single ability was displayed in gen 3 randbats for some reason. 
The alternative would have been implementing that check in the ability rendering code, and I thought it was neater to do it here.

Zarel:
Implementing that check in the ability rendering code seems to be the right call.

How else would you distinguish "we don't know what the base ability is" from "we know what the base ability is"?

asgdf:
But in which situation is "this pokemon had Levitate, but now it has Levitate" useful to distinguish from "this pokemon has Levitate"?

Zarel:
It is certainly useful to distinguish "This Pokémon has the ability it started with, which is Levitate" from "This Pokémon has Levitate, but we don't know what it started with".

PR #1308, opened 2019-06-20
Update PSPL thread URL

asgdf:
It links to last year's. This is this year's finals' thread.


HoeenCoder:
Already updated in #1306 

Also generally the link goes to the annoucement thread rather than the finals (which 1306 handles).

asgdf:
Ah. Well, that's the thread Kala pointed me towards.

PR #1309, opened 2019-06-20
Add analytics to track battle sound preferences

scheibo:
It seems like the mute preference (or any preference really) should be a [User-level scoped Custom Dimension](https://support.google.com/analytics/answer/2709828?hl=en) that gets updated when the pref gets updated (ie. set in `Storage.whenPrefsLoaded()` and when it's toggled), only this doesn't accomplish much because we need an event for the custom dimension to be sent with. We can't send this dimension with the `ga('send', 'pageview')` metric because that happens on page load, before we have prefs available (and also doesnt cover changes in the pref's value over the session). 

Instead, I've created a 'join battle' event (event category `'Battle'`, event action `'join'`, [per the documentation](https://developers.google.com/analytics/devguides/collection/analyticsjs/events#event_fields)), so that we could track what the sound preference was when the battle was joined. This doesn't cover the case where a user immediately mutes after joining a battle (something I often do), but I'm not sure what a good solution to that is.

**NOTE:** The custom dimension requires [additional setup in the Admin UI](https://support.google.com/analytics/answer/2709829).

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -47,6 +47,8 @@
 			this.battle.startCallback = function () { self.updateControls(); };
 			this.battle.stagnateCallback = function () { self.updateControls(); };
 
+			if (window.ga) ga('send', 'event', 'Battle', 'join');
+
 			this.battle.play();
 		},
 		events: {
--- a/js/client-topbar.js
+++ b/js/client-topbar.js
@@ -62,6 +62,7 @@
 			var muted = !Dex.prefs('mute');
 			Dex.prefs('mute', muted);
 			BattleSound.setMute(muted);
+			if (window.ga) ga('set', /* mute */ 'dimension1', muted);
 			app.topbar.$('button[name=openSounds]').html('<i class="' + (muted ? 'fa fa-volume-off' : 'fa fa-volume-up') + '"></i>');
 		},
 
@@ -404,6 +405,7 @@
 			var muted = !!e.currentTarget.checked;
 			Dex.prefs('mute', muted);
 			BattleSound.setMute(muted);
+			if (window.ga) ga('set', /* mute */ 'dimension1', muted);
 
 			if (!muted) {
 				this.$('.effect-volume').html('<label class="optlabel">Effect volume:</label><input type="range" min="0" max="100" step="1" name="effectvolume" value="' + (Dex.prefs('effectvolume') || 50) + '" />');
--- a/js/client.js
+++ b/js/client.js
@@ -451,6 +451,7 @@ function toId() {
 
 				var muted = Dex.prefs('mute');
 				BattleSound.setMute(muted);
+				if (window.ga) ga('set', /* mute */ 'dimension1', muted);
 
 				$('html').toggleClass('dark', !!Dex.prefs('dark'));
 


Slayer95:
Collecting user data should have a proper justification IMO.
What's the goal for tracking these?

scheibo:
https://www.smogon.com/forums/threads/have-mute-sounds-checked-by-default.3650625/

Slayer95:
Seems sensible!

Zarel:
Cool! Can you walk me through this implementation on Discord?

scheibo:
> Cool! Can you walk me through this implementation on Discord?

I'm not sure how to elaborate much further on the 5 lines than my description above - it might be easier if you had explicit questions I could answer? Either here or on Discord works, just ping me. :)

Zarel:
Cool, it's added; rest is up to you.

scheibo:
@Zarel 

I'd like to preregister hypotheses for what kind of data we expect from this setting/determine a number which would be sufficient to cause us to change our default. 

The analytics that have been added can't track when a user changes preferences in the middle of the battle (eg. muting immediately the first time a user enters a battle), but this is should be somewhat offset by the fact that more active battlers are going to be counted more heavily (ie. we're tracking % of battles muted, not % of players who use mute). It's going to be about new users (either those who just drop in for a look or those who haven't figured out about the sound settings yet) vs. power users IMO.

I don't think this is just the case that if its simply >50% of battles have mute turned on then we would change the default - hearing sounds and trying to figure out how to turn them off is a lot more discoverable than *not* hearing sounds and wondering if you could turn them on. Furthermore, we want to give new users the 'full' experience as much as possible while they're exploring/evaluating the simulator.

**What cutoff should we be looking for here?** Somewhere in the >60-75% range as a cutoff seems reasonable to me for determining whether we would change the default. My personal hypothesis is that >30% of battles will be muted, but I'm not confident enough to make further predictions (people on the interent are weird, and I feel those like myself who are complaining about sounds are more likely to be niche power users than the other way around)

Zarel:
My personal hypothesis is that everyone is significantly overestimating the percentage of users who change defaults. I'd also estimate that a significant proportion of people who don't like sound already have their computer muted. I'd guess closer to 10-20% mute usage.

60% would be enough for me to change the default, yes.

Also consider that we have sound effects, music, and notifications as separate settings. We should definitely also have in-battle sounds as a separate setting from notification sounds.

scheibo:
To be clear, we're talking about "proportion of battles", not "proportion of users". My hypothesis is that the people who play the most are more likely to turn off sounds. But I agree, I imagine I'm in the minority, but at least we'll have data to prove it.

> Also consider that we have sound effects, music, and notifications as separate settings. We should definitely also have in-battle sounds as a separate setting from notification sounds.

This PR only added tracking for mute. However, you're right, users may just turn off battle sounds and leave on notification sounds, causing the number muted fully to skew lower. Maybe my >30% was a high guess. We'll see. :)

Zarel:
I ended up just implementing a working tracker myself.

And it turns out, holy shit, more battles start muted than unmuted. It's around 55% muted. I definitely didn't expect that many people to fiddle with defaults. But I guess a lot of people play in public without headphones, and a lot of people prefer their own music.

This is definitely worth doing _something_ about. Maybe make the sound option more prominent on first launch? A "sound on" / "sound off" button in the "Battle!" box?

scheibo:
> This is definitely worth doing _something_ about. Maybe make the sound option more prominent on first launch? A "sound on" / "sound off" button in the "Battle!" box?

In addition to also making it not the default?

Zarel:
I don't know what to do about the default. I still lean against defaulting sound to off, for discoverability reasons. But it might make sense for the default to be "force the user to choose" instead of "on".

Or some compromise between "on" and "force the user to choose", like how the replay player is designed.

scheibo:
I'd really like to default to off so that we can then only have to load the large sound files on demand. The important question here - **is having sounds on by default attracting/retaining enough new users to justify the cost of both shipping the sound file and any users who bounce due to being annoyed by the sounds.** I say 'no', but I'm obviously biased by both not liking songs and not liking the idea of downloading assets I don't want.

Discoverability is another issue. Easiest fix in my mind is to just add a callout (like a tool-tip tutorial-esque hover) on the sound controls based on a pref - if the user has ever toggled sound we dont show the callout, but new users would be draw to the sound controls so they're aware its an option. That's just the first thing I thought of, I'm sure we could think of other ways to make it discoverable if we wanted to.

55% is pretty close to your original pre-registered threshold ("60% would be enough for me to change the default, yes."), FWIW

Slayer95:
> 55% is pretty close to your original pre-registered threshold ("60% would be enough for me to change the default, yes."), FWIW

That's only if you measure it in a scale of 0% to 60%, but imo it would make more sense to measure it in a scale of 50% to 60%, as anything less goes *against* the switching defaults option.

Other than that, if it's only 55% of *battles* and not 55% of *users* who have this setting, that would make switch a much less attractive option, as it has been argued that "people who play the most are more likely to turn off sounds" -so that would translate to a lower amount of users with muted setting, likely less than 50%.

> users who bounce due to being annoyed by the sounds

You come to a *game*, to play Pokémon, just like in the cartridge... What would they expect? If they instaleave rather than muting, I don't think we should concern ourselves about that.

Most browsers have a browser-level setting to mute either a tab (or the whole site -thanks Chrome for the update, now I can't open multiple Twitch / YT channels), and PS sound controls have a quite reasonable UI imo. Still, even more discoverability would be welcome, surely -for that, and for many other features!

> justify the cost of both shipping the sound file

Sounds are already loaded lazily. So in fact, the code should already not load BGM if PS is muted. Is that not the case? That could have been fixed without any need for analytics...

Zarel:
It's worth noting that I don't know a single other game which defaults sounds to off. I tend to have at least a bit of belief in Chesterton's Fence.

I guess you can blame that as "most games spend a lot on music and don't want that effort to go to waste; sunk cost fallacy".

Slayer95:
> This is definitely worth doing something about

It seems, however, that the performed research suggests action that IMO might not actually be needed. Earlier I fiddled around with [some numbers](https://gist.github.com/Slayer95/26af8ea104fbe37d33bf1f29bc436a50) and confirmed that effectively it's quite feasible that a population where less than half of users (40% in that case) use the muted setting would yield a frequency of 55% of battles muted.

However, since it's mainstream data to refute data with data, let me propose another experiment -without persistent tracking, to fit my flavor. So, it would go like this:
Define a set date (say, next Saturday), and a chance (10%). Then, when a user starts PS without cookie ``soundtest=1``, if a biased coinflip with said chance results in head, schedule for the next time among 03, 09, 15 and 21 ± 10 min (random deviation) a function that does the following:
- If the user has 3 or more *chat* rooms open, don't do anything (*define power users by that criteria, and exclude them*).
- Otherwise, register with Analytics whether sounds are muted or not.
- Read the ratio of non-power users who mute sounds after the survey is over.
- If that ratio was higher than some threshold (say, 40%, to be generous), I would agree that we need to do "something".

Zarel:
I feel like a visible "mute" checkbox on the challenge UI might be a good compromise?

Slayer95:
That wouldn't do anything for those whose first few dozen battles are in the ladder. A fair amount of them might not even know we support challenges, if playing on mobile.

Also, in the end, how do you know if the measures taken were effective or not? Repeating @scheibo 's or mine experiment afterwards would yield the same result -or it won't. Then if it's the same, are we done? Or is there still "something" to be done? Or maybe it's different, and it's *then* that we verify that there is still "something" to be done?

My point is, if the aim is to figure out whether sound controls' discoverability is enough or not, the whole approach we are following is misguided. It does answer, however, whether the current default is right or not.

We could theoretically run actual UX experiments for a randomized sample of users.
*Clicks Look for a battle*
Alert: *Hey, please wait a moment. We are executing research to improve Pokemon Showdown. To look for a battle, please mute the sounds (you can turn them on again afterwards!)*

Zarel:
I mean, by "the challenge UI" I mean "the search and challenge UIs". Right below the team selector on most of them.

It's not so much about discoverability than convenience.

Slayer95:
Makes sense. *hijacks thread* While at it, can we add a "View in Teambuilder" button for teams in in the ``TeamPopup``? (I have received that convenient feature request more times than I can count).

scheibo:
I'd be fine with a muted search button, because it seems like regardless of what additional data we gather that's probably the best I can hope for. And then only loading battle and effect sounds when the normal button (non muted button) is pressed

Zarel:
One possibility might be "Sounds on" and "Sounds off" buttons instead of a checkbox, which would go away when you chose one. Keeps clutter lowish.

scheibo:
> One possibility might be "Sounds on" and "Sounds off" buttons instead of a checkbox, which would go away when you chose one. Keeps clutter lowish.

We're talking here?

6mbM63wab2u
Where there's a 'Sounds On' and 'Sounds Off' buttons (that disappear when either is clicked)? Probably together spanning the same width and height as the format dropdown? This would seem to give it too much prominence?

Should it instead be a toggle? (with sounds on selected)?

Zarel:
Yeah.

If it were a toggle, it wouldn't be obvious when to hide it.

Having them as separate buttons makes the hiding clearer.

scheibo:
I might be misinterpreting your vision, but I feel these buttons would place too much prominence on the sound setting in the UI? I think the Replay example where `> Play` is larger and more prominent compared to the `Play (music off)` button is what we'd want to go for here. To copy that design, it seems like we'd want a smaller `Battle! (music off)` button under the `Battle!` button for the ladder and a separate `Challenge (sounds off)` button as well, not weird large "`Sounds On / Sounds Off` buttons that actually default to `Sounds On` if you dont click it but thats not indicated in any way" situation. I can see this being slightly undesirable though because it would muddle the obvious call-to-action that the current `Battle!` button presents.

Zarel:
"Play (music off)" is temporary; would "Battle (music off)" also be temporary? If yes, I doubt it would actually be more convenient than the permanent mute button in the upper right (which has fine discoverability). If no, it would work differently from replays, and I would worry about misleading design.

PR #1311, opened 2019-06-26
Improve teambuilder searchbox

TheJetOU:
\- Only run `this.updateTeamList` when the user has 'finished' typing

\- Fix searching when Pokemon have nicknames

\- Only focus the search box when a search has been made

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -112,6 +112,12 @@
 		curSet: null,
 		curSetLoc: 0,
 
+		// This is used in the search bar in attempt of figuring
+		// out when the user has 'finished' typing, it has a timer
+		// of 5ms which starts on a 'keyup' and on every 'keyup'
+		// it is reset
+		searchTimer: null,
+
 		// curFolder will have '/' at the end if it's a folder, but
 		// it will be alphanumeric (so guaranteed no '/') if it's a
 		// format
@@ -393,8 +399,14 @@
 					if (filterFormat && filterFormat !== (team.format || 'gen7')) continue;
 					if (filterFolder !== undefined && filterFolder !== team.folder) continue;
 
-					if (this.curSearchVal !== '' && team.team.indexOf(this.curSearchVal) == -1) {
-						continue;
+					if (this.curSearchVal) {
+						var pokemon = team.team.split(']').map(function (el) {
+							return toID(splitFirst(el, '|')[0]);
+						});
+						var curSearchValID = toID(this.curSearchVal);
+						if (!pokemon.includes(curSearchValID) && team.team.indexOf(curSearchValID) === -1) {
+							continue;
+						}
 					}
 
 					if (!atLeastOne) atLeastOne = true;
@@ -452,7 +464,7 @@
 				this.teamScrollPos = 0;
 			}
 
-			//reset focus to searchbar
+			// reset focus to searchbar
 			var teamSearchBar = this.$("#teamSearchBar");
 			var strLength = teamSearchBar.val().length;
 			teamSearchBar.focus();
@@ -2626,10 +2638,11 @@
 			this.chartSet(val, selectNext);
 		},
 		searchChange: function (e) {
-			//91 for right CMD / 93 for left CMD / 17 for CTL
+			// 91 for right CMD / 93 for left CMD / 17 for CTL
 			if (e.keyCode !== 91 && e.keyCode !== 93 && e.keyCode !== 17) {
 				this.curSearchVal = e.currentTarget.value;
-				this.updateTeamList();
+				clearTimeout(this.searchTimer);
+				this.searchTimer = setTimeout(this.updateTeamList.bind(this), 500);
 			}
 		},
 		chartSetCustom: function (val) {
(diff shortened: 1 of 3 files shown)


js/client-teambuilder.js
```diff
@@ -112,6 +112,12 @@
 		curSet: null,
 		curSetLoc: 0,
 
+		// This is used in the search bar in attempt of figuring
```
scheibo: "Instead of running a search after each character typed in the search bar we instead use a timer to only kick off a search 500ms after the last 'keyup' event (which is intended to approximate running a search only after the user has 'finished' typing)."

js/client-teambuilder.js
```diff
@@ -393,8 +399,14 @@
 					if (filterFormat && filterFormat !== (team.format || 'gen7')) continue;
 					if (filterFolder !== undefined && filterFolder !== team.folder) continue;
 
-					if (this.curSearchVal !== '' && team.team.indexOf(this.curSearchVal) == -1) {
-						continue;
+					if (this.curSearchVal) {
+						var pokemon = team.team.split(']').map(function (el) {
```
scheibo: Let's add a comment here explaining why we're doing this? Something like:

"If a Pokemon hasn't been given a nickname, species is omitted from the packed team.team in favor of the name field since the name defaults to the species' display name. While eliminating this redundancy between name and species helps with packed team size, the display name unfortunately won't match the ID search term and so we need to special case searching for Pokemon here"

scheibo:
This LGTM, Zarel can you please let us know if `src/battle-dex.ts` is an appropriate place for the `Chat.splitFirst` function we've stolen from the server codebase?

Zarel:
I really want a `splitFirst` function, but we have no place to put it.

`BattleTextParser` might be the best place on client, but it's annoyingly long to type.

scheibo:
Does it make sense to create a `Tools` or `Utils` class for these things (with the obvious caveat that classes like these need to be closely watched to ensure they don't because a dumping ground)?

Zarel:
I don't like dumping grounds. I explicitly got rid of the old Tools and I don't care to bring it back.

`splitFirst` is the only generically useful function we currently have; it can really be stuck anywhere, I think.

scheibo:
>  it can really be stuck anywhere, I think.

Including where this PR places it in `Dex`? Or do you mean "any namespaced location", ie/ "no new globals"

Zarel:
Any namespaced location. I didn't even realize this PR made it `splitFirst` and not `Dex.splitFirst`.

Zarel:
I actually don't _really_ mind leaving it as a global, either. I can't think of a place for it.

js/client-teambuilder.js
```diff
@@ -112,6 +112,13 @@
 		curSet: null,
 		curSetLoc: 0,
 
+		// Instead of running a search after each character typed
+		// in the search bar we instead use a timer to only kick off
+		// a search 500ms after the last 'keyup' event
+		// (which is intended to approximate running a search only
+		//  after the user has 'finished' typing).
```
Zarel: Why? This makes the search feel laggy for no real benefit.

scheibo: I mean, the timer can definitely be adjusted?

The main concern @TheJetOU  raised in <<dev>> was the performance of having to parse through potentially a very large amount of teams each keystroke. Maybe its less of a concern now that we've optimized it to just be doing `splitFirst` and search names, but adding some sort of timer is a fairly standard search-as-you-type optimization?

Zarel: The perf isn't actually a problem. On the other hand, render perf is... You'd want to only render, say, the top 12 teams or so. For the rest, you'd only want to display them on scroll.

Zarel: (If "on scroll" is too hard, you can also just display a clickable "More" button.)

scheibo:
Possibly (probably) out of scope for this PR, but do we want to maybe support multiple search terms to further narrow down searches? ie. `clefable,gliscor` for teams with both a Clefable and a Gliscor etc? It doesnt seem like it would be too difficult to add (split search term on `,`, iterate over terms and make sure all match) and its pretty much what was requested [here](https://www.smogon.com/forums/threads/what-can-ps-do-for-you.3622358/post-8161678).

scheibo:
This LGTM, but Zarel has outstanding concerns around the timer. Do we want to remove it for now to just land the fix, and if perf proves to be an issue we can talk about whether to add it back in/change rendering etc?

TheJetOU:
>This LGTM, but Zarel has outstanding concerns around the timer. Do we want to remove it for now to just land the fix, and if perf proves to be an issue we can talk about whether to add it back in/change rendering etc?

I'll probably work on changing the timer into a on-scroll/button sometime this week, it just depends on if @Zarel wants to ship this out as-is (the timer is definitely an improvement over the last system) or is willing to wait for my commit before merging.

Zarel:
The timer isn't an improvement over the last system on most metrics. :[

In particular, it feels "laggier" unless you have a lot of teams.

scheibo:
> I'll probably work on changing the timer into a on-scroll/button sometime this week, it just depends on if @Zarel wants to ship this out as-is (the timer is definitely an improvement over the last system) or is willing to wait for my commit before merging.

Let's ship this now to fix things for most users and look into optimizing the 'large number of teams' case as needed.

PR #1295, opened 2019-06-12
Include user's group in battle join message

scheibo:
Regression introduced by 8e9bc3fe1e739a92c8b25075a6a0479ffc56e0aa

PR #1312, opened 2019-06-28
Include group in join/leave messages

scheibo:
Battle join/leave was fixed to include group in #1295, but apparently chat join/leave should show group as well.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1472,7 +1472,7 @@
 				this.$chat.append('<div class="message"><small>Loading...</small></div>');
 				this.$joinLeave = this.$chat.children().last();
 			}
-			this.joinLeave[action].push(user.name);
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -1472,7 +1472,7 @@
 				this.$chat.append('<div class="message"><small>Loading...</small></div>');
 				this.$joinLeave = this.$chat.children().last();
 			}
-			this.joinLeave[action].push(user.name);
+			this.joinLeave[action].push(`${user.group}${user.name}`);
```
Zarel: ESLint is complaining because IE9 doesn't support template strings: `` ` ``

Zarel:
```suggestion
			this.joinLeave[action].push(user.group + user.name);
```

PR #1313, opened 2019-07-01
Don't overwrite issuing a challenge with accepting one

scheibo:
Currently, an incoming challenge replaces the a pending challenge, causing the 'Challenge' button to turn into an 'Accept', resulting in accidentally accepting the challenge with the wrong team.

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -631,12 +631,14 @@
 						self.notifyOnce("Challenge from " + name, "Format: " + BattleLog.escapeFormat(format), 'challenge:' + userid);
 					}
 					var $challenge = self.openChallenge(name, $pmWindow);
-					var buf = '<form class="battleform"><p>' + BattleLog.escapeHTML(name) + ' wants to battle!</p>';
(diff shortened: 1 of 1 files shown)


js/client-mainmenu.js
```diff
@@ -664,9 +664,7 @@
 		openChallenge: function (name, $pmWindow) {
 			if (!$pmWindow) $pmWindow = this.openPM(name, true);
 			var $challenge = $pmWindow.find('.challenge');
-			if (!$challenge.length) {
-				$challenge = $('<div class="challenge"></div>').insertAfter($pmWindow.find('h3'));
-			}
+			$challenge = $('<div class="challenge"></div>').insertAfter($challenge.length ? $challenge : $pmWindow.find('h3'));
```
Zarel: Err, the window having multiple challenges is not a supported state...

Zarel: The code assumes the challenge window a single window that can be found with `$pmWindow.find('.challenge')`.

A lot of code will break if that assumption is false, such as:

https://github.com/Zarel/Pokemon-Showdown-Client/blob/89a3ec78c0c3d6128135f73919c66ab768490862/js/client-mainmenu.js#L641-L658

PR #1315, opened 2019-07-03
Support player count in |tournament|start

scheibo:
Server: Zarel/Pokemon-Showdown#5581

### DIFF
--- a/js/client-chat-tournament.js
+++ b/js/client-chat-tournament.js
@@ -315,7 +315,8 @@
 					} else if (this.info.teambuilderFormat.substr(0, 4) === 'gen5' && !Dex.loadedSpriteData['bw']) {
 						Dex.loadSpriteData('bw');
 					}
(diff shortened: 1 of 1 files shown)


js/client-chat-tournament.js
```diff
@@ -315,7 +315,8 @@
 					} else if (this.info.teambuilderFormat.substr(0, 4) === 'gen5' && !Dex.loadedSpriteData['bw']) {
 						Dex.loadSpriteData('bw');
 					}
-					this.room.$chat.append("<div class=\"notice tournament-message-start\">The tournament has started!</div>");
+					var participants = data[0] ? " (" + data[0] + " participants)" : "";
```
Zarel:
```suggestion
					var participants = data[0] ? " (" + data[0] + " players)" : "";
```

"Participants" is a less common word, and probably harder for non-native speakers.

PR #1317, opened 2019-07-06
Support new status protocol in client

whalemer:
Required for zarel/pokemon-showdown#5591


Zarel:
We ended up not going for this approach.

PR #1318, opened 2019-07-07
Display Elo in rated battles

TheJetOU:
Server Pull Request: Zarel/Pokemon-Showdown#5595

Suggested and approved here: https://www.smogon.com/forums/threads/3651687/

I did not implement the "make `/rank` only display the tier of the battle you're playing in" suggestion because you can use `/rank [player], [format]` or with this change it's even easier because you can just scroll up.

### DIFF
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -625,8 +625,8 @@ class BattleScene {
 		pokemonhtml = '<div class="teamicons">' + pokemonhtml + '</div>';
 		const $sidebar = (side.n ? this.$rightbar : this.$leftbar);
 		if (side.name) {
-			const ratinghtml = side.rating ? `title="Rating: ${BattleLog.escapeHTML(side.rating)}` : ``;
-			$sidebar.html(`<div class="trainer" ${ratinghtml}><strong>${BattleLog.escapeHTML(side.name)}</strong><div class="trainersprite" style="background-image:url(${Dex.resolveAvatar(side.avatar)})"></div>${pokemonhtml}</div>`);
+			const ratinghtml = side.rating ? ` title="Rating: ${BattleLog.escapeHTML(side.rating)}` : ``;
+			$sidebar.html(`<div class="trainer"${ratinghtml}><strong>${BattleLog.escapeHTML(side.name)}</strong><div class="trainersprite" style="background-image:url(${Dex.resolveAvatar(side.avatar)})"></div>${pokemonhtml}</div>`);
 			$sidebar.find('.trainer').css('opacity', 1);
 		} else {
 			$sidebar.find('.trainer').css('opacity', 0.4);
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -625,8 +625,8 @@ class BattleScene {
 		pokemonhtml = '<div class="teamicons">' + pokemonhtml + '</div>';
 		const $sidebar = (side.n ? this.$rightbar : this.$leftbar);
 		if (side.name) {
-						const ratinghtml = side.rating ? `title="Rating: ${BattleLog.escapeHTML(side.rating)}` : ``;
-						$sidebar.html(`<div class="trainer" ${ratinghtml}><strong>${BattleLog.escapeHTML(side.name)}</strong><div class="trainersprite" style="background-image:url(${Dex.resolveAvatar(side.avatar)})"></div>${pokemonhtml}</div>`);
+			const ratinghtml = side.rating ? `title="Rating: ${BattleLog.escapeHTML(side.rating)}` : ``;
+			$sidebar.html(`<div class="trainer" ${ratinghtml}><strong>${BattleLog.escapeHTML(side.name)}</strong><div class="trainersprite" style="background-image:url(${Dex.resolveAvatar(side.avatar)})"></div>${pokemonhtml}</div>`);
 			$sidebar.find('.trainer').css('opacity', 1);
 		} else {
 			$sidebar.find('.trainer').css('opacity', 0.4);
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -625,7 +625,8 @@ class BattleScene {
 		pokemonhtml = '<div class="teamicons">' + pokemonhtml + '</div>';
 		const $sidebar = (side.n ? this.$rightbar : this.$leftbar);
 		if (side.name) {
-			$sidebar.html('<div class="trainer" ' + (side.rating ? 'title="Rating: ' + side.rating + "\">" : '>') + '<strong>' + BattleLog.escapeHTML(side.name) + '</strong><div class="trainersprite" style="background-image:url(' + Dex.resolveAvatar(side.avatar) + ')"></div>' + pokemonhtml + '</div>');
+						const ratinghtml = side.rating ? `title="Rating: ${BattleLog.escapeHTML(side.rating)}` : ``;
+						$sidebar.html(`<div class="trainer" ${ratinghtml}><strong>${BattleLog.escapeHTML(side.name)}</strong><div class="trainersprite" style="background-image:url(${Dex.resolveAvatar(side.avatar)})"></div>${pokemonhtml}</div>`);
 			$sidebar.find('.trainer').css('opacity', 1);
 		} else {
 			$sidebar.find('.trainer').css('opacity', 0.4);
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -625,7 +625,7 @@ class BattleScene {
 		pokemonhtml = '<div class="teamicons">' + pokemonhtml + '</div>';
 		const $sidebar = (side.n ? this.$rightbar : this.$leftbar);
 		if (side.name) {
-			$sidebar.html('<div class="trainer"><strong>' + BattleLog.escapeHTML(side.name) + '</strong><div class="trainersprite" style="background-image:url(' + Dex.resolveAvatar(side.avatar) + ')"></div>' + pokemonhtml + '</div>');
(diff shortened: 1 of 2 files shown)


Zarel:
We usually use "rating" rather than "Elo".

I also don't know if I approve of this. Most games don't do this, probably for good reason.

TheJetOU:
> I also don't know if I approve of this. Most games don't do this, probably for good reason.

It seems to be relatively harmless to me; if there's a large backlash from the community, we can just remove it.

Maybe we should poll the community? I'm not sure how that would even occur though.

scheibo:
This is a popular suggestion and simplifies a common use case. What fallout do we expect? People whining about the matchmaking system?

Zarel:
If nothing else, it clutters up the battle screen. People can use the commands if they really care.

Zarel:
How about it only appears on hover?

scheibo:
On hover of what? The avatar? The name in the text log?

I don't really think its clutter - rating is useful information. Clutter is users using `/rank opponent` at the start of each battle (because they're too lazy to type in `, format` as well) and getting a large table in their logs as opposed to a 2 lines.

I'd guess few players care about rank to that extent (we're talking about the <5% power user probably), but there's already a ton of frontmatter to the output logs, and listing the clauses which never change is a lot spammier than rating which can _sometimes_ be interesting to the player.

I'm not opposed to it being 'on hover'/more hidden, but I also don't really see the need. Curious what others think

Zarel:
I do also think the clauses should be consolidated in some way that makes them easier to skip over. Like, put them in a box or something.

TheJetOU:
I wouldn't consider clauses _spammy_ they're useful information, maybe not to the person who has already played 3000 games, but to the new people, and putting it in a box would conceal that handy data.

As for rating being spammy, they only appear at team preview, and there's nothing too important in the logs to look at in team preview.  The information become more valuable when the battle actually starts.

Zarel:
"Clutter" is probably a better word here. Clutter makes it hard to find specific things you're looking for.

scheibo:
Can we all agree that theres clutter in the battle intro with clauses already and that this adds to it, but potentially is not enough to push it over the edge? I'd suggest not blocking this and instead filing a followup issue to figure out a good UI for collapsing the less essential info? (maybe a hover over the format lists the closes, and hover over players names lists the ELO?)

Zarel:
Either way, this should be implemented in `battle-log`, not in `battle`. Look at how everything else in the battle log is implemented.

Zarel:
We currently don't support adding things to the battle log outside of the order shown. Unless we tracked them in the individual `Side` objects, which might be necessary.

image

Here's what the start of a battle looks like. The clauses should probably be put in a box. "Next to the teams" might not be a bad place to put the users' ratings. But I'd really rather they just appear on hover on the avatars or something.

It's probably true that once we finalize the protocol, we don't need to block the server-side. But I don't think "it's already cluttered, just make it more cluttered" is a good reason to merge the client-side change. I think this change is a net decrease in usability, and I don't think any of the code in this PR will be useful to the final implementation.

scheibo:
> Here's what the start of a battle looks like. The clauses should probably be put in a box

A box is the bare minimum as it would separate them from visually at least (I'm interpreting box as 'surround the entire set of clauses as one unit with a border'). Can we collapse them all down into an on-hover for **[Gen 7] OU** though? I'm not sure `Format:` adds a ton of value either. I'd do

```
Player A joined
Player B joined
**[Gen 7] OU**
Player A's team
Player B's team
Battle started between Player A and Player B!
```
Where the format can be hovered over or perhaps is actually a button which can be clicked to expand to reveal what all the clauses are. 

> But I'd really rather they just appear on hover on the avatars or something.

I support adding the rating on hover over the player's avatar. They don't need to show up in the log, this is fine to display on demand with a quick mouse over or long press for the power user IMO.

Zarel:
I agree that "Format:" is unnecessary. I don't agree with collapsing the clauses.

scheibo:
Why exactly? The format defines which clauses are active. They don't change very often if ever, and people could still see them if they cared by hovering over format?

scheibo:
Anyway, @TheJetOU - how do you feel about showing the rank only when hovering over the trainer avatar? Decluttering the front matter wrt formats and clauses is out of scope for what you're trying to do and we can handle that in a follow up.

TheJetOU:
> Anyway, @TheJetOU - how do you feel about showing the rank only when hovering over the trainer avatar? Decluttering the front matter wrt formats and clauses is out of scope for what you're trying to do and we can handle that in a follow up.

Sure.

Zarel:
> Why exactly? The format defines which clauses are active. They don't change very often if ever, and people could still see them if they cared by hovering over format?

We get new players relatively often. I'd rather not make them less discoverable. I was thinking maybe some sort of collapse button for old users, but honestly as long as clauses are properly offset, I don't actually think they're very cluttery at all.

Rating, on the other hand, would be pretty cluttered in the current design. Maybe something like

image

src/battle.ts
```diff
@@ -626,6 +627,9 @@ class Side {
 	setAvatar(avatar: string) {
 		this.avatar = avatar;
 	}
+	setRating(rating: string) {
```
Zarel: I wouldn't make a setter for this. `setAvatar` exists as an external API, but unless someone needs to set `rating` externally, you can just set `rating` directly; it's defined in the same file, after all.

src/battle-animations.ts
```diff
@@ -625,7 +625,7 @@ class BattleScene {
 		pokemonhtml = '<div class="teamicons">' + pokemonhtml + '</div>';
 		const $sidebar = (side.n ? this.$rightbar : this.$leftbar);
 		if (side.name) {
-			$sidebar.html('<div class="trainer"><strong>' + BattleLog.escapeHTML(side.name) + '</strong><div class="trainersprite" style="background-image:url(' + Dex.resolveAvatar(side.avatar) + ')"></div>' + pokemonhtml + '</div>');
+			$sidebar.html('<div class="trainer" ' + (side.rating ? 'title="Rating: ' + side.rating + "\">" : '>') + '<strong>' + BattleLog.escapeHTML(side.name) + '</strong><div class="trainersprite" style="background-image:url(' + Dex.resolveAvatar(side.avatar) + ')"></div>' + pokemonhtml + '</div>');
```
scheibo: Native `title` attributes generally don't work on mobile because you can't hover. Does the tooltip logic PS uses on mobile work based on `title`? If not, perhaps we want to be the PS tooltips instead

Zarel: PS tooltips are extremely hard to use, sadly.

Zarel: This isn't sanitized.

```suggestion
			$sidebar.html(`<div class="trainer" ${side.rating ? 'title="Rating: ' + BattleLog.escapeHTML(side.rating) + '"' : ''}><strong>${BattleLog.escapeHTML(side.name)}</strong><div class="trainersprite" style="background-image:url(${Dex.resolveAvatar(side.avatar)})"></div>${pokemonhtml}</div>`);
```

scheibo: So our story on mobile here is "tough luck"? I don't know how many power users who care about ELO of each of their opponents are also using PS on a touch device, so maybe its an irrelevant concern?

TheJetOU: I guess the only other solutions is adding hacky jQuery to show a message on-click? But then the box probably wouldn't be at the place of the avatar so we would need to also fix that. I don't think it's _too_ worth it.

scheibo: I think its probably fine to punt.

Zarel: It's not _completely_ unusable on mobile, I think it appears on the long-press menu?

Zarel: For the record, I know PS tooltips are hard to use because I've wanted to use them for everything else ever (appearing instantly instead of on a delay is really good for UX).

Maybe I could write some code replacing all titles with PS tooltips...

scheibo: Some tooltip libraries do that, which is why I asked. As for long press - was not aware of that, but I'll take your word for it.

src/battle-animations.ts
```diff
@@ -625,7 +625,8 @@ class BattleScene {
 		pokemonhtml = '<div class="teamicons">' + pokemonhtml + '</div>';
 		const $sidebar = (side.n ? this.$rightbar : this.$leftbar);
 		if (side.name) {
-			$sidebar.html('<div class="trainer"><strong>' + BattleLog.escapeHTML(side.name) + '</strong><div class="trainersprite" style="background-image:url(' + Dex.resolveAvatar(side.avatar) + ')"></div>' + pokemonhtml + '</div>');
+			const ratinghtml = side.rating ? `title="Rating: ${BattleLog.escapeHTML(side.rating)}` : ``;
```
Zarel:
```suggestion
			const ratinghtml = side.rating ? ` title="Rating: ${BattleLog.escapeHTML(side.rating)}` : ``;
```

src/battle-animations.ts
```diff
@@ -625,7 +625,8 @@ class BattleScene {
...
+			$sidebar.html(`<div class="trainer" ${ratinghtml}><strong>${BattleLog.escapeHTML(side.name)}</strong><div class="trainersprite" style="background-image:url(${Dex.resolveAvatar(side.avatar)})"></div>${pokemonhtml}</div>`);
```
Zarel:
```suggestion
			$sidebar.html(`<div class="trainer"${ratinghtml}><strong>${BattleLog.escapeHTML(side.name)}</strong><div class="trainersprite" style="background-image:url(${Dex.resolveAvatar(side.avatar)})"></div>${pokemonhtml}</div>`);
```

Zarel: (Minor nitpick, but this moves the space over to its correct location.)

scheibo:
> Everything else looks good. This is a good first step to making it visible in the log.

Wait, it seemed like you didn't want it visible in the log at all? I thought this was the end state?

Zarel:
> Wait, it seemed like you didn't want it visible in the log at all? I thought this was the end state?

I didn't want it to clutter the log. I'm open to showing it with a good design.

PR #1319, opened 2019-07-07
Place ranks/groups in a pillbox to further distinguish from status

scheibo:
>> Can we better differentiate statuses from rank/group display such that its more difficult to confuse the two?
>
> Rank/group could be pills? Like how Discord does them.

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown/issues/5590#issuecomment-508913502_

Screen Shot 2019-07-07 at 14 35 37

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -2494,8 +2494,12 @@ function toId() {
 			var status = offline ? '(Offline)' : '(Online)';
 			if (data.status && !offline) status = data.status.startsWith('!') ? data.status.slice(1) : data.status;
 			buf += '<span class="userstatus' + (offline ? ' offline' : '') + '">' + BattleLog.escapeHTML(status) + '</span><br />';
-			buf += '<small>' + (group || '&nbsp;') + '</small>';
-			if (globalgroup) buf += '<br /><small>' + globalgroup + '</small>';
+			if (group) {
+				buf += '<small class="usergroup roomgroup">' + group + '</small>';
+			} else {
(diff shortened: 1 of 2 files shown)


bumbadadabum:
I feel like this design takes up slightly too much vertical space. Maybe
decrease the vertical padding by 1px?

scheibo:
Screen Shot 2019-07-07 at 15 02 27

scheibo:
Screen Shot 2019-07-07 at 15 05 12

Zarel:
Hm, that doesn't look very good. I might need to play around with the styling more.

scheibo:
> Hm, that doesn't look very good. I might need to play around with the styling more.

Yeah. I mean, it grows on you, but its not very sexy. I tried slightly wider borders and well as box-shadows, but feel free to tinker.

scheibo:
This is what it currently looks like: 
Screen Shot 2019-07-08 at 17 52 56

switched to `1px 5px` and changed the border-radius from Discord's `11px` to PS's `5px`

Also tested out a box-shadow, but didn't commit these changes (looks on the verge of being clickable IMO)
Screen Shot 2019-07-08 at 17 54 33

scheibo:
@Zarel - have you had a chance to tinker with this? It's a fairly easy thing to play around with

Zarel:
It's not so much easiness, I just have a huge backlog and this is low-priority.

PR #1320, opened 2019-07-08
Update the build to allow for configurable routes

scheibo:
Obviously `Config.routes` will need to be used in the other 200+ places routes are hardcoded in the client, but there's enough going on here that I felt like just creating a PR for this first step.

### DIFF
--- a/build-tools/update
+++ b/build-tools/update
@@ -41,6 +41,9 @@ configBuf = configBuf.replace(/\/\* version \*\/[^;\n]*;/, `/* version */ Config
 
 fs.writeFileSync('config/config.js', configBuf);
 console.log("DONE");
+const context = {Config: {}};
+require('vm').runInNewContext(configBuf, context);
+const Config = context.Config;
 
 /*********************************************************
  * Compile TS files
@@ -82,31 +85,42 @@ if (!ignoreGraphics) {
  * Update cachebuster and News
  *********************************************************/
 
+const URL_REGEX = /(src|href)="\/(.*?)(\?[a-z0-9]*?)?"/g;
+
+function updateURL(a, b, c, d) {
+	c = c.replace('/replay.pokemonshowdown.com/', '/' + Config.routes.replays + '/');
+	c = c.replace('/dex.pokemonshowdown.com/', '/' + Config.routes.dex + '/');
+	c = c.replace('/play.pokemonshowdown.com/', '/' + Config.routes.client + '/');
+	c = c.replace('/pokemonshowdown.com/', '/' + Config.routes.root + '/');
+
+	if (d) {
+		let hash = Math.random(); // just in case creating the hash fails
+		try {
+			const filename = c.replace('/' + Config.routes.client + '/', '');
+			const fstr = fs.readFileSync(filename, {encoding: 'utf8'});
+			hash = crypto.createHash('md5').update(fstr).digest('hex').substr(0, 8);
+		} catch (e) {}
+
+		return b + '="/' + c + '?' + hash + '"';
+	} else {
+		return b + '="/' + c + '"';
+	}
+}
+
+function writeIndex(indexContents, indexContents2) {
+	process.stdout.write("Writing new `index.html` file... ");
+	fs.writeFileSync('index.html', indexContents);
+	fs.writeFileSync('preactalpha.html', indexContents2);
+	console.log("DONE");
+}
+
 function updateIndex() {
-	// add hashes to js and css files
-	process.stdout.write("Updating hashes... ");
+	// add hashes to js and css files and rewrite URLs
+	process.stdout.write("Updating hashes and URLs... ");
 	let indexContents = fs.readFileSync('index.template.html', {encoding: 'utf8'});
-	indexContents = indexContents.replace(/(src|href)="\/(.*?)\?[a-z0-9]*?"/g, function (a, b, c) {
-		let hash = Math.random(); // just in case creating the hash fails
-		try {
-			const filename = c.replace('/play.pokemonshowdown.com/', '');
-			const fstr = fs.readFileSync(filename, {encoding: 'utf8'});
-			hash = crypto.createHash('md5').update(fstr).digest('hex').substr(0, 8);
-		} catch (e) {}
-
-		return b + '="/' + c + '?' + hash + '"';
-	});
+	indexContents = indexContents.replace(URL_REGEX, updateURL);
 	let indexContents2 = fs.readFileSync('preactalpha.template.html', {encoding: 'utf8'});
-	indexContents2 = indexContents2.replace(/(src|href)="\/(.*?)\?[a-z0-9]*?"/g, function (a, b, c) {
-		let hash = Math.random(); // just in case creating the hash fails
-		try {
-			const filename = c.replace('/play.pokemonshowdown.com/', '');
-			const fstr = fs.readFileSync(filename, {encoding: 'utf8'});
-			hash = crypto.createHash('md5').update(fstr).digest('hex').substr(0, 8);
-		} catch (e) {}
-
-		return b + '="/' + c + '?' + hash + '"';
-	});
+	indexContents2 = indexContents2.replace(URL_REGEX, updateURL);
 	console.log("DONE");
 
 	// add news, only if it's actually likely to exist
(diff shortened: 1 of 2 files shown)


build-tools/update
```diff
@@ -41,6 +39,7 @@ configBuf = configBuf.replace(/\/\* version \*\/[^;\n]*;/, `/* version */ Config
 
 fs.writeFileSync('config/config.js', configBuf);
 console.log("DONE");
+eval(configBuf);
```
scheibo: That moment when the `eval` hater uses `eval`...

build-tools/update
```diff
@@ -82,31 +81,42 @@ if (!ignoreGraphics) {
...
-			hash = crypto.createHash('md5').update(fstr).digest('hex').substr(0, 8);
-		} catch (e) {}
+const URL_REGEX = /(src|href)="\/(.*?)(\?[a-z0-9]*?)?"/g;
 
-		return b + '="/' + c + '?' + hash + '"';
-	});
-	let indexContents2 = fs.readFileSync('preactalpha.template.html', {encoding: 'utf8'});
-	indexContents2 = indexContents2.replace(/(src|href)="\/(.*?)\?[a-z0-9]*?"/g, function (a, b, c) {
+function updateURL(a, b, c, d) {
+	c = c.replace('/replay.pokemonshowdown.com/', '/' + Config.routes.replays + '/');
```
scheibo: Not a huge fan of using multiple `replace` lines. Also, anchoring with `/` because otherwise the routes will interfere with each other). Kind of interesting to use the canonical routes as the placeholders to be replaced, but it means `index.template.html` is still valid on its own (as opposed to like `<%= Config.routes.root %>` or whatever), which is nice.

build-tools/update
```diff
@@ -128,11 +138,10 @@ function updateIndex() {
 			indexContents = indexContents.replace(/<!-- news -->/g, newsData[1]);
 			console.log("DONE");
 
-			process.stdout.write("Writing new `index.html` file... ");
-			fs.writeFileSync('index.html', indexContents);
-			fs.writeFileSync('preactalpha.html', indexContents2);
-			console.log("DONE");
+			writeIndex(indexContents, indexContents2);
 		});
+	} else {
```
scheibo: I don't think it hurts to write `index.html` even if the repo isn't housed under a directory named `play.pokemonshowdown.com`... 

Out of scope for this change, but we should consider `build`-ing the `testclient.html` as well so that it actually can use cache busting...

build-tools/update
Slayer95: It would probably be better to do

```js
const context = {Config: {}}
require('vm').runInNewContext(configBuf, context);
const {Config} = context;
```

scheibo: Nice! Thanks!

Slayer95:
The following files need to be updated as well:
- ``crossprotocol.html`` <- For support of shared teambuilders between HTTP and HTTPS
- ``src/battle-dex.ts`` (or whatever it builds to) <- Particularly ``resourcePrefix``, ``fxPrefix```, and ``soundManager``)
``js/storage.js`` <- ``Config.origindomain``, ``Storage.origin``. ``Config.defaultserver`` requires ``Config.routes.app``

---------
Other notes
- Some CSS files refer ``play.pokemonshowdown.com`` -that's for replays support. Gotta update that too.
- I don't see a fallback in case Zarel (or other consumers) forget to add ``Config.routes`` to the Config.
- Building is already a requirement, so we might as well use templates rather than replacing the canonical routes. Not a blocker, though, as it can be added if/when we actually need to preserve ``*.pokemonshowdown.com`` somewhere independently of where the client is hosted.
- I only read until now that you consider this a "first step", but at least the instances mentioned above the line are rather important.

Slayer95:
The following files need to be updated as well:
- ``crossprotocol.html`` <- For support of shared teambuilders between HTTP and HTTPS
- ``src/battle-dex.ts`` (or whatever it builds to) <- Particularly ``resourcePrefix``, ``fxPrefix```, and ``soundManager``)
``js/storage.js`` <- ``Config.origindomain``, ``Storage.origin``. ``Config.defaultserver`` requires ``Config.routes.app``

---------
Other notes
- Some CSS files refer ``play.pokemonshowdown.com`` -that's for replays support. Gotta update that too.
- I don't see a fallback in case Zarel (or other consumers) forget to add ``Config.routes`` to the Config.
- Building is already a requirement, so we might as well use templates rather than replacing the canonical routes. Not a blocker, though, as it can be added if/when we actually need to preserve ``*.pokemonshowdown.com`` somewhere independently of where the client is hosted.
- I only read until now that you consider this a "first step", but at least the instances mentioned above the line are rather important.

Zarel:
This is the client; there are assumed to be no other consumers; a fallback is unnecessary.

Slayer95:
> This is the client; there are assumed to be no other consumers; a fallback is unnecessary.

Except that there used to be 3 consumers, including xfix and myself -I wouldn't know if any of them are still active, or if there are any others now.

Zarel:
I edited: No other consumer is officially supported. It's not like server, where zero-config updates are mostly guaranteed. I'm fine with breaking updates with client; anyone running a client should know what they're doing.

Zarel:
I also think most places routes are used can be left alone. Heck, we really only need it in one place, `index.html`, which can actually be consolidated into a single `<base>` tag:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base

Slayer95:
> I also think most places routes are used can be left alone. Heck, we really only need it in one place, `index.html`, which can actually be consolidated into a single `<base>` tag:
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base

I wouldn't be that optimistic considering we must support ``testclient.html`` Though if it could be worked out, it would be great.

Zarel:
`testclient.html` is already designed to be opened from file system, and it's already supported...

scheibo:
> The following files need to be updated as well:
>     `...`
>
> * Some CSS files refer `play.pokemonshowdown.com` -that's for replays support. Gotta update that too.
> * `...`
> * I only read until now that you consider this a "first step", but at least the instances mentioned above the line are rather important.

To be clear, as stated in the original description, this PR is not intended to fully convert the client to using `Config.routes` in one blow. @Slayer95  - the files you mention in https://github.com/Zarel/Pokemon-Showdown-Client/pull/1320#issuecomment-509896695, are they required as part of this change, or can I not migrate them in a followup(s)? There are places in the PHP etc as well which need to be updated, I just wanted to get a foothold in this change and introduce `Config.routes`.

> * I don't see a fallback in case Zarel (or other consumers) forget to add `Config.routes` to the Config.

I think Zarel answered this.

> * Building is already a requirement, so we might as well use templates rather than replacing the canonical routes. Not a blocker, though, as it can be added if/when we actually need to preserve `*.pokemonshowdown.com` somewhere independently of where the client is hosted.

What does using templates entail in your mind? Something like Mustache (picking a templating lib at random), or just a more 'formal' search and replace like we're currently doing? I think Zarel is not super fond of templating libraries, which is why I was proceeding with this simplistic approach

> I also think most places routes are used can be left alone. Heck, we really only need it in twoish places; index.html can be consolidated into a single `<base>` tag

... I assume you mean in `index.html`. We play.pokemonshowdown.com is hard coded in like 200+ places, and will definitely need to be replaced elsewhere.

I'm fine to use a `<base>` tag in the HTML, I just wanted to make sure we're on the same page about what future PRs will do.

Slayer95:
> To be clear, as stated in the original description, this PR is not intended to fully convert the client to using Config.routes in one blow. @Slayer95 - the files you mention in #1320 (comment), are they required as part of this change, or can I not migrate them in a followup(s)? There are places in the PHP etc as well which need to be updated, I just wanted to get a foothold in this change and introduce Config.routes.

The PHP files are part of a different system (login server). The files I mentioned are just client-side, so I think they should be part of this change.

> What does using templates entail in your mind? Something like Mustache (picking a templating lib at random), or just a more 'formal' search and replace like we're currently doing? I think Zarel is not super fond of templating libraries, which is why I was proceeding with this simplistic approach

Just this sort of thing: "(<%= Config.routes.root %>)". I would be against libraries for this as well.

Slayer95:
It seems this is waiting for @Zarel to reply to scheibo's earlier comment.

scheibo:
No, this is mostly waiting on me.

PR #1321, opened 2019-07-12
Tournaments: Support replace message

whalemer:


whalemer:
(Required for Zarel/pokemon-showdown#5600)

js/client-chat-tournament.js
```diff
@@ -307,7 +307,9 @@
 					if (leaves.length > 0) message.push(arrayToPhrase(leaves) + " left the tournament");
 					this.$lastJoinLeaveMessage.text(message.join("; ") + ".");
 					break;
-
+				case 'replace':
```
Zarel: Looks like the surrounding code double spaces cases.

```suggestion

				case 'replace':
```

js/client-chat-tournament.js
```diff
@@ -307,7 +307,9 @@
...
+					this.room.$chat.append("<div class=\"notice\">" + BattleLog.escapeHTML(data[1]) + " has joined the tournament, replacing " + BattleLog.escapeHTML(data[0]) + ".</div>");
+					break;
```
Zarel:
```suggestion
					break;

```

Zarel:
I'm surprised we've never implemented this in the past.

Zarel:
Do we actually want to keep supporting the tournament message ID system? All other server messages just directly send text to the client.

whalemer:
The only reason I can really see is to make it easier for bots to track state, but I'm not sure if that's worth the maintenance effort.

whalemer:
(also battle starts giving a nice message, and batching join / leave messages)

There is a lot that could be moved out, but maybe it's better to have them all as |tournament messages for consistency?

Zarel:
They should probably be kept in there, yes, for ease of `ignoretournaments` ignoring them.

They could all be `|tournament|message|` or something, though.

PR #1325, opened 2019-07-14
Fix Castform-Rainy

tmagicturtle:
Castform-Rainy's sprite is defined twice. Further, the second definition actually points to Castform-Snowy. This causes Castform-Rainy to appear as Castform-Snowy.

https://i.imgur.com/K9wS7K4.png


tmagicturtle:
On second thought, is it better to remove the line, or is it better to correct it to Castform-Snowy?

Marty-D:
I was about to correct it and comment, yes. This section is for the left-facing sprites which are designed for asymmetrical Pokemon. Thanks for noticing the typo!

PR #1326, opened 2019-07-15
Add the ability to filter the ladder by a common prefix (for OLT)

scheibo:
**Context:** https://www.smogon.com/forums/threads/ladder-filter.3652321/

Having trouble testing this ('just test in prod'), but this is a timely and popular feature request that I think we should look to implement.

### DIFF
--- a/js/client-ladder.js
+++ b/js/client-ladder.js
@@ -150,7 +150,7 @@
 						if (self.curFormat !== format) return;
 						var buf = '<div class="ladder pad"><p><button name="selectFormat"><i class="fa fa-chevron-left"></i> Format List</button></p><p><button class="button" name="refresh"><i class="fa fa-refresh"></i> Refresh</button>';
 						buf += '<input type="text" id="ladderSearchBar" name="search" class="textbox searchinput" value="' + (this.curSearchVal || '') + '" placeholder="username prefix"/></p>';
-						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top 500</h3>';
+						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top ' + (this.curSearchVal ? "- '" + this.curSearchValue + "'" : '500') + '</h3>';
 						buf += data + '</div>';
 						self.$el.html(buf);
 					}, 'html');
--- a/js/client-ladder.js
+++ b/js/client-ladder.js
@@ -146,7 +146,7 @@
 					}, function (data) {
 						if (self.curFormat !== format) return;
 						var buf = '<div class="ladder pad"><p><button name="selectFormat"><i class="fa fa-chevron-left"></i> Format List</button></p><p><button class="button" name="refresh"><i class="fa fa-refresh"></i> Refresh</button>';
-						buf += '<form><input type="text" id="ladderSearchBar" name="search" class="textbox searchinput" value="' + (self.curSearchVal || '') + '" placeholder="username prefix"/><button name="search" type="submit"> Search</button></form></p>';
+						buf += '<form class="search"><input type="text" name="searchval" class="textbox searchinput" value="' + (self.curSearchVal || '') + '" placeholder="username prefix"/><button type="submit"> Search</button></form></p>';
 						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top ' + (self.curSearchVal ? "- '" + self.curSearchVal + "'" : '500') + '</h3>';
 						buf += data + '</div>';
 						self.$el.html(buf);
@@ -154,8 +154,8 @@
 				}
 			}
 		},
-		search: function (e) {
-			this.curSearchVal = $('#ladderSearchBar').val();
+		submitSearch: function (e) {
+			this.curSearchVal = this.$('input[name=searchval]').val();
 			this.update();
 		},
 		showHelp: function () {
--- a/js/client-ladder.js
+++ b/js/client-ladder.js
@@ -87,7 +87,7 @@
 
 	this.LadderRoom = HTMLRoom.extend({
 		events: {
-			'keyup .searchinput': 'searchChange',
+			'input .searchinput': 'searchChange',
 		},
 		type: 'ladder',
 		title: 'Ladder',
@@ -150,7 +150,7 @@
 						if (self.curFormat !== format) return;
 						var buf = '<div class="ladder pad"><p><button name="selectFormat"><i class="fa fa-chevron-left"></i> Format List</button></p><p><button class="button" name="refresh"><i class="fa fa-refresh"></i> Refresh</button>';
 						buf += '<input type="text" id="ladderSearchBar" name="search" class="textbox searchinput" value="' + (this.curSearchVal || '') + '" placeholder="username prefix"/></p>';
-						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top ' + (this.curSearchVal ? "- '" + this.curSearchValue + "'" : '500') + '</h3>';
+						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top ' + (this.curSearchVal ? "- '" + this.curSearchVal + "'" : '500') + '</h3>';
 						buf += data + '</div>';
 						self.$el.html(buf);
 					}, 'html');
@@ -158,8 +158,7 @@
 			}
 		},
 		searchChange: function (e) {
-			// 91 for right CMD / 93 for left CMD / 17 for CTL
-			if (e.keyCode !== 91 && e.keyCode !== 93 && e.keyCode !== 17) {
+			if (this.curSearchVal !== e.currentTarget.value) {
 				this.curSearchVal = e.currentTarget.value;
 				this.update();
 			}
--- a/js/client-ladder.js
+++ b/js/client-ladder.js
@@ -86,6 +86,9 @@
 	});
 
 	this.LadderRoom = HTMLRoom.extend({
+		events: {
+			'submit .search': 'submitSearch'
+		},
 		type: 'ladder',
 		title: 'Ladder',
 		initialize: function () {
@@ -146,7 +149,7 @@
 					}, function (data) {
 						if (self.curFormat !== format) return;
 						var buf = '<div class="ladder pad"><p><button name="selectFormat"><i class="fa fa-chevron-left"></i> Format List</button></p><p><button class="button" name="refresh"><i class="fa fa-refresh"></i> Refresh</button>';
-						buf += '<form class="search"><input type="text" name="searchval" class="textbox searchinput" value="' + BattleLog.escapeHTML(self.curSearchVal || '') + '" placeholder="username prefix"/><button type="submit"> Search</button></form></p>';
+						buf += '<form class="search"><input type="text" name="searchval" class="textbox searchinput" value="' + BattleLog.escapeHTML(self.curSearchVal || '') + '" placeholder="username prefix" /><button type="submit"> Search</button></form></p>';
 						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top ' + BattleLog.escapeHTML(self.curSearchVal ? "- '" + self.curSearchVal + "'" : '500') + '</h3>';
 						buf += data + '</div>';
 						self.$el.html(buf);
--- a/js/client-ladder.js
+++ b/js/client-ladder.js
@@ -86,6 +86,9 @@
 	});
 
 	this.LadderRoom = HTMLRoom.extend({
+		events: {
+			'keyup .searchinput': 'searchChange',
+		},
 		type: 'ladder',
 		title: 'Ladder',
 		initialize: function () {
@@ -104,6 +107,7 @@
 			}, this);
 		},
 		curFormat: '',
+		curSearchVal: '',
 		join: function () {},
 		leave: function () {},
 		update: function () {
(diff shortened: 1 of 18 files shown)


lib/ntbb-ladder.lib.php
```diff
@@ -224,7 +224,11 @@ function getTop() {
 			// 	$limit = 1000;
 			// }
 
-			$res = $ladderdb->query("SELECT * FROM `{$ladderdb->prefix}ladder` WHERE `formatid` = '{$this->formatid}' ORDER BY `elo` DESC LIMIT $limit");
+			if ($prefix) {
+				$res = $ladderdb->query("SELECT * FROM `{$ladderdb->prefix}ladder` WHERE `formatid` = '{$this->formatid}' AND `userid` LIKE '{$prefix}%' ORDER BY `elo` DESC LIMIT $limit");
```
Zarel: This is not an indexed query, and the ladder database is already at maximum stress despite zero unindexed queries, so I'm not particularly happy with this.

Ideally this would be done by manually flagging accounts as ladder-tour accounts, and then indexing by that flag.

scheibo: AIUI, there are 'multiple' ladder tours, even in OLT there is something like LT1...LT8 or whatever, and other metas etc can have their own tours/competitions (`bhlttar` is a current balanced hackmons ladder tour I think?). Is this something users would do at name registration? Is it something we automate based on a prefix? I don't know exactly how this scales.

I'm not against adding a tour bit to accounts - one use case for flagging accounts that came up recently is that **we could disable `/offertie` for ladder tours** in the same way we do on Smogtours/room tours. Similarly, being able to **flag an account as participating in a suspect test** would also make some Tier Leaders happy, given there was concern over how they would treat users who `/offertie`.

scheibo: Per discussion in chat I've switched to the overfetch + filter model. Does it need a `FORCE INDEX (formatid)` as well?

lib/ntbb-ladder.lib.php
```diff
@@ -224,7 +224,15 @@ function getTop() {
 			// 	$limit = 1000;
 			// }
 
-			$res = $ladderdb->query("SELECT * FROM `{$ladderdb->prefix}ladder` WHERE `formatid` = '{$this->formatid}' ORDER BY `elo` DESC LIMIT $limit");
+			if ($prefix) {
+				// The ladder database can't really handle large queries which aren't indexed, so we instead perform
+				// an indexed query for additional rows and filter them down further. This is obviously *not* guaranteed
+				// to return exactly $limit results, but should be 'good enough' in practice.
+				$overfetch = $limit * 4;
+				$res = $ladderdb->query("SELECT * (SELECT * FROM `{$ladderdb->prefix}ladder` WHERE `formatid` = '{$this->formatid}' ORDER BY `elo` DESC LIMIT $overfetch) WHERE `userid` LIKE '{$prefix}%' LIMIT $limit)");
```
Zarel: This should be using prepared statements or whatever it's called...

Zarel:
```suggestion
				$res = $ladderdb->query(
					"SELECT * (SELECT * FROM `{$ladderdb->prefix}ladder` WHERE `formatid` = ? ORDER BY `elo` DESC LIMIT ?) WHERE `userid` LIKE ? LIMIT ?)",
					[$this->formatid, $overfetch, "{$prefix}%", $limit]
				);
```

lib/ntbb-ladder.lib.php
```diff
@@ -224,7 +224,15 @@ function getTop() {
...
+			} else {
+				$res = $ladderdb->query("SELECT * FROM `{$ladderdb->prefix}ladder` WHERE `formatid` = '{$this->formatid}' ORDER BY `elo` DESC LIMIT $limit");
```
Zarel:
```suggestion
				$res = $ladderdb->query(
					"SELECT * FROM `{$ladderdb->prefix}ladder` WHERE `formatid` = ? ORDER BY `elo` DESC LIMIT ?",
					[$this->formatid, $limit]
				);
```

lib/ntbb-ladder.lib.php
Zarel: In hindsight, it doesn't need a `FORCE INDEX`; it's only the replay database with broken indexes.

js/client-ladder.js
```diff
@@ -135,22 +139,31 @@
...
+						buf += '<input type="text" id="ladderSearchBar" name="search" class="textbox searchinput" value="' + (this.curSearchVal || '') + '" placeholder="username prefix"/></p>';
+						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top ' + (this.curSearchVal ? "- '" + this.curSearchValue + "'" : '500') + '</h3>';
 						buf += data + '</div>';
 						self.$el.html(buf);
 					}, 'html');
 				}
 			}
 		},
+		searchChange: function (e) {
+			// 91 for right CMD / 93 for left CMD / 17 for CTL
```
Zarel: Why this set of detections? Why not just test `this.curSearchVal !== e.currentTarget.value`?

js/client-ladder.js
```diff
@@ -86,6 +86,9 @@
 	});
 
 	this.LadderRoom = HTMLRoom.extend({
+		events: {
+			'keyup .searchinput': 'searchChange',
```
Zarel: Also, if you're looking for changes, `input` is probably the right event to listen to? It includes non-keyboard-based inputs, like right-click -> Paste.

js/client-ladder.js
scheibo: I coped both of this (and below) from the teambuilder search. Not entirely sure why its the case either, which probably means I shouldnt have added it. But is it relevant in teambuilder even?

js/client-ladder.js
scheibo: see above. theres a very non-zero change this was the wrong choice, i copied and pasted from teambuilder

js/client-ladder.js
Zarel: The teambuilder search isn't implemented by me and so is worse than things implemented by me. ;)

js/client-ladder.js
```diff
@@ -135,22 +139,30 @@
 				var self = this;
 				this.$el.html('<div class="ladder pad"><p><button name="selectFormat"><i class="fa fa-chevron-left"></i> Format List</button></p><p><em>Loading...</em></p></div>');
 				if (app.localLadder) {
-					app.send('/cmd laddertop ' + format);
+					app.send('/cmd laddertop ' + format + (this.curSearchVal ? ' ,' + this.curSearchVal : ''));
 				} else {
 					$.get('/ladder.php', {
```
Zarel: On the other hand, I'm not sure we want to send a new HTTP request for every keypress. Unlike local searches, it probably _does_ make sense to either wait until the user presses Enter, or at least on a debounce.

Zarel: (Especially not an HTTP request that kicks off a slow SQL query.)

Zarel: I would explicitly wait until the user presses Enter, actually. With a "Filter" submit button.

scheibo:
FECK

js/client-ladder.js
```diff
@@ -135,22 +136,28 @@
...
 						server: Config.server.id.split(':')[0],
-						output: 'html'
+						output: 'html',
+						prefix: this.curSearchVal
 					}, function (data) {
 						if (self.curFormat !== format) return;
-						var buf = '<div class="ladder pad"><p><button name="selectFormat"><i class="fa fa-chevron-left"></i> Format List</button></p><p><button class="button" name="refresh"><i class="fa fa-refresh"></i> Refresh</button></p>';
-						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top 500</h3>';
+						var buf = '<div class="ladder pad"><p><button name="selectFormat"><i class="fa fa-chevron-left"></i> Format List</button></p><p><button class="button" name="refresh"><i class="fa fa-refresh"></i> Refresh</button>';
+						buf += '<form><input type="text" id="ladderSearchBar" name="search" class="textbox searchinput" value="' + (self.curSearchVal || '') + '" placeholder="username prefix"/><button name="search" type="submit"> Search</button></form></p>';
```
Zarel:
```suggestion
						buf += '<form><input type="text" id="ladderSearchBar" name="search" class="textbox searchinput" value="' + (self.curSearchVal || '') + '" placeholder="username prefix"/><button type="submit"> Search</button></form></p>';
```

Zarel: Oh, wait, I understand why your submit button is named `search` now.

scheibo: Yeah, its kind of weird that theyre both named `search`? Suggestions?

Zarel:
```suggestion
						buf += '<form class="search"><input type="text" name="searchval" class="textbox searchinput" value="' + (self.curSearchVal || '') + '" placeholder="username prefix"/><button type="submit"> Search</button></form></p>';
```

- Form submission should rely on the `submit` event. Relying on the `click` event of the submit button is unreliable.
- You should not have both an `<input>` and a `<button>` named `search` – names within a single form should be unique.
- You should not use the `id` attribute, ever (with rare exceptions). It's HTML/CSS's equivalent of a global variable. If you want to make an element findable, `class` is the usual alternative, and specifically `name` is the alternative for form properties.

js/client-ladder.js
```diff
@@ -135,22 +136,28 @@
...
+						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top ' + (self.curSearchVal ? "- '" + self.curSearchVal + "'" : '500') + '</h3>';
 						buf += data + '</div>';
 						self.$el.html(buf);
 					}, 'html');
 				}
 			}
 		},
+		search: function (e) {
```
Zarel:
```suggestion
		submitSearch: function (e) {
```

This will need to be wired up in the events table as `"submit .search": "submitSearch"`

js/client-ladder.js
```diff
@@ -135,22 +136,28 @@
...
+			this.curSearchVal = $('#ladderSearchBar').val();
```
Zarel:
```suggestion
			this.curSearchVal = this.$('input[name=searchval]').val();
```

Zarel: `this.$` will do a jQuery search in the Ladder – this is how we do the equivalent of local variables in HTML/CSS.

js/client-ladder.js
```diff
@@ -135,22 +136,28 @@
...
-						output: 'html'
+						output: 'html',
+						prefix: this.curSearchVal
 					}, function (data) {
 						if (self.curFormat !== format) return;
-						var buf = '<div class="ladder pad"><p><button name="selectFormat"><i class="fa fa-chevron-left"></i> Format List</button></p><p><button class="button" name="refresh"><i class="fa fa-refresh"></i> Refresh</button></p>';
-						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top 500</h3>';
+						var buf = '<div class="ladder pad"><p><button name="selectFormat"><i class="fa fa-chevron-left"></i> Format List</button></p><p><button class="button" name="refresh"><i class="fa fa-refresh"></i> Refresh</button>';
+						buf += '<form class="search"><input type="text" name="searchval" class="textbox searchinput" value="' + (self.curSearchVal || '') + '" placeholder="username prefix"/><button type="submit"> Search</button></form></p>';
+						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top ' + (self.curSearchVal ? "- '" + self.curSearchVal + "'" : '500') + '</h3>';
```
Zarel:
```suggestion
						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top ' + BattleLog.escapeHTML(self.curSearchVal ? "- '" + self.curSearchVal + "'" : '500') + '</h3>';
```

js/client-ladder.js
```diff
@@ -135,22 +136,28 @@
...
 						server: Config.server.id.split(':')[0],
-						output: 'html'
+						output: 'html',
+						prefix: this.curSearchVal
 					}, function (data) {
 						if (self.curFormat !== format) return;
-						var buf = '<div class="ladder pad"><p><button name="selectFormat"><i class="fa fa-chevron-left"></i> Format List</button></p><p><button class="button" name="refresh"><i class="fa fa-refresh"></i> Refresh</button></p>';
-						buf += '<h3>' + BattleLog.escapeFormat(format) + ' Top 500</h3>';
+						var buf = '<div class="ladder pad"><p><button name="selectFormat"><i class="fa fa-chevron-left"></i> Format List</button></p><p><button class="button" name="refresh"><i class="fa fa-refresh"></i> Refresh</button>';
+						buf += '<form class="search"><input type="text" name="searchval" class="textbox searchinput" value="' + (self.curSearchVal || '') + '" placeholder="username prefix"/><button type="submit"> Search</button></form></p>';
```
Zarel:
```suggestion
						buf += '<form class="search"><input type="text" name="searchval" class="textbox searchinput" value="' + BattleLog.escapeHTML(self.curSearchVal || '') + '" placeholder="username prefix"/><button type="submit"> Search</button></form></p>';
```

js/client-ladder.js
```diff
@@ -104,6 +104,7 @@
 			}, this);
```
Zarel: You don't seem to have added the events table entry yet.

scheibo: i couldnt commit it through the Web UI... I'm doing it locally and testing

scheibo:
@Zarel PTAL

PR #1246, opened 2019-03-09
Update nwjs API for later version

whalemer:

PR #1327, opened 2019-07-16
Desktop: Small package.json changes

NotTsunami:
Restrict node-remote to https as per suggestion in https://github.com/Zarel/Pokemon-Showdown-Client/pull/1246#discussion_r290869251

no-edit-menu defaults to false since v0.7.3 (https://github.com/nwjs/nw.js/wiki/Manifest-format#no-edit-menu) to current (http://docs.nwjs.io/en/latest/References/Manifest%20Format/#no-edit-menu-mac)

### DIFF
--- a/desktop/package.json
+++ b/desktop/package.json
@@ -3,8 +3,7 @@
   "version": "0.3.0",
 
   "main": "index.html",
-  "node-remote": "*://play.pokemonshowdown.com/*",
-  "no-edit-menu": false,
+  "node-remote": "https://play.pokemonshowdown.com/*",
   "window": {
 	  "icon": "icons/icon_32x32.png",
     "title": "Loading...",


Slayer95:
Does this mean the app might no longer work in China?

NotTsunami:
> Does this mean the app might no longer work in China?

According to [this](https://webmasters.stackexchange.com/a/100891) link, HTTPS itself isn't blocked, but sites are blocked by IP, or by DNS hijacking.

NotTsunami:
https://www.websitepulse.com/tools/china-firewall-test returns 200. If you test HTTP, you get the expected redirect to HTTPS, and testing directly with HTTPS also returns 200. You can test it on the 4 different China servers.

Slayer95:
Thanks for the clarification!

PR #1329, opened 2019-07-17
Do not change IVs for Hidden Power in Gen 7

jamesqo:
This addresses https://github.com/Zarel/Pokemon-Showdown/issues/4026. If the user has indicated they're building a Gen 7 team and the Pokemon in question is level 100, then the IVs are not adjusted.

### DIFF
--- a/js/storage.js
+++ b/js/storage.js
@@ -1071,12 +1071,12 @@ Storage.getPackedTeam = function (team) {
 	return team.team;
 };
 
-Storage.importTeam = function (buffer, teams, format) {
+Storage.importTeam = function (buffer, format, teams) {
 	var text = buffer.split("\n");
-	var team = teams ? null : [];
 	if (!format) {
 		format = 'gen7';
 	}
+	var team = teams ? null : [];
 	var curSet = null;
 	if (teams === true) {
 		Storage.teams = [];
@@ -1508,8 +1508,7 @@ Storage.nwLoadTeamFile = function (filename, localApp) {
 			self.teams.push({
 				name: line,
 				format: format,
-				// XXX
-				team: Storage.packTeam(Storage.importTeam('' + data)),
+				team: Storage.packTeam(Storage.importTeam('' + data), format),
 				folder: folder,
 				iconCache: '',
 				filename: filename
--- a/js/storage.js
+++ b/js/storage.js
@@ -1071,9 +1071,12 @@ Storage.getPackedTeam = function (team) {
 	return team.team;
 };
 
-Storage.importTeam = function (buffer, teams) {
+Storage.importTeam = function (buffer, teams, format) {
 	var text = buffer.split("\n");
 	var team = teams ? null : [];
+	if (!format) {
+		format = 'gen7';
+	}
 	var curSet = null;
 	if (teams === true) {
 		Storage.teams = [];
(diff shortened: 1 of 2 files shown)


Zarel:
This approach is really messy and leads to bugs.

My plan for the Preact client was a better approach that just treats implicitly and explicitly set IVs differently.

Zarel:
I would recommend implementing that approach (having a data structure for "implicitly set IVs").

js/storage.js
```diff
@@ -1196,7 +1198,8 @@ Storage.importTeam = function (buffer, teams) {
 			if (line.substr(0, 14) === 'Hidden Power [') {
 				var hptype = line.substr(14, line.length - 15);
 				line = 'Hidden Power ' + hptype;
-				if (!curSet.ivs && window.BattleTypeChart && window.BattleTypeChart[hptype]) {
+				var shouldChangeIVs = !curSet.ivs && (!format.startsWith('gen7') || (curSet.level && curSet.level !== 100));
+				if (shouldChangeIVs && window.BattleTypeChart && window.BattleTypeChart[hptype]) {
```
Slayer95: I am thinking ``!curSet.ivs || !canHyperTrain`` would lead to a better explanation of why this check is needed.

Slayer95: ... I guess Zarel's comment makes this point moot.

DaWoblefet:
If you can carry over these checks for if Gen 7 and if VGC, that'd be awesome too. VGC automatically scales to level 50 and we also have this issue.

jamesqo:
@Zarel

> I would recommend implementing that approach (having a data structure for "implicitly set IVs").

So something like this?

```js
class IVsWrapper {
    _implicit: true;
    _hp: 31;
    ...

    get hp() {
        if (!this._implicit) return this._hp;
        if (set has a Hidden Power move && can't do hyper training) {
          return 30 or 31 according to HP type;
        }
        return 31;
    }

    set hp(value) {
        this._implicit = true;
        this._hp = value;
    }
    ...
}
```

Zarel:
> Do you mean a wrapper class like this:

No. I mean a representation in which something like:

`ivs: null,` means "figure out IVs from context"

and then

`ivs: {hp: 31, atk: 31, def: 31, spa: 31, spd: 31, spe: 31}` means "explicitly-set maxed-out IVs"

> Also, are you saying in your first comment that I should implement this in the Preact client only, and leave the old client as-is?

You can do it in either client. The Preact client doesn't have a working teambuilder yet, so it might be a bit premature to implement this in that one.

Zarel:
There are several ways to do this, but one way might be to have an "automatic IVs" checkbox, which is checked by default, corresponding to `ivs: null`, in which IVs will be auto-filled, and auto-updated as generation and level are changed.

jamesqo:
@Zarel Ok, I'm working on a new PR to implement that. Since we're no longer going to be reflecting it in the export if the IVs are changed because of HP type, does that mean we should stop inserting `IVs: 0 Atk` for no attacking moves, `IVs: 0 Spe` for Gyro Ball, etc as well? The user would have to manually uncheck the box to get that in the export again.

Zarel:
Yes, that would probably be ideal.

jamesqo:
Closing in favor of https://github.com/Zarel/Pokemon-Showdown-Client/pull/1333

PR #1333, opened 2019-07-21
Implement automatic IVs

jamesqo:
This PR adds a checkbox labeled "Use automatic IVs" to the teambuilder.

- When it is checked, all of the Pokemon's IVs are inferred from context and there will be no `IVs:` line in the export, even if we adjusted it for stuff like Hidden Power.
- When it is unchecked, all non-31 IVs will be reflected in the export via `IVs`. This includes IVs that the user manually specified, as well as Hidden Power IVs, `0 Atk`, and `0 Spe` that we automatically insert. The user can manually uncheck the box, or it will automatically uncheck if the user modifies one of the IVs.
- When we import a set, we will check the box according to whether or not there is an `IVs:` line.
  - We erase IVs that are set to 31 via the textbox GUI, so that `IVs: 31 HP` doesn't randomly show up if the user accidentally adjusts the HP IV and readjusts it back. However, if they import a set with an explicit `IVs: 31 HP` (for example, if they want a Ferrothorn w/ Gyro Ball to have max speed for w/e reason) then we will preserve it and continue to show that when the user exports the set.
- `ivs` now represents only IVs that are explicitly specified by the user; if the box is checked it corresponds to `ivs: {}`. `getAutoIVs()` gets all of the IVs by inferring the missing ones.

/cc @Zarel

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -2231,7 +2231,24 @@
 			var hpTypes = ['Fighting', 'Flying', 'Poison', 'Ground', 'Rock', 'Bug', 'Ghost', 'Steel', 'Fire', 'Water', 'Grass', 'Electric', 'Psychic', 'Ice', 'Dragon', 'Dark'];
 			var hpType;
 			if (this.curTeam.gen <= 2) {
-				// TODO: Old gens
+				var autoIVs = this.getAutoIVs();
+				var hpDV = Math.floor(autoIVs.hp / 2);
+				var atkDV = Math.floor(autoIVs.atk / 2);
+				var defDV = Math.floor(autoIVs.def / 2);
+				var speDV = Math.floor(autoIVs.spe / 2);
+				var spcDV = Math.floor(autoIVs.spa / 2);
+				hpType = hpTypes[4 * (atkDV % 4) + (defDV % 4)];
+
+				var expectedHpDV = (atkDV % 2) * 8 + (defDV % 2) * 4 + (speDV % 2) * 2 + (spcDV % 2);
+				if (expectedHpDV !== hpDV) {
+					var newHpIV = expectedHpDV * 2;
+					if (newHpIV === 30) {
+						delete set.ivs.hp;
+					} else {
+						set.ivs.hp = newHpIV;
+					}
+					this.$chart.find('input[name=iv-hp]').val(expectedHpDV);
+				}
 			} else {
 				var hpTypeX = 0;
 				var i = 1;
@@ -2904,20 +2921,17 @@
 			if (hpType) {
 				if (gen > 2) {
 					minValue = (exports.BattleTypeChart[hpType].HPivs[stat] || 31) % 2;
-				} else {
-					// TODO: Old gens
+					if (!canHyperTrain) {
+						// Must have matching parity for the correct Hidden Power type
+						maxValue = 30 + minValue;
+					}
+				} else if (stat === 'atk' || stat === 'def') { // Gen 2 only uses atk/def to calc HP type
+					minValue = ((exports.BattleTypeChart[hpType].HPdvs[stat] || 15) % 4) * 2; // Results in 0, 2, 4, or 6
+					maxValue = (minValue === 6 ? 31 : 24 + minValue);
 				}
 			}
 
-			// TODO: Old gens
-			if (hpType && !canHyperTrain) {
-				// Must have matching parity for the correct Hidden Power type
-				maxValue = 30 + minValue;
-			}
-
-			if (stat === 'hp') {
-				// TODO: Old gens
-			} else if (stat === 'atk') {
+			if (stat === 'atk') {
 				if (gen > 2) {
 					useMaxValue = false; // Default to 0 Atk IVs if we don't have a physical attacking move
 					for (var i = 0; i < moves.length; ++i) {
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -19,7 +19,7 @@
 		focus: function () {
 			if (this.curTeam) {
 				this.curTeam.iconCache = '!';
-				this.curTeam.gen = this.getGen(this.curTeam.format);
+				this.curTeam.gen = getGen(this.curTeam.format);
 				Storage.activeSetList = this.curSetList;
 			}
 		},
@@ -665,7 +665,7 @@
 			i = +i;
 			this.curTeam = teams[i];
 			this.curTeam.iconCache = '!';
-			this.curTeam.gen = this.getGen(this.curTeam.format);
+			this.curTeam.gen = getGen(this.curTeam.format);
 			Storage.activeSetList = this.curSetList = Storage.unpackTeam(this.curTeam.team);
 			this.curTeamIndex = i;
 			this.update();
@@ -1293,7 +1293,7 @@
 		},
 		changeFormat: function (format) {
 			this.curTeam.format = format;
-			this.curTeam.gen = this.getGen(this.curTeam.format);
+			this.curTeam.gen = getGen(this.curTeam.format);
 			this.save();
 			if (this.curTeam.gen === 5 && !Dex.loadedSpriteData['bw']) Dex.loadSpriteData('bw');
 			this.update();
@@ -1956,7 +1956,7 @@
 
 			if (this.curTeam.gen > 2) {
 				buf += '<div class="col ivcol"><div><strong>IVs</strong></div>';
-				var ivs = set.ivs || this.getAutoIVs();
+				var ivs = set.ivs || getAutoIVs(set, this.curTeam.format);
 				for (var i in stats) {
 					var val = '' + (ivs[i]);
 					buf += '<div><input type="number" name="iv-' + i + '" value="' + BattleLog.escapeHTML(val) + '" class="textbox inputform numform" min="0" max="31" step="1" ' + (set.ivs ? '' : 'disabled') + ' /></div>';
@@ -1970,7 +1970,7 @@
 						}
 					}
 				}
-				if (hpType && !this.canHyperTrain(set)) {
+				if (hpType && !canHyperTrain(set, this.curTeam.format)) {
 					var hpIVs;
 					switch (hpType) {
 					case 'dark':
@@ -2079,7 +2079,7 @@
 				buf += '</div>';
 			} else {
 				buf += '<div class="col ivcol"><div><strong>DVs</strong></div>';
-				var ivs = set.ivs || this.getAutoIVs();
+				var ivs = set.ivs || getAutoIVs(set, this.curTeam.format);
 				for (var i in stats) {
 					var val = '' + Math.floor(ivs[i] / 2);
 					buf += '<div><input type="number" name="iv-' + i + '" value="' + BattleLog.escapeHTML(val) + '" class="textbox inputform numform" min="0" max="15" step="1" ' + (set.ivs ? '' : 'disabled') + ' /></div>';
@@ -2214,7 +2214,7 @@
 		},
 		updateIVs: function () {
 			var set = this.curSet;
-			if (!set.moves || this.canHyperTrain(set)) return;
+			if (!set.moves || canHyperTrain(set, this.curTeam.format)) return;
 			var hasHiddenPower = false;
 			for (var i = 0; i < set.moves.length; i++) {
 				if (toID(set.moves[i]).slice(0, 11) === 'hiddenpower') {
@@ -2226,7 +2226,7 @@
 			var hpTypes = ['Fighting', 'Flying', 'Poison', 'Ground', 'Rock', 'Bug', 'Ghost', 'Steel', 'Fire', 'Water', 'Grass', 'Electric', 'Psychic', 'Ice', 'Dragon', 'Dark'];
 			var hpType;
 			if (this.curTeam.gen <= 2) {
-				var ivs = set.ivs || this.getAutoIVs();
+				var ivs = set.ivs || getAutoIVs(set, this.curTeam.format);
 				var atkDV = Math.floor(ivs.atk / 2);
 				var defDV = Math.floor(ivs.def / 2);
 				hpType = hpTypes[4 * (atkDV % 4) + (defDV % 4)];
@@ -2248,7 +2248,7 @@
 			} else {
 				var hpTypeX = 0;
 				var i = 1;
-				var ivs = set.ivs || this.getAutoIVs();
+				var ivs = set.ivs || getAutoIVs(set, this.curTeam.format);
 				for (var s in ivs) {
 					hpTypeX += i * (ivs[s] % 2);
 					i *= 2;
@@ -2372,7 +2372,7 @@
 			if (newValue) {
 				// Reset the IVs to their defaults
 				set.ivs = null;
-				var autoIVs = this.getAutoIVs();
+				var autoIVs = getAutoIVs(set, this.curTeam.format);
 				for (var stat in autoIVs) {
 					var val = this.curTeam.gen > 2 ? autoIVs[stat] : Math.floor(autoIVs[stat] / 2);
 					var input = this.$chart.find('input[name=iv-' + stat + ']');
@@ -2386,7 +2386,7 @@
 				// We are still using the same IVs, but this causes them to become "hardcoded" instead
 				// of automatically inferred. For example, if the user adds Gyro Ball to the moveset afterwards,
 				// we will not adjust their speed. Also, we will include any non-31 IVs in the export.
-				if (!set.ivs) set.ivs = this.getAutoIVs();
+				if (!set.ivs) set.ivs = getAutoIVs(set, this.curTeam.format);
 
 				for (var stat in set.ivs) {
 					var input = this.$chart.find('input[name=iv-' + stat + ']');
@@ -2777,16 +2777,6 @@
 			}
 			this.save();
 		},
-		canHyperTrain: function (set) {
-			if (this.curTeam.gen < 7 || this.curTeam.format === 'gen7hiddentype') return false;
-			var format = this.curTeam.format;
-			if (!set.level || set.level === 100) return true;
-			if (format.substr(0, 3) === 'gen') format = format.substr(4);
-			if (format.substr(0, 10) === 'battlespot' || format.substr(0, 3) === 'vgc' || format === 'ultrasinnohclassic') {
-				if (set.level === 50) return true;
-			}
-			return false;
-		},
 		chooseMove: function (moveName) {
 			var set = this.curSet;
 			if (moveName === 'Return') {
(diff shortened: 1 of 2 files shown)


js/client-teambuilder.js
```diff
@@ -2216,25 +2234,16 @@
...
+				var autoIVs = this.getAutoIVs();
+				var atkDV = Math.floor(autoIVs.atk / 2);
+				var defDV = Math.floor(autoIVs.def / 2);
 				hpType = hpTypes[4 * (atkDV % 4) + (defDV % 4)];
-				var expectedHpDV = (atkDV % 2) * 8 + (defDV % 2) * 4 + (speDV % 2) * 2 + (spcDV % 2);
-				if (expectedHpDV !== hpDV) {
-					set.ivs.hp = expectedHpDV * 2;
-					if (set.ivs.hp === 30) set.ivs.hp = 31;
-					this.$chart.find('input[name=iv-hp]').val(expectedHpDV);
-				}
```
jamesqo: I removed the behavior where we automatically adjust the HP here. It will cause problems with the new way we store the IVs, and this behavior is only triggered when there's Hidden Power in the moveset anyway.

js/client-teambuilder.js
```diff
@@ -2732,23 +2776,6 @@
 			}
 			this.save();
 		},
-		unChooseMove: function (moveName) {
```
jamesqo: This method was just cleanup logic to undo any IV changes we may have made when selecting a move, but now we don't make any such changes in the first place so there's no point in keeping it around

js/client-teambuilder.js
```diff
@@ -2759,85 +2786,12 @@
...
-				this.curSet.happiness = 0;
-			} else if (moveName === 'Gyro Ball') {
-				minSpe = true;
-			}
-
-			if (this.curTeam.format === 'gen7hiddentype') return;
-
-			var minAtk = true;
-			if (set.ability === 'Battle Bond') minAtk = false; // only available through an event with 31 Atk IVs
-			var hpModulo = (this.curTeam.gen >= 6 ? 2 : 4);
```
jamesqo: I didn't carry this part over to the new method `getAutoIVs`. Shouldn't it always be 2 since Hidden Power only depends on the parity of the IV?

Zarel:
>  We erase IVs that are set to 31 via the textbox GUI, so that `IVs: 31 HP` doesn't randomly show up if the user accidentally adjusts the HP IV and readjusts it back. However, if they import a set with an explicit `IVs: 31 HP` (for example, if they want a Ferrothorn w/ Gyro Ball to have max speed for w/e reason) then we will preserve it and continue to show that when the user exports the set.

I'm not sure what this means. Ideally, if "Automatic IVs" is not checked, IVs not mentioned should default to 31.

> `ivs` now represents only IVs that are explicitly specified by the user; if the box is checked it corresponds to `ivs: {}`. `getAutoIVs()` gets all of the IVs by inferring the missing ones.

I would rather have it correspond to `ivs: null`. `ivs: {}` should never exist (I regret letting it exist in the past; it leads to more code complexity with very little benefit).

js/client-teambuilder.js
Zarel: What's the new plan for auto HP adjustment?

Zarel:
In other words, if `ivs` exists at all, it should have all six properties: hp, atk, def, spa, spd, spe. That would significantly simplify the rest of your code, I think.

Zarel:
In a packed team, `||` would mean "automatic IVs", and `|,,,,,|` would mean "full 31s". We would have to manually fill in the IVs when sending a packed team to the server, though.

Zarel:
(By the way, thank you! This is pretty close to what I was looking for.)

jamesqo:
> In other words, if ivs exists at all, it should have all six properties: hp, atk, def, spa, spd, spe. That would noticeably simplify the rest of your code, I think, and also fix a few bugs it currently has, as well as completely resolve your implicit/explicit 31 IV problem.

The problem with this is if we see `spe: 0` in `ivs`, we can't tell whether we added that ourselves because of Gyro Ball, or the user explicitly set it to 0 speed. So we don't know whether or not to include `IVs: 0 Spe` when the set is exported with auto IVs turned on. My approach solves this by having `ivs.spe === undefined` if we auto-insert it and `ivs.spe === 0` if the user did.

js/client-teambuilder.js
jamesqo: I did not add replacement code for this. However, we only trigger this code if the Pokemon knows Hidden Power anyway, so most people never see this code execute in the first place. (If you mean Hidden Power and not hit points, we still adjust the HP type in the moveset after this)

Zarel:
> The problem with this is if we see `spe: 0` in `ivs`, we can't tell whether we added that ourselves because of Gyro Ball, or the user explicitly set it to 0 speed.

Yes, we can: We know for a fact that the user explicitly set it to 0 speed.

If we added it ourselves, it would be `ivs: null`, remember?

scheibo:
why `ivs: null` instead of `ivs?: ...`? Seems the first is more explicit (in which case, id rather see `| 'infer'` or something, but the second better represents that the IVs are simply not specified.

Zarel:
> why `ivs: null` instead of `ivs?: ...`? Seems the first is more explicit (in which case, id rather see `| 'infer'` or something, but the second better represents that the IVs are simply not specified.

We've gone over this, the short answer is that the rest of the codebase has standardized on `null` for nullable values, with the exception of JSON meant to be sent over the wire, which uses `undefined` for conciseness.

`PokemonSet` is no longer meant to be sent over the wire (we have packed teams for that), so `null` it is.

I'm not _unwilling_ to change the standard, but it should probably be done all at once instead of relitigating the decision in every PR. :p

jamesqo:
@Zarel I have implemented it as you suggested, now `null` is used to represent auto IVs.

js/client-teambuilder.js
```diff
@@ -2201,7 +2201,7 @@
 
 				var changed = !set.ivs || set.ivs[stat] !== val;
 				if (changed) {
```
Zarel: Hm, this might make it too easy to accidentally turn off auto IVs? Maybe the IV choosers should be grayed out if auto-IVs is checked.

js/storage.js
```diff
@@ -1283,20 +1283,22 @@ Storage.exportTeam = function (team) {
...
+				if (firstIV) {
+					text += 'IVs: ';
+					firstIV = false;
+				} else {
+					text += ' / ';
+				}
+				text += '' + val + ' ' + BattleStatNames[j];
+			}
+			if (!firstIV) {
+				text += "  \n";
```
Zarel: What kind of value would this give if all IVs were manually set to `31`?

I'm thinking it should say `IVs: 31 all`?

Zarel: Maybe just `IVs: maxed`.

jamesqo: Well since we're wiping 31s anyway, why not just treat them as automatic for this particular case instead of introducing a new keyword? That seems to be what [`packTeam` currently does](https://github.com/Zarel/Pokemon-Showdown-Client/blob/8582b6cf52221ccc170956f883ae4293819171c6/js/storage.js#L736-L737)

Zarel: Because we need to distinguish "automatic IVs" from "all 31 IVs".

jamesqo: @Zarel I realized that but I didn't think it was worth it to add support for such a rare scenario. ig it's your call though.

Personally, I think `31 all` sounds better than `maxed`, and it makes it clearer what the difference is between that and using the default IVs.

Zarel: I think it's a really common scenario.

I don't think I'm strongly opinionated between `31 all` and `maxed` – but I think they're equally clear about the difference between that and the default IVs.

js/storage.js
```diff
@@ -1283,18 +1281,24 @@ Storage.exportTeam = function (team) {
...
-					text += ' / ';
+			var all31 = (ivs.hp === 31 && ivs.atk === 31 && ivs.def === 31 && ivs.spa === 31 && ivs.spd === 31 && ivs.spe === 31);
+			if (all31) {
+				text += 'IVs: 31 all';
+			} else {
+				for (var j in BattleStatNames) {
+					var val = curSet.ivs[j];
+					if (val === undefined || isNaN(val) || val === 31) continue;
+					if (firstIV) {
+						text += 'IVs: ';
```
Zarel: This is unnecessary now. `IVs:` will always be displayed, so you can just move this up to where `if (ivs) {` is.

jamesqo: Ok

js/storage.js
```diff
@@ -1268,78 +1260,64 @@ Storage.exportTeam = function (team) {
...
 			}
-		}
-		if (!first) {
-			text += "  \n";
-		}
-		if (curSet.moves) for (var j = 0; j < curSet.moves.length; j++) {
-			var move = curSet.moves[j];
-			if (move.substr(0, 13) === 'Hidden Power ') {
-				move = move.substr(0, 13) + '[' + move.substr(13) + ']';
+			if (!firstIV) {
```
Zarel: This is also no longer necessary, now that `IVs:` is always displayed.

Zarel: On the other hand, `text += '31 all';` can be moved here, so you don't need to manually check for all 31's.

Zarel: In other words, this should be:

```js
if (!firstIV) {
  text += '31 all';
}
text += "  \n";
```

jamesqo: We need it to tell whether to add a slash, we don't add a slash before the first IV.

jamesqo: Done

jamesqo:
@Zarel I have implemented all the feedback you suggested, so PTAL when you have time

js/storage.js
scheibo: sorry, late to the party here, but not a fan of `31 all`/`maxed` at all. why not just list out the IVs (`31 HP / 31 Atk / ...`)? if we think its that common compared to the default option (which *should* be superior), then we nee to change the default behavior.

adding new keywords here is silly

jamesqo: @scheibo Even if automatic IVs is turned off, we will omit IVs with the value 31 while exporting. If all of them are explicitly set to 31 and this code was removed, we would not export any `IVs:` line, so we would read them back in as automatic IVs. We need a keyword to handle this specific case.

scheibo: I'm proposing you just don't omit them. omitted = automatic default, specified = not default Then you don't need to make the parser any more sophisticated.

jamesqo: @scheibo If they are not omitted though, then in ie. [this case](https://www.smogon.com/dex/sm/pokemon/scizor/uber/) instead of exporting it as `IVs: 0 Spe`, the user will see `IVs: 31 HP / 31 Atk / 31 Def / 31 SpA / 31 SpD / 0 Spe` all just because they wanted to change a single IV.

scheibo: Can we do:

1) If IVs are not specified at all, do auto behavior
2) If any IVs are specified, the rest that arent specified are 31

This means the set can still be just `IVs: 0 Spe` for `31/31/31/31/31/0`, and you can actually get `31/31/31/31/31/31` behavior with just listing `31 Atk` or whatever (because if any are specified, all get filled in). You can change the implicit fill behavior to instead be the automatic defaults instead of filling with 31s, in which case to specify 'all 31s' you just need to specify 31 on the specific fields that the automatic chooser would get wrong. Either way, you don't need to introduce a new keyword

jamesqo: > Can we do:
> 
> 1. If IVs are not specified at all, do auto behavior
> 2. If any IVs are specified, the rest that arent specified are 31
> 
> This means the set can still be just `IVs: 0 Spe` for `31/31/31/31/31/0`, and you can actually get `31/31/31/31/31/31` behavior with just listing `31 Atk` or whatever (because if any are specified, all get filled in).

This is what is implemented in the current version of the PR.

> You can change the implicit fill behavior to instead be the automatic defaults instead of filling with 31s, in which case to specify 'all 31s' you just need to specify 31 on the specific fields that the automatic chooser would get wrong. Either way, you don't need to introduce a new keyword

That was the original approach I took, inferring any IVs that were not explicitly specified (eg. if you put 30 Defense IVs and added Gyro Ball, it would still adjust to 0 Spe, but only say `IVs: 30 Def` in the export). [This](https://github.com/Zarel/Pokemon-Showdown-Client/pull/1333/commits/2415e1deb82418f6903cb07f925f11657f297cbc) is the latest commit before it was rolled back. @Zarel thought it added too much complexity because you have to account for the possibility of `ivs` being only partially filled. I also think it kind of conflicts with expectations a little, if "Use automatic IVs" is unchecked then the implication should be _none_ of them are automatic, as opposed to "they will stay automatic if you don't touch them".

scheibo: First off, I want to apologize for randomly ghosting in here and not following the full commit history/context of this change. I really appreciate your patience and your contribution!

I've slept on this, and I think my main issue is actually with (1). ie/ "if IVs are not specified, do the auto behavior". **I think the default for `IVs` being absent should be 'fill with all 31s' , and nothing sophisticated whatsoever**. That behavior is easy to specify and easy for third party developers to implement. We should not require programs parsing our set export format to ape our 'magic' auto inference logic - if the teambuilder automagically sets a Pokemon with no physical attacks `Atk IV` to 0 or `Happiness` to 0 if they're using `Frustration` or whatever, we **should export that explicitly**. Yes, this means if a Pokemon has `Hidden Power` we should list the `IVs` when they differ from `31` (this means downstream set importers can be super basic!).

For backwards compatibility, this is going to require leaving in additional complexity to deal with legacy behavior, but I think taking on extra complexity here due to a poor past design decision is worth simplifying things for external developers.

I think we're going to want to wait for @Zarel to weigh in here. Sorry for dragging you into this debate :)

Zarel: Sorry, I've had a pretty busy few days.

So one thing we could do is just have an `(auto)` in parentheses next to IVs, if IVs should be auto-inferred. That preserves PS state, while allowing downstream importers to be simpler.

scheibo: I'd prefer that option to `31 all` or `maxed`.

jamesqo: @Zarel If I understand you correctly, you're saying have a special keyword for the inferred IVs case rather than the all explicitly 31 case. So if there is no IVs line, then that implicitly means `IVs: 31 HP / 31 Atk / ... / 31 Spe`. But wouldn't that mean there is a change in behavior for ie. a Pokemon with Gyro Ball since we will "see" an explicit 31 Speed, so we can't adjust it to 0 Speed?

Zarel: I would still want a `31 all` or `maxed` keyword. :[ I'm very opposed to actually listing all the 31 IVs.

Zarel: No IVs line would still correspond to auto IVs.

Zarel: In other words, while exporting:

- Export as usual.
- If they're all 31s, `IVs: 31 all`.
- If they're auto and all 31s, leave off the `IVs` line.
- If they're auto and not all 31s, render all the IVs, but add `(auto)` to the end.

While importing:

- If there's no IV line or it ends in `(auto)`, take auto IVs (don't actually read the auto IVs)
- Otherwise, read as normal, using 31 for any skipped IVs

jamesqo: @scheibo 

> I think the default for IVs being absent should be 'fill with all 31s' , and nothing sophisticated whatsoever.

Every set exported from the Smogon analyses and all of the pastes on PokePaste omit the `IVs` line under the assumption that the IVs will be automatically adjusted once they are imported into the teambuilder. If we change what it means to not have an `IVs` line, then aren't we changing the meaning of all of those imports and breaking backwards compatibility?

scheibo: > If they're all 31s, `IVs: 31 all.`

I'd have this be no `IVs` line at all

> If they're auto and all 31s, leave off the `IVs` line.

if theyre all 31, how is this not the same as the case above?

> If they're auto and not all 31s, render all the IVs, but add `(auto)` to the end.

I would change this to 'render all the IVs that aren't 31', and don't bother with auto at all.

Basically, remove auto from importing/exporting. Auto as a teambuilder concept, not one which is encoded in the persisted output at all.

Zarel: > I'd have this be no `IVs` line at all

You seem to have missed the comment right above yours for why that's impossible.

> if theyre all 31, how is this not the same as the case above?

With no IV line, edits to the moveset will change the IVs, with `IVs: 31 all`, they wouldn't.

> I would change this to 'render all the IVs that aren't 31', and don't bother with auto at all.
>
> Basically, remove auto from importing/exporting. Auto as a teambuilder concept, not one which is encoded in the persisted output at all.

Auto is a teambuilder concept we expect to be able to preserve when importing and exporting from the teambuilder. People expect to be able to backup/restore their teams without changing them.

jamesqo:
Ok, I managed implement the suggested behavior with regards to `(auto)` in import/exportTeam ([diff](https://github.com/Zarel/Pokemon-Showdown-Client/compare/bb2e06f3ab6cc884435e246ec7634244426fed8f...b3eca2fb0a4ab88bac6c465ea2ba2946f79d46be)). The meat of the change is only about ~20 lines, the rest of the diff is just moving the auto IVs code over from `client-teambuilder.js` into `storage.js` so I can access the auto IVs code from `exportTeam`, and refactoring the call sites.

Zarel:
A lot of what I wanted to avoid was having to pass the `format` to `exportTeam`, but it looks like that won't be possible with the newest approach. :(

Zarel:
Maybe sets can have an optional `autoIVs` property?

js/storage.js
```diff
@@ -1,3 +1,97 @@
+// TODO: Move these helper fns elsewhere
```
jamesqo: I'm not sure where would be a better place to move these functions so that they are accessible both in this file and `client-teambuilder.js`. It seems sloppy to make them global like this (also, ESLint is complaining and saying it can't find them from `client-teambuilder.js` causing CI to fail, though I can run these changes locally with no errors).

scheibo:
> A lot of what I wanted to avoid was having to pass the format to exportTeam, but it looks like that won't be possible with the newest approach. :(

If we're adding format to the export term, can we use that to CAP negotiate to break backwards compatibilty with `IVs`?

js/storage.js
scheibo: > You seem to have missed the comment right above yours for why that's impossible. 

It was posted 45s before mine was, so its safe to say I was composing my comment as it was posted and I didn't see it.


> Every set exported from the Smogon analyses and all of the pastes on PokePaste omit the IVs line under the assumption that the IVs will be automatically adjusted once they are imported into the teambuilder.

This is regrettable - a poor decision was made in the past, tying our hands now.

jamesqo:
@scheibo We're not adding the format in the export, but we need to know it for the purposes of telling whether it's possible to hyper train the current Pokemon and decide whether certain auto IVs are 30 or 31.

jamesqo:
@Zarel I undid the previous commits and implemented a shorter/more hackish solution based on your suggestion.

Zarel:
I think auto-IV calculation should maybe still be done in Storage, so it can be done upon unpacking.

Your previous mistake was making them top-level functions, instead of putting them in the `Storage` namespace.

jamesqo:
@Zarel `getAutoIVs` needs to know the format, so if you want to move the functions back to `Storage` we'll have to pass `format` as a parameter to `exportTeam` again. Do you prefer this approach or the previous one?

I don't think unpacking a team should involve calculating auto IVs-- it just hands `ivs: null` to the teambuilder.

Zarel:
Either unpacking or exporting a team should involve calculating auto IVs. I guess either stage works, but unpacking allows us to guarantee that `set.autoIVs` is always kept in sync with the rest of the set.

Zarel:
On the other hand, calculating on export eliminates the need for `set.autoIVs` in the first place.

Either choice has advantages, but we _should_ choose one or the other.

jamesqo:
We are calculating them on export, but we still need the `autoIVs` field around so that we can do so in the teambuilder and avoid having to pass the format to storage.

Zarel:
Yeah, I'm saying that's not a great compromise, it's a bad tradeoff. `autoIVs` should either be always up-to-date, or it shouldn't be a part of the unpacked team, I think.

jamesqo:
@Zarel I reverted to the previous approach and put the functions under `Storage`.

jamesqo:
@Zarel Do you have any time to look this over?

jamesqo:
@Zarel Do you have time to look at this?

jamesqo:
@Zarel Can you take a look at this sometime

Zarel:
Sorry, I forgot about this.

PR #1335, opened 2019-07-22
Implement HP Type and Pokeball selector in Teambuilder UI

scheibo:
Fixes https://github.com/Zarel/Pokemon-Showdown/issues/3073

Makes progress towards https://github.com/Zarel/Pokemon-Showdown/issues/1355 (still need team validation and ball protocol/messages/animation etc)

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -37,6 +37,7 @@
 
 			// details
 			'change .detailsform input': 'detailsChange',
+			'change .detailsform select': 'detailsChange',
 			'click .changeform' : 'altForm',
 			'click .altform' : 'altForm',
 
@@ -2392,6 +2393,15 @@
 				buf += '</div></div>';
 			}
 
+			if (this.curTeam.gen > 6) {
+				buf += '<div class="formrow"><label class="formlabel">HP Type:</label><div><select name="hptype">';
+				buf += '<option value=""' + (!set.hpType ? ' selected="selected"' : '') + '></option>'; // unset
+				for (var type in exports.BattleTypeChart) {
+					buf += '<option value="' + type + '"' + (set.hpType === type ? ' selected="selected"' : '') + '>'+ type + '</option>';
+				}
+				buf += '</select></div></div>';
+			}
+
 			buf += '</form>';
 			if (template.otherForms && template.baseSpecies !== 'Unown') {
 				buf += '<button class="altform">Change sprite</button>';
@@ -2424,11 +2434,20 @@
 				delete set.shiny;
 			}
 
+			// gender
 			var gender = this.$chart.find('input[name=gender]:checked').val();
 			if (gender === 'M' || gender === 'F') {
 				set.gender = gender;
 			} else {
 				delete set.gender;
+			}
+
+			// HP type
+			var hpType = this.$chart.find('select[name=hptype]').val();
+			if (hpType && exports.BattleTypeChart[hpType]) {
+				set.hpType = hpType;
+			} else {
+				delete set.hpType;
 			}
 
 			// update details cell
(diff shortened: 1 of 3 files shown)


js/client-teambuilder.js
```diff
@@ -37,6 +37,7 @@
 
 			// details
 			'change .detailsform input': 'detailsChange',
+			'change .detailsform select': 'detailsChange',
```
scheibo: Technically the selects could be in a more optimized version of `detailsChange` (like `findDetailsChange`) because they dont need to update the UI, but I dont know that optimization matters.

js/client-teambuilder.js
```diff
@@ -2424,13 +2444,31 @@
...
 			if (gender === 'M' || gender === 'F') {
 				set.gender = gender;
 			} else {
 				delete set.gender;
 			}
 
+			// pokeball
+			var pokeball = this.$chart.find('select[name=pokeball]').val();
+			// TODO: add isPokeball and use getItem(pokeball).isPokeball instead
+			if (pokeball && Dex.forGen(this.curTeam.gen).getPokeballs()[pokeball]) {
```
scheibo: This is the main place performance wise where an `isPokeball` would be useful

src/battle-dex.ts
```diff
@@ -755,6 +755,17 @@ const Dex = new class implements ModdedDex {
 		let sanitizedType = type.replace(/\?/g, '%3f');
 		return '<img src="' + Dex.resourcePrefix + 'sprites/types/' + sanitizedType + '.png" alt="' + type + '" height="14" width="32"' + (b ? ' class="b"' : '') + ' />';
 	}
+
+	getPokeballs() {
```
scheibo: I think we should probably cache these instead of looping through 300+ items each time?

Zarel:
I'd block on `isPokeball`. It's not like this is in all that much hurry.

src/battle-dex.ts
Zarel: Possibly? Stick them in `Dex.pokeballs`?

scheibo:
> I'd block on `isPokeball`. It's not like this is in all that much hurry.

Done.

scheibo:
Added 'DO NOT MERGE' conservatively because we want the server changes to land first (though maybe thats just needs `/hotpatch formats`?) When talking to old servers, the client will just display no options for the Pokeball, but won't crash or anything - I assume thats enough backwards compatibility?

src/battle-dex.ts
```diff
@@ -215,6 +215,8 @@ const Dex = new class implements ModdedDex {
 	readonly statNames: ReadonlyArray<StatName> = ['hp', 'atk', 'def', 'spa', 'spd', 'spe'];
 	readonly statNamesExceptHP: ReadonlyArray<StatNameExceptHP> = ['atk', 'def', 'spa', 'spd', 'spe'];
 
+	pokeballs: {[k: string]: Item} = null!;
```
Zarel: I would make this a `string[] | null` or `ID[] | null`

scheibo: I guess `string[] | null` because we need the `.name`, and can get `.id` from `.name`

Zarel:
There's no need for the server changes to land first, though? The client doesn't get its item database from the server.

(There are plans to change this, but they haven't currently been implemented.)

scheibo:
Right, its just a `node build full`. Forgot about this - the server changes just had to *land*, not be *deployed*

src/battle-dex.ts
```diff
@@ -215,6 +215,8 @@ const Dex = new class implements ModdedDex {
 	readonly statNames: ReadonlyArray<StatName> = ['hp', 'atk', 'def', 'spa', 'spd', 'spe'];
 	readonly statNamesExceptHP: ReadonlyArray<StatNameExceptHP> = ['atk', 'def', 'spa', 'spd', 'spe'];
 
+	pokeballs: string[] = null!;
```
Zarel:
```suggestion
	pokeballs: string[] | null = null;
```

?

Zarel: I use `null!` for "null during initialization only". This one stays null until `getPokeballs` is called.

src/battle-dex.ts
```diff
@@ -765,6 +778,7 @@ class ModdedDex {
 		Items: {} as any as {[k: string]: Item},
 		Templates: {} as any as {[k: string]: Template},
 	};
+	pokeballs: string[] = null!;
```
Zarel:
```suggestion
	pokeballs: string[] | null = null;
```

see above

PR #1338, opened 2019-07-25
Fix /hl (list/roomlist) not returning response

TheJetOU:
It would not return a response if it's empty due to the faulty if condition

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -778,10 +778,10 @@
 					} else if (['show', 'list', 'roomshow', 'roomlist'].includes(target)) {
 						// Shows a list of the current highlighting words
 						var key = target.startsWith('room') ? (Config.server.id + '#' + this.id) : 'global';
-						if (highlights[key].length > 0) {
+						if (highlights[key] && highlights[key].length > 0) {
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -778,10 +778,10 @@
 					} else if (['show', 'list', 'roomshow', 'roomlist'].includes(target)) {
 						// Shows a list of the current highlighting words
 						var key = target.startsWith('room') ? (Config.server.id + '#' + this.id) : 'global';
-						if (highlights[key].length > 0) {
+						if (highlights[key] && highlights[key].length > 0) {
 							this.add("Current highlight list " + (key === 'global' ? "(everywhere): " : "(in " + key + "): ") + highlights[key].join(", "));
 						} else {
-							this.add('Your highlight list is empty.');
+							this.add('Your highlight list ' + (key === 'global' ? 'everywhere' : 'in ' + this.id) + ' is empty.');
```
Zarel: I think "everywhere" could be misinterpreted.

I would just leave it off.

```suggestion
							this.add('Your highlight list' + (key === 'global' ? '' : ' in ' + this.id) + ' is empty.');
```

PR #1340, opened 2019-07-27
Support importing/exporting Pokeballs/Hidden Power

scheibo:


js/storage.js
```diff
@@ -1268,6 +1274,12 @@ Storage.exportTeam = function (team) {
 		if (typeof curSet.happiness === 'number' && curSet.happiness !== 255 && !isNaN(curSet.happiness)) {
 			text += 'Happiness: ' + curSet.happiness + "  \n";
 		}
+		if (curSet.pokeball) {
+			text += 'Pokeball: ' + curSet.pokeball + "  \n";
+		}
+		if (curSet.hpType) {
+			text += 'Hidden Power: ' + curSet.hpType + "  \n";
```
Zarel: I would call this `Hidden Power type`

scheibo: Thats pretty verbose? Also, this current text has the advantage of matching the UI. I'd also go with `Type` over `type`

Zarel: I suppose I'm not strongly opinionated here. My main worry is people reading at a glance might think Hidden Power is one of the moves in the moveset.

Zarel: I might even rather something like "HP Type"

scheibo: Users complained about [HP Type](https://www.smogon.com/forums/threads/rename-hp-type-to-hidden-power-type.3652893/).

Zarel: I'm tentatively fine with this.

PR #1341, opened 2019-07-31
Use fervent wish message for Rayquaza

urkerab:
Otherwise you just get "Rayquaza is reacting to the Key Stone!"


DaWoblefet:
This message also occurs for Illusioned Rayquaza Mega Evolving.

Zarel:
I don't think this PR handles that case. That sounds like that would require a protocol change to handle.

PR #1342, opened 2019-08-01
Allow filtering by userid prefix in battle search

TheJetOU:
(Server PR: Zarel/Pokemon-Showdown#5662)

Suggested and apporved here: https://www.smogon.com/forums/threads/3653073/

I made it run on button click instead of on input, because sending `/cmd roomlist` on every character is kind of ludicrous.

The button technically does have the same function as the refresh button, but it's less confusing to users.

### DIFF
--- a/js/client-rooms.js
+++ b/js/client-rooms.js
@@ -127,7 +127,7 @@
 
 			buf += '<p><label class="label">Format:</label><button class="select formatselect" name="selectFormat">(All formats)</button></p>';
 			buf += '<label>Minimum Elo: <select name="elofilter"><option value="none">None</option><option value="1100">1100</option><option value="1300">1300</option><option value="1500">1500</option><option value="1700">1700</option><option value="1900">1900</option></select></label>';
-			buf += '<p><label>Username prefix: <input type="text" name="prefixsearch" class="textbox" value="' + BattleLog.escapeHTML(this.usernamePrefix) + '" placeholder="username prefix"/><button class="button search">Search</button></label></p>';
+			buf += '<p><input type="text" name="prefixsearch" class="textbox" value="' + BattleLog.escapeHTML(this.usernamePrefix) + '" placeholder="username prefix"/><button class="button search">Search</button></p>';
 			buf += '<div class="list"><p>Loading...</p></div>';
 			buf += '</div></div>';
 
--- a/js/client-rooms.js
+++ b/js/client-rooms.js
@@ -127,7 +127,7 @@
 
 			buf += '<p><label class="label">Format:</label><button class="select formatselect" name="selectFormat">(All formats)</button></p>';
 			buf += '<label>Minimum Elo: <select name="elofilter"><option value="none">None</option><option value="1100">1100</option><option value="1300">1300</option><option value="1500">1500</option><option value="1700">1700</option><option value="1900">1900</option></select></label>';
-			buf += '<p><label>Username prefix: <input type="text" name="prefixsearch" class="textbox" value="' + BattleLog.escapeHTML(this.usernamePrefix) + '"/><button class="button search">Search</button></label></p>';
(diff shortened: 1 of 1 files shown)


js/client-rooms.js
```diff
@@ -118,14 +118,16 @@
...
 			this.$el.addClass('ps-room-light').addClass('scrollable');
 			var buf = '<div class="pad"><button class="button" style="float:right;font-size:10pt;margin-top:3px" name="close"><i class="fa fa-times"></i> Close</button><div class="roomlist"><p><button class="button" name="refresh"><i class="fa fa-refresh"></i> Refresh</button> <span style="' + Dex.getPokemonIcon('meloetta-pirouette') + ';display:inline-block;vertical-align:middle" class="picon" title="Meloetta is PS\'s mascot! The Pirouette forme is Fighting-type, and represents our battles."></span></p>';
 
 			buf += '<p><label class="label">Format:</label><button class="select formatselect" name="selectFormat">(All formats)</button></p>';
 			buf += '<label>Minimum Elo: <select name="elofilter"><option value="none">None</option><option value="1100">1100</option><option value="1300">1300</option><option value="1500">1500</option><option value="1700">1700</option><option value="1900">1900</option></select></label>';
+			buf += '<p><label>Username prefix: <input type="text" name="prefixsearch" class="textbox" value="' + BattleLog.escapeHTML(this.usernamePrefix) + '"/><button class="button search">Search</button></label></p>';
```
scheibo: I think this should match the ladder filter ('username prefix' placeholder instead of label)

TheJetOU: It looks out of place with no label. I'll show you what I mean in a couple minutes.

TheJetOU: https://i.imgur.com/vec0uat.png vs https://i.imgur.com/ykNmsJa.png.

TheJetOU: On second though, yours doesn't look too bad, but I feel like it's easier to read a label than reading the faded placeholder text.

scheibo: I'm just about consistently. How about we go with the placeholder for now to match and then in a followup  we can potentially change both of them to use labels?

TheJetOU: Sure.

scheibo: Sorry, I don't think I was clear. I didn't mean "use both", I meant, use the option from the second screenshot (no label, just placeholder) and then we can later discuss if we want to remove the placeholder in favor of labels for both filters

js/client-rooms.js
```diff
@@ -209,10 +212,10 @@
 			return this.$list.html('<p>' + buf.length + (buf.length === 100 ? '+' : '') + ' ' + BattleLog.escapeFormat(this.format) + ' ' + (buf.length === 1 ? 'battle' : 'battles') + '</p>' + buf.join(""));
 		},
 		refresh: function () {
-			var elofilter = '';
-			var elo = this.$('select[name=elofilter]').val();
-			if (elo !== 'none') elofilter = ', ' + elo;
-			app.send('/cmd roomlist ' + this.format + elofilter);
+			var usernamePrefix = this.$('input[name=prefixsearch]').val();
+			var elofilter = this.$('select[name=elofilter]').val();
+			var searchParams = [this.format, elofilter, usernamePrefix];
```
scheibo: I don't think the space in `', '` is significant looking at the server side code.

js/client-rooms.js
```diff
@@ -118,14 +118,16 @@
 		title: 'Battles',
 		isSideRoom: true,
 		events: {
-			'change select[name=elofilter]': 'refresh'
+			'change select[name=elofilter]': 'refresh',
+			'click .search': 'refresh'
```
scheibo: Not sure if `.search` is the best choice (is there a more precise way of accessing this node in the DOM like with `select[name=elofilter]`?

PR #1343, opened 2019-08-03
Added FolderToggle to the team selector for battling.

kanmingxu00:
It doesn't affect the regular selector at all, everything
functions normally. Changes only occur upon toggle on.
All the features are kept, including the current selected team.
No data formatting changes made.

UI can and should be improved upon, that's for a later date. This is simply a POC, and something my PS friends wanted to see, so I cranked out the idea.

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -1241,19 +1241,18 @@
 			}
 		},
 		selectFolder: function (key) {
-			var folderNotExpanded = this.folderNotExpanded;
 			var keyExists = false;
-			for (var i = folderNotExpanded.length - 1; i > -1; i--) {
-				if (folderNotExpanded[i] === key) {
+			for (var i = this.folderNotExpanded.length - 1; i > -1; i--) {
+				if (this.folderNotExpanded[i] === key) {
 					keyExists = true;
-					folderNotExpanded.splice(i, 1);
+					this.folderNotExpanded.splice(i, 1);
 				}
 			}
 			if (!keyExists) {
-				folderNotExpanded.push(key);
+				this.folderNotExpanded.push(key);
 			}
 			this.close();
-			app.addPopup(TeamPopup, {team: this.team, format: this.format, sourceEl: this.sourceEl, room: this.room, isMoreTeams: this.isMoreTeams, folderToggleOn: this.folderToggleOn, folderNotExpanded: folderNotExpanded});
+			app.addPopup(TeamPopup, {team: this.team, format: this.format, sourceEl: this.sourceEl, room: this.room, isMoreTeams: this.isMoreTeams, folderToggleOn: this.folderToggleOn, folderNotExpanded: this.folderNotExpanded});
 		},
 		foldersToggle: function () {
 			this.close();
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -1106,7 +1106,7 @@
 			this.team = data.team;
 			this.format = data.format;
 			this.room = data.room;
-			this.moreTeamsTemp = data.moreTeamsTemp || false;
+			this.isMoreTeams = data.isMoreTeams || false;
 			var format = BattleFormats[data.format];
 
 			var teamFormat = (format.teambuilderFormat || (format.isTeambuilderFormat ? data.format : false));
@@ -1139,53 +1139,56 @@
 						}
 					}
 					if (this.folderToggleOn) {
-						var keys = Object.keys(folders);
-						for (var i = 0; i < keys.length; i++) {
-							var folderData = folders[keys[i]];
+						for (var key in folders) {
+							var folderData = folders[key];
 							var exists = false;
 							for (var j = 0; j < this.folderNotExpanded.length; j++) {
-								if (this.folderNotExpanded[j] === keys[i]) {
+								if (this.folderNotExpanded[j] === key) {
 									exists = true;
 									break;
 								}
 							}
 							if (!exists) {
-								bufs[curBuf] += '<li><button name="selectFolder" class="folderButtonOpen folderButtonOver" value="' + keys[i] + '"><i class="fa fa-folder-open" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML(keys[i]) + '</button></li>';
+								bufs[curBuf] += '<li><button name="selectFolder" class="folderButtonOpen folderButtonOver" value="' + key + '"><i class="fa fa-folder-open" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML(key) + '</button></li>';
 								count++;
 								if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
 								for (var j = 0; j < folderData.length; j++) {
 									var selected = (folderData[j].id === curTeam);
-									bufs[curBuf] += '<li><button name="selectTeam" value="' + folders[keys[i]][j].id + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(folderData[j].name) + '</button></li>';
+									bufs[curBuf] += '<li><button name="selectTeam" value="' + folders[key][j].id + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(folderData[j].name) + '</button></li>';
 									count++;
 									if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
 								}
 							} else {
-								bufs[curBuf] += '<li><button name="selectFolder" class="folderButton" value="' + keys[i] + '"><i class="fa fa-folder" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML(keys[i]) + '</button></li>';
+								bufs[curBuf] += '<li><button name="selectFolder" class="folderButton" value="' + key + '"><i class="fa fa-folder" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML(key) + '</button></li>';
 								count++;
 								if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
 							}
 						}
-						var noFolderTemp = false;
+						var isNoFolder = false;
 						for (var i = 0; i < this.folderNotExpanded.length; i++) {
 							if (this.folderNotExpanded[i] === "(No Folder)") {
-								noFolderTemp = true;
+								isNoFolder = true;
 								break;
 							}
 						}
-						if (!noFolderTemp) {
+						if (!isNoFolder) {
 							bufs[curBuf] += '<li><button name="selectFolder" class="folderButtonOpen folderButtonOver" value="(No Folder)"><i class="fa fa-folder-open" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML("(No Folder)") + '</button></li>';
+							count++;
+							if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
 						} else {
 							bufs[curBuf] += '<li><button name="selectFolder" class="folderButton" value="(No Folder)"><i class="fa fa-folder" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML("(No Folder)") + '</button></li>';
+							count++;
+							if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
 						}
-						if (!noFolderTemp) {
+						if (!isNoFolder) {
 							for (var i = 0; i < teams.length; i++) {
 								if ((!teams[i].format && !teamFormat) || teams[i].format === teamFormat) {
 									var selected = (i === curTeam);
 									if (teams[i].folder === "") {
 										bufs[curBuf] += '<li><button name="selectTeam" value="' + i + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(teams[i].name) + '</button></li>';
 										count++;
+										if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
 									}
-									if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
 								}
 							}
 						}
@@ -1196,9 +1199,9 @@
 				} else {
 					bufs[curBuf] = '<li><button name="teambuilder" class="button"><strong>Teambuilder</strong></button></li>';
 					bufs[curBuf] += '<li><h3>All teams</h3></li>';
-					this.moreTeamsTemp = true;
+					this.isMoreTeams = true;
 				}
-				if (this.moreTeamsTemp) {
+				if (this.isMoreTeams) {
 					for (var i = 0; i < teams.length; i++) {
 						if (teamFormat && teams[i].format === teamFormat) continue;
 						var selected = (i === curTeam);
@@ -1222,11 +1225,11 @@
 			}
 		},
 		events: {
-			'click [type="checkbox"]':'foldersToggle',
+			'click [type="checkbox"]': 'foldersToggle',
 		},
 		moreTeams: function () {
 			this.close();
-			app.addPopup(TeamPopup, {team: this.team, format: this.format, sourceEl: this.sourceEl, room: this.room, moreTeamsTemp: true, folderToggleOn: this.folderToggleOn, folderNotExpanded: this.folderNotExpanded});
+			app.addPopup(TeamPopup, {team: this.team, format: this.format, sourceEl: this.sourceEl, room: this.room, isMoreTeams: true, folderToggleOn: this.folderToggleOn, folderNotExpanded: this.folderNotExpanded});
 		},
 		teambuilder: function () {
 			var teamFormat = this.teamFormat;
@@ -1250,11 +1253,11 @@
 				folderNotExpanded.push(key);
 			}
 			this.close();
-			app.addPopup(TeamPopup, {team: this.team, format: this.format, sourceEl: this.sourceEl, room: this.room, moreTeamsTemp: this.moreTeamsTemp, folderToggleOn: this.folderToggleOn, folderNotExpanded: folderNotExpanded});
+			app.addPopup(TeamPopup, {team: this.team, format: this.format, sourceEl: this.sourceEl, room: this.room, isMoreTeams: this.isMoreTeams, folderToggleOn: this.folderToggleOn, folderNotExpanded: folderNotExpanded});
 		},
 		foldersToggle: function () {
 			this.close();
-			app.addPopup(TeamPopup, {team: this.team, format: this.format, sourceEl: this.sourceEl, room: this.room, moreTeamsTemp: this.moreTeamsTemp, folderToggleOn: !this.folderToggleOn, folderNotExpanded: this.folderNotExpanded});
+			app.addPopup(TeamPopup, {team: this.team, format: this.format, sourceEl: this.sourceEl, room: this.room, isMoreTeams: this.isMoreTeams, folderToggleOn: !this.folderToggleOn, folderNotExpanded: this.folderNotExpanded});
 		},
 		selectTeam: function (i) {
 			i = +i;
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -1126,7 +1126,7 @@
 							if (!this.folderToggleOn) {
 								bufs[curBuf] += '<li><button name="selectTeam" value="' + i + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(teams[i].name) + '</button></li>';
 								count++;
-								if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+								if (count % bufBoundary === 0 && curBuf < 4) curBuf++;
 							} else {
 								var folderName = teams[i].folder || "";
 								if (folderName) {
@@ -1151,17 +1151,17 @@
 							if (!exists) {
 								bufs[curBuf] += '<li><button name="selectFolder" class="folderButtonOpen folderButtonOver" value="' + key + '"><i class="fa fa-folder-open" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML(key) + '</button></li>';
 								count++;
-								if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+								if (count % bufBoundary === 0 && curBuf < 4) curBuf++;
 								for (var j = 0; j < folderData.length; j++) {
 									var selected = (folderData[j].id === curTeam);
 									bufs[curBuf] += '<li><button name="selectTeam" value="' + folders[key][j].id + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(folderData[j].name) + '</button></li>';
 									count++;
-									if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+									if (count % bufBoundary === 0 && curBuf < 4) curBuf++;
 								}
 							} else {
 								bufs[curBuf] += '<li><button name="selectFolder" class="folderButton" value="' + key + '"><i class="fa fa-folder" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML(key) + '</button></li>';
 								count++;
-								if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+								if (count % bufBoundary === 0 && curBuf < 4) curBuf++;
 							}
 						}
 						var isNoFolder = false;
@@ -1172,11 +1172,11 @@
 							}
 						}
 						if (!isNoFolder) {
-							bufs[curBuf] += '<li><button name="selectFolder" class="folderButtonOpen folderButtonOver" value="(No Folder)"><i class="fa fa-folder-open" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML("(No Folder)") + '</button></li>';
+							bufs[curBuf] += '<li><button name="selectFolder" class="folderButtonOpen folderButtonOver" value="(No Folder)"><i class="fa fa-folder-open" style="margin-right: 7px; margin-left: 4px;"></i>(No Folder)</button></li>';
 						} else {
-							bufs[curBuf] += '<li><button name="selectFolder" class="folderButton" value="(No Folder)"><i class="fa fa-folder" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML("(No Folder)") + '</button></li>';
+							bufs[curBuf] += '<li><button name="selectFolder" class="folderButton" value="(No Folder)"><i class="fa fa-folder" style="margin-right: 7px; margin-left: 4px;"></i>(No Folder)</button></li>';
 							count++;
-							if (count % bufBoundary == 0 && count != 0 && curBuf < 4) curBuf++;
+							if (count % bufBoundary === 0 && count != 0 && curBuf < 4) curBuf++;
 						}
 						if (!isNoFolder) {
 							for (var i = 0; i < teams.length; i++) {
@@ -1185,7 +1185,7 @@
 									if (teams[i].folder === "") {
 										bufs[curBuf] += '<li><button name="selectTeam" value="' + i + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(teams[i].name) + '</button></li>';
 										count++;
-										if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+										if (count % bufBoundary === 0 && curBuf < 4) curBuf++;
 									}
 								}
 							}
@@ -1205,7 +1205,7 @@
 						var selected = (i === curTeam);
 						bufs[curBuf] += '<li><button name="selectTeam" value="' + i + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(teams[i].name) + '</button></li>';
 						count++;
-						if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+						if (count % bufBoundary === 0 && curBuf < 4) curBuf++;
 					}
 				} else {
 					bufs[curBuf] += '<li><button name="moreTeams" class="button">Show all teams</button></li>';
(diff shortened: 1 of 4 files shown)


kanmingxu00:
Comments are left in there for future use, can be removed and force pushed over if needed.

kanmingxu00:
https://drive.google.com/file/d/1Nz7uA9DTPmOOON7LJe_fGikCTipekdIv/view?usp=sharing

kanmingxu00:
Description: Ability to Toggle between folders and show all in teambuilder, within the format type.
folders only display teams in that certain format (can be changed)

js/client-mainmenu.js
```diff
@@ -1175,6 +1225,29 @@
 				teambuilder.selectFolder(teamFormat);
 			}
 		},
+		selectFolder: function (key) {
+			var sourceEl = this.sourceEl;
```
scheibo: why are you declaring variables for each of these as opposed to just doing `{team: this.team, ...}` etc when you call `addPopup` (ditto in `foldersToggle`)?

scheibo: I guess because youre copying the pattern from `moreTeams`? I dont know the pattern was that great to begin with.

js/client-mainmenu.js
```diff
@@ -1158,13 +1206,15 @@
 				this.$el.html('<ul class="popupmenu">' + bufs[0] + '</ul>');
 			}
 		},
-		moreTeams: function () {
+		moreTeamsPress: function () {
```
scheibo: Not sure about this rename - I don't think we need action verbs (press, click, toggle, etc) in the names?

js/client-mainmenu.js
```diff
@@ -1122,12 +1122,60 @@
 					for (var i = 0; i < teams.length; i++) {
 						if ((!teams[i].format && !teamFormat) || teams[i].format === teamFormat) {
 							var selected = (i === curTeam);
-							bufs[curBuf] += '<li><button name="selectTeam" value="' + i + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(teams[i].name) + '</button></li>';
-							count++;
-							if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+							if (this.folderToggleOn !== true) {
```
scheibo: whats wrong with `(!this.folderToggleOn)`?

js/client-mainmenu.js
```diff
@@ -1122,12 +1122,60 @@
...
+								if (folderName !== "") {
+									if (folders[folderName] === undefined) folders[folderName] = []; // create empty array to push to
+									var thisTeam = teams[i];
+									thisTeam.id = i; // create this ID for future sorting purposes, with respect to storage
+									folders[folderName].push(thisTeam); // shows the folders with only the content of format, move out if needed
+								}
+							}
 						}
 					}
+					if (this.folderToggleOn) {
```
scheibo: this should just be an `} else {` block, no?

js/client-mainmenu.js
```diff
@@ -1122,12 +1122,60 @@
...
+						for (var i = 0; i < keys.length; i++) {
+							var folderData = folders[keys[i]];
+							var exists = false;
+							for (var j = 0; j < this.folderExpanded.length; j++) {
+								if (this.folderExpanded[j] === keys[i]) {
+									exists = true;
+								}
+							}
+							if (exists) {
+								for (var j = 0; j < folderData.length; j++) { // fuck i need to compare with teams
```
scheibo: ?

js/client-mainmenu.js
```diff
@@ -1122,12 +1122,60 @@
...
-							count++;
-							if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+							if (this.folderToggleOn !== true) {
+								bufs[curBuf] += '<li><button name="selectTeam" value="' + i + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(teams[i].name) + '</button></li>';
+								count++;
+								if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+							} else {
+								var folderName = teams[i].folder || "";
+								if (folderName !== "") {
+									if (folders[folderName] === undefined) folders[folderName] = []; // create empty array to push to
```
scheibo: I think maybe you dont need to be commenting each line?

js/client-mainmenu.js
```diff
@@ -1122,12 +1122,60 @@
...
-							bufs[curBuf] += '<li><button name="selectTeam" value="' + i + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(teams[i].name) + '</button></li>';
-							count++;
-							if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+							if (this.folderToggleOn !== true) {
+								bufs[curBuf] += '<li><button name="selectTeam" value="' + i + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(teams[i].name) + '</button></li>';
+								count++;
+								if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+							} else {
+								var folderName = teams[i].folder || "";
+								if (folderName !== "") {
```
scheibo: just `if (folderName)` should be fine

js/client-mainmenu.js
scheibo: similarly, you can just do an `if (folders[folderName])` check. theres nothing *wrong* per se with being super explicit, but its just added verbosity and makes the reader question 'why is the code so explicit here, could `folders[folderName]` really be falsey and not undefined?'

js/client-mainmenu.js
```diff
@@ -1122,12 +1122,60 @@
...
+									var thisTeam = teams[i];
+									thisTeam.id = i; // create this ID for future sorting purposes, with respect to storage
+									folders[folderName].push(thisTeam); // shows the folders with only the content of format, move out if needed
+								}
+							}
 						}
 					}
+					if (this.folderToggleOn) {
+						var keys = Object.keys(folders);
+						for (var i = 0; i < keys.length; i++) {
```
scheibo: Do we ban `for..in`?

js/client-mainmenu.js
```diff
@@ -1122,12 +1122,60 @@
...
+							var folderData = folders[keys[i]];
+							var exists = false;
+							for (var j = 0; j < this.folderExpanded.length; j++) {
+								if (this.folderExpanded[j] === keys[i]) {
+									exists = true;
```
scheibo: `break;`

js/client-mainmenu.js
```diff
@@ -1122,12 +1122,60 @@
...
+									count++;
+									if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+								}
+							} else {
+								bufs[curBuf] += '<li><button name="selectFolder" class="button" value="' + keys[i] + '">' + BattleLog.escapeHTML(keys[i]) + '</button></li>';
+								count++;
+								if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+							}
+						}
+						// for (var i = 0; i < teams.length; i++) { // shows the other teams not in folders, for later changes
```
scheibo: Please don't leave in commented code. Either include it in this PR or remove and add in a followup PR.

js/client-mainmenu.js
```diff
@@ -1144,7 +1192,7 @@
 						if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
 					}
 				} else {
-					bufs[curBuf] += '<li><button name="moreTeams" class="button">Show all teams</button></li>';
+					bufs[curBuf] += '<li><button name="moreTeamsPress" class="button">Show all teams</button></li>'; // its here
```
scheibo:
```suggestion
					bufs[curBuf] += '<li><button name="moreTeamsPress" class="button">Show all teams</button></li>';
```

js/client-mainmenu.js
scheibo:
```suggestion
								for (var j = 0; j < folderData.length; j++) {
```

scheibo:
Thanks for your contribution - the video demo is super helpful! @Zarel can comment on the actual feature, but I've added some first pass comments regarding the implementation

js/client-mainmenu.js
kanmingxu00: Yessir, simply to follow design pattern @scheibo

js/client-mainmenu.js
kanmingxu00: Changed the name because this.moreTeams was erroring, as there was the this.moreTeam instance and the this.moreTeams method, and it was picking up the method. Not too familiar with backbone js, so if there's an alternative let me know

kanmingxu00: Notice that I also created the instance field moreTeams, so I could manipulate it in the other action methods (before it was literally just false -> true, with no toggle. this also makes it a lot easier to toggle other teams, if needed)

js/client-mainmenu.js
kanmingxu00: I can change this all if needed

js/client-mainmenu.js
kanmingxu00: No, this call comes after the for loop

js/client-mainmenu.js
kanmingxu00: This is the call inside, and folders is the container for the folders. This starts out as an empty json object. When I first initialize that parameter of folderName, I want to create an array. So I'm declaring as an empty array when it's undefined. I'm not being needlessly explicit here, but is there a better way to write that?

js/client-mainmenu.js
kanmingxu00: Oh, nice javascript. I tried forEach and it failed so I gave up on all the iterable stuff. Going to implement that now... thanks

kanmingxu00:
Functionality exactly the same, cleaned up code. Ran into strange problem where for..in was giving me the indexes instead of the values of array.

kanmingxu00:
As I said, UI can be changed @Zarel as this is simply a proof of concept. Let me know if there's a specific way you want me to format the buttons or whatever

Zarel:
Hm. I improved the team selector in Preact but I don't think I remembered to implement folders.

Part of me wants the default view to show folders instead of all teams. But it would interfere with "the top team is also your most recent team, if you add new teams to the top".

I guess showing the most recent _folder_, intermixed with teams, might make sense. (Use the folder icon).

kanmingxu00:
A bit confused by this comment. My original idea was to show all the folders first, and then the teams without folders when u toggle folders on. I can do that immediately, I have the code for that. Is this desired?

Zarel:
On further reflection, I think the simplest and easiest way to do this would be to treat "no folder" as a folder, and then show folders as sections of one list.

(No folder)
- Team 1
- Team 3

Stall teams
- Team 2
- Team 4

Like that.

kanmingxu00:
Got it, I'd had it like that before. I think No folder should go after though, no?

Zarel:
Yeah, that might be better.

Zarel:
I was originally thinking they should be in order of teams, so if your newest team didn't have a folder, "no folder" would be on top.

But I think that's probably unnecessary complexity and would make "no folder" harder to find, too.

kanmingxu00:
I can go through that all eventually down the road. Thanks for your input. I'll upload a video of what the feature looks like when I'm done.

kanmingxu00:
https://drive.google.com/file/d/1bXRrd86RRlCYVCLKJP3f63q5C71jCle_/view?usp=sharing

kanmingxu00:
Looks godawful in darkmode, but that's easy fix. Let me know what you think @Zarel

kanmingxu00:
As commit message description, top folders are the ones with the most recent team.

kanmingxu00:
This is a finished product, I think. I will link a video. Full compatibility with dark mode and light mode, everything works as expected.

kanmingxu00:
https://drive.google.com/file/d/1Vu_VI4uZGrCAJyvCgdmOHzdwY_TuLqUT/view?usp=sharing
Finished product video.

Zarel:
I was thinking the folders would be open by default.

kanmingxu00:
I see.

kanmingxu00:
Alright, Ill push those changes when I get home. Anything else? I would like to be finished before I have to go work tmr. Thank you!

Zarel:
The "Show all" button seems unclearly named.

Zarel:
I think it might be clearer as a checkbox? "Group by folders".

kanmingxu00:
I see, ok

style/client.css
```diff
@@ -197,19 +197,44 @@ select {
...
+.dark .popupmenu button.folderButton,
 .dark .popupmenu button.button {
 	background: #484848;
 	box-shadow: inset 0 3px 4px rgba(255, 255, 255, 0.25);
 	border-color: #A9A9A9;
 	text-shadow: none;
 	color: #F9F9F9;
 }
+
+.dark .popupmenu button.folderButtonOpen {
```
scheibo: this can be added to line 203 and then you can just make this:

```
background: linear-gradient(to bottom, #2a2a2a, #313131);
```

because I believe thats the only difference?

style/client.css
```diff
@@ -197,19 +197,44 @@ select {
...
+	box-shadow: inset 0 3px 4px rgba(255, 255, 255, 0.25);
+	border-color: #A9A9A9;
+	text-shadow: none;
+	color: #F9F9F9;
+	background: linear-gradient(to bottom, #2a2a2a, #313131);
+}
+
+.dark .popupmenu button.folderButtonOpen:hover {
+	color: #F9F9F9;
+	background: #606060;
```
scheibo: similarly, merge this wit the normal button hover and then just change the `color` for this rule

style/client.css
```diff
@@ -197,19 +197,44 @@ select {
...
+	border-color: #EEEEEE;
+}
+
 .dark .button:hover,
 .dark .tabbar a.button:hover {
 	background: #606060;
 	border-color: #EEEEEE;
 }
+
+.dark .folderButton:hover {
```
scheibo: ditto

scheibo: in fact, you can do:

```css
.dark .popupmenu button.folderButtonOpen:hover,
.dark .folderButton:hover {
  color: #F9F9F9;
}
```

style/client.css
```diff
@@ -375,6 +400,58 @@ select {
...
 	padding: 3px 8px;
 }
+
+.popupmenu i {
+	display: inline-block;
+	width: 16px;
+	color: #777777;
+}
+
+.popupmenu button.folderButton {
```
scheibo: please share with `folderButtonOpen` and only add a second rule for the diffs

js/client-rooms.js
```diff
@@ -182,7 +182,7 @@
 			}
 			return '<div><a href="' + app.root + id + '" class="ilink">' + roomDesc + '</a></div>';
 		},
-		submitSearch: function(e) {
+		submitSearch: function (e) {
```
scheibo: please sync with master, @Zarel fixed my style error already

js/client-mainmenu.js
scheibo: @Zarel - can you explain this pattern of defining `var`s for seemingly no reason (here and below)

js/client-mainmenu.js
```diff
@@ -1175,6 +1244,38 @@
...
+		selectFolder: function (key) {
+			var sourceEl = this.sourceEl;
+			var team = this.team;
+			var format = this.format;
+			var room = this.room;
+			var folderToggleOn = this.folderToggleOn;
+			var folderExpanded = this.folderExpanded;
+			var moreTeams = this.moreTeams;
+			var keyExists = false;
+			for (var i = folderExpanded.length - 1; i > -1; i--) {
```
scheibo: This is a place that might be worth commenting on. Though [`Array#filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) is probably a better bet here. It probably doesn't need a polyfill in `battle-dex.ts`.

js/client-mainmenu.js
Zarel: I don't remember; I assume it was just to make their use easier. It's probably something someone else did to copy my earlier code,  which defines `team` and `format` because they're used a lot and omitting `this.` makes them more readable?

Zarel: I would definitely recommend not defining short variables if they're only used once.

js/client-mainmenu.js
scheibo: Ah yes, I missed that you added a new instance field, which would necessitate renaming this.

js/client-mainmenu.js
scheibo: ...

js/client-mainmenu.js
scheibo: Ah sorry, had trouble parsing the indentation!

js/client-mainmenu.js
Zarel: If it needs a polyfill, I'll add it myself later.

Zarel:
Also can you send me a team export so I can test it?

js/client-mainmenu.js
```diff
@@ -1101,33 +1100,101 @@
 			} else if (teams.length > 27) {
 				bufBoundary = Math.ceil(teams.length / 2);
 			}
-
+			this.folderExpanded = data.folderExpanded || [];
+			if (data.folderToggleOn === undefined) {
```
scheibo: this seems error prone. if you want this new field to default to true, make it `dataFolderOff` and then just do `!!dataFolderOff`

kanmingxu00:
https://pokepast.es/4eff34f3bc50bc6c
@Zarel full paste of my teams.

kanmingxu00:
https://drive.google.com/file/d/1mVv8Rck1DynCPIHhA3XVG9U-j2ny7ibM/view?usp=sharing
Updated version of the 
Made some last second changes in the CSS for colors, so test it for the final final result.

js/client-mainmenu.js
kanmingxu00: i just took it off completely, the other variables dont have this either. just make sure we declare it every time.

js/client-mainmenu.js
```diff
@@ -1144,7 +1207,7 @@
 						if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
 					}
 				} else {
-					bufs[curBuf] += '<li><button name="moreTeams" class="button">Show all teams</button></li>';
+					bufs[curBuf] += '<li><button name="moreTeamsPress" class="button">Show all teams</button></li>';
```
Zarel: I'm confused why this was renamed?

style/client.css
```diff
@@ -197,24 +197,42 @@ select {
...
+}
+
 .dark .button:hover,
+.dark .popupmenu button.folderButtonOpen:hover,
+.dark .folderButton:hover,
+.dark .popupmenu button.folderButton:hover,
+.dark .popupmenu button.button:hover,
 .dark .tabbar a.button:hover {
-	background: #606060;
+	background: #606060 !important;
```
Zarel: Would rather not use `!important`

js/client-mainmenu.js
kanmingxu00: Created instance with moreTeams, it was getting confused with this one in the call

kanmingxu00: Could just change instance name...

style/client.css
kanmingxu00: Alright 👍

js/client-mainmenu.js
kanmingxu00: yea sorry, ihad to create this because i'm reinitializing the instance upon all the button calls, and didn't want the moreteams things to just close every time. but i'm changing the naming right now

kanmingxu00:
Ignore last commit, multiple dom elements with same ID...

kanmingxu00:
Any other comments, @Zarel ?

js/client-mainmenu.js
kanmingxu00: I resolved this.

js/client-mainmenu.js
```diff
@@ -1101,30 +1100,94 @@
...
 			}
-
+			this.folderNotExpanded = data.folderNotExpanded || [];
+			this.folderToggleOn = data.folderToggleOn;
+			var folders = {};
 			this.team = data.team;
 			this.format = data.format;
 			this.room = data.room;
-
+			this.moreTeamsTemp = data.moreTeamsTemp || false;
```
scheibo: `moreTeamsTemp` is a very poor name. this is a boolean, so a name beginning with `is` or `should` would probably be a good place to start.

js/client-mainmenu.js
scheibo: I'm still confused as to why this isn't `for (var key in folders)` ?

js/client-mainmenu.js
```diff
@@ -1175,6 +1237,25 @@
 				teambuilder.selectFolder(teamFormat);
 			}
 		},
+		selectFolder: function (key) {
+			var folderNotExpanded = this.folderNotExpanded;
```
scheibo: it seems like this `var` is unnecessary - saving having to type `this.` doesn't buy all that much? (this doesnt perform a copy, youd need to `.slice(0)` for that)

js/client-mainmenu.js
```diff
@@ -1158,13 +1221,12 @@
 				this.$el.html('<ul class="popupmenu">' + bufs[0] + '</ul>');
 			}
 		},
+		events: {
+			'click [type="checkbox"]':'foldersToggle',
```
scheibo:
```suggestion
			'click [type="checkbox"]': 'foldersToggle',
```

js/client-mainmenu.js
```diff
@@ -1101,30 +1100,94 @@
...
+									count++;
+									if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+								}
+							} else {
+								bufs[curBuf] += '<li><button name="selectFolder" class="folderButton" value="' + keys[i] + '"><i class="fa fa-folder" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML(keys[i]) + '</button></li>';
+								count++;
+								if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+							}
+						}
+						var noFolderTemp = false;
```
scheibo: also a terrible name

js/client-mainmenu.js
```diff
@@ -1101,30 +1100,95 @@
...
+					bufs[curBuf] = '<li><h3 style="margin-bottom: 5px;">' + BattleLog.escapeFormat(teamFormat) + ' teams</h3></li>';
+					bufs[curBuf] += '<li style="padding-bottom: 5px;"><input type="checkbox"' + (this.folderToggleOn ? ' checked' : '') + '><strong>Group by folders</strong></button></li>';
 					for (var i = 0; i < teams.length; i++) {
 						if ((!teams[i].format && !teamFormat) || teams[i].format === teamFormat) {
 							var selected = (i === curTeam);
-							bufs[curBuf] += '<li><button name="selectTeam" value="' + i + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(teams[i].name) + '</button></li>';
+							if (!this.folderToggleOn) {
+								bufs[curBuf] += '<li><button name="selectTeam" value="' + i + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(teams[i].name) + '</button></li>';
+								count++;
+								if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
```
scheibo:
```suggestion
								if (count % bufBoundary === 0 && curBuf < 4) curBuf++;
```

js/client-mainmenu.js
```diff
@@ -1101,30 +1100,95 @@
...
+							for (var j = 0; j < this.folderNotExpanded.length; j++) {
+								if (this.folderNotExpanded[j] === key) {
+									exists = true;
+									break;
+								}
+							}
+							if (!exists) {
+								bufs[curBuf] += '<li><button name="selectFolder" class="folderButtonOpen folderButtonOver" value="' + key + '"><i class="fa fa-folder-open" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML(key) + '</button></li>';
+								count++;
+								if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
```
scheibo:
```suggestion
								if (count % bufBoundary === 0 && curBuf < 4) curBuf++;
```

js/client-mainmenu.js
scheibo: Does the linter not complain about this because `%` forces things to be a number? Always using `===` is just more consistent.

Also, where does the number 4 come from?

js/client-mainmenu.js
```diff
@@ -1101,30 +1100,95 @@
...
+								for (var j = 0; j < folderData.length; j++) {
+									var selected = (folderData[j].id === curTeam);
+									bufs[curBuf] += '<li><button name="selectTeam" value="' + folders[key][j].id + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(folderData[j].name) + '</button></li>';
+									count++;
+									if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
```
scheibo:
```suggestion
									if (count % bufBoundary === 0 && curBuf < 4) curBuf++;
```

js/client-mainmenu.js
```diff
@@ -1101,30 +1100,95 @@
...
+								}
+							} else {
+								bufs[curBuf] += '<li><button name="selectFolder" class="folderButton" value="' + key + '"><i class="fa fa-folder" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML(key) + '</button></li>';
+								count++;
+								if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
```
scheibo:
```suggestion
								if (count % bufBoundary === 0 && curBuf < 4) curBuf++;
```

js/client-mainmenu.js
```diff
@@ -1101,30 +1100,95 @@
...
+								break;
+							}
+						}
+						if (!isNoFolder) {
+							bufs[curBuf] += '<li><button name="selectFolder" class="folderButtonOpen folderButtonOver" value="(No Folder)"><i class="fa fa-folder-open" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML("(No Folder)") + '</button></li>';
+						} else {
+							bufs[curBuf] += '<li><button name="selectFolder" class="folderButton" value="(No Folder)"><i class="fa fa-folder" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML("(No Folder)") + '</button></li>';
 							count++;
-							if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
+							if (count % bufBoundary == 0 && count != 0 && curBuf < 4) curBuf++;
```
scheibo:
```suggestion
							if (count % bufBoundary === 0 && count !== 0 && curBuf < 4) curBuf++;
```

js/client-mainmenu.js
```diff
@@ -1101,30 +1100,95 @@
...
+						}
+						var isNoFolder = false;
+						for (var i = 0; i < this.folderNotExpanded.length; i++) {
+							if (this.folderNotExpanded[i] === "(No Folder)") {
+								isNoFolder = true;
+								break;
+							}
+						}
+						if (!isNoFolder) {
+							bufs[curBuf] += '<li><button name="selectFolder" class="folderButtonOpen folderButtonOver" value="(No Folder)"><i class="fa fa-folder-open" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML("(No Folder)") + '</button></li>';
```
scheibo:
```suggestion
							bufs[curBuf] += '<li><button name="selectFolder" class="folderButtonOpen folderButtonOver" value="(No Folder)"><i class="fa fa-folder-open" style="margin-right: 7px; margin-left: 4px;"></i>(No Folder)</button></li>';
```

js/client-mainmenu.js
```diff
@@ -1101,30 +1100,95 @@
...
+						} else {
+							bufs[curBuf] += '<li><button name="selectFolder" class="folderButton" value="(No Folder)"><i class="fa fa-folder" style="margin-right: 7px; margin-left: 4px;"></i>' + BattleLog.escapeHTML("(No Folder)") + '</button></li>';
```
scheibo:
```suggestion
							bufs[curBuf] += '<li><button name="selectFolder" class="folderButton" value="(No Folder)"><i class="fa fa-folder" style="margin-right: 7px; margin-left: 4px;"></i>(No Folder)</button></li>';
```

js/client-mainmenu.js
```diff
@@ -1101,30 +1100,95 @@
...
+							if (count % bufBoundary == 0 && count != 0 && curBuf < 4) curBuf++;
+						}
+						if (!isNoFolder) {
+							for (var i = 0; i < teams.length; i++) {
+								if ((!teams[i].format && !teamFormat) || teams[i].format === teamFormat) {
+									var selected = (i === curTeam);
+									if (teams[i].folder === "") {
+										bufs[curBuf] += '<li><button name="selectTeam" value="' + i + '"' + (selected ? ' class="sel"' : '') + '>' + BattleLog.escapeHTML(teams[i].name) + '</button></li>';
+										count++;
+										if (count % bufBoundary == 0 && curBuf < 4) curBuf++;
```
scheibo:
```suggestion
										if (count % bufBoundary === 0 && curBuf < 4) curBuf++;
```

js/client-mainmenu.js
```diff
@@ -1175,6 +1238,24 @@
 				teambuilder.selectFolder(teamFormat);
 			}
 		},
+		selectFolder: function (key) {
+			var keyExists = false;
+			for (var i = this.folderNotExpanded.length - 1; i > -1; i--) {
```
scheibo: What happened to the suggestion of using `Arrary.prototype.filter` here?

js/client-mainmenu.js
scheibo: `if (folders[foldersName])` will work if  `folders[foldersName] === undefined`. `[]` is not falsey.

js/client-mainmenu.js
scheibo: @kanmingxu00

js/client-mainmenu.js
kanmingxu00: The Max rows that the team selector wants. That was not my code, I pasted it around.

kanmingxu00:
thanks @scheibo 👍 made all the changes

js/client-mainmenu.js
```diff
@@ -1240,17 +1240,19 @@
 		},
 		selectFolder: function (key) {
 			var keyExists = false;
-			for (var i = this.folderNotExpanded.length - 1; i > -1; i--) {
-				if (this.folderNotExpanded[i] === key) {
+			var folderNotExpandedTemp = this.folderNotExpanded.filter(function (folder) {
```
scheibo: I though we went over `temp` being a bad name? Please just call this `folderNotExpanded`.

scheibo:
@Zarel 

Screen Shot 2019-08-14 at 09 21 01

Issue #1307, opened 2019-06-20
Replay search improvements


scheibo:
> Currently you can only search by either username or format, but not both at the same time. Also searching with two users to find specific battles between the two.

_**wuhoodude:** [Replay search improvements](https://www.smogon.com/forums/threads/replay-search-improvements.3650623/)_

---

Started looking into this:

-  Is there a publicly available schema for the `ps_replays` table? The other tables on the server have `.sql` files describing them, but I can't find a similar one for the table `replays/replays.lib.php`?
- It looks like we're currently [deliberately](https://github.com/Zarel/Pokemon-Showdown-Client/blob/master/replays/search.php#L10) not allowing `username` and `format` search together, presumably because we dont have a query that supports it.
- Thoughts on UI? My first thought is to remove the double submit button and add a second user textbox to have:

  ```
  User(s): ________ ________
  Format: _______

  <Submit>
  ```

  But I'm not not sure what the best way to denote only one of the text boxes is required outside of literally adding a message (like smalltext which says "At least one field must be filled") which feels kludgey.

- Whats the story behind this `FORCE INDEX... UNION` query the replaces the commented out code which is much more obvious? 
https://github.com/Zarel/Pokemon-Showdown-Client/blob/e4e3798fc720f44c3916fa7aa015857c4ce4c93f/replays/replays.lib.php#L73-L75
Presumably this is for performance - I'm surprised this is at all necessary? This doesn't generalize well - if I start allowing for both format and player for example it becomes more difficult to `FORCE INDEX` on both. Are we very concerned about database load if we revert to allowing the query optimizer to execute the query without hints?
- Is there a specific reason username doesn't support ordering by rating or formats don't support `$isPrivate`?
- I think there should be one `search` method, which results in the same logic as today if called with the same arguments (ie. should be equivalent to today's `searchFormat` if just called with `$format`)
  ```php
  function search($usera = NULL, $userb = NULL, $format = NULL, $page = 0, $isPrivate = 
  false, $byRating = false) { 
   // validate at least one of $usera / $userb / $format is defined
   ...
  }
  ```
  This is the main important thing to get right, and would probably be where I'd start - I'd write this function and have the existing UI call it, ensure there's not change in functionality/regression and then change the UI to allow fully exercising the new options.
- Finally, which version of PHP do I need/is there any special setup required?

scheibo:
Strawman code to illustrate the approach:

```php
[...]

PR #1346, opened 2019-08-07
Unify username and format replay search internals

scheibo:
This is a step towards #1307, where eventually we will support calling `search` with any combination of player username, opponent username and format.

### DIFF
--- a/replays/replays.lib.php
+++ b/replays/replays.lib.php
@@ -78,7 +78,7 @@ class Replays {
 	}
 
 	function search($args) {
-		$page = isset(args["page"]) ? args["page"] : 0;
+		$page = args["page"] ?? 0;
 
 		if (!$this->db) return [];
 		if ($page > 100) return [];
@@ -86,15 +86,15 @@ class Replays {
 		$limit1 = intval(50*($page-1));
 		if ($limit1 < 0) $limit1 = 0;
 
-		$isPrivate = (isset($args["isPrivate"]) && args["isPrivate"]) ? 1 : 0;
-		$byRating = isset(args["byRating"]) && args["byRating"];
+		$isPrivate = ($args["isPrivate"] ?? null) ? 1 : 0;
+		$byRating = args["byRating"] ?? null;
 
-		if (isset($args["p1"])) {
+		if ($args["p1"] ?? null) {
 			$order = $byRating ? "rating" : "uploadtime";
 			$p1id = $this->toId($args["p1"]);
-			if (isset($args["p2"])) {
+			if ($args["p2"] ?? null) {
 				$p2id = $this->toId($args["p2"]);
-				if (isset($args["format"])) {
+				if ($args["format"] ?? null) {
 					$format = $this->toId($args["format"]);
 					$res = $this->db->prepare("(SELECT uploadtime, id, format, p1, p2, password FROM ps_replays FORCE INDEX (p1) WHERE private = ? AND p1id = ? AND p2id = ? AND format = ? ORDER BY ? DESC) UNION (SELECT uploadtime, id, format, p1, p2, password FROM ps_replays FORCE INDEX (p1) WHERE private = ? AND p1id = ? AND p2id = ? AND format = ? ORDER BY ? DESC) ORDER BY ? DESC LIMIT ?, 51;");
 					$res->execute([$isPrivate, $p1id, $p2id, $format, $order, $isPrivate, $p2id, $p1id, $format, $order, $limit1]);
@@ -103,7 +103,7 @@ class Replays {
 					$res->execute([$isPrivate, $p1id, $p2id, $order, $isPrivate, $p2id, $p1id, $order, $limit1]);
 				}
 			} else {
-				if (isset($args["format"])) {
+				if ($args["format"] ?? null) {
 					$res = $this->db->prepare("(SELECT uploadtime, id, format, p1, p2, password FROM ps_replays FORCE INDEX (p1) WHERE private = ? AND p1id = ? AND format = ? ORDER BY ? DESC) UNION (SELECT uploadtime, id, format, p1, p2, password FROM ps_replays FORCE INDEX (p2) WHERE private = ? AND p2id = ? AND format = ? ORDER BY uploadtime DESC) ORDER BY ? DESC LIMIT ?, 51;");
 					$res->execute([$isPrivate, $p1id, $this->toId($args["format"]), $order, $isPrivate, $p1id, $format, $order, $limit1]);
 				} else {
--- a/replays/replays.lib.php
+++ b/replays/replays.lib.php
@@ -89,26 +89,26 @@ class Replays {
 		$isPrivate = ($args["isPrivate"] ?? null) ? 1 : 0;
 		$byRating = args["byRating"] ?? null;
 
-		if ($args["p1"] ?? null) {
+		if ($args["username"] ?? null) {
 			$order = $byRating ? "rating" : "uploadtime";
-			$p1id = $this->toId($args["p1"]);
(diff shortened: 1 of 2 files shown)


replays/replays.lib.php
```diff
@@ -77,17 +77,47 @@ function toID($name) {
 		return preg_replace('/[^a-z0-9]+/','',$name);
 	}
 
-	function search($term, $page = 0, $isPrivate) {
+	function search($p1 = NULL, $p2 = NULL, $format = NULL, $byRating = false, $isPrivate = false, $page = 0) {
```
Zarel: This seems like a significantly less readable API than my previous one?

I guess you could pass search terms by associative array.

scheibo: I wasn't going for readability, I was going for power. But sure

Slayer95: Would it make sense to make the old API a wrapper around this function?

Zarel: I'm fine with just dropping support for the old API.

replays/replays.lib.php
```diff
@@ -77,17 +77,50 @@ function toID($name) {
 		return preg_replace('/[^a-z0-9]+/','',$name);
 	}
 
-	function search($term, $page = 0, $isPrivate) {
+	function search($args) {
+		$page = isset(args["page"]) ? args["page"] : 0;
```
Zarel:
```suggestion
		$page = args["page"] ?? 0;
```

replays/replays.lib.php
```diff
@@ -77,17 +77,50 @@ function toID($name) {
...
 
 		$limit1 = intval(50*($page-1));
 		if ($limit1 < 0) $limit1 = 0;
-		$term = $this->toID($term);
-		$isPrivate = $isPrivate ? 1 : 0;
-		// $res = $this->db->prepare("SELECT uploadtime, id, format, p1, p2 FROM ps_replays WHERE private = 0 AND (p1id = ? OR p2id = ?) ORDER BY uploadtime DESC LIMIT ?, 51");
-		$res = $this->db->prepare("(SELECT uploadtime, id, format, p1, p2, password FROM ps_replays FORCE INDEX (p1) WHERE private = ? AND p1id = ? ORDER BY uploadtime DESC) UNION (SELECT uploadtime, id, format, p1, p2, password FROM ps_replays FORCE INDEX (p2) WHERE private = ? AND p2id = ? ORDER BY uploadtime DESC) ORDER BY uploadtime DESC LIMIT ?, 51;");
-		$res->execute([$isPrivate, $term, $isPrivate, $term, $limit1]);
+
+		$isPrivate = (isset($args["isPrivate"]) && args["isPrivate"]) ? 1 : 0;
```
Zarel: So, in case it was unclear, `a ?? b` is equivalent to `isset(a) ? a : b`. Like `isset`, it bypasses the "variable not set" error.

```suggestion
		$isPrivate = ($args["isPrivate"] ?? null) ? 1 : 0;
```

replays/replays.lib.php
```diff
@@ -77,17 +77,50 @@ function toID($name) {
...
+		$byRating = isset(args["byRating"]) && args["byRating"];
```
Zarel:
```suggestion
		$byRating = args["byRating"] ?? null;
```

replays/replays.lib.php
```diff
@@ -77,17 +77,50 @@ function toID($name) {
...
+
+		if (isset($args["p1"])) {
```
Zarel:
```suggestion
		if ($args["p1"] ?? null) {
```

replays/replays.lib.php
```diff
@@ -77,17 +77,50 @@ function toID($name) {
...
+			$order = $byRating ? "rating" : "uploadtime";
+			$p1id = $this->toId($args["p1"]);
+			if (isset($args["p2"])) {
```
Zarel:
```suggestion
			if ($args["p2"] ?? null) {
```

replays/replays.lib.php
```diff
@@ -77,17 +77,50 @@ function toID($name) {
...
+				$p2id = $this->toId($args["p2"]);
+				if (isset($args["format"])) {
```
Zarel:
```suggestion
				if ($args["format"] ?? null) {
```

replays/search.php
```diff
@@ -92,12 +92,12 @@
 		// no
 	} else if ($username) {
 		if (!$isPrivate || $isPrivateAllowed) {
-			$replays = $Replays->search($username, $page, $isPrivate);
+			$replays = $Replays->search(array("p1" => $username, "isPrivate" => $isPrivate, "page" => $page));
```
Zarel:
```suggestion
			$replays = $Replays->search(["p1" => $username, "isPrivate" => $isPrivate, "page" => $page]);
```

replays/search.php
```diff
@@ -92,12 +92,12 @@
...
 		}
 	} else if ($contains) {
 		$replays = $Replays->fullSearch($contains, $page);
 	} else {
-		$replays = $Replays->searchFormat($format, $page, $byRating);
+		$replays = $Replays->search(array("format" => $format, "byRating" => $byRating, "page" => $page));
```
Zarel:
```suggestion
		$replays = $Replays->search(["format" => $format, "byRating" => $byRating, "page" => $page]);
```

replays/replays.lib.php
```diff
@@ -77,17 +77,50 @@ function toID($name) {
...
+					$format = $this->toId($args["format"]);
+					$res = $this->db->prepare("(SELECT uploadtime, id, format, p1, p2, password FROM ps_replays FORCE INDEX (p1) WHERE private = ? AND p1id = ? AND p2id = ? AND format = ? ORDER BY ? DESC) UNION (SELECT uploadtime, id, format, p1, p2, password FROM ps_replays FORCE INDEX (p1) WHERE private = ? AND p1id = ? AND p2id = ? AND format = ? ORDER BY ? DESC) ORDER BY ? DESC LIMIT ?, 51;");
+					$res->execute([$isPrivate, $p1id, $p2id, $format, $order, $isPrivate, $p2id, $p1id, $format, $order, $limit1]);
+				} else {
+					$res = $this->db->prepare("(SELECT uploadtime, id, format, p1, p2, password FROM ps_replays FORCE INDEX (p1) WHERE private = ? AND p1id = ? AND p2id = ? ORDER BY ? DESC) UNION (SELECT uploadtime, id, format, p1, p2, password FROM ps_replays FORCE INDEX (p1) WHERE private = ? AND p1id = ? AND p2id = ? ORDER BY ? DESC) ORDER BY ? DESC LIMIT ?, 51;");
+					$res->execute([$isPrivate, $p1id, $p2id, $order, $isPrivate, $p2id, $p1id, $order, $limit1]);
+				}
+			} else {
+				if (isset($args["format"])) {
```
Zarel:
```suggestion
				if ($args["format"] ?? null) {
```

replays/search.php
```diff
@@ -92,12 +92,12 @@
 		// no
 	} else if ($username) {
 		if (!$isPrivate || $isPrivateAllowed) {
-			$replays = $Replays->search($username, $page, $isPrivate);
+			$replays = $Replays->search(["p1" => $username, "isPrivate" => $isPrivate, "page" => $page]);
```
Zarel: On second thought, I don't know if I like `p1` instead of just `userid` to mean "either p1 or p2". Let's go with `userid`?

scheibo: I chose `p1` for symmetry. Also, `p1` vs. `p2` is a mater of perspective...

I'd prefer `player` and `opponent` to `userid` and `userid2` or `p2` or whatever else, but `p1` and `p2` is most concise.

PR #1352, opened 2019-08-16
Improve /(show|hide)joins

thejetou:
Changes:
- /(show|hide)joins defaults to the current room
- /(show|hide)joins global, /(show|hide)joins battles, /(show|hide)joins chatrooms were added, their roles are self-explanatory

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -635,17 +635,22 @@
 			case 'showjoins':
 				var showjoins = Dex.prefs('showjoins') || {};
 				var serverShowjoins = showjoins[Config.server.id] || {};
-				if (target) {
-					var room = toID(target);
-					if (serverShowjoins['global']) {
-						delete serverShowjoins[room];
-					} else {
-						serverShowjoins[room] = 1;
-					}
-					this.add('Join/leave messages on room ' + room + ': ALWAYS ON');
-				} else {
-					serverShowjoins = {global: 1};
-					this.add('Join/leave messages: ALWAYS ON');
+				switch (target) {
+				case 'global':
+					serverShowjoins['global'] = 1;
+					this.add('Join/leave messages' + ': ALWAYS ON');
+					break;
+				case 'chatrooms':
+					serverShowjoins['chatrooms'] = 1;
+					this.add('Join/leave messages in chatrooms' + ': ALWAYS ON');
+					break;
+				case 'battles':
+					serverShowjoins['battles'] = 1;
+					this.add('Join/leave messages in battle' + ': ALWAYS ON');
+					break;
+				default:
+					serverShowjoins[this.id] = 1;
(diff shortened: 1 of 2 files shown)


Zarel:
Are you planning on switching it to actually hide joins in a future PR?

js/client-chat.js
```diff
@@ -635,35 +635,45 @@
...
+				case 'chatrooms':
+					serverShowjoins['chatrooms'] = 1;
+					this.add('Join/leave messages in chatrooms' + ': ALWAYS ON');
+					break;
+				case 'battles':
+					serverShowjoins['battles'] = 1;
+					this.add('Join/leave messages in battle' + ': ALWAYS ON');
+					break;
+				default:
+					serverShowjoins[this.id] = 1;
```
Zarel: It might be excessive to store these settings forever in battles that auto-expire after 10 minutes.

Zarel:
In hindsight, I might want to default to changing chatrooms/games/groupchats, depending on whichever is currently open.

TheJetOU:
> Are you planning on switching it to actually hide joins in a future PR?

Do you mean `/forcehidejoins` instead of automatic? Yeah, I guess I can add that here.

TheJetOU:
Okay, so just so we're on the same page here: 
- the default state will not change (1),
- `/showjoins` will not change (0).
- `/hidejoins` will hide all instead of being the "automatic setting" (-1).

> In hindsight, I might want to default to changing chatrooms/games/groupchats, depending on whichever is currently open.

Changing it to what? I don't think chatrooms should ever show join / leave messages if not explicitly set by the user.

Zarel:
> Changing it to what? I don't think chatrooms should ever show join / leave messages if not explicitly set by the user.

Changing it when explicitly set by the user.

Zarel:
Like, changing

"/(show|hide)joins defaults to the current room"

to

"/(show|hide)joins defaults to whichever of chatrooms|games|groupchats the current room is"

Zarel:
> Okay, so just so we're on the same page here:
> 
> * the default state will not change (1),
> * `/showjoins` will not change (0).
> * `/hidejoins` will hide all instead of being the "automatic setting" (-1).

You still need a way to change it back to the default setting. I recommend `/autoshowjoins`. It should also be mentioned whenever you use `/showjoins` or `/hidejoins`.

TheJetOU:
> Like, changing
> 
> "/(show|hide)joins defaults to the current room"
> 
> to
> 
> "/(show|hide)joins defaults to whichever of chatrooms|games|groupchats the current room is"

Hm, that would certainly solve the issue of having room battles that expire after 10 minutes in `localStorage` for the rest of time. +1.

TheJetOU:
> You still need a way to change it back to the default setting. I recommend /autoshowjoins. It should also be mentioned whenever you use /showjoins or /hidejoins.

SGTM.

PR #1355, opened 2019-08-26
Runner-ups -> Runners-up

MathyFurret:


xfix:
If that counts for anything, Trivia script uses "runners-up", because the room staff requested this form.

PR #1358, opened 2019-09-03
Implement connection timeouts

Kaiepi:
This makes the client use SockJS' `timeout` config option (which isn't used when the connection's using the WebSocket protocol, FTR) and updates SockJS to v1.4.0 for fixes to bugs that made it impossible to test the client locally.

If this doesn't actually get rid of ghost connections, then implementing heartbeats as talked about before would be needed.


Zarel:
What's the host check?

Zarel:
Also, how could this possibly fix server-side ghost connections, which are caused by the server and client being unreachable by each other?

Kaiepi:
> What's the host check?

I'm not entirely sure. Existing code already had it, so I kept it around.

> Also, how could this possibly fix server-side ghost connections, which are caused by the server and client being unreachable by each other?

These changes depend on ghost connections being a bug in SockJS rather than half-open/half-closed TCP sockets.

I had @scheibo run a shell script I wrote to check if there were an unreasonable number of half-open/half-closed TCP sockets around, and that ended up not being the case. I don't think this was terribly long after a restart though, so they probably didn't have much of a chance to pop up yet. On second thought, it'd probably be a better idea to wait a week or two longer to check a couple more times, and maybe log information about connections on the server when they're unreasonably old before actually attempting to fix it.

If it turns out to really be a bug in SockJS, then this would help determine whether it's a bug in its client or server. If the server is where the bug is, then you're right, this wouldn't solve anything, but if not, this may be good enough on its own to fix it.

Zarel:
Interesting. Thanks!

PR #1360, opened 2019-09-04
Fix SockJS when using the desktop client

Kaiepi:
This hasn't been tested, since I'm not really sure how you build the desktop client on OpenBSD, but I diffed sockjs-1.1.1-nwjsfix.min.js with the unpatched version and applied the same changes to the upgraded version of SockJS.


Zarel:
Heh, I maybe should have asked, but I just assumed SockJS fixed the NWjs issue in the last three minor versions. :/

PR #1362, opened 2019-09-08
Tooltips: Properly prevent duplicate ability display

asgdf:
This is, again, for "Ability: Levitate (base ability: Levitate)".
The last time I thought I'd remove the duplication in the get-the-data code, but it turns out the ability data is sometimes an ID and sometimes a name, throwing everything off. I don't want to have to deal with having to toID all over the place, so I put the check in the display code after we got the proper name regardless so we can compare freely.

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -1577,7 +1577,7 @@ class BattleTooltips {
 		if (clientPokemon) {
 			if (clientPokemon.ability) {
 				abilityData.ability = clientPokemon.ability || clientPokemon.baseAbility;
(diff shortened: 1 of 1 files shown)


DaWoblefet:
I'd check this against Skill Swap in Doubles to make sure it's working right - it's notorious for doing this sort of "double Ability" there, as well as displaying the wrong ability.

asgdf:
I tested it with Skill Swap in doubles. It seemed to be working fine.

Zarel:
We should also probably standardize it to always being ability name.

PR #1365, opened 2019-09-09
Fix name in trainer card

whalemer:
[02:42:41] peach: name never starts with a rank symbol on initialization
[02:42:50] peach: and it never starts with a rank symbol in userdetails
[02:43:10] peach: so i think you can just delete the line i linked as well as L2467 and be fine

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -2464,7 +2464,6 @@ function toId() {
 		initialize: function (data) {
 			data.userid = toID(data.name);
 			var name = data.name;
-			if (/[a-zA-Z0-9]/.test(name.charAt(0))) name = ' ' + name;
 			this.data = data = _.extend(data, UserPopup.dataCache[data.userid]);
 			data.name = name;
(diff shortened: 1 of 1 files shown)


js/client.js
```diff
@@ -2497,7 +2496,6 @@ function toId() {
 					globalgroup = "Global " + globalgroup;
 				}
 			}
-			if (group || name.charAt(0) === ' ') name = name.substr(1);
```
Zarel: I would replace this with an ID detector.

Zarel:
> [02:42:41] peach: name never starts with a rank symbol on initialization

I don't know if I believe this. I'm pretty sure it does.

Zarel:
Oh, huh, apparently it was separated into `roomgroup` at some point. I suppose that's fine by me.

whalemer:
As far as i can tell, the only time that'd be meaningful would be for the
moment after I did `/user ~Zarel` before the server responded, so I'm not
sure what it really achieves.

PR #1367, opened 2019-09-09
UserPopup: Don't show IDs for expired users

thejetou:
Previously, clicking a username whose user object has been destroyed would lead to the name field showing the userid instead of the username because the server can't find the username.

~This is the simplest solution, although I guess it's more load on the server since it's not receiving pure IDs, but I seriously doubt that matters.~

~I'm not too sure of how it ever was showing the real username, or if it ever even was doing that.~

This is because `data.name` would get shadowed by the server's `name` field.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -178,7 +178,7 @@
 			var name = $(e.currentTarget).data('name') || $(e.currentTarget).text();
 			var away = $(e.currentTarget).data('away') || false;
 			var status = $(e.currentTarget).data('status');
-			app.addPopup(UserPopup, {roomGroup: roomGroup, name: name, away: away, status: status, sourceEl: e.currentTarget, position: position});
+			app.addPopup(UserPopup, {roomGroup: roomGroup, username: name, away: away, status: status, sourceEl: e.currentTarget, position: position});
 		},
 		openPM: function (e) {
 			e.preventDefault();
@@ -486,7 +486,7 @@
 			case 'user':
 			case 'open':
 				var openUser = function (target) {
-					app.addPopup(UserPopup, {name: target});
+					app.addPopup(UserPopup, {username: target});
 				};
 				target = toName(target);
 				if (!target) {
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -458,7 +458,7 @@
 		clickUsername: function (e) {
 			e.stopPropagation();
 			var name = $(e.currentTarget).data('name') || $(e.currentTarget).text();
-			app.addPopup(UserPopup, {name: name, sourceEl: e.currentTarget});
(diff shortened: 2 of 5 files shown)


TheJetOU:
> I'm not too sure of _how it ever was showing the real username_, or if it ever even was doing that.

I realized how just after I created this.

Apparently it was showing the real username before but Zarel/Pokemon-Showdown#5766 made it send `name` so `data.name` was being shadowed by the server's response.

TheJetOU:
On second thought, maybe we should show the real 'username' (the one received from the server) when available instead of the inputted username / username at the time of that message?

Zarel:
It's probably useful to show "Name (renamed from other name)" where appropriate.

TheJetOU:
> It's probably useful to show "Name (renamed from other name)" where appropriate.

I think this would only make sense to show if it's different IDs, it'd take too much space just to convey 'this user changed one letter to capital'.

Zarel:
Fair.

Zarel:
I'm still confused, though: `name` _should_ get shadowed, right?

TheJetOU:
> I'm still confused, though: `name` _should_ get shadowed, right?

It _does_ get shadowed. I was just contemplating whether we should `name` if it's a username instead of `data-name`.

Zarel:
Okay, can you clarify what this is intended to fix? Like CONTRIBUTING says, commit names should describe what you're fixing, not what you changed in the code to make the fix happen.

Zarel:
Oh, wait, that _is_ described, just crossed out.

Anyway, there are two possible fixes here:

1. Send the username to the server, so it can echo it back to us.

2. Make the server not send a username when it doesn't know it.

I'd be fine with either. But renaming this is probably the wrong solution.

TheJetOU:
> Make the server not send a username when it doesn't know it.

Now that I took a closer look at the server code, I'm leaning towards that solution. I find it kind of weird that we send the same value twice just with different field names. https://github.com/Zarel/Pokemon-Showdown/blob/master/server/chat-commands.js#L4527-L4530.

Zarel:
@Morfent was the one who wanted it there; does he have an opinion?

TheJetOU:
(I think his GitHub account is @Kaiepi now.)

Zarel:
@Kaiepi???

Kaiepi:
I'm not sure what I'm being asked...?

TheJetOU:
> I'm not sure what I'm being asked...?

Is there a reason you need to send `name` when the user doesn't exist when it's the same as `id`?

Kaiepi:
> Is there a reason you need to send name when the user doesn't exist when it's the same as id?

I just wanted the object sent to have a consistent shape. It can be removed if it really needs to go.

Zarel:
Hm, we could probably look at other aspects of the message to distinguish "offline" from "expired"...

Issue #1357, opened 2019-09-02
Provide a way to easily import sets into the teambuilder


scheibo:
![idea](https://media.discordapp.net/attachments/617978210323529728/618135384345542656/Screen_Shot_2019-09-02_at_10.21.25.png?width=800&height=294)

Elaborating on this:

![VZVN7wNQnEG](https://user-images.githubusercontent.com/117249/64130583-1abeeb00-cd78-11e9-9f26-921fd3f7504d.png)

<img width="631" alt="Screen Shot 2019-09-02 at 11 52 41" src="https://user-images.githubusercontent.com/117249/64130592-31fdd880-cd78-11e9-82ab-094dcab1a22b.png">

https://github.com/Zarel/Pokemon-Showdown/issues/5732 will provide the set import utilities and package the sets together in a `@pokemon-showdown/sets` package. The client build process can pull the JSON files from this package out of `node_modules/` and host them (https://play.pokemshowdown.com/data/sets?). 

The UI should only load sets for the format selected (`gen7uu` etc) to minimize size concerns. The set data will contain sets from multiple source (`smogon.com/dex`, `smogon.com/stats`, etc) and the UI can decide which, if any, to display.

**Sets which are displayed from `smogon.com/dex` must link back to the correct Smogon analysis** (and as mentioned in the screenshot, the entire set should probably be clickable and link back to the analysis - there will be an 'Import' button to actually replace the contents of the Import/Export field with the selected set).

scheibo:
I was about to close this, but '**Sets which are displayed from `smogon.com/dex` must link back to the correct Smogon analysis**' doesn't really seem to be done? I imagine we can just stick a link to the analysis under the set import buttons and call it a day?

PR #1368, opened 2019-09-12
Import @pokemon-showdown/sets during build

scheibo:
Required for #1357. We're installing with `--no-save` because we pretty much always want the latest version and don't want to deal with the hassle of updating deps in `package.json`.

### DIFF
--- a/build
+++ b/build
@@ -37,7 +37,6 @@ case 'full':
 	execSync(`node ./build-tools/build-indexes`, options);
 	execSync(`node ./build-tools/build-learnsets`, options);
 	execSync(`node ./build-tools/build-minidex`, options);
-	execSync(`node ./build-tools/build-sets`, options);
 	execSync(`node ./replays/build`, options);
 	full = ' full';
 	break;
--- a/build
+++ b/build
@@ -37,6 +37,7 @@ case 'full':
 	execSync(`node ./build-tools/build-indexes`, options);
 	execSync(`node ./build-tools/build-learnsets`, options);
 	execSync(`node ./build-tools/build-minidex`, options);
+	execSync(`node ./build-tools/build-sets`, options);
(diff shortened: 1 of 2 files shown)


build-tools/build-sets
```diff
@@ -0,0 +1,25 @@
...
+const fs = require("fs");
+const child_process = require('child_process');
+const path = require("path");
+
+process.stdout.write("Importing sets from @pokemon-showdown/sets... ");
+
+const shell = cmd => child_process.execSync(cmd, {stdio: 'inherit', cwd: path.resolve(__dirname, '../..')});
+shell(`npm install --no-save @pokemon-showdown/sets`);
+
+const src = path.resolve(__dirname, '../node_modules/@pokemon-showdown/sets');
```
Zarel: Wait have we actually published `@pokemon-showdown/sets`?

scheibo: Yes. 0.0.1

build
```diff
@@ -37,6 +37,7 @@ case 'full':
 	execSync(`node ./build-tools/build-indexes`, options);
 	execSync(`node ./build-tools/build-learnsets`, options);
 	execSync(`node ./build-tools/build-minidex`, options);
+	execSync(`node ./build-tools/build-sets`, options);
```
Zarel: I would not add this to the full build until after we started using it, I think.

scheibo: Done. TBH, `npm install` is pretty slow (seems to also do dependency checks on other things in the package.json instead of just install? I wonder if I can turn that off), so maybe we'll always want to update sets piecemeal

scheibo: Found it. `--no-audit`. And `--force` to always go remote

scheibo: Hmm
maybe I dont want force

PR #1371, opened 2019-09-13
Don't give private smogtours replays passwords

scheibo:
Smogon TDs would like to go back to the old replay approach, as they're "having trouble gathering usage statistics an experiencing issues being unable to open games and replays".

### DIFF
--- a/replays/replays.lib.php
+++ b/replays/replays.lib.php
@@ -231,7 +231,7 @@ class Replays {
 			return 'not found';
 		}
 
-		$smogtours = preg_match('/^[a-z0-9]+-smogtours-[0-9]+$/', $id);
+		$smogtours = preg_match('/^smogtours-[a-z0-9]+-[0-9]+$/', $id);
 		$password = null;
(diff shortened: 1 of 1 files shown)


replays/replays.lib.php
```diff
@@ -231,11 +231,12 @@ function upload(&$reqData) {
 			return 'not found';
 		}
 
+		$smogtours = preg_match('/^[a-z0-9]+-smogtours-[0-9]+$/', $id);
```
TheJetOU: Replay IDs are in the form of `<serverid>-<format>-<num>` (as evidenced by https://replay.pokemonshowdown.com/smogtours-ou-39893), so this seems to be a faulty regex.
```suggestion
		$smogtours = preg_match('/^smogtours-[a-z0-9]+-[0-9]+$/', $id);
```

scheibo: Thanks, I misread m[1]

Zarel:
This can be done purely server-side, by setting `private` to `2`, the way it's done for punishment-lock replays.

Issue #1369, opened 2019-09-12
Add support for alternative past gen sprites


scheibo:
**Context:** https://www.smogon.com/forums/threads/dppt-sprites.3653313/

Contributors are working on providing the 96x96 transparent PNG resources for past gens, but we need to provide a way to allow for users to select their preference (see NetBattle screenshot below) 

![NetBattle](https://images-ext-1.discordapp.net/external/womGZI-gCkArKlgQhRrJlT1WIYRFsdnpBxlK7Ha_gVU/https/puu.sh/Eg87w/adc8dc0a51.png?width=274&height=300)

A new preference, `spritegfx`, will be added which will be populated with the keys from [`SPRITE_SOURCES`](https://gist.github.com/scheibo/ba515d907803aad88e051bd594c68fc1), with the value in that mapping being the value that becomes persisted in local storage. This select box contains information that will replace or alter the following preferences

- `noanim` ('Disable animations') becomes 'Disable battle animations', as sprite animation preference is now moved to `spritegfx`
- `bwgfx` ('Use BW sprites instead of XY models') is removed (just select 'Black/White' instead)
- `nopastgens` ('Use modern sprites for past generations') is removed (just select 'X/Y' instead)
'Disable animations' (no anim) and sprite animations

Which sprites to use will depend on the gen of the battle taking place, the sprite preference and the Pokemon's generation (to be able to determine availability).

```ts
function getSpritePreference(template: Template, gen: number) {
	const pref = Dex.prefs('spritegfx');
	if (pref && SPRITE_GENS[pref]) {
		if (SPRITE_GENS[pref] >= template.gen) return {dir: pref, gen: SPRITE_GENS[pref]};
		gen = 5; // fallback
	}
	const ani = !pref || !pref.endsWith('ani');
	gen = Math.max(gen, Math.min(template.gen, 5));
	return {dir: SPRITE_GEN_DEFAULTS[ani ? 'animated' : 'static'][gen], gen};
}
```

Whether we *actually* support animating old gen sprites (only on switch, by toggling back and forth between frames 1 and 2) is out of scope for this change (NetBattle just allowed selecting whether you prefer the static frame 1 or frame 2, but Zarel seems to think that would complicate the UI, though its already a long select box...).

[...]

PR #1372, opened 2019-09-15
Stop scaling BW sprites

scheibo:
This effectives reverts f8001f3 by making it so that the sprites are
no longer transformed by 2x.

### DIFF
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -1874,8 +1874,6 @@ class PokemonSprite extends Sprite {
 			statbarOffset += this.isBackSprite ? 1 : 20;
 		} else if (this.scene.gen <= 3) {
 			statbarOffset += this.isBackSprite ? 5 : 30;
-		// TODO: is this mechanicsGen? graphicsGen? spriteGen?
-		// TODO: if we're scaling gen5 backsprites still we likely still need some offset?
 		} else if (this.scene.gen !== 5) {
 			statbarOffset += this.isBackSprite ? 20 : 30;
 		}
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -1872,7 +1872,7 @@ class PokemonSprite extends Sprite {
 			statbarOffset += this.isBackSprite ? 1 : 20;
 		} else if (this.scene.gen <= 3) {
 			statbarOffset += this.isBackSprite ? 5 : 30;
-		} else {
+		} else if (this.scene.gen !== 5) {
 			statbarOffset += this.isBackSprite ? 20 : 30;
 		}
 
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -340,7 +340,9 @@ class BattleScene {
 
 		let left = 210;
 		let top = 245;
-		let scale = 1.5 - 0.5 * ((loc.z!) / 200);
+		let scale = (obj.gen === 5
+			? 2.0 - ((loc.z!) / 200)
+			: 1.5 - 0.5 * ((loc.z!) / 200));
 		if (scale < .1) scale = .1;
 
 		left += (410 - 190) * ((loc.z!) / 200);
@@ -1872,6 +1874,8 @@ class PokemonSprite extends Sprite {
 			statbarOffset += this.isBackSprite ? 1 : 20;
 		} else if (this.scene.gen <= 3) {
 			statbarOffset += this.isBackSprite ? 5 : 30;
+		// TODO: is this mechanicsGen? graphicsGen? spriteGen?
+		// TODO: if we're scaling gen5 backsprites still we likely still need some offset?
 		} else if (this.scene.gen !== 5) {
 			statbarOffset += this.isBackSprite ? 20 : 30;
 		}
--- a/src/battle-dex.ts
+++ b/src/battle-dex.ts
@@ -601,7 +601,7 @@ const Dex = new class implements ModdedDex {
 
 		if (animationData[facing + 'f'] && options.gender === 'F') facing += 'f';
 		let allowAnim = !Dex.prefs('noanim') && !Dex.prefs('nogif');
-		if (fieldGenNum === 5 || allowAnim && genNum >= 6) spriteData.pixelated = false;
+		if (allowAnim && genNum >= 6) spriteData.pixelated = false;
 		if (allowAnim && animationData[facing] && genNum >= 5) {
 			if (facing.slice(-1) === 'f') name += '-f';
 			dir = gen + 'ani' + dir;
--- a/src/battle-dex.ts
+++ b/src/battle-dex.ts
@@ -636,7 +636,7 @@ const Dex = new class implements ModdedDex {
 		}
 
 		if (!options.noScale) {
-			if (graphicsGen > 4) { // TODO: should this be spriteData.gen?
+			if (graphicsGen > 4) {
 				// no scaling
 			} else if (!spriteData.isBackSprite) {
 				spriteData.w *= 2;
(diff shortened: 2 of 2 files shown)


scheibo:
Screen Shot 2019-09-14 at 21 22 28

src/battle-dex.ts
```diff
@@ -601,7 +601,7 @@ const Dex = new class implements ModdedDex {
 
 		if (animationData[facing + 'f'] && options.gender === 'F') facing += 'f';
 		let allowAnim = !Dex.prefs('noanim') && !Dex.prefs('nogif');
-		if (allowAnim && genNum >= 6) spriteData.pixelated = false;
+		if (fieldGenNum === 5 || allowAnim && genNum >= 6) spriteData.pixelated = false;
```
scheibo: Can you help me understand `fieldGenNum` vs `genNum`? Its not clear to me why we check one vs the other in this function (like below - most code in the `options.noScale` block checks `fieldGenNum` but not L640). Understanding the distinction here (and how it relates to the `BattleScene.gen`) is important for https://github.com/Zarel/Pokemon-Showdown-Client/issues/1369 as well

Zarel: I'm surprised it wasn't clear from the definition.

`genNum` is the actual generation of the battle, for the purposes of game mechanics.

`fieldGenNum` is the generation we pull the sprites for the battle from (it also affects what backdrop we're using).

scheibo: `options.gen` is the battle gen AFAICT?

```
let genNum = Math.max(fieldGenNum, Math.min(template.gen, 5));
```

is gen num, and I'm not entirely sure what it does. Also, why would the battle gen effect y position etc?

scheibo: Also, I feel `spriteGenNum` is perhaps more clear than `fieldGenNum`?

Zarel: Oh, I didn't look at `genNum`. That's the sprite generation for the individual sprite. It's capped so we don't try to grab e.g. a gen-2 sprite for a gen-6 pokemon.

Zarel: So yeah `spriteGenNum` would be a better name for clarity there.

Zarel:
Well, this explains why I rushed a gen 5 sprite scale I wasn't happy with: I didn't like the alternatives, either.

I still think backsprites at 2x is probably ideal.

scheibo:
I'm very confident the userbase would prefer the old approach, and I guarantee 2x backsprites is not a good solution. I understand if you want to keep the current solution, but can I at least make the legacy behavior from this PR a toggle option?

src/battle-dex.ts
Zarel: Are we going to do the rename?

Zarel:
I'm not sure there's a need for a non-pixelated thing. I was planning on a 1x option which wouldn't scale any sprite at all, which should probably be sufficient.

scheibo:
> I'm not sure there's a need for a non-pixelated thing.

Removed

>  I was planning on a 1x option which wouldn't scale any sprite at all, which should probably be sufficient.

Do you want this included as well?

src/battle-dex.ts
scheibo: I'll rename this during #1369

Zarel:
Or we could pixelate the front-sprite only!!!!!

That's actually a cool idea, it would give the scene some pseudo-tilt-shift.

scheibo:
I'm more willing to consider that than 2x backsprite, but like, can we just ship it as is and consider further tweaks as necessary?

Zarel:
I don't like this as-is enough to _approve_ it in its current state, tbh.

If you feel really strong about this, I guess...?

scheibo:
Per chat:

- I will rename the gen variables and add doc comments explaining how they differ
- I will change the 3D engine algorithm to handle 2.0 scale for BW backsprites (as well as allow for 0 scaling)
- I will add a toggle to turn off all scaling (all old gens)

src/battle-animations.ts
```diff
@@ -1872,7 +1874,9 @@ class PokemonSprite extends Sprite {
 			statbarOffset += this.isBackSprite ? 1 : 20;
 		} else if (this.scene.gen <= 3) {
 			statbarOffset += this.isBackSprite ? 5 : 30;
-		} else {
+		// TODO: is this mechanicsGen? graphicsGen? spriteGen?
+		// TODO: if we're scaling gen5 backsprites still we likely still need some offset?
+		} else if (this.scene.gen !== 5) {
```
scheibo: @Zarel

src/battle-dex.ts
```diff
@@ -539,21 +541,31 @@ const Dex = new class implements ModdedDex {
...
+		//     This will default to mechanicsGen, but may be altered depending on user preferences.
+		//   - spriteData.gen: the generation number of a the specific Pokemon sprite in question.
+		//     This defaults to graphicsGen, but if the graphicsGen doesn't have a sprite for the Pokemon
+		//     (eg. Darmanitan in graphicsGen 2) then we go up gens until it exists.
+		//
+		let graphicsGen = mechanicsGen;
+		if (Dex.prefs('nopastgens')) graphicsGen = 6;
+		if (Dex.prefs('bwgfx') && graphicsGen >= 6) graphicsGen = 5;
+		spriteData.gen = Math.max(graphicsGen, Math.min(template.gen, 5));
+		const baseDir = ['', 'gen1', 'gen2', 'gen3', 'gen4', 'gen5', '', ''][spriteData.gen];
```
scheibo: I renamed this as well. `gen` is too confusing when theres already 3 other gens at play and its not actually always a generation in the case of `''`.

src/battle-dex.ts
```diff
@@ -601,45 +613,43 @@ const Dex = new class implements ModdedDex {
...
+			if (spriteData.gen >= 4 && miscData['frontf'] && options.gender === 'F') {
 				name += '-f';
 			}
 
 			spriteData.url += dir + '/' + name + '.png';
 		}
 
 		if (!options.noScale) {
-			if (fieldGenNum > 5) {
+			if (graphicsGen > 4) { // TODO: should this be spriteData.gen?
```
scheibo: Is this a graphics level thing or a sprite level thing? I *think* its fine as is, but I just want to confirm.

src/battle-dex.ts
```diff
@@ -601,45 +613,43 @@ const Dex = new class implements ModdedDex {
...
+				// old gen backsprites are multiplied by 1.5x by the 3D engine
 				spriteData.w *= 2 / 1.5;
 				spriteData.h *= 2 / 1.5;
 				spriteData.y += -11;
 			}
-			if (fieldGenNum === 5) spriteData.y = -35;
-			if (fieldGenNum === 5 && spriteData.isBackSprite) spriteData.y += 40;
-			if (genNum <= 2) spriteData.y += 2;
+			if (spriteData.gen <= 2) spriteData.y += 2;
+			if (spriteData.gen === 5 && spriteData.isBackSprite) spriteData.y += 40; // TODO: determine correct offset
```
scheibo: as above, i need to test this given we're still planning to scale the backsprite

scheibo:
On second though, the DO NOT MERGE label is too scary, WIP seems more forgiving :)

src/battle-animations.ts
Zarel: `scene.gen` is mechanics gen. We don't technically need it since we _can_ just read `scene.battle.gen`, I suppose.

Zarel: lol, it turns out I'm wrong:

https://github.com/Zarel/Pokemon-Showdown-Client/blob/21fa46b9a60079902fbe8080a52ddd61a98e6f67/src/battle-animations.ts#L38-L39

We should probably rename it to `graphicsGen`.

src/battle-dex.ts
Zarel: I'd probably make this a graphicsGen thing.

src/battle-animations.ts
scheibo: I'm punting on this renaming for now because it actually is used in a lot of places and I dont want to cause breakages. I'd like to ship as is and fix in my followup adding the noScale and alternative gen support.

scheibo:
2x backsprite in the 3D engine looks shippable to me.




I'm going to move the noScale work into the PR where i do alternative gen because they both require pref changes and I've decided I'd like to keep the scope of this change small. I think this is ready to ship.

PR #1375, opened 2019-09-27
Stop prefixing battle names with a space

whalemer:
As far as I can tell, this comes from #1071, which was from a much, much older name format and isn't necessary anymore (rank isn't included as any part of the name string anymore)


Zarel:
It's not...?

Zarel:
I suppose if we're doing `parseNameParts`, it's already been separated.

PR #1376, opened 2019-09-28
Support custom terrains being overwritten

HoeenCoder:
Did not use `endsWith` since I'm pretty sure thats unsupported by older browsers.


Zarel:
We have an `endsWith` polyfill in `battle-dex` (and if we don't, it should be added)

HoeenCoder:
Checked, it is polyfilled so I made the change here.

PR #1377, opened 2019-09-29
Fix battle joins in modchat not updating user list

asgdf:


Zarel:
lol wow that was the mistake?

PR #1378, opened 2019-09-29
Tooltips: Properly round displayed factors

asgdf:
To fix this: 

image

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -101,7 +101,7 @@ class ModifiableValue {
 			this.maxValue = 0;
 			return true;
 		}
-		if (name) this.comment.push(` (${factor}&times; from ${name})`);
+		if (name) this.comment.push(` (${this.round(factor)}&times; from ${name})`);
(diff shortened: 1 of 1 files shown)


src/battle-tooltips.ts
```diff
@@ -118,10 +118,13 @@ class ModifiableValue {
...
 	}
+	round(value: number) {
+		return value ? Number(value.toFixed(2)) : 0;
+	}
 	toString() {
 		let valueString;
-		const roundedValue = this.value ? Number(this.value.toFixed(2)) : 0;
-		const roundedMaxValue = this.maxValue ? Number(this.maxValue.toFixed(2)) : 0;
+		const roundedValue = this.round(this.value);
+		const roundedMaxValue = this.round(this.maxValue);
```
Zarel: Do we still need these intermediate values? I'd guess no.

asgdf: Oh, yeah, fair, changed.

Zarel:
Cool, thanks!

PR #1379, opened 2019-09-30
Fix broken avatar change messsage when using avatar numbers

asgdf:
When changing your avatar with /avatar [number], or when using the popup menu to select one, the success message has a broken image:
image
because it sends the avatar number directly to the server, and avatars are organized by name instead of number now, so the server has no way to display the new avatar. This way we'd just always send the name.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -923,8 +923,12 @@
 			case 'avatar':
 				var parts = target.split(',');
 				var avatar = parts[0].toLowerCase().replace(/[^a-z0-9-]+/g, '');
+				// Replace avatar number with name before sending it to the server, only the client knows what to do with the numbers
+				if (window.BattleAvatarNumbers && Object.prototype.hasOwnProperty.call(window.BattleAvatarNumbers, avatar)) {
+					avatar = window.BattleAvatarNumbers[avatar];
+				}
 				Dex.prefs('avatar', avatar);
-				return text; // Send the /avatar command through to the server.
+				return '/avatar ' + avatar; // Send the command through to the server.
 
 			case 'afd':
 				var cleanedTarget = toID(target);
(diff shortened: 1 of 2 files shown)


asgdf:
Oh uhhh
I tested this and it did work, so I'm not sure where the travis complaint is coming from about it being undefined
And what's the alternative to `hasOwnProperty `here? I didn't want issues with `/avatar constructor.`

asgdf:
Okay I guess I can use window.BattleAvatarNumbers instead, although that seems clunky.
Do I need to use `Object.hasOwnProperty.call(window.BattleAvatarNumbers, avatar)` or something for the other issue

TheJetOU:
I think `Object.getOwnPropertyDescriptor` would work here too.

Zarel:
It's `Object.prototype.hasOwnProperty.call`, not `Object.hasOwnProperty.call`.

asgdf:
Right, thanks. But is that the best available solution for doing `x in y` without taking y's prototype into account?

Zarel:
Sorry, I'm taking a while to comment because I'm trying to think of the best solution here.

Yes, I think `Object.prototype.hasOwnProperty.call` is probably the best solution here.

asgdf:
Not very pretty, but it works. Thanks! (didn't mean to come off as snippy in the previous message)

Slayer95:
> It's Object.prototype.hasOwnProperty.call, not Object.hasOwnProperty.call.

Both work, since ``Object`` is an instance of ``Object`` too.

Zarel:
> > It's Object.prototype.hasOwnProperty.call, not Object.hasOwnProperty.call.
> 
> Both work, since `Object` is an instance of `Object` too.

!!!!!!!!!!!

That makes total sense and now I'm really confused why no one ever thought to shorten it that way.

Zarel:
I suppose if you were actually trying to minimize space, you'd do something like `''.hasOwnProperty.call`.

PR #1380, opened 2019-10-01
Use hex colors instead of hsl

scheibo:


Zarel:
Wait, I already have a hex conversion a few lines above!

```js
		let C = (100 - Math.abs(2 * L - 100)) * S / 100 / 100;
		let X = C * (1 - Math.abs((H / 60) % 2 - 1));
		let m = L / 100 - C / 2;
		let R1;
		let G1;
		let B1;
		switch (Math.floor(H / 60)) {
		case 1: R1 = X; G1 = C; B1 = 0; break;
		case 2: R1 = 0; G1 = C; B1 = X; break;
		case 3: R1 = 0; G1 = X; B1 = C; break;
		case 4: R1 = X; G1 = 0; B1 = C; break;
		case 5: R1 = C; G1 = 0; B1 = X; break;
		case 0: default: R1 = C; G1 = X; B1 = 0; break;
		}
		let R = R1 + m;
		let G = G1 + m;
		let B = B1 + m;
```

Maybe these should be consolidated?

scheibo:
> Maybe these should be consolidated?

I try that in the commit i just pushed, but im seeing different results. Is there something obvious I'm missing?

src/battle-log.ts
```diff
@@ -412,6 +412,30 @@ class BattleLog {
...
+		if (HLmod > 18) HLmod = (HLmod - 18) * 2.5;
+		else if (HLmod < 0) HLmod = (HLmod - 0) / 3;
+		else HLmod = 0;
+		// let mod = ';border-right: ' + Math.abs(HLmod) + 'px solid ' + (HLmod > 0 ? 'red' : '#0088FF');
+		let Hdist = Math.min(Math.abs(180 - H), Math.abs(240 - H));
+		if (Hdist < 15) {
+			HLmod += (15 - Hdist) / 3;
+		}
+
+		L += HLmod;
```
Zarel: You don't run `HSLtoRBG` again after modding `L` here.

scheibo: Fuck, I even knew I had to call it twice but then didn't do it. *sigh*. TY.

Zarel:
I wonder if creating/destructuring an array is faster than creating/destructuring an object...

scheibo:
> I wonder if creating/destructuring an array is faster than creating/destructuring an object...

I was about to say it doesnt matter, especially given theyre so small, but we do actually call this function enough...

PR #1381, opened 2019-10-09
Fix grassy terrain battle text

Alumn2000:
https://www.smogon.com/forums/threads/fix-grassy-terrain-battle-text.3652044/

testclient html#battle-gen7ou-989733005

### DIFF
--- a/data/text-afd.js
+++ b/data/text-afd.js
@@ -70,7 +70,7 @@ exports.BattleTextAFD = {
 		damageFromPartialTrapping: "  [POKEMON] is hurt by [MOVE]!",
 		heal: "  [POKEMON] restored its HP.",
 		healFromZEffect: "  [POKEMON] restored its HP using its Z-Power!",
-		healFromEffect: "  [POKEMON]'s HP was restored.",
+		healFromEffect: "  [POKEMON] restored HP using its [EFFECT]!",
 
 		boost: "  [POKEMON]'s [STAT] increased!",
 		boost2: "  [POKEMON]'s [STAT] increased twice!",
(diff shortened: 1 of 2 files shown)


Alumn2000:
Problem: It works for Grassy terrain but I don't know if I broke something (even if I checked locally) + we have "The opposing [Pokemon]'s HP was restored." and not "The opposing [Pokemon] had its HP restored." because I don't know how to implement that, but I think that's ok.

Zarel:
@DaWoblefet is this message only used for Grassy Terrain? If so, it should be associated with Grassy Terrain, not the generic `healFromEffect`.

Zarel:
And yes, you did break something.

Zarel:
You see this:

```js
	grassyterrain: {
		start: "  Grass grew to cover the battlefield!",
		end: "  The grass disappeared from the battlefield.",
	},
```

Your message should have been added there:

```js
	grassyterrain: {
		start: "  Grass grew to cover the battlefield!",
		end: "  The grass disappeared from the battlefield.",
		heal: "  [POKEMON]'s HP was restored.",
	},
```

Alumn2000:
evidence1 horn leech + wish messages works (so drain punch, ... will work too)
evidence2 soft-boiled message works (so recover, roost, ... will work too)
evidence3 Wiki berry messages works (so Sitrus berry, Aguav berry, berry juice, ... will work too)

It was all under Grassy terrain (mb). Will try to see tomorrow if it works if grassy terrain isn't there but I was checking the local and normal client, and everything seemed ok.

Alumn2000:
Ah, I didn't saw your comment Zarel. Ok, will do it tomorrow.

DaWoblefet:
I would have to review all the healing effects to make sure, but offhand I think this is Grassy Terrain only.

Zarel:
Thanks!

PR #1383, opened 2019-10-15
Tooltips: Fix type/base power of Z-Hidden Power

asgdf:
The first issue was that Z-Hidden Power wrongly displayed its type as that of the base Hidden Power instead of normal (in the move selection, not the tooltips). So I made it base the type of the actual Z move, if a separate object is available, and only fall back on the base move if there is none (like for Z status moves).
The second issue was that Z-Hidden Power did not have a base power displayed in the tooltips, because only the base HP has a zMovePower, the typed ones don't. Instead of adding that property to all hidden powers I opted to make it fall back on the base move here. I can do the former instead if you'd prefer.


Zarel:
This looks good! Thanks!

Issue #1384, opened 2019-10-18
Add feature to challenge user using name at top of pm window


jordanhui19:
I was battling a friend during a voice call, and noticed there was no way to send them another them without searching their name again, since we weren't using the in-game chat.

I want to add a feature where clicking their name in the top of the pm window brings up the challenge menu the same way clicking their name in a pm message does.

Seems like this pretty simple to implement so I should be able to do it myself.

Zarel:
At the end of the battle, there's a "Rematch" button.

Zarel:
Anyway, the new Preact client fixes this with user lists in PMs. If someone else wants to fix this in the old client, feel free.

jordanhui19:
I after each match we made new teams, so we couldn't just rematch. I thought it was inconvenient to have to search for someone you already have a pm window open with, and I figured it'd make sense to be able to rematch from there.

Zarel:
The Rematch button lets you choose a new team and format.

jordanhui19:
Can you access teambuilder from rematch?

Zarel:
Yes... The Rematch button literally does the same thing as the Challenge button.

jordanhui19:
Huh okay, didn't realize the challenge window stays open if you go into teambuilder.
Still, I think its weird that clicking a username anywhere but the header of the pm lets you challenge them.

Zarel:
The team selector in the challenge window (and everywhere) also contains a Teambuilder link.

I do agree there should be some way to challenge from a PM window, though.

DaWoblefet:
This was done in https://github.com/smogon/pokemon-showdown-client/pull/1385

PR #1385, opened 2019-10-18
Add "Challenge" and "..." buttons to top of chat windows

jordanhui19:
Placed the username from the header of pm-window in a span with the class "header-username", added a click event for it that brings up the challenge/chat window.
Adds feature requested in issue #1384.

osdchallege

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -189,7 +189,7 @@
 				var buf = '<div class="pm-window pm-window-' + userid + '" data-userid="' + userid + '" data-name="' + name + '">';
 				buf += '<h3><button class="closebutton" href="' + app.root + 'teambuilder" tabindex="-1" aria-label="Close"><i class="fa fa-times-circle"></i></button>';
 				buf += '<button class="minimizebutton" href="' + app.root + 'teambuilder" tabindex="-1" aria-label="Minimize"><i class="fa fa-minus-circle"></i></button>';
-				buf += group + BattleLog.escapeHTML(name.substr(1)) + '</h3>';
+				buf += group + '<span class="username">' + BattleLog.escapeHTML(name.substr(1)) + '</span> </h3>';
 				buf += '<div class="pm-log"><div class="inner" role="log"></div></div>';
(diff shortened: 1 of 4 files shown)


jordanhui19:
I wasn't able to figure out how to un-highlight the minimize button when the username was hovered. Any tips?

Zarel:
Maybe it should be a separate box below the titlebar, instead of being part of the titlebar.

jordanhui19:
Do you mean something like having a "Challenge" button under the title bar instead of having to click the name to bring up the popup?

Zarel:
Yeah.

Zarel:
(Maybe also have the "More" button next to it.)

Zarel:
I would put it under the header. It also makes sense for it to scroll offscreen when there's a lot of chat, leaving more room for chat.

jordanhui19:
But if you have a lot of chat, wouldn't it be a hassle to have to scroll all the way back up to get the buttons again?

jordanhui19:
Also the chat display is the same size as before, so right now you can still see the same number of messages. The buttons are just in between the header and the pm log, so the overall window is bigger.

Zarel:
> But if you have a lot of chat, wouldn't it be a hassle to have to scroll all the way back up to get the buttons again?

If you have a lot of chat, you can just click on their username. :p

> the overall window is bigger

Yes, this is what I'm trying to avoid.

jordanhui19:
chatscrollhide
chatscrollhide2

So something like this?

jordanhui19:
I updated it so that the buttons to behave the same way as messages do, so scrolling up hides them.
I still think having a hideable section under the header makes more sense since I think losing the buttons while messaging is weird, but either way works.

js/client-mainmenu.js
```diff
@@ -8,8 +8,11 @@
 			'keydown textarea': 'keyDown',
 			'keyup textarea': 'keyUp',
 			'click .username': 'clickUsername',
+			'click .header-username': 'clickUsername',
```
Zarel: This part can be reverted, then?

jordanhui19: Yeah it can be taken out.
I forgot to remove it.

Zarel:
CI seems to be failing

jordanhui19:
Alright its fixed.

Zarel:
Cool, thanks!

PR #1386, opened 2019-10-21
Make session lengths consistently 2 weeks

xfix:


lib/ntbb-session.lib.php
```diff
@@ -72,8 +72,6 @@ function __construct() {
 		}
 		if (intval($sess['timeout'])<$ctime) {
 			// session expired
-			// delete all sessions that will expire within 30 minutes
-			$ctime += 60 * 30;
 			$psdb->query("DELETE FROM `{$psdb->prefix}sessions` WHERE `timeout` < ?", [$ctime]);
```
Zarel: Wait, let's make sure this database has a `timeout` index.

Zarel: Okay, good, it does.

Issue #1331, opened 2019-07-21
Disable minimizing Attack IVs for Ditto


jamesqo:
If you copy and paste Ditto's set from [here](https://www.smogon.com/dex/sm/pokemon/ditto/uber/) into the teambuilder, it will insert 0 Attack IVs since it doesn't see an attacking move. This causes Ditto's HP type to be Psychic instead of Ice contrary to what is suggested in the analysis, currently the user has to manually go in and insert 29 Attack IVs to work around this.

PR #1388, opened 2019-11-01
Do not change attack IVs for Pokemon with Transform

jamesqo:
This is an easy fix for https://github.com/Zarel/Pokemon-Showdown-Client/issues/1331. Note that it will still auto adjust the Speed IV, ie. if you have both Transform and Gyro Ball on Mew for example, so the behavior is not completely fixed. That should be easier to deal with once https://github.com/Zarel/Pokemon-Showdown-Client/pull/1333 is merged.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -2844,9 +2844,11 @@
 			var moves = set.moves;
 			for (var i = 0; i < moves.length; ++i) {
 				if (!moves[i]) continue;
-				if (moves[i].substr(0, 13) === 'Hidden Power ' || move.id === 'transform') hasHiddenPower = true;
+				if (moves[i].substr(0, 13) === 'Hidden Power ') hasHiddenPower = true;
 				var move = Dex.forGen(this.curTeam.gen).getMove(moves[i]);
-				if (move.category === 'Physical' &&
+				if (move.id === 'transform') {
+					hasHiddenPower = true;
+				} else if (move.category === 'Physical' &&
 						!move.damage && !move.ohko && move.id !== 'rapidspin' && move.id !== 'foulplay' && move.id !== 'endeavor' && move.id !== 'counter') {
 					minAtk = false;
 				} else if (move.id === 'metronome' || move.id === 'assist' || move.id === 'copycat' || move.id === 'mefirst') {
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -2844,12 +2844,12 @@
 			var moves = set.moves;
 			for (var i = 0; i < moves.length; ++i) {
 				if (!moves[i]) continue;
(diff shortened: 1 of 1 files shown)


Zarel:
0 Atk makes HP Ice still possible with Hyper Training. The correct answer might be to default to Ice here?

jamesqo:
@Zarel When we set the Attack IVs to 0 in the builder, we're saying that the IVs post-Hyper Training are 0.

Zarel:
Yes, and with a pre-hyper training IV set of `31/0/30/31/31/31`, you can Hyper Train Defense only to get `31/0/31/31/31/31`.

jamesqo:
OK, you're saying that if we see an EV spread of `31/31/31/31/31/31`, since we don't know if any of the stats were hyper trained and therefore can't say what the Hidden Power should be, we should arbitrarily set it to Ice?

I think it's a cleaner approach to assume that whatever EVs are given are not hyper trained and infer the HP type off of that, unless `Hidden Power: X` is specified or `Hidden Power X` is given as a move. This will allow for more fine-grained control; for example, Ditto sometimes wants to have HP Ground instead of Ice, so if we were to arbitrarily default to Ice regardless of EVs then we would break the instructions [here](https://www.smogon.com/dex/sm/pokemon/ditto/ou/) which say
```
30 Special Attack, 30 Special Defense, and 1 Speed IV can be used for Hidden Power Ground, which allows Ditto to countertrap Magnezone.
```

Zarel:
Oh, I see what you're saying.

That shouldn't matter for the Attack IV, though. If the Attack IV should be odd and `minAtk = true`, Atk will be set to 1 instead of 0.

Zarel:
See this line:

image

Maybe instead we should set `hasHiddenPower` instead of `minAtk`.

jamesqo:
@Zarel Sorry, didn't see your comment earlier. This issue came up for me again though so I remembered this PR.

I think the current solution is simpler... your solution only makes it so that Ditto is left with 1 Attack IV instead of 31 Attack IVs. Which almost never comes into play anyway since the stat is erased by Imposter, and when it does it's better to have the Attack stat higher (when Ditto is forced to use Struggle).

jamesqo:
Ah, nvm. I realized that changing the other variable covers this for both Attack and Speed rather than just Attack, which is probably why you wanted it that way. I've implemented this how you wanted.

I run into this bug all the time, so it'd be great if you could consider merging this.

jamesqo:
Ah, nvm. I realized that changing the other variable covers this for both Attack and Speed rather than just Attack, which is probably why you wanted it that way. I've implemented this how you wanted.

I run into this bug all the time, so it'd be great if you could consider merging this @Zarel.

Zarel:
Looks good, thanks!

If you want to, feel free to PR a change to maximize the Atk IV for a Transform-only moveset. That should also be a one-line change.

jamesqo:
:pray:

PR #1391, opened 2019-11-11
Tooltips: Fix Gyro Ball/Electro Ball base power while paralyzed

asgdf:
This way it will use the speed calculated clientside with all factors, including things like paralysis, which serverPokemon.stats does not include.


asgdf:
I think the build failure isn't my fault, I don't get that locally?

Zarel:
Yeah, I've heard other people have this problem but I can't repro.

asgdf:
Oh did it fix itself, neat

Zarel:
No, I fixed it in 651f875cf2ccf9e

Issue #1387, opened 2019-10-29
Resume button visible behind play button


JacenBoy:
On downloaded replays, when the page first loads, the "Resume" button is visible behind the "Play" button.

![replay](https://user-images.githubusercontent.com/11392691/67805650-bcf10b80-fa67-11e9-8a52-f9dae9efdd1f.png)

jordanhui19:
I can work on this one.
@Zarel how would I test replay-embed.js with a downloaded replay?
Would I have to make a new replay with testclient?

Zarel:
You can just edit the downloaded replay source to point to your own copy of `replay-embed.js`

Zarel:
Fixed by 6b9ff7a3c4256e0

PR #1393, opened 2019-11-13
Stopped replay button from being shown on start

jordanhui19:
Fix for issue #1387 
Stopped resume button from being shown when opening the replay, instead it only shows up when the replay is paused.
Play
resume

### DIFF
--- a/js/replay-embed.js
+++ b/js/replay-embed.js
@@ -49,7 +49,7 @@ var Replays = {
 		this.battle = new Battle(this.$('.battle'), this.$('.battle-log'));
 		//this.battle.preloadCallback = updateProgress;
 		this.battle.errorCallback = this.errorCallback.bind(this);
-		this.battle.resumeButton = this.resume.bind(this);
+		//this.battle.resumeButton = this.resume.bind(this);
 
 		this.setlog(log);
 	},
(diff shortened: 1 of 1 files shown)


Zarel:
What was the problem in the code, and how does this fix it?

jordanhui19:
So originally, this.battle.resumeButton = this.resume.bind(this); was called in init, placing a resume button on the battle screen when the replay is loaded, then it would put the play button on top of it like this:
67805650-bcf10b80-fa67-11e9-8a52-f9dae9efdd1f

I moved this.battle.resumeButton = this.resume.bind(this); to the pause() function so that the resume button is only shown when the replay is paused during playback.

Zarel:
Do we actually need to set `battle.resumeButton` at all?

I would guess no. We might be able to clear everything in there.

jordanhui19:
Technically no cause you can press the play button at the bottom to resume the match, but I think it makes sense to keep it there cause the Play button also appears in the battle window when the replay is paused before beginning.
Having the resume button there makes things more consistent.

Zarel:
Are you sure that not setting `battle.resumeButton` prevents the resume button from showing up? That seems doubtful to me.

Zarel:
`resumeButton` is an event handler! How could it lead to two buttons showing up?

jordanhui19:
Yeah I tested it out and it works.
Also battle.resumeButton seems to be the only reference to a resume button in replay-embed.

Zarel:
Don't worry, I already figured out the real issue and pushed it.

PR #1398, opened 2019-11-17
Filter out non-usable moves in the teambuilder

KrisXV:


js/search.js
```diff
@@ -968,6 +968,7 @@
 						} else */ if (learnsetEntry.indexOf(gen) < 0) {
 							continue;
 						}
+						if (this.gen === 8 && BattleMovedex[moveid].isNonstandard === "Past") continue;
```
Zarel: `if (learnsetEntry.indexOf(gen) < 0) {` should cover this...

KrisXV: It doesnt

Zarel: Oh, of course it doesn't.

Zarel: My mistake.

Zarel: Well, we'll need to add an exception for NatDex formats.

Can we do that now, actually? Will make it easier later.

Zarel: Well, we'll need to add an exception for NatDex formats.

Can we do that now, actually? Will make it easier later.

KrisXV: I mean, it is possible, but I believe @TheImmortal said somewhere that illegal moves wouldn't be usable in NatDex metas, just the Pokemon. Correct me if im wrong.

Zarel: I was planning on including illegal moves, though...

Zarel: I was going to just blanket-allow `Past` in the NatDex meta.

KrisXV: do you know what the formats will be named? maybe if its like [Gen 8 NatDex] i can add a format based exception

Zarel: `[Gen 8] National Dex` is the tentative name

KrisXV: did

PR #1402, opened 2019-11-20
Change default format to Gen 8 Random Battle

KrisXV:


Zarel:
I wonder if our default randbat should be NatDex?

KrisXV:
I guess we'll just have to wait for the feedback to start rolling in, but I don't see any issue with it.

PR #1407, opened 2019-11-22
Fix CAP and STABmons teambuilder learnsets

KrisXV:


js/search.js
```diff
@@ -1047,7 +1048,7 @@
 						for (var j = 0; j < baseTemplate.otherFormes.length; j++) {
 							var forme = Dex.getTemplate(baseTemplate.otherFormes[j]);
 							for (var h = 0; h < forme.types.length; h++) {
-								if (template.forme.indexOf('Alola') >= 0 || template.forme.indexOf('Galar') > 0 || forme.baseSpecies === 'Wormadam') continue;
+								if (template.forme.indexOf('Alola') >= 0 || template.forme.indexOf('Galar') > 0 || forme.baseSpecies === 'Wormadam' || forme.battleOnly) continue;
```
Zarel: ?

KrisXV: the teambuilder was displaying illegal moves on battleOnly formes, so Charizard-Mega-Y was appearing with Dragon-type moves

PR #1410, opened 2019-11-28
Add fail messages for weight-based moves

PartMan7:
Unsure if Sky Drop has the ``fail`` flag activated against Dynamax Pokemon.

Oh, and I agree to MIT license this code.

### DIFF
--- a/data/text.js
+++ b/data/text.js
@@ -558,6 +558,12 @@ exports.BattleText = {
 	healingwish: {
 		heal: "  The healing wish came true for [POKEMON]!",
 	},
+	heatcrash: {
+		fail: "#grassknot",
+	},
+	heavyslam: {
+		fail: "#grassknot",
+	},
 	helpinghand: {
 		start: "  [SOURCE] is ready to help [POKEMON]!",
 	},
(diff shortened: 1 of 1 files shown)


KrisXV:
The client doesn't use the MIT license, it uses the AGPLv3 license

PartMan7:
License

Your submitted code should be MIT licensed. The GitHub ToS (and the fact that your fork also contains our LICENSE file) ensures this, so we won't ask when you submit a pull request, but keep this in mind.

For simplicity (mostly to make relicensing easier), client code should be also be MIT licensed. The first time you make a client pull request, we'll ask you to explicitly state that you agree to MIT license it.

-CONTRIBUTING.md

KrisXV:
License
Pokémon Showdown's client is distributed under the terms of the AGPLv3.

The reason is mostly because I don't want low-effort proprietary forks that add bad code that steals everyone's passwords, or something like that.

If you're doing anything else other than forking, especially if you want to some client code files in your own open-source project that you want to release under a more permissive license (like, if you want to make your own multiplayer open-source game client for a different game), please ask at staff@pokemonshowdown.com. I hold all the copyright to the AGPLv3 parts and can relicense them to MIT for you.

WARNING: This is NOT the same license as Pokémon Showdown's server.

-README.md

data/text.js
```diff
@@ -779,6 +785,7 @@ exports.BattleText = {
 	skydrop: {
 		prepare: "[POKEMON] took [TARGET] into the sky!",
 		end: "  [POKEMON] was freed from the Sky Drop!",
+		fail: "#grassknot",
```
KrisXV: This is wrong. As SadisticMystic said in the discord, Sky Drop just gets permanently grayed out against a dynamax Pokemon, similarly to Belch pre-berry-consumption.

PartMan7:
Understood; will fix when I get back home.

Issue #1417, opened 2019-12-03
Add central team storage and team browser


marksg07:
This feature would be super cool to have from a user perspective. It solves two issues that have often annoyed me, and (I'm guessing) many other users:

1. Using PS on multiple devices. I play PS on my home desktop computer, my work laptop, and my phone at different times during the day. When I play, I often will make modifications to my teams. When I want to switch to a different device, I have to export my team to pokepastes or email it to myself from the original device, and then get the team and import it on the other device. This is a pain in the ass to do multiple times per day.

2. Sharing teams with other people. There isn't a central repository of teams that I can look through if I want to try using a new team. Also, if I want to share a team with a specific person on PS, I have to export to pokepastes and message them either through PS, which I believe only works if they are online, or through some other medium, which (again) is a pain in the ass. And in order to use someone else's team that they have shared with me or the world, I have to grab their team export and import it into my teambuilder. 

Both of these issues could be solved by maintaining a central team database on a PS server that could be directly accessed through the client. Essentially, users could upload teams to this database publicly or privately -- publicly would mean that the team could be found by anyone through a public team browser; privately would mean that only certain users could see the team (by default, just the person who uploaded the team). This solves the first issue, because I can upload my changed team privately before I switch device, and then download it from my "uploaded teams" screen on my new device. As for the second issue, I could share my team with someone else using their PS username (even if they were offline), and they could download that team from their "shared teams" screen. There would also be a "public teams" screen. All of these lists of teams could be searchable by team name, user name, and possibly even pokemon or pokemon sets.

joebangara43903:
![image](https://user-images.githubusercontent.com/131228993/232965491-01a24040-2b6b-4659-9274-066aa35fe67a.png)

PR #1418, opened 2019-12-03
Add team storage to login server (See #1417)

marksg07:
Database/backend part of #1417. See also the PR on the game server side that uses the `uploadteam` and `shareteam` actions: smogon/pokemon-showdown#6140.

This PR adds five new actions for the login server, which all relate to storing teams in a database. It also includes schemas for new tables that would store teams and team-related information. The actions are:

- `getuploadedteams`: Gets the user's uploaded teams. User must be logged in.
- `getsharedteams`: Gets teams shared directly with the user. User must be logged in.
- `getpublicteams`: Gets teams public to everyone.

The above three actions are all available directly to the client, and they are all paged -- the server enforces a maximum page size of 50, so the user can only fetch 50 teams per request. This is to prevent against the user downloading a ton of teams from the server all at once. 

- `uploadteam`: Upload a new team to the team database.
- `shareteam`: Share one user's team with a different user.

The above two actions are server-only; if the client attempts them they will immediately fail. This is to prevent against the user being able to add arbitrary stuff to the DB tables. The game server will prevent against invalid teams for given format being uploaded and other forms of invalid input.

`ntbb_teams` and `ntbb_sharelist` are the two tables that should be added to the PS database in order for these actions to work.

This is a first crack, future improvements would probably relate to filtering options for the `getXXXteams` actions (i.e. filter by username, teamname, pokemon, ...).

### DIFF
--- a/action.php
+++ b/action.php
@@ -32,6 +32,7 @@ include_once 'lib/dispatcher.lib.php';
 
 $dispatcher = new ActionDispatcher(array(
 	new DefaultActionHandler(),
-	new LadderActionHandler()
+	new LadderActionHandler(),
+	new TeamsActionHandler()
 ));
 $dispatcher->executeActions();
--- a/lib/dispatcher.lib.php
+++ b/lib/dispatcher.lib.php
@@ -605,7 +605,6 @@ class TeamsActionHandler {
 	/**
 	 * Fetches all published teams uploaded by the user.
 	 * Mandatory request args: userid: user id; page, pagesize: control the range of data that is returned.
-	 * Optional request args: match: string to regex match teamname against.
 	 * Server only.
 	 */
 	public function getuploadedteams($dispatcher, &$reqData, &$out) {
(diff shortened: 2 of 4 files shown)


Zarel:
Quite impressive. Unfortunately, MySQL sucks, so we might need to complete the Postgres transition before this is usable at all.

marksg07:
Is that difficulty related to the way the PR is structured, or is it something more general about MYSQL that makes it suck?

Zarel:
It's a more general problem about our databases being in a pretty bad state right now.

marksg07:
@Zarel Is there any chance that this gets merged pre-database fix? If not, can I do something like make a "database transition" issue so that I can get notified when that's done? I'm eager to continue work on this.

Zarel:
An issue tracking it makes sense, yes.

Zarel:
Depends on #1423

PR #1419, opened 2019-12-07
Added favoriting system

jordanhui19:
Pull request for issue #6158

I created a feature where users can 'favorite' other users which allows them to be quickly found later, without having to search up their username.
The favorited users are stored in local storage and parsed into an array when needed.

favebuttonadded
favebuttontoggles
I added a favorite/unfavorite button to the user popup which toggles depending on if the selected user is already favorited.

viewallfave
I also added a button under "Find a user" to display all your favorited users. I copied the css properties of button6 to this button and named it button7 since I noticed that 6 wasn't being used but I wasn't sure if i should attach my own feature to it.

viewallfavewindow
viewallfavpopup
Clicking the button opens a pm style window under the news window which shows all the saved usernames. Clicking them opens the user popup like the other instances of username.
The list also updates itself when a user is added or removed.

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -52,7 +52,7 @@
 
 			buf += '<div class="menugroup"><p><button class="button mainmenu4 onlineonly disabled" name="joinRoom" value="battles">Watch a battle</button></p>';
 			buf += '<p><button class="button mainmenu5 onlineonly disabled" name="finduser">Find a user</button></p>';
-			buf += '<p><button class="button mainmenu7 onlineonly disabled" name="favoritedUsers">Favorited users</button></p></div>'
+			buf += '<p><button class="button mainmenu7 onlineonly disabled" name="favoritedUsers">Favorited users</button></p></div>';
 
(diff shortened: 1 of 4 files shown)


jordanhui19:
Issue #1420

Lusamine:
This is different from a friend system on Discord, for example, which requires you to agree to add someone.  While I can see the benefits, I can see people using this to better track people who don't want to be tracked.  What is going to stop bad users from favoriting me in order to track when I am online to harass?

AeonicX7:
This feature could make it extremely easy to target and harass users (especially staff). For instance, if I lock a troll and they decide to favorite me, they can then easily PM me and know when I'm online to bother me constantly. I agree with Anubis @ above re: both users need to consent for you to be favorited (and that shouldn't ping the requested user because trolls will then spam me with that).

monsanto:
Thanks for the PR, sorry it never got merged. A friends system was added in the time since. I don't know if that will fully address what this PR is trying to do but if it doesn't we'd need another PR that builds on the existing friends system.

PR #1424, opened 2019-12-18
Teambuilder: Import teams from pokepast.es or gist urls

BlueWinds:
This PR is based on https://github.com/smogon/pokemon-showdown-client/issues/1349, which is a feature I've wanted for some time myself. It only attempts to import if a valid-looking URL is the *only* thing in the team textbox, and it only looks for a single URL. I figure this meets 95% of actual use, and so is a good place to start.

I've attempted to match the style of the surrounding code, but please let me know if there's a different preferred style for anything.

I used $.ajax rather than $.get in order to bypass the testclient CORS helper - it doesn't play nicely with multiline return values. Due to the fact that pokepast.es insists on https, you'll also have to either host testclient.html on https localhost, or temporarily disable your browser's content security policy. This of shouldn't be an issue when it's on a real server, but it does unfortunately make it a pain to work with locally.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1092,7 +1092,7 @@
 					buf += '<li><button name="undeleteSet"><i class="fa fa-undo"></i> Undo Delete</button></li>';
 				}
 				if (i === 0) {
-					buf += '<li><button name="import" class="button big"><i class="fa fa-upload"></i> Import from text</button></li>';
+					buf += '<li><button name="import" class="button big"><i class="fa fa-upload"></i> Import from text or URL</button></li>';
 				}
 				if (i < 6) {
 					buf += '<li><button name="addPokemon" class="button big"><i class="fa fa-plus"></i> Add Pok&eacute;mon</button></li>';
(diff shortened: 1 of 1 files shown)


BlueWinds:
First time working with the repo, and hadn't realized there was a linter. I amended the initial commit to fix the lint errors.

js/client-teambuilder.js
```diff
@@ -1213,8 +1213,43 @@
...
+				$.ajax({
+					type: 'GET',
+					url: url,
+					success: function (data) {
+						Storage.activeSetList = self.curSetList = Storage.importTeam(data);
+						self.$('.teamedit textarea, .teamedit .savebutton').attr('disabled', null);
+						self.back();
+					},
+					error: function () {
+						self.$('.teamedit textarea, .teamedit .savebutton').attr('disabled', null);
```
Zarel: No error message?

BlueWinds: Doh. Added popup message.

Zarel:
This actually looks pretty good. Can I get you to license your contribution under the MIT license? Just say "I hereby license all my contributions to Smogon, past and future, under the MIT license".

BlueWinds:
I hereby license all my contributions to Smogon, past and future, under the MIT license.
:)

PR #1426, opened 2019-12-19
Add support for NatDex metas

KrisXV:


js/search.js
```diff
@@ -802,6 +803,12 @@
 		if (format.includes('doubles')) this.isDoubles = true;
 		var isLetsGo = format.startsWith('letsgo');
 		if (isLetsGo) format = format.slice(6);
+		var isNatDex = format.startsWith('nationaldex');
+		if (isNatDex) {
+			format = format.slice(11);
+			this.isNatDex = true;
+		}
+		if (format === 'beta') format = 'ou';
```
KrisXV: this can be removed once (beta) is removed from natdex, and is preferrably renamed National Dex OU

PR #1428, opened 2019-12-20
Add viability checks for various moves

KrisXV:


js/search.js
```diff
@@ -1093,10 +1093,12 @@
 				var isViable = BattleMovedex[id] && BattleMovedex[id].isViable;
 				if (id === 'aerialace') isViable = (toID(set.species) in {scyther:1, aerodactylmega:1, kricketune:1});
 				if (id === 'ancientpower') isViable = (toID(set.ability) === 'technician' || (toID(set.ability) === 'serenegrace') || (template.types.indexOf('rock') > 0 && moves.indexOf('powergem') < 0));
+				if (id === 'aurawheel') isViable = (toID(set.species).startsWith('morpeko'));
```
Zarel: Isn't it always viable? Just add the flag to moves.js.

js/search.js
```diff
@@ -1121,10 +1124,11 @@
 				if (id === 'skyattack') isViable = (toID(set.species) === 'hawlucha');
 				if (id === 'smackdown') isViable = (template.types.indexOf('ground') > 0);
 				if (id === 'smartstrike') isViable = (template.types.indexOf('steel') > 0 && moves.indexOf('ironhead') < 0);
-				if (id === 'solarbeam') isViable = (toID(set.abilities) in {drought:1, chlorophyll:1});
+				if (id === 'solarbeam' || id === 'solarblade') isViable = (toID(set.ability) in {desolateland:1, drought:1, chlorophyll:1});
 				if (id === 'stompingtantrum') isViable = ((moves.indexOf('earthquake') < 0 && moves.indexOf('drillrun') < 0) || (toID(set.ability) === 'toughclaws' && moves.indexOf('drillrun') < 0 && moves.indexOf('earthquake') < 0));
 				if (id === 'storedpower') isViable = (toID(set.species) in {necrozma:1, espeon:1, sigilyph:1});
 				if (id === 'stunspore') isViable = (moves.indexOf('thunderwave') < 0);
+				if (id === 'teleport') isViable = (this.gen > 7);
```
Zarel: Drypass is extremely rare... how many people actually use Teleport competitively?

js/search.js
```diff
@@ -1121,10 +1124,11 @@
 				if (id === 'skyattack') isViable = (toID(set.species) === 'hawlucha');
 				if (id === 'smackdown') isViable = (template.types.indexOf('ground') > 0);
 				if (id === 'smartstrike') isViable = (template.types.indexOf('steel') > 0 && moves.indexOf('ironhead') < 0);
-				if (id === 'solarbeam') isViable = (toID(set.abilities) in {drought:1, chlorophyll:1});
+				if (id === 'solarbeam' || id === 'solarblade') isViable = (toID(set.ability) in {desolateland:1, drought:1, chlorophyll:1});
```
Zarel:
```suggestion
				if (id === 'solarbeam' || id === 'solarblade') isViable = ['desolateland', 'drought', 'chlorophyll'].includes(toID(set.ability));
```

js/search.js
Zarel: nvm, apparently it fails unless the user is Morpeko.

Weird restriction.

js/search.js
KrisXV: people use it as a slow pivoting move to get a guaranteed switch into a pokemon that a teammate checks

Zarel: I guess I won't judge. PS's standards for usable moves are supposed to be really low.

js/search.js
Zarel: I would still flag it in `moves.js`, for the record.

PR #1430, opened 2019-12-20
Import sample sets from smogondex for teambuilder

BlueWinds:
Based on https://github.com/smogon/pokemon-showdown-client/issues/1357, this PR causes the teambuilder to import sets from smogon (based on team format) and provides buttons to easily fill them in on the "Import Pokemon" window.

It currently offers a button for all available sets, both official and "Showdown Usage". This has looked decent on the 'mons I've tested so far. It doesn't show any error on a failed import, because there aren't sets available for all formats (eg, gen8vgc2020), and it didn't end up feeling right to show anything visually.

image

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1487,16 +1487,18 @@
 			var formatSets = this.smogonSets[this.curTeam.format];
 			var species = this.curSet.species;
 
-			var setDiv = this.$('.teambuilder-pokemon-import .teambuilder-import-smogon-sets');
-			setDiv.empty();
+			var $setDiv = this.$('.teambuilder-pokemon-import .teambuilder-import-smogon-sets');
+			$setDiv.empty();
 
 			if (!formatSets) {
 				return;
 			}
 
 			var sets = $.extend({}, formatSets['smogon.com/dex'][species], formatSets['smogon.com/stats'][species]);
+
+			$setDiv.text('Sample sets: ');
 			for (var set in sets) {
-				setDiv.append('<button name="importSmogonSet">' + set + '</button>');
+				$setDiv.append('<button name="importSmogonSet">' + BattleLog.escapeHTML(set) + '</button>');
 			}
 		},
 		importSmogonSet: function (i, button) {
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1454,6 +1454,61 @@
 				.val(Storage.exportTeam([this.curSet]).trim())
 				.focus()
 				.select();
+
+			this.getSmogonSets();
+		},
+		getSmogonSets: function () {
+			this.$('.teambuilder-pokemon-import .teambuilder-import-smogon-sets').empty();
+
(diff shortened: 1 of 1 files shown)


BlueWinds:
Also, it looks like  the preact/typescript client in src/ is in active development. Do you prefer features like this be implemented there as well, and if so, in the same or a separate PR?

Zarel:
Here is probably more important. The Preact client is a decent ways off, but feel free to also PR there.

js/client-teambuilder.js
```diff
@@ -1454,6 +1454,61 @@
...
+			var setDiv = this.$('.teambuilder-pokemon-import .teambuilder-import-smogon-sets');
+			setDiv.empty();
+
+			if (!formatSets) {
+				return;
+			}
+
+			var sets = $.extend({}, formatSets['smogon.com/dex'][species], formatSets['smogon.com/stats'][species]);
+			for (var set in sets) {
+				setDiv.append('<button name="importSmogonSet">' + set + '</button>');
```
```suggestion
				setDiv.append('<button name="importSmogonSet">' + BattleLog.escapeHTML(set) + '</button>');
```

js/client-teambuilder.js
```diff
@@ -1454,6 +1454,61 @@
...
+			var species = this.curSet.species;
+
+			var setDiv = this.$('.teambuilder-pokemon-import .teambuilder-import-smogon-sets');
+			setDiv.empty();
+
+			if (!formatSets) {
+				return;
+			}
+
+			var sets = $.extend({}, formatSets['smogon.com/dex'][species], formatSets['smogon.com/stats'][species]);
```
Zarel: I suspect this will be less confusing labeled with something:

```suggestion
			var sets = $.extend({}, formatSets['smogon.com/dex'][species], formatSets['smogon.com/stats'][species]);
			setDiv.text('or import from: ');
```

js/client-teambuilder.js
```diff
@@ -1454,6 +1454,61 @@
...
+					self.smogonSets[format] = false;
+				}
+				self.importSetButtons();
+			});
+		},
+		importSetButtons: function () {
+			var formatSets = this.smogonSets[this.curTeam.format];
+			var species = this.curSet.species;
+
+			var setDiv = this.$('.teambuilder-pokemon-import .teambuilder-import-smogon-sets');
```
```suggestion
			var $setDiv = this.$('.teambuilder-pokemon-import .teambuilder-import-smogon-sets');
```

Just a habit to make it clearer that a variable is jQuery.

js/client-teambuilder.js
BlueWinds: Done.

js/client-teambuilder.js
BlueWinds: Good call, added "Sample sets:"

js/client-teambuilder.js
BlueWinds: :+1:

Zarel:
Looks really good!

PR #1432, opened 2019-12-23
Add gen 8 Illusion support

KrisXV:


Zarel:
Can you clarify what this changes?

src/battle-animations.ts
```diff
@@ -626,10 +626,10 @@ class BattleScene {
 				if (speciesOverage && speciesTable.includes(species)) {
 					for (const sidebarIcon of sidebarIcons) {
 						if (side.pokemon[sidebarIcon[1]!].getBaseTemplate().baseSpecies === species) {
-							sidebarIcon[0] = 'pokemon-illusion';
+							if (this.gen < 8) sidebarIcon[0] = 'pokemon-illusion';
```
Zarel: For one thing, why would you put an `if` this deep into an inner loop? You're running the loop for no reason if `this.gen >= 8`.

Zarel: For another thing, do you know what this does? This makes it so that if Species Clause is in effect but there are two Pokémon of the same species on the field, one of them is marked as an Illusion. This seems obviously correct.

src/battle-animations.ts
```diff
@@ -669,7 +669,7 @@ class BattleScene {
 				} else {
 					pokemonhtml += `<span${tooltipCode} style="` + Dex.getPokemonIcon('pokeball') + `" aria-label="Non-statused"></span>`;
 				}
-			} else if (iconType === 'pseudo-zoroark') {
+			} else if (iconType === 'pseudo-zoroark' && this.gen < 8) {
```
Zarel: In addition to being an incorrect interpretation of mechanics, just adding this condition will make the `gen >= 8` condition fall down to `if (!poke)`, turning a flag into an extra pokeball which is clearly not your intent.

PR #1433, opened 2019-12-27
Tooltips: Remove Choice item boost from Dynamaxed Pokemon

KrisXV:
I also added tooltips for the defender-based Utility Umbrella changes

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -1204,7 +1204,7 @@ class BattleTooltips {
 		if (move.id === 'blizzard') {
 			value.weatherModify(0, 'Hail');
 		}
-		if ((move.id === 'hurricane' || move.id === 'thunder')) {
+		if (move.id === 'hurricane' || move.id === 'thunder') {
 			value.weatherModify(0, 'Rain Dance');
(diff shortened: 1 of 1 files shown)


Zarel:
I don't think this affects the first turn of dynamaxing yet?

src/battle-tooltips.ts
```diff
@@ -1192,7 +1204,7 @@ class BattleTooltips {
 		if (move.id === 'blizzard') {
 			value.weatherModify(0, 'Hail');
 		}
-		if (move.id === 'hurricane' || move.id === 'thunder') {
+		if ((move.id === 'hurricane' || move.id === 'thunder')) {
```
Zarel: ?

KrisXV: hehe oops

KrisXV:
no it doesn't @Zarel because the pokemon doesn't have the volatile applied yet

Zarel:
I mean, it should if you check the "Dynamax" button.

KrisXV:
how do i do that

Zarel:
It's not currently possible, you'd have to pass in another parameter or grab it from somewhere.

PR #1435, opened 2019-12-27
Fix import of smogon sets

BlueWinds:
The code comment should explain what went wrong and how it's fixed.

Basically, I forgot that jQuery - trying to be helpful - would parse the response into an object unless asked not to.  Inconsistency between the CORs helper and the actual implementation.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1528,6 +1528,9 @@
 				this.importSetButtons();
 				return;
 			}
+			// We fetch this as 'text' and JSON.parse it ourserves in order to have consistent behavior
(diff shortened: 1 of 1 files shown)


Zarel:
Wait, I thought it had been tested. Does `testclient` not work for this?

BlueWinds:
> Wait, I thought it had been tested. Does `testclient` not work for this?

It did work on testClient and was tested there. The problem is that the behavior *outside of* test client was broken. This makes the test client and 'production' work the same (by telling jQuery not to parse the object, the same way testClient wasn't parsing the object).

Zarel:
Oh, weird.

PR #1438, opened 2020-01-03
Fix unreleased hidden abilities in past gens

KrisXV:
Currently, Pokemon with no access to their hidden ability in gen 8 (such as clefable) can still use them in gen 7, etc; however, the teambuilder still crosses them out.


KrisXV:
Bump

Zarel:
I don't think this deserves a teambuilder table. This is only relevant for the brief period of time following a generation before Pokémon Home is out, and is only relevant for the most recent generation. Teambuilder tables are sophisticated tables designed for things that can change every generation.

I'd probably just implement this as an array of mons somewhere.

Zarel:
Maybe teambuilder tables themselves should be revised to be smarter, though. The current situation is a bit untenable.

Zarel:
I actually think the "Past" flag is enough by itself. The teambuilder table itself is unnecessary.

KrisXV:
done @Zarel

PR #1443, opened 2020-01-17
Add teambuilder support for Metronome Battle

KrisXV:


KrisXV:
bump?

Zarel:
rebase?

KrisXV:
done @Zarel

Zarel:
Do we actually need a table for just "metronome battles"?

Zarel:
It's just, Metronome's legal, right?

Zarel:
Oh, it has its own tiers?

KrisXV:
it doesnt have its own tiers, but it requires its own teambuilder table because it allows every single Pokemon and the teambuilder currently display that, so a lot of users think the meta is Galar dex only.

Zarel:
That shouldn't be a table, that should just be a setting in search.js to consider it a NatDex format.

Zarel:
Or a hackable format.

KrisXV:
I'm not sure what you mean. In search.js it is labeled as a natdex format, but more can be used than just natdex Pokemon. Not to mention the tiers for it are also used to filter out the bans

Zarel:
I'm saying it's simple enough to hardcode rather than generating a teambuilder table for it.

KrisXV:
There's also a plethora of banned items that could be added to a table. if i did that could the generated table stay? I don't really like the idea of hardcoding a teambuilder list for a format that constantly has changing bans.

Zarel:
Oh, huh.

Zarel:
That's annoying. The current systems aren't really set up for this sort of thing...

Zarel:
Hey wait, there aren't even any teambuilder item tables in this PR. What are you actually trying to do?

KrisXV:
There aren’t any because I forgot to add them. I can change that though

KrisXV:
bump

Zarel:
Sooner or later we need to genericize this code instead of piling on more special cases.

PR #1444, opened 2020-01-17
Add a 'hide' property for Pokemon

KrisXV:
Needs https://github.com/smogon/pokemon-showdown/pull/6291 to be merged first


KrisXV:
Should I add a comment above the line that notes that `hide` is implemented server-side?

monsanto:
The linked issue was closed so I am assuming this is no longer something we're doing.

PR #1445, opened 2020-01-17
Fix build

KrisXV:


KrisXV:
this is probably wrong

test/battle.test.js
```diff
@@ -95,12 +95,12 @@ describe('Text parser', () => {
 	it('should process messages correctly', () => {
 		let parser = new BattleTextParser();
 
-		assert.equal(parser.extractMessage(`|-activate|p2a: Cool.|move: Skill Swap|Speed Boost|Cute Charm|[of] p1a: Speedy`), `  [The opposing Cool.'s Speed Boost]
-  [Speedy's Cute Charm]
```
Zarel: This is also wrong, the two leading spaces mean "display this in small font".

data/text.js
```diff
@@ -5,7 +5,7 @@ exports.BattleText = {
 		tieBattle: "Tie between [TRAINER] and [TRAINER]!",
 
 		pokemon: "[NICKNAME]",
-		opposingPokemon: "the opposing [NICKNAME]",
+		opposingPokemon: "The opposing [NICKNAME]",
```
Zarel: This part is definitely wrong, yes. It should be lowercase in the middle of a sentence; there's other code which capitalizes this when it's at the beginning of a sentence.

PR #1447, opened 2020-01-19
Fix App#renameRoom

TheJetOU:
It was missing some features like updating title bar, autojoins, etc.

Server PR: smogon/pokemon-showdown#6295.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -879,7 +879,8 @@ function toId() {
 						self.send('/join ' + roomid);
 					});
 				} else if (data === 'rename') {
-					this.renameRoom(roomid, errormessage);
+					var parts = errormessage.split('|');
+					this.renameRoom(roomid, parts[0], parts[1]);
 				} else if (data === 'nonexistent' && Config.server.id && roomid.slice(0, 7) === 'battle-' && errormessage) {
 					var replayid = roomid.slice(7);
 					if (Config.server.id !== 'showdown') replayid = Config.server.id + '-' + replayid;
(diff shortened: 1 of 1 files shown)


Zarel:
`roomid === toRoomid(title)` is not necessarily guaranteed.

For instance, `battle-ou-6234532` might have title `[Gen 7] OU: Zarel vs Yuyuko`

And `wifi` might have title `Wi-Fi`.

Neither of these match `toRoomid`.

TheJetOU:
I suppose we need to make the server send `roomid|title`.

js/client.js
```diff
@@ -1722,18 +1723,25 @@ function toId() {
 			if (room.requestLeave && !room.requestLeave(e)) return false;
 			return this.removeRoom(id);
 		},
-		renameRoom: function (id, newid) {
+		renameRoom: function (id, newtitle, newid) {
```
```suggestion
		renameRoom: function (id, newid, newtitle) {
```

Bikeshedding, but `newid` should be first for backwards compatibility reasons.

js/client.js
```diff
@@ -1722,18 +1723,25 @@ function toId() {
...
+			var newid = newid || toRoomid(newtitle);
```
Zarel: And `|| toRoomid(newtitle)` should be dropped; it should instead be `newtitle = newtitle || newid`.

js/client.js
```diff
@@ -879,7 +879,9 @@ function toId() {
 						self.send('/join ' + roomid);
 					});
 				} else if (data === 'rename') {
-					this.renameRoom(roomid, errormessage);
+					// |newtitle|newid
```
Zarel: This should also be `|newid|newtitle`.

Unless it wasn't in the past?

TheJetOU: > Unless it wasn't in the past?

It didn't exist in the past. I just appended roomid to it in the server code when I realized it was necessary. Is there a difference?

Zarel: Was it accepting titles in the past?

Zarel: It seems like `|rename|newid` was the old signature. If you want to support titles, it should go after the id, not before.

PR #1448, opened 2020-01-20
Switch Aura Wheel display type when Morpeko-Hangry

patrickbahr:
Just a quick change to display Aura Wheel as Dark type when in Morpeko-Hangry form. It already calcs the move as Dark type, but in the move selection it showed as Electric type and that could confuse some players. 
If there's anything that needs to change or if this is not a valid feature let me know, this is my first time contributing to the repo =D

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -1172,14 +1172,14 @@ class BattleTooltips {
 		}
 
 		// Aura Wheel as Morpeko-Hangry changes the type to Dark
-		if(move.id === 'aurawheel') {
-			if(value.pokemon.getTemplate().species === 'Morpeko-Hangry') {
+		if (move.id === 'aurawheel') {
+			if (value.pokemon.getTemplate().species === 'Morpeko-Hangry') {
 				moveType = 'Dark';
 			} else {
 				moveType = 'Electric';
 			}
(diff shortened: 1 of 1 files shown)


TheImmortal:
Uh, I think you need to exclude transformed Pokemon as Ditto transformed into Morpeko doesn't follow this mechanic exactly.

patrickbahr:
@TheImmortal according to the issue smogon/pokemon-showdown#6084, Ditto's Aura Wheel type respects the form that Ditto transforms into (Electric if Morpeko, Dark if Morpeko-Hangry), even though in-game it always shows as Electric-Type... Should we keep showing that is Electric as well, even if it is misleading?

src/battle-tooltips.ts
```diff
@@ -1170,6 +1170,16 @@ class BattleTooltips {
 				break;
 			}
 		}
+
+		// Aura Wheel as Morpeko-Hangry changes the type to Dark
+		if (move.id === 'aurawheel') {
+			if (value.pokemon.getTemplate().species === 'Morpeko-Hangry') {
+				moveType = 'Dark';
+			} else {
+				moveType = 'Electric';
```
Zarel: This part seems unnecessary.

Issue #1408, opened 2019-11-23
Hovering over Gmax moves shows tooltip of their dynamax counterpart. 


TZubiri:
For example, Gmax resonance shows the info of Max Hailstorm.

PR #1454, opened 2020-01-23
Add support for Max Guard and G-Max Moves to tooltips

HoeenCoder:


src/battle-tooltips.ts
```diff
@@ -476,14 +479,9 @@ class BattleTooltips {
 				});
 				zEffect = this.getStatusZMoveEffect(move);
 			} else {
-				let moveid;
-				if (item.zMoveType) {
-					moveid = BattleTooltips.zMoveTable[item.zMoveType as TypeName];
-				} else {
-					moveid = BattleTooltips.maxMoveTable[move.type as TypeName];
-				}
+				let moveid = BattleTooltips.zMoveTable[item.zMoveType as TypeName];
```
KrisXV: nitpick but this should be called `moveName` since the moves in the zMoveTable aren't IDs

HoeenCoder:
Fixes #1408

HoeenCoder:
Bump @Zarel

src/battle-dex-data.ts
```diff
@@ -1177,7 +1177,7 @@ class Move implements Effect {
 				} else if (this.basePower >= 45) {
 					this.gmaxPower = 75;
 				} else  {
-					this.gmaxPower = 10;
+					this.gmaxPower = 70;
```
Zarel: Does this need to be fixed on server, too, or is it implemented differently there?

Zarel: Looks like SadisticMystic fixed it here, but also fixed some other things which should probably be propagated here.

https://github.com/smogon/pokemon-showdown/commit/b2e9336c327efbd22d36eb69f9b7905cd3955fb4

src/battle-tooltips.ts
```diff
@@ -268,13 +268,16 @@ class BattleTooltips {
 		let buf: string;
 		switch (type) {
 		case 'move':
-		case 'zmove': { // move|MOVE|ACTIVEPOKEMON
+		case 'zmove':
+		case 'maxmove': { // move|MOVE|ACTIVEPOKEMON
```
Zarel:
```suggestion
		case 'maxmove': { // move|MOVE|ACTIVEPOKEMON|[GMAXMOVE]
```

HoeenCoder:
I have not forgotten about this, I'll finish it up when I get off LOA in a few days/get a bit of time to do it.

Zarel:
Thanks!

PR #1455, opened 2020-01-24
Fix teambuilder tiers for previous generations

KrisXV:
This fixes the issue where Pokemon in past-gens doubles formats would have their tiers display as their gen 8 singles tiers (corsola in gen 4 doubles would display as unreleased) and also where Mega Pokemon in past gens would have their tiers appear as their base forme's

### DIFF
--- a/src/battle-dex.ts
+++ b/src/battle-dex.ts
@@ -443,29 +443,27 @@ const Dex = new class implements ModdedDex {
 		return template;
 	}
 
-	/** @deprecated */
-	getTier(pokemon: Template, gen = 7, isDoubles = false): string {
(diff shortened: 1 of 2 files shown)


KrisXV:
bump @Zarel

src/battle-dex.ts
```diff
@@ -443,29 +443,27 @@ const Dex = new class implements ModdedDex {
 		return template;
 	}
 
-	/** @deprecated */
```
Zarel: This is still deprecated. :[ The Preact client doesn't use this.

KrisXV: oh my bad, i didnt mean to remove it

PR #1459, opened 2020-02-05
Update desktop client

Zarel:
Last time we built the desktop client, in 2015, NW.js was called node-webkit.

Since then:

- NW.js has added NaCl and some other things, ballooning the installer from 30 MB to 120 MB. (The NaCl stuff on Windows seems to be deletable, helping keep its installed size slightly smaller.)

  (If anyone wants to make a custom build without NaCl, that'd help me a lot)

- NW.js also has a lot more files, and NSIS makes adding/removing files annoying enough that I wrote a script to automate it for me.

- Apple has made Gatekeeper requirements a lot stricter, requiring setting up entitlements, Hardened Runtime, notarization, and stapling, (in addition to just signing in 2015). I documented the process because there are zero online guides for it right now.

But on the other hand, this update hopefully fixes the crashes that the old desktop client has. It also fixes:

- the Chrome bold emoji bug is gone

- MP3 is no longer under patent, allowing us to have only MP3 files for everything

- honestly I don't think we use any other Chrome feature added in the last four years


Zarel:
The new desktop client is already deployed on the website, this is just the code changes for it.

PR #1463, opened 2020-02-11
Update battle-animations.ts for battle crys in digimon mode

josephjoestar:
need to pass this.mod as argument in order for it to work

### DIFF
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -1738,7 +1738,7 @@ class PokemonSprite extends Sprite {
 		if (this.$sub) return;
 		const subsp = Dex.getSpriteData('substitute', this.siden, {
 			gen: this.scene.gen,
-			mod: this.mod,
+			mod: this.scene.mod,
 		});
 		this.subsp = subsp;
 		this.$sub = $('<img src="' + subsp.url + '" style="display:block;opacity:0;position:absolute"' + (subsp.pixelated ? ' class="pixelated"' : '') + ' />');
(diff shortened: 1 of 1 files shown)


Zarel:
Do you see that Travis build error?

josephjoestar:
> Do you see that Travis build error?

Yep had it fixed

Zarel:
Hm, it seems like having both `gen` and `mod` in the same call is probably unnecessary. I wonder if there's some way to merge them.

josephjoestar:
do you think concatenating mod after gen. and splitting it after it is passed is a good idea?

josephjoestar:
All the digimon have num<2000. Is it accepatable if I change
`	if (options.mod) {
			spriteData.cryurl = `sprites/${options.mod}/audio/${toID(template.baseSpecies)}`;
			spriteData.cryurl += (window.nodewebkit ? '.ogg' : '.mp3');
		}`

to
`if (pokemon.num<2000) {
			spriteData.cryurl = `sprites/digimon/audio/${toID(template.baseSpecies)}`;
			spriteData.cryurl += (window.nodewebkit ? '.ogg' : '.mp3');
		}``

and just get rid of mod

Zarel:
Mod should already include gen, if you're using the correct format... PS's own mods probably need to be updated, though. :/

josephjoestar:
as far as I checked the battlescene.mod was only set when battle.id.includes('digimon'). I wasn't aware there were any other uses outside of setting the audio for digimon.

Zarel:
Well, I wish we had a better architecture, but I don't have time to design one, and this isn't too bad. I'll accept it as-is.

PR #1464, opened 2020-02-12
Add teambuilder support for NFE

KrisXV:


Zarel:
If we're going to add teambuilder support for everything ever, we probably want a way more generic system than this one, which requires so many haphazard changes in so many different places.

KrisXV:
Yep; I'm working on one.

Zarel:
I suppose this is fine for now.

PR #1466, opened 2020-02-20
Add Heavy Duty Boots, Expert Belt, and Room Service to Popular items

petuuuhhh:
Heavy Duty Boots should go in the Popular items section. I discussed this briefly on the Discord and the reception was unanimously positive.

I would like Expert Belt and Room Service to be added to the Popular items section as well.


Zarel:
Which Discord?

Also if you want the others to be added, just add them here, don't make three PRs for them. :/

PR #1468, opened 2020-02-26
Add link back to Smogon analysis for sample sets

### LINKED ISSUE #1357: Provide a way to easily import sets into the teambuilder

scheibo:
Closes #1357

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1510,9 +1510,7 @@
 
 			var format = this.curTeam.format;
 			// If we don't have a specific format, don't try and guess which sets to use.
-			if (format.match(/gen\d$/)) {
-				return false;
-			}
+			if (format.match(/gen\d$/)) return;
 
 			var self = this;
 			this.smogonSets = this.smogonSets || {};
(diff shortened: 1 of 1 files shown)


Zarel:
Thanks!

PR #1473, opened 2020-03-08
Add Sinistea and Polteageist minisprite indexes

KrisXV:
The regular sprites can also be used for sinistea-antique and polteageist-antique


Zarel:
Are we using the new sprite sheets? What's going on with the sprite sheets?

KrisXV:
They’re pinned in #general-dev

PR #1475, opened 2020-03-11
Attribute /raw messages, fix /uhtml in chat

whalemer:
(allowing for `|c|user|/raw` messages to be hidden, among other things)

clientside half of https://github.com/smogon/pokemon-showdown/pull/6444

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1611,7 +1611,7 @@
 			}
 
 			var isHighlighted = userid !== app.user.get('userid') && this.getHighlight(message);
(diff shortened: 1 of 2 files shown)


Zarel:
Is this for misbehaving bots?

Zarel:
I think it might be a good idea to keep uhtml and chat messages separate...

whalemer:
>Is this for misbehaving bots?

Yes, and misbehaving / accidents from users.

>I think it might be a good idea to keep uhtml and chat messages separate...

Not sure what you mean by this.

Zarel:
"Users"? Only bots should have access to `/addhtml`, right?

whalemer:
Yeah, the html attribution is for misbehaving bots (or for when ashe accidentally links death grips covers). The !code attribution is for accidents from users.

PR #1476, opened 2020-03-11
Begin adding sprite tests

scheibo:
Confidently implementing #1369 is fairly intractable currently as there are literally thousands of possible combinations. This introduces some testing (fairly comprehensive for `getTeambuilderSpriteData`, much *much* more required for even basic levels of confidence in `getSpriteData`).


scheibo:
The current game plan for #1369 is:

- this PR with initial tests
- add more thorough `getSpriteData` tests
- implement new logic for #1369, doesn't actually get used because no way to select in the UI
- add support for the selector in the UI

Zarel:
I always thought this would be more of a burden than anything, since any refactor to this code would want to change its output, too.

scheibo:
> I always thought this would be more of a burden than anything, since any refactor to this code would want to change its output, too.

Yes, but you need to be able to verify the output youre doing with the refactor as well.

scheibo:
can the github workflow not write to disk or something?

.github/workflows/test.yml
```diff
@@ -25,6 +25,6 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
     - run: npm install
-    - run: npm run full-test
+    - run: npm run test
```
Zarel: ?

Zarel:
Incidentally: I would probably rather pad all dex sprites to 128x128 than have to maintain two sizes for them.

.github/workflows/test.yml
scheibo: > EDIT: oh spawnSync /bin/sh ENOENT
> what fun

scheibo:
> Incidentally: I would probably rather pad all dex sprites to 128x128 than have to maintain two sizes for them.

Out of scope for this change. :)

.github/workflows/test.yml
scheibo: basically, im fine with these tests not running through CI due to not being able to download the data files. we could probably force git to do it through a submodule init or possibly by including it in the run instructions here, but its not important. having tests which you can run  at all is better than the status quo of them not existing

Zarel: Is there not an easy way to e.g. switch to `/bin/bash`?

scheibo: lmao you think bash is available but not sh?

Zarel: In hindsight yeah that's probably unlikely.

I wonder how they get away with not providing `sh`, though. I bet that breaks a lot of CI. I wouldn't be surprised if it's possible to enable somewhere in the GitHub Actions config.

Issue #890, opened 2017-03-24
Displaying information provided by Gen 7's in-game battle UI


MacChaeger:
In Sun & Moon, the battle interface provides the player with quite a bit more information than it did last gen, and has actually outdone PS's own client in that regard. Information provided to the player in-game that the client doesn't display includes (but is not necessarily limited to) the following.

- [ ] The effectiveness of attacking moves on active opponents is displayed to the player as `Super effective`, `Not very effective`, `Effective`, or `No effect` when they are selecting their moves. This information is entirely based on typing and does not account for items, abilities, or other effects except Soak, Protean, etc., and Illusion (but more on that later). If a move is doubly or triply super effective or resisted, they will appear the same as moves that are only singly so. For example, Earthquake appears as `Super Effective` against Tyranitar, Eelektross, and Magnezone. Moves that deal a fixed amount of damage, such as Night Shade, will not have any information supplied about their effectiveness, even against targets that are immune. In doubles, the move itself will display the highest effectiveness possible between the two opponents, but when selecting a target, individual effectiveness for each possible target is displayed.
- [ ] The actual type of Arceus and Silvally that have been hacked to appear as a different type than they really are is revealed to both players through the game's interface. However, a Pokémon disguised through Illusion will appear to be the actual type of the Pokémon it's disguised as, even so far as to not display type changes made through the likes of Soak.
- [ ] The number of turns remaining until Future Sight and Doom Desire hit are telegraphed to the players as a status effect on the target, similar to how Perish Song works now.

If there's any other information that the in-game UI displays that PS doesn't, tell me and I'll add it.

Zarel:
Yeah, I want to implement this eventually. I want to take abilities/items into account, though, which is putting this off.

Marty-D:
If you Transform into an opposing Pokemon, the Transform user's Ability will be displayed as the actual Ability of the target (the summary screen still shows your original Ability as in every other gen). Obviously, if it's not possible to check the battle screen before you get forced out or knocked out then you don't learn anything, so it would be an information leak if we updated the mouseover instantly.

MacChaeger:
@Marty-D I assume it's the same with Imposter?

MacChaeger:
For Doom Desire and Future Sight, would it be better to add a volatile with a duration à la Perish Song to the initial target of the move as well as any pokemon that enters that field position, or would it be better to use the same system as Reflect and other side conditions?

Zarel:
[...]

PR #1478, opened 2020-03-15
Add getType to ModdedDex and populate correctly

scheibo:
Required for #890 in past gens. Verified by patching in #1476 and adding to the Dex test (please merge it already...).

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -913,6 +913,39 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 				overrideAbilityDesc[id] = (pastEntry.shortDesc || pastEntry.desc);
 			}
 		}
+
+		const overrideTypeChart = {};
+		BattleTeambuilderTable[gen].overrideTypeChart = overrideTypeChart;
+		const removeType = {};
+		BattleTeambuilderTable[gen].removeType = removeType;
+		for (const id in nextGenData.TypeChart) {
+			const nextEntry = nextGenData.TypeChart[id];
+			const pastEntry = genData.TypeChart[id];
+			if (!pastEntry) {
+				removeType[id] = true;
+				continue;
+			}
+			if (Object.keys(pastEntry).length !== Object.keys(nextEntry).length) {
+				overrideTypeChart[id] = pastEntry;
+				continue;
+			}
+			let diff = false;
+			for (const key in pastEntry) {
+				if ((typeof pastEntry[key] === 'string' && pastEntry[key] !== nextEntry[key]) ||
+					Object.keys(pastEntry[key]).length !== Object.keys(nextEntry[key]).length) {
+					diff = true;
+					break;
+				}
+				for (const k in pastEntry[key]) {
+					if (nextEntry[key][k] !== pastEntry[key][k]) {
+						diff = true;
+						break;
+					}
+				}
+				if (diff) break;
+			}
+			if (diff) overrideTypeChart[id] = pastEntry;
+		}
 	}
 
 	buf += 'exports.BattleTeambuilderTable = ' + JSON.stringify(BattleTeambuilderTable) + ';\n\n';
--- a/src/battle-dex.ts
+++ b/src/battle-dex.ts
@@ -812,6 +812,7 @@ class ModdedDex {
 		Items: {} as any as {[k: string]: Item},
 		Abilities: {} as any as {[k: string]: Ability},
 		Templates: {} as any as {[k: string]: Template},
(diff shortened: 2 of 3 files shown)


Zarel:
Sad, you can't get GitHub to run the CI on the auto-rebased version. :(

scheibo:
> Sad, you can't get GitHub to run the CI on the auto-rebased version. :(

Doesn't really matter, until you accept #1476 there are no tests for this code anyway

scheibo:
```ts
	it('getAbility', () => {
		assert.equal(Dex.getAbility('Sturdy').shortDesc, 'If this Pokemon is at full HP, it survives one hit with at least 1 HP. Immune to OHKO.');
		assert.equal(Dex.forGen(3).getAbility('s turdy').shortDesc, 'OHKO moves fail when used against this Pokemon.');
	});
	it('getItem', () => {
		assert(Dex.getItem('Aerodactylite').megaEvolves, 'Aerodactyl');
	});
	it('getMove', () => {
		assert(Dex.getMove('Draco Meteor').basePower, 130);
		assert(Dex.forGen(4).getMove('DracoMeteor').basePower, 140);
		assert(Dex.getMove('Crunch').category, 'Physical');
		assert(Dex.forGen(2).getMove('CRUNCH').category, 'Special');
	});
	it('getTemplate', () => {
		assert.equal(Dex.getTemplate('Alakazam').baseStats.spd, 95);
		assert.equal(Dex.forGen(3).getTemplate('Alakazam').baseStats.spd, 85);
	});
	it('getType', () => {
		assert(Dex.getType('Fairy').exists);
		assert(!Dex.forGen(1).getType('steel').exists);
		assert.equal(Dex.forGen(1).getType('Psychic').damageTaken['Ghost'], 3);
		assert.equal(Dex.getType('Psychic').damageTaken['Ghost'], 1);
		assert.equal(Dex.getType('Fire').damageTaken['Water'], 1);
		assert.equal(Dex.getType('Water').damageTaken['Fire'], 2);
		assert.equal(Dex.getType('Ground').damageTaken['Electric'], 3);
	});
});
```

if you want to try locally

scheibo:
Yay, now we even have tests

build-tools/build-indexes
```diff
@@ -913,6 +913,39 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
+				}
+				for (const k in pastEntry[key]) {
+					if (nextEntry[key][k] !== pastEntry[key][k]) {
+						diff = true;
+						break;
+					}
+				}
+				if (diff) break;
+			}
+			if (diff) overrideTypeChart[id] = pastEntry;
```
Zarel: I guess this is a deep compare? I agree that it sucks JS doesn't have this built-in, but honestly I would use a hack for readability:

```suggestion
			if (JSON.stringify(nextEntry) !== JSON.stringify(pastEntry)) {
				overrideTypeChart[id] = pastEntry;
			}
```

Zarel: (This is in a build script and will be executed 8 times ever, so I'm prioritizing readability over perf here.)

scheibo: I agree perf isnt important here, and I wanted to do JSON.stringify myself, only it breaks if the keys are in different orders.

Zarel: Are the keys commonly in different orders but otherwise identical?

Zarel: That sounds like a bug in the data files rather than here.

scheibo: I'm not changing this code to introduce subtle bugs if someone lists `spd: 30, spe: 30` vs. `spe: 30, spd: 30`. I'm unable to package the client anyway so I'm happy to drop this change because I don't depend on its correctness anymore, but I don't want my name showing up in `git blame` when theres some stupid and easily preventable bug because you depended on key ordering.

Zarel: TypeChart doesn't contain stats...

I also think you're going backwards? My change preserves key ordering, your change goes out of your way to normalize it.

Zarel: Like, the only reason to prefer your version over mine is if you prioritize a smaller bundle size for `BattleTeambuilderTable` over preserving key ordering, but that's not what you're arguing for at all.

And the smaller bundle size only comes up if someone messes up key ordering in the data files because they are not currently bugged and never intentionally be.

Zarel: This conversation will come up in `git blame`, anyway, but if you _really_ care, I can merge and commit this separately.

scheibo: Youre right, but I dont care to fix this.

PR #1479, opened 2020-03-15
Remove getTemplate call from Pokemon constructor

scheibo:
- `tsc` presumably covers us within `src`
- per `#sim-dev` I checked `replay.js`, `replay-embed.js` and `client-battle.js`
	- the two replay files are short and I have high confidence that they are not relying on fields from `Template`
  - I searched for `'okemon'` and looked at the 142 usages in `client-battle.js` and none appear to be using `Template` fields. no strong guarantee here given nothing is typed, but this is an easy enough change to rollback

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -106,9 +106,6 @@ class Pokemon implements PokemonDetails, PokemonHealth {
 		this.side = side;
 		this.species = data.species;
 
-		// TODO: stop doing this
-		Object.assign(this, Dex.getTemplate(data.species));
-
 		this.details = data.details;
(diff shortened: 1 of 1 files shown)


Zarel:
You're right about "easily reversible", but I'm going to hold this off while I push a higher-priority refactor, so it's easier to tell what broke what.

This could even give you some time to rebase and fix that CI error. :p

scheibo:
The CI error is still your `nw` error...

Zarel:
Yes, and it's fixed by rebasing.

PR #1480, opened 2020-03-16
Introduce a generated client engine package

scheibo:
`build-package` has been added to generate a new self contained and generally useful TypeScript`@pokemon-showdown/client` package on demand. Some small changes have been made outside of just the build process to the code in question, either to facilitate generation or to make things more useful when packaged as a library.

- `Battle` and `Side` switch to using dependency injection, taking a 'provider' / factory method for creating their dependent classes so that a developer can hook into existing behavior and extend the functionality.
- `Battle`'s parameters have changed - most downstream uses are not going to be using a battle scene
- use of browser-only globals are wrapped
- `BattleSceneStub`'s interface has been corrected to match `BattleScene`, and `BattleScene` now extends `BattleSceneStub`. Being able to extend `BattleSceneStub` is useful for developers building on top of this new package as they are able to just fill in the methods they wish to handle
- `"types"` was added to the `tsconfig.json` as a workaround for `tsc` errors that occur if `@types/node` happens to be in `node_modules/`  (which is the case after you run `build package 0.0.1`)

The package can be tested by commenting out the existing front matter in the tests and uncommenting the code which imports from the package.

### DIFF
--- a/build-tools/build-package
+++ b/build-tools/build-package
@@ -1,0 +1,527 @@
+#!/usr/bin/env node
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const child_process = require('child_process');
+
+const shell = cmd => {
+	try {
+		child_process.execSync(
+			cmd, {
+				stdio: ['ignore', 'ignore', 'pipe'],
+				cwd: path.resolve(__dirname, '..'),
+			});
+	} catch (err) {
+		console.log("FAILED");
+		console.error(err.stderr.toString());
+		process.exit(1);
+	}
+};
+
+const PACKAGE = 'client';
+
+process.stdout.write(`Initializing package @pokemon-showdown/${PACKAGE}... `);
+
+// A version number is required (or inferred) because there needs to
+// be a publishable package at the end of all of this
+let version = process.argv[2];
+if (!version || version.match(/^[^\d]/)) {
+	try {
+		const current = require(`../package/package.json`).version;
+		const [major, minor, patch] = current.split('.');
+		if (version === 'major' || version === 'breaking') {
+			version = `${Number(major) + 1}.0.0`;
+		} else if (version === 'minor') {
+			version = `${major}.${Number(minor) + 1}.0`;
+		} else {
+			version = `${major}.${minor}.${Number(patch) + 1}`;
+		}
+	} catch (err) {
+		console.log("FAILED");
+		console.error(`Version required to create @pokemon-showdown/${PACKAGE} package`);
+		process.exit(1);
+	}
+}
+
+const ROOT = path.resolve(__dirname, '..');
+const DIR = path.resolve(ROOT, 'package');
+
+if (!fs.existsSync(DIR)) fs.mkdirSync(DIR);
+// Delete the existing package.json so that if this process fails it makes
+// it more difficult to accidentally publish the corrupted state
+if (fs.existsSync(path.resolve(DIR, 'package.json'))) {
+	fs.unlinkSync(path.resolve(DIR, 'package.json'));
+}
+
+const DATA = path.resolve(DIR, 'data');
+if (!fs.existsSync(DATA)) fs.mkdirSync(DATA);
+
+// Due to circular dependencies caused by the nature of how PS is architected around
+// globals the sanest way to create a usable package is to concatenate all of the
+// logic together. The data files are bulkier and would benefit from being downloaded
+// in parallel (or elided completely), so we break these out into their own data files
+// to let the consumer decide how/if they should be included.
+const cpdata = (name, type) => {
+	const o = `Battle${type}`;
+	const t = require(`../data/${Array.isArray(name) ? name[0] : name}`, 'utf8')[o];
+	fs.writeFileSync(path.resolve(DATA, `${Array.isArray(name) ? name[1] : name}.json`), JSON.stringify(t));
+};
+
+cpdata('abilities', 'Abilities');
+cpdata('aliases', 'Aliases');
+cpdata('items', 'Items');
+cpdata('moves', 'Movedex');
+cpdata(['pokedex', 'species'], 'Pokedex');
+cpdata('text', 'Text');
+cpdata(['typechart', 'types'], 'TypeChart');
+
+// BattleTeambuilderTable gets special treatment to remove the fields only relevant to PS's
+// teambuilder internals and preserve the more general fields required for ModdedDex to work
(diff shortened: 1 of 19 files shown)


build-tools/build-package
```diff
@@ -0,0 +1,527 @@
...
+fs.writeFileSync(path.resolve(DATA, 'formats-data.json'), JSON.stringify(table));
+fs.writeFileSync(path.resolve(DATA, 'learnsets.json'), JSON.stringify(learnsets));
+
+// TypeScript assumes each individual entry is its own type and doesn't attempt
+// to perform unification to end up with a more general type. As such, the JSON
+// data is cast to slightly simplified versions of the types lifted from the
+// server code in our main bundle. When then set up each of the constants on
+// window where some of the code looks for it.
+let index =
+`import * as abilitiesJSON from './data/abilities.json';
```
Zarel: This is a huge string; shouldn't it just get its own file?

src/battle.ts
```diff
@@ -994,8 +1001,11 @@ interface ServerPokemon extends PokemonDetails, PokemonHealth {
 	pokeball: string;
 }
 
+type SceneProvider = (b: Battle) => BattleSceneStub;
+type SideProvider = (b: Battle, n: number) => Side;
+
 class Battle {
-	scene: BattleScene | BattleSceneStub;
+	scene: BattleSceneStub;
```
Zarel: Actually do we even need a stub now that we have optional chaining and can just replace `scene.` with `scene?.`?

I guess having the stub is nice for other people using it, but maybe it can be an interface?

src/battle.ts
```diff
@@ -3180,8 +3201,8 @@ class Battle {
 			break;
 		}
 		case 'join': case 'j': case 'J': {
-			if (this.roomid) {
-				let room = app!.rooms[this.roomid];
+			if (this.roomid && window.app) {
```
Zarel: `this.roomid` should not exist if `window.app` doesn't, at least that's the theory behind the `!`. But I guess the extra safety couldn't hurt.

Zarel:
I don't have a good idea of dependency injection. Why are we passing functions instead of just objects? I can't see the advantage of doing it that way.

scheibo:
> I don't have a good idea of dependency injection. Why are we passing functions instead of just objects? I can't see the advantage of doing it that way.

Because the objects have circular dependencies on the Battle/Side. Side takes a battle in its constructor, so you cant pass in a newed up Side to Battle.

src/battle.ts
scheibo: you have no idea how external users will set `roomid` - your logic only holds in the PS UI

Zarel: There's no reason to set `roomid` except to communicate back with a PS room. I might as well have no idea that people will override random functions with functions that throw, and defend against that or something. There's no way to be secure against enemies with write access to your own code; why bother?

src/battle.ts
Zarel: @scheibo accidentally edited my post and wrote

> Actually do we even need a stub now that we have optional chaining and can just replace `scene.` with `scene?.`?

Perhaps not, but changing that is large in scope than this. I could do it in a follow up if thats the pattern you wanted to switch to

> I guess having the stub is nice for other people using it, but maybe it can be an interface?

I thought about making it an interface, and that may be the way to go if we switch to an interface (and have all of the methods options), but as it stands its basically an abstract base class type of pattern that means people only need to override the methods they want to

build-tools/build-package
scheibo: I thought about that, but its not really a valid ts file on its own. I dont really have a strong preference though, i liked it here because then everything is contained in one place and we dont add a 'package-prelude.ts` which we also add to the tsconfig exclude and have to read in here instead.

src/battle.ts
scheibo: LOL 'enemies'? are you kidding? Of course there are valid reasons to use the roomid field. obviously your general point that anything could be overriden is valid, but if thats only logical if theyre breaking the contract laid out by the class. Right now, the fields contract is "this only has meaning in  PS" which i disagree with, as thats entirely against the point of what I'm trying to do here with this package.

> But I guess the extra safety couldn't hurt.

lets just leave this as the reason then

src/battle.ts
scheibo: > Actually do we even need a stub now that we have optional chaining and can just replace `scene.` with `scene?.`?

Perhaps not, but changing that is large in scope than this. I could do it in a follow up if thats the pattern you wanted to switch to

> I guess having the stub is nice for other people using it, but maybe it can be an interface?

I thought about making it an interface, and that may be the way to go if we switch to an interface (and have all of the methods options), but as it stands its basically an abstract base class type of pattern that means people only need to override the methods they want to

Zarel:
I don't think Battle/Side/Pokemon need to be split up. I don't like the idea of needing to maintain the interfaces between them.

Scene I could see, I guess. Although I wouldn't call myself a fan of doing anything more than "render to web or not".

scheibo:
> I don't think Battle/Side/Pokemon need to be split up. I don't like the idea of needing to maintain the interfaces between them.
> 
> Scene I could see, I guess. Although I wouldn't call myself a fan of doing anything more than "render to web or not".

Unfortunately this is the full ball game. If I can't override Battle/Side/Pokemon I must fork. If thats the case, please close this PR and I'll just publish the package under @pkmn from my own repo.

Zarel:
You've linked to your repo but it doesn't appear to do anything? I think everything in that repo can be done with the regular `Battle` class?

I don't think you need to fork, I just need to understand what your use-case is before I can merge.

src/battle.ts
Zarel: > Of course there are valid reasons to use the roomid field.

Can you give an example of a valid reason to use the `roomid` field other than to interface with `app.rooms`?

> Right now, the fields contract is "this only has meaning in PS" which i disagree with

I don't understand how you can disagree with a field's contract? I went back and checked, and it literally does nothing except interface with `app.rooms`. Go make your own field if you want one with a different contract.

I don't understand how this is any different from typing `throw new Error()` and being mad when it throws an error.

> lets just leave [extra safety] as the reason then

Actually, no, that was before I took a second look at the code. I'm now actively confused. This isn't safety, it's insanity.

`battle.roomid` does nothing _except_ report joins/leaves back to `app.rooms[roomid]`. It could certainly be documented better, but whether or not it is, it seems insane to `console.log("Hello world")` and then complain when `"Hello world"` shows up in your console.

Throwing an error when you set `battle.roomid` without `app.rooms[roomid]` existing is what _should_ happen. You shouldn't silently suppress errors.

PR #1481, opened 2020-03-18
Make burn tooltips more accurate

KrisXV:


src/battle-tooltips.ts
```diff
@@ -1484,6 +1484,9 @@ class BattleTooltips {
 		if (['psn', 'tox'].includes(pokemon.status) && move.category === 'Physical') {
 			value.abilityModify(1.5, "Toxic Boost");
 		}
+		if (this.battle.gen > 2 && serverPokemon.status === 'brn' && move.id !== 'facade' && move.category === 'Physical') {
+			if (!value.tryAbility("Guts")) value.modify(0.5, 'Burn');
```
Zarel: The reason this isn't here is because burn modifies the Attack stat, not the move's Base Power.

Zarel: Is that wrong since Gen 2?

src/battle-tooltips.ts
```diff
@@ -1512,7 +1515,7 @@ class BattleTooltips {
 		if (move.category !== 'Status' && !noTypeOverride.includes(move.id)) {
 			if (move.type === 'Normal') {
 				value.abilityModify(this.battle.gen > 6 ? 1.2 : 1.3, "Aerilate");
-				value.abilityModify(this.battle.gen > 6 ? 1.2 : 1.3, "Galvanize");
+				value.abilityModify(1.2, "Galvanize");
```
Zarel: I guess Galvanize didn't exist in Gen 6?

Zarel: Honestly, why not match the others? Maybe it'll come in handy for OMs that want to backport abilities?

src/battle-tooltips.ts
Zarel: Apparently yes. Did not know. Oh well, I definitely massively prefer it this way, because it's much more obvious what's going on in the tooltip.

src/battle-tooltips.ts
KrisXV: I guess

src/battle-tooltips.ts
KrisXV: Except this is more accurate; Burn cuts the damage of Body Press/Foul Play, which isn't immediately obvious with the way it's currently implemented

DaWoblefet:
I was also hesitant about this Zarel, since I worry about the number of people who will report that burn isn't halving their attack or whatever more than burn's interaction with Body Press being wrong. There's no way to represent "this move halves damage" any more than we can represent a super effective attack is 2x powerful or rain is a 1.5x boost (since those are damage mods too), so I think base power is an _acceptable_ improvement.

Zarel:
Well, let's see how this goes!

PR #1482, opened 2020-03-19
Support /hidetext linecount client-side

tmagicturtle:
Reference: https://github.com/smogon/pokemon-showdown/pull/6471

This is a client-side patch that modifies hidetext messages to support an integer parameter of lines to hide. Default is still all lines, but now a supplied number of lines can be hidden as well.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1427,7 +1427,7 @@
 						// but it's now always applied
 						$messages = this.$chat.find('.chatmessage-' + user);
 						if (!$messages.length) break;
-						$messages.hide().addClass('revealed').find('button').parent().remove();
+						$messages.slice(-row[3]).hide().addClass('revealed').find('button').parent().remove();
 						this.$chat.children().last().append(' <button name="toggleMessages" value="' + user + '" class="subtle"><small>(' + $messages.length + ' line' + ($messages.length > 1 ? 's' : '') + ' from ' + user + ' hidden)</small></button>');
 					}
 					break;
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1427,7 +1427,7 @@
 						// but it's now always applied
 						$messages = this.$chat.find('.chatmessage-' + user);
 						if (!$messages.length) break;
-						if (row[3]) $messages = $messages.slice(-parseInt(row[3]);
+						if (row[3]) $messages = $messages.slice(-parseInt(row[3]));
 						$messages.hide().addClass('revealed').find('button').parent().remove();
 						this.$chat.children().last().append(' <button name="toggleMessages" value="' + user + '" class="subtle"><small>(' + $messages.length + ' line' + ($messages.length > 1 ? 's' : '') + ' from ' + user + ' hidden)</small></button>');
 					}
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1428,7 +1428,7 @@
 						$messages = this.$chat.find('.chatmessage-' + user);
 						if (!$messages.length) break;
 						$messages.slice(-row[3]).hide().addClass('revealed').find('button').parent().remove();
-						this.$chat.children().last().append(' <button name="toggleMessages" value="' + user + '" class="subtle"><small>(' + $messages.length + ' line' + ($messages.length > 1 ? 's' : '') + ' from ' + user + ' hidden)</small></button>');
+						this.$chat.children().last().append(' <button name="toggleMessages" value="' + user + '" class="subtle"><small>(' + $messages.slice(-row[3]).length + ' line' + ($messages.slice(-row[3]).length > 1 ? 's' : '') + ' from ' + user + ' hidden)</small></button>');
 					}
 					break;
 
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1427,8 +1427,9 @@
 						// but it's now always applied
 						$messages = this.$chat.find('.chatmessage-' + user);
 						if (!$messages.length) break;
-						$messages.slice(-row[3]).hide().addClass('revealed').find('button').parent().remove();
-						this.$chat.children().last().append(' <button name="toggleMessages" value="' + user + '" class="subtle"><small>(' + $messages.slice(-row[3]).length + ' line' + ($messages.slice(-row[3]).length > 1 ? 's' : '') + ' from ' + user + ' hidden)</small></button>');
+						if (row[3]) $messages = $messages.slice(-parseInt(row[3]);
+						$messages.hide().addClass('revealed').find('button').parent().remove();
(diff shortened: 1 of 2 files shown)


js/client-chat.js
```diff
@@ -1427,8 +1427,8 @@
 						// but it's now always applied
 						$messages = this.$chat.find('.chatmessage-' + user);
 						if (!$messages.length) break;
-						$messages.hide().addClass('revealed').find('button').parent().remove();
-						this.$chat.children().last().append(' <button name="toggleMessages" value="' + user + '" class="subtle"><small>(' + $messages.length + ' line' + ($messages.length > 1 ? 's' : '') + ' from ' + user + ' hidden)</small></button>');
+						$messages.slice(-row[3]).hide().addClass('revealed').find('button').parent().remove();
```
Zarel: `row[3]` is a string; please don't do implicit conversion here.

js/client-chat.js
```diff
@@ -1427,8 +1427,8 @@
...
+						this.$chat.children().last().append(' <button name="toggleMessages" value="' + user + '" class="subtle"><small>(' + $messages.slice(-row[3]).length + ' line' + ($messages.slice(-row[3]).length > 1 ? 's' : '') + ' from ' + user + ' hidden)</small></button>');
```
Zarel: You're using `$messages.slice(-row[3])` three times here, you should really cache the result instead of making three new jQuery objects.

```suggestion
						if (row[3]) $messages = $messages.slice(-parseInt(row[3]);
						$messages.hide().addClass('revealed').find('button').parent().remove();
						this.$chat.children().last().append(' <button name="toggleMessages" value="' + user + '" class="subtle"><small>(' + $messages.length + ' line' + ($messages.length > 1 ? 's' : '') + ' from ' + user + ' hidden)</small></button>');
```

Zarel:
Also this needs to be done to `src/battle-log.ts`.

tmagicturtle:
I'm struggling a lot with this one. It says the two lines I've changed are missing a radix parameter, but when I add them, it says expected 1 argument and got 2. The battle-log file is written very very differently and I'm not super familiar with TypeScript (and client builds are failing for me for one reason or another).

src/battle-log.ts
```diff
@@ -141,9 +141,9 @@ class BattleLog {
 
 		case 'unlink': {
 			const user = toID(args[2]) || toID(args[1]);
-			this.unlinkChatFrom(user);
+			this.unlinkChatFrom(user, parseInt(args[args.length], 10));
```
```suggestion
			this.unlinkChatFrom(user);
```

src/battle-log.ts
```diff
@@ -141,9 +141,9 @@ class BattleLog {
...
 			if (args[2]) {
-				this.hideChatFrom(user);
+				this.hideChatFrom(user, parseInt(args[args.length], 10));
```
```suggestion
				const lineCount = parseInt(args[args.length], 10);
				this.hideChatFrom(user, true, lineCount);
```

src/battle-log.ts
```diff
@@ -315,6 +315,7 @@ class BattleLog {
 				node.style.display = 'none';
 				node.className = 'revealed ' + node.className;
 				count++;
+				if (count >= lineCount && lineCount !== 0) break;
```
Zarel: This is going to remove the first `lineCount ` lines; we want the last `lineCount` lines.

src/battle-log.ts
Zarel: (You can't only use the third argument, you have to use the second and then the third.)

src/battle-log.ts
Zarel: Hiding a line count only matters for `|unlink|hide|`, so you don't have to deal with this if it's _not_ a hide instruction.

You didn't change the `unlinkChatFrom` function to take a line count in the first place.

src/battle-log.ts
Zarel: The line count is always `args[3]`, because this is `|unlink|hide|USERNAME|LINECOUNT`.

src/battle-log.ts
Zarel: You'll need to change the `for (const node of this.innerElem.childNodes as any) {` loop, but sadly it's too far up for me to suggest a change.

Zarel: This is probably best done in two loops – one to grab the matches, and one to modify the last `n` matches.

That's how your jQuery solution works, it's just that you have to do it manually here.

Zarel: ...I'll write it for you.

Zarel:
```ts
	hideChatFrom(userid: ID, showRevealButton = true, lineCount = 0) {
		const classStart = 'chat chatmessage-' + userid + ' ';
		let nodes = [];
		for (const node of this.innerElem.childNodes as any) {
			if (node.className && (node.className + ' ').startsWith(classStart)) {
				nodes.push(node);
			}
		}
		if (this.preemptElem) {
			for (const node of this.preemptElem.childNodes as any) {
				if (node.className && (node.className + ' ').startsWith(classStart)) {
					nodes.push(node);
				}
			}
		}
		if (lineCount) nodes = nodes.slice(-lineCount);

		for (const node of nodes) {
			if (node.className && (node.className + ' ').startsWith(classStart)) {
				node.style.display = 'none';
				node.className = 'revealed ' + node.className;
			}
		}
		if (!nodes.length || !showRevealButton) return;
		const button = document.createElement('button');
		button.name = 'toggleMessages';
		button.value = userid;
		button.className = 'subtle';
		button.innerHTML = `<small>(${nodes.length} line${nodes.length > 1 ? 's' : ''} from ${userid} hidden)</small>`;
		const lastNode = nodes[nodes.length - 1];
		lastNode.appendChild(document.createTextNode(' '));
		lastNode.appendChild(button);
	}
```

tmagicturtle: Thank you, apologies.

js/client-chat.js
```diff
@@ -1427,6 +1427,7 @@
 						// but it's now always applied
 						$messages = this.$chat.find('.chatmessage-' + user);
 						if (!$messages.length) break;
+						if (row[3]) $messages = $messages.slice(-parseInt(row[3], 10));
```
```suggestion
						var lineCount = parseInt(row[3], 10) || 0;
						if (lineCount) $messages = $messages.slice(-lineCount);
```

Might as well improve args readability here.

src/battle-log.ts
```diff
@@ -306,37 +307,40 @@ class BattleLog {
...
 			}
 		}
-		if (!count || !showRevealButton) return;
+		if (lineCount) nodes = nodes.slice(-lineCount);
+
+		for (const node of nodes) {
+			if (node.className && (node.className + ' ').startsWith(classStart)) {
+				node.style.display = 'none';
+				node.className = 'revealed ' + node.className;
+			}
```
```suggestion
			node.style.display = 'none';
			node.className = 'revealed ' + node.className;
```

Whoopsie, the `if` block is redundant here.

PR #1485, opened 2020-03-22
Stop htmlboxes from notifying

whalemer:
......whoops


Zarel:
What is this fixing? Were htmlboxes highlighting? Or subtle-notifying?

whalemer:
Giving the highlight notification, yes.

PR #1486, opened 2020-03-28
Fix missed instances of renaming Pokemon#species

KrisXV:
discovered in https://replay.pokemonshowdown.com/gen8challengecup1v1-1086619961-ba1cz8csx1k0n5ytxa5opldq0h6eys5pw

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -193,7 +193,7 @@
 					switch (args[0]) {
 					case 'trapped':
 						requestData.trapped = true;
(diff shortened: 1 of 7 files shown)


Zarel:
I already said no to `forme`. :|

Zarel:
Is the Species rename live? There's a critical data-loss bug in it. I guess I'll have to push my fix:

https://github.com/smogon/pokemon-showdown-client/commit/c9dd7fbe36b9a71e567d17d67f2e701aed782e8d

Zarel:
Anyway, if you can't get consensus on anything else, just make it `speciesForme`. It's not currently used so it can actually be find/replaced, unlike `forme`.

KrisXV:
done @Zarel

PR #1497, opened 2020-04-27
Teambuilder: implement exporting to PokePaste

AnnikaCodes:
Implemented as per [this suggestion](https://www.smogon.com/forums/threads/create-option-for-a-team-to-be-pokepasted-from-the-teambuilder.3663121/). I opted to use a `<form>` to make the request since pokepast.es' CORS setup prevents jQuery POST requests from PS! 
I spread the html out over multiple lines because I feel it is more readable; if you want me to revert that let me know.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1071,11 +1071,6 @@
 				buf = '<div class="pad"><button name="back"><i class="fa fa-chevron-left"></i> List</button> ';
 				buf += '<input class="textbox teamnameedit" type="text" class="teamnameedit" size="30" value="' + BattleLog.escapeHTML(this.curTeam.name) + '" /> ';
 				buf += '<button name="import"><i class="fa fa-upload"></i> Import/Export</button> ';
-				buf += '<form id="pokepasteForm" style="display:inline" method="post" action="https://pokepast.es/create" target="_blank">';
-				buf += '<input type="hidden" name="title" id="pasteTitle">';
-				buf += '<input type="hidden" name="paste" id="pasteData">';
-				buf += '<input type="hidden" name="author" id="pasteAuthor">';
-				buf += '<button name="pokepasteExport" type="submit"><i class="fa fa-upload"></i> Export to PokePaste</button></form></div>';
 				buf += '<div class="teamchartbox">';
 				buf += '<ol class="teamchart">';
 				buf += '<li>' + this.clipboardHTML() + '</li>';
@@ -1114,6 +1109,11 @@
 					buf += '<li><button name="addPokemon" class="button big"><i class="fa fa-plus"></i> Add Pok&eacute;mon</button></li>';
 				}
 				buf += '</ol>';
+				buf += '<form id="pokepasteForm" style="display:inline" method="post" action="https://pokepast.es/create" target="_blank">';
+				buf += '<input type="hidden" name="title" id="pasteTitle">';
+				buf += '<input type="hidden" name="paste" id="pasteData">';
+				buf += '<input type="hidden" name="author" id="pasteAuthor">';
+				buf += '<button name="pokepasteExport" type="submit" class="button exportbutton"><i class="fa fa-upload"></i> Upload to PokePaste</button></form>';
 				buf += '</div>';
 			}
 			this.$el.html('<div class="teamwrapper">' + buf + '</div>');
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -797,7 +797,6 @@
 			this.update();
 		},
 		pokepasteExport: function () {
-			console.log("HI");
 			document.getElementById("pasteData").value = BattleLog.escapeHTML(Storage.exportTeam(this.curSetList));
 			document.getElementById("pasteTitle").value = BattleLog.escapeHTML(this.curTeam.name);
 			document.getElementById("pasteAuthor").value = BattleLog.escapeHTML(app.user.get('name'));
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1062,7 +1062,14 @@
 				buf = '<div class="pad"><button name="back"><i class="fa fa-chevron-left"></i> List</button> <input class="textbox teamnameedit" type="text" class="teamnameedit" size="30" value="' + BattleLog.escapeHTML(this.curTeam.name) + '" /> <button name="saveImport"><i class="fa fa-upload"></i> Import/Export</button> <button name="saveImport" class="savebutton"><i class="fa fa-floppy-o"></i> Save</button></div>';
 				buf += '<div class="teamedit"><textarea class="textbox" rows="17">' + BattleLog.escapeHTML(Storage.exportTeam(this.curSetList)) + '</textarea></div>';
 			} else {
-				buf = '<div class="pad"><button name="back"><i class="fa fa-chevron-left"></i> List</button> <input class="textbox teamnameedit" type="text" class="teamnameedit" size="30" value="' + BattleLog.escapeHTML(this.curTeam.name) + '" /> <button name="import"><i class="fa fa-upload"></i> Import/Export</button></div>';
+				buf = '<div class="pad"><button name="back"><i class="fa fa-chevron-left"></i> List</button> ';
+				buf += '<input class="textbox teamnameedit" type="text" class="teamnameedit" size="30" value="' + BattleLog.escapeHTML(this.curTeam.name) + '" /> ';
+				buf += '<button name="import"><i class="fa fa-upload"></i> Import/Export</button> ';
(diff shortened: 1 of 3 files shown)


Zarel:
This is on the setlist? Can I get a screenshot?

js/client-teambuilder.js
```diff
@@ -1062,7 +1062,14 @@
...
 				buf += '<div class="teamedit"><textarea class="textbox" rows="17">' + BattleLog.escapeHTML(Storage.exportTeam(this.curSetList)) + '</textarea></div>';
 			} else {
-				buf = '<div class="pad"><button name="back"><i class="fa fa-chevron-left"></i> List</button> <input class="textbox teamnameedit" type="text" class="teamnameedit" size="30" value="' + BattleLog.escapeHTML(this.curTeam.name) + '" /> <button name="import"><i class="fa fa-upload"></i> Import/Export</button></div>';
+				buf = '<div class="pad"><button name="back"><i class="fa fa-chevron-left"></i> List</button> ';
+				buf += '<input class="textbox teamnameedit" type="text" class="teamnameedit" size="30" value="' + BattleLog.escapeHTML(this.curTeam.name) + '" /> ';
+				buf += '<button name="import"><i class="fa fa-upload"></i> Import/Export</button> ';
+				buf += '<form style="display:inline" method="post" action="https://pokepast.es/create" target="_blank">';
+				buf += '<button type="submit"><input type="hidden" name="paste" value="' + BattleLog.escapeHTML(Storage.exportTeam(this.curSetList)) + '">';
```
Zarel, line 1069: These input elements should definitely not be _inside_ the button.

AnnikaCodes:
looks like the checks are failing in an unrelated part of the code, which is odd, but it doesn't seem to be caused by my code?

js/client-teambuilder.js
AnnikaCodes: most definitely, just fixed that

AnnikaCodes:
And yes, it's next to the text-based single-team Import/Export button as seen at https://prnt.sc/s7hioi.

Zarel:
Yeah, the test failures were due to a TypeScript update; I fixed them in ba52aab0ef9ab2. The next time you make a change, they'll disappear.

js/client-teambuilder.js
```diff
@@ -1062,7 +1062,14 @@
...
+				buf = '<div class="pad"><button name="back"><i class="fa fa-chevron-left"></i> List</button> ';
+				buf += '<input class="textbox teamnameedit" type="text" class="teamnameedit" size="30" value="' + BattleLog.escapeHTML(this.curTeam.name) + '" /> ';
+				buf += '<button name="import"><i class="fa fa-upload"></i> Import/Export</button> ';
+				buf += '<form style="display:inline" method="post" action="https://pokepast.es/create" target="_blank">';
+				buf += '<input type="hidden" name="title" value="' + BattleLog.escapeHTML(this.curTeam.name) + '">';
+				buf += '<input type="hidden" name="author" value="' + BattleLog.escapeHTML(app.user.get('name')) + '">';
+				buf += '<button type="submit"><input type="hidden" name="paste" value="' + BattleLog.escapeHTML(Storage.exportTeam(this.curSetList)) + '">';
+				buf += '<i class="fa fa-upload"></i> Export to PokePaste</button></form></div>';
```
Zarel, line 1072:
```suggestion
				buf += '<input type="hidden" name="paste" value="' + BattleLog.escapeHTML(Storage.exportTeam(this.curSetList)) + '">';
				buf += '<button type="submit"><i class="fa fa-upload"></i> Export to PokePaste</button>';
				buf += '</form></div>';
```

You still have an input element inside the button.

I looked it up and apparently it's allowed, but I think it makes the button less readable not to all be on one line.

Zarel: I'm also pretty concerned by the `paste` property of the form. You're storing it here, which means it will only be updated by `updateTeamView()` – not all team edits call `updateTeamView()`, so this form could easily send an outdated team.

The easiest way to fix this is to leave these three inputs blank in the HTML, and only fill them out when the button is clicked and an `onSubmit` JavaScript handler is run.

js/client-teambuilder.js
```diff
@@ -1062,7 +1062,14 @@
...
 				buf += '<div class="teamedit"><textarea class="textbox" rows="17">' + BattleLog.escapeHTML(Storage.exportTeam(this.curSetList)) + '</textarea></div>';
 			} else {
-				buf = '<div class="pad"><button name="back"><i class="fa fa-chevron-left"></i> List</button> <input class="textbox teamnameedit" type="text" class="teamnameedit" size="30" value="' + BattleLog.escapeHTML(this.curTeam.name) + '" /> <button name="import"><i class="fa fa-upload"></i> Import/Export</button></div>';
+				buf = '<div class="pad"><button name="back"><i class="fa fa-chevron-left"></i> List</button> ';
+				buf += '<input class="textbox teamnameedit" type="text" class="teamnameedit" size="30" value="' + BattleLog.escapeHTML(this.curTeam.name) + '" /> ';
+				buf += '<button name="import"><i class="fa fa-upload"></i> Import/Export</button> ';
+				buf += '<form style="display:inline" method="post" action="https://pokepast.es/create" target="_blank">';
+				buf += '<input type="hidden" name="title" value="' + BattleLog.escapeHTML(this.curTeam.name) + '">';
+				buf += '<input type="hidden" name="paste" value="' + BattleLog.escapeHTML(Storage.exportTeam(this.curSetList)) + '">';
```
Zarel, line 1070: I'm also pretty concerned by the `paste` property of the form. You're storing it here, which means it will only be updated by `updateTeamView()` – not all team edits call `updateTeamView()`, so this form could easily send an outdated team.

The way I'd fix this is to leave these three inputs blank in the HTML, and only fill them out when the button is clicked and an `onSubmit` JavaScript handler is run.

AnnikaCodes: is something like the `pokepasteExport` function I wrote what you meant?

js/client-teambuilder.js
```diff
@@ -796,6 +796,13 @@
 			this.exportMode = true;
 			this.update();
 		},
+		pokepasteExport: function () {
+			console.log("HI");
+			document.getElementById("pasteData").value = BattleLog.escapeHTML(Storage.exportTeam(this.curSetList));
+			document.getElementById("pasteTitle").value = BattleLog.escapeHTML(this.curTeam.name);
+			document.getElementById("pasteAuthor").value = BattleLog.escapeHTML(app.user.get('name'));
```
Zarel, line 802: You don't need the `escapeHTML` because you're setting it directly rather than through HTML.

Zarel:
```suggestion
			document.getElementById("pasteData").value =Storage.exportTeam(this.curSetList);
			document.getElementById("pasteTitle").value =this.curTeam.name;
			document.getElementById("pasteAuthor").value =app.user.get('name');
```

js/client-teambuilder.js
```diff
@@ -796,6 +796,13 @@
 			this.exportMode = true;
 			this.update();
 		},
+		pokepasteExport: function () {
+			console.log("HI");
```
```suggestion
```

something something debug

js/client-teambuilder.js
Zarel: Yep! Looks good to me.

Zarel: I mean, assuming you've tested it and it works. I don't know how I feel about intercepting the submit button and manually submitting the form, but as long as it works, it's fine.

Zarel:
Okay, now I think the main thing is that the topbar is a bit busy now. I wonder if it might make more sense to appear only after clicking Import/Export.

js/client-teambuilder.js
AnnikaCodes: it does work, yes.

AnnikaCodes:
I can certainly make it appear after clicking Import/Export, but the presence of the Save button there causes the Export to PokePaste button to be moved onto a new line in left/right panel mode (for my device anyway). In my opinion, that is more cluttered, as seen below. However, if you want me to move it, I can commit that.
Skärmavbild 2020-04-28 kl  6 40 08 em

Zarel:
It could go below the team, as an "Upload to PokePaste" button.

Zarel:
How does that look now? I would imagine you would need to make the text box slightly less tall, so it would fit.

AnnikaCodes:
> How does that look now? I would imagine you would need to make the text box slightly less tall, so it would fit.
Oh, I thought you meant below the team before Import/Export was clicked.
image

Zarel:
Oh, that works, too.

scheibo:
Thank you @AnnikaCodes for implementing this!

AnnikaCodes:
> Thank you @AnnikaCodes for implementing this!

no problem!

PR #1498, opened 2020-04-27
Implement a /news command

mia-pi-git:
from https://www.smogon.com/forums/threads/news-command.3653715/

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -561,7 +561,8 @@
 				return false;
 
 			case 'news':
-				var newsId = Number(Storage.prefs('newsid'));
+				var newsId = '1990';
+				if (newsId === '' + Dex.prefs('readnews')) return;
 				app.rooms[''].addPseudoPM({
 					title: 'News',
 					html: '<iframe src="/news-embed.php?news' + (window.nodewebkit || document.location.protocol === 'https:' ? '&amp;https' : '') + '" width="270" height="400" border="0" style="border:0;width:100%;height:100%;display:block"></iframe>',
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -560,6 +560,16 @@
 				}
 				return false;
 
+			case 'news':
+				var newsId = '1990';
+				app.rooms[''].addPseudoPM({
+					title: 'News',
+					html: '<iframe src="/news-embed.php?news' + (window.nodewebkit || document.location.protocol === 'https:' ? '&amp;https' : '') + '" width="270" height="400" border="0" style="border:0;width:100%;height:100%;display:block"></iframe>',
```
Zarel, line 567: How is News currently implemented? We should probably piggy-back off the existing implementation, rather than creating a new imitation one.

mia-pi-git: news is currently implemented in mainmenu as 
`		addNews: function () {
			var newsId = '1990';
			if (newsId === '' + Dex.prefs('readnews')) return;
			this.addPseudoPM({
				title: 'Latest News',
				html: '<iframe src="/news-embed.php?news' + (window.nodewebkit || document.location.protocol === 'https:' ? '&amp;https' : '') + '" width="270" height="400" border="0" style="border:0;width:100%;height:100%;display:block"></iframe>',
				attributes: 'data-newsid="' + newsId + '"',
				cssClass: 'news-embed',
				height: 400
			});
		},`

mia-pi-git: and added within the loading process

mia-pi-git: Actually, i can just call it directly it seems.

PR #1500, opened 2020-04-29
Update evo/prevo/etc. to match their server-side types

KrisXV:


src/battle-dex.ts
```diff
@@ -484,7 +484,7 @@ const Dex = new class implements ModdedDex {
 			window.BattlePokedex[id] = species;
 		}
 
-		if (formid === id || !species.cosmeticFormes || !species.cosmeticFormes.includes(formid)) {
+		if (formid === id || !species.cosmeticFormes || !species.cosmeticFormes.map(toID).includes(formid)) {
```
Zarel, line 487: Can you just do it the way I implemented it on server? That way's probably better.

PR #1503, opened 2020-05-05
Use DexSearch for the teambuilder search engine

Zarel:
This will prevent us from needing to maintain two search engines in the future.

The new DexSearch doesn't have too much in the way of new features so far, but it does show filtered things in categories, and list illegal results in-filter even after sorting.

These are really minor differences, though; the main thing is just that its architecture should make it much easier to maintain, and in particular, to add more filters.

### DIFF
--- a/index.template.html
+++ b/index.template.html
@@ -151,6 +151,7 @@ ga('send', 'pageview');
 
 <script src="//play.pokemonshowdown.com/data/search-index.js?"></script>
 <script src="//play.pokemonshowdown.com/data/teambuilder-tables.js?"></script>
+<script src="//play.pokemonshowdown.com/js/battle-dex-search.js?"></script>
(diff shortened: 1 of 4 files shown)


Zarel:
BEFORE:

image

AFTER:

image

Zarel:
Sorting a filtered result:

BEFORE:

image

AFTER:

image

scheibo:
It seems like this would be useful to repurpose for the formats search dropdown. eg.`[OU], [Singles]` or `[VGC]`. Preferably coupled with buttons to quick apply certain popular filters on mobile.

Zarel:
We have a formats search dropdown?

Zarel:
I'm not confident there'll be no bugs, but whatever, we can fix them live.

PR #1504, opened 2020-05-05
Fix DPP Doubles teambuilder

KrisXV:
Fixes https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8455568


Zarel:
Can you update the new search engine (`src/battle-dex-search.js`)?

PR #1505, opened 2020-05-07
Fix various bugs with the new teambuilder

KrisXV:


src/battle-dex-search.ts
```diff
@@ -542,8 +542,7 @@ abstract class BattleTypedSearch<T extends SearchType> {
 	 */
 	set: PokemonSet | null = null;
 
-	protected isDoubles = false;
-	protected isLetsGo = false;
+	protected mod: 'natdex' | 'letsgo' | 'doubles' | null = null;
```
Zarel, line 545: I don't like `mod` as a name for this because it already means something else.

Zarel: `formatType`?

src/battle-dex-search.ts
```diff
@@ -520,7 +520,7 @@ class DexSearch {
 abstract class BattleTypedSearch<T extends SearchType> {
 	searchType: T;
 	/**
-	 * Dex for the mod/generation to search.
+	 * Dex for the formatType/generation to search.
```
Zarel, line 523:
```suggestion
	 * Dex for the mod/generation to search.
```

This one actually means "mod", which is why I didn't want you to do the same name as that. :p

KrisXV: not sure how this still got changed.... i very intentionally skipped this

PR #1506, opened 2020-05-07
Add room categories

KrisXV:


Zarel:
As I mentioned in PM, I've resisted this every time because it gives too much disadvantage to rooms in lower categories.

The only design that I've come up with so far without that problem is to keep the default list sorted by usercount, but also allow you to click category buttons to filter the rooms.

scheibo:
> As I mentioned in PM, I've resisted this every time because it gives too much disadvantage to rooms in lower categories.

Does it make sense to instead start with categories (except general i guess) in some sort of collapsed state ("Competitive", "Social"... what other types of categories are there lol?) so that all of the categories fit on the users screen and then they can decide what type of room theyre interested in? It seems like people would usually be in the mood for certain types of rooms?

KrisXV:
I'm taking this out of a draft state since the code already works fully, there just needs to be discussion on the category names/how we decide to sort them.

Zarel:
> Does it make sense to instead start with categories (except general i guess) in some sort of collapsed state ("Competitive", "Social"... what other types of categories are there lol?) so that all of the categories fit on the users screen and then they can decide what type of room theyre interested in? It seems like people would usually be in the mood for certain types of rooms?

That seems like my idea but without the usercount-sorted list under it, which I don't think is an improvement?

KrisXV:
@Zarel I don't want this to be a filter-only kind of thing because then it somewhat defeats the purpose of room categories existing.

TheImmortal:
> As I mentioned in PM, I've resisted this every time because it gives too much disadvantage to rooms in lower categories..

What if it were collapsible? Then no category would be at a disadvantage as all would be immediately visible and accessible.

scheibo:
> That seems like my idea but without the usercount-sorted list under it, which I don't think is an improvement?

I don't believe its similar to your idea. There are no filters, the categories are just collapsed.

> What if it were collapsible? Then no category would be at a disadvantage as all would be immediately visible and accessible.

^ This is the idea I was advocating for.

Zarel:
> I don't believe its similar to your idea. There are no filters, the categories are just collapsed.

What's the difference between a filter button and a collapsed category?

scheibo:
Zodiax brought this up with me in PMs so I actually went back and looked at this - I think I was partially misunderstanding Zarel's design earlier (and still might be). Time for some shitty sketches:



I think the design most of us are looking for when we talk about room categories is that the room list is replaced entirely such that it no longer shows rooms at all but instead shows top level categories which can be expanded to reveal rooms of a specific type (in the above sketch, the "Language" category is expanded to reveal "Francais", "Espanol", etc).

What I *believe* Zarel's argument was is that you could just add filter buttons on top of the existing activity sorted room list that could be used to filter the list by category.



Once again, the "Language" category is selected and the list is then filtered down just to language rooms.

I think when this was last discussed I was misunderstanding the filters to work more like the filters in the teambuilder/PS dex as opposed to buttons (though Zarel very clearly wrote button, so if there was a  misunderstanding it was clearly on my side).

I think both designs are equally effective for solving the "I want to see rooms of type X" problem, though only the former solves the "clutter" problem (adding filter buttons to an already overwhelming massive list of rooms exacerbates the clutter issue in the latter design). With respect to the "I've resisted this every time because it gives too much disadvantage to rooms in lower categories." argument (which isn't invalid!), I would posit that the list of collapsed categories would be sufficiently small to be able to fit on any display size such that there's not really any 'above the fold' issues and rooms in a category sorted higher to the top than another would not be unfairly advantaged in absolute terms.

PR #1507, opened 2020-05-07
Fix Pokemon that inherit movesets from base formes

KrisXV:


src/battle-dex-search.ts
```diff
@@ -682,7 +682,10 @@ abstract class BattleTypedSearch<T extends SearchType> {
 		if (!species) return '' as ID;
 		if (species.prevo) return toID(species.prevo);
 		let baseSpecies = species.baseSpecies;
-		if (baseSpecies !== species.name && (baseSpecies === 'Rotom' || baseSpecies === 'Pumpkaboo')) {
+		if (species.changesFrom) {
+			return toID(species.changesFrom);
+		}
+		if (baseSpecies !== species.name && baseSpecies === 'Pumpkaboo') {
```
Zarel, line 688: Why does this still need to be hardcoded?

Zarel: Let me guess: A Pumpkaboo event for a specific size.

Hah! I was right.

Zarel:
```suggestion
		if (baseSpecies !== species.name && baseSpecies === 'Pumpkaboo') {
			// Pumpkaboo-Super event
```

Zarel: I would probably add an `inheritFrom` flag in `learnsets` instead of hardcoding, but if you want the easy way out I'll accept the comment.

KrisXV: id rather just have the comment for now yea

PR #1509, opened 2020-05-08
Fix sorting move because of header

Raymondsquared:
Fix sorting move because of header.

the results to be sorted includes headers such as
```
["header", "Moves"]
["header", "Usually useless moves"]
```

and it caused the sorting function when selecting pokemon's moves to not function at all.
Screen Shot 2020-05-08 at 3 46 58 pm


Zarel:
Thanks for helping out! Unfortunately, this removes headers even when not sorting, and doesn't fix the crash when filters and sorts are combined.

As this is a high-priority crash, I've pushed my own fix.

PR #1511, opened 2020-05-08
Update "not useless" moves

Zarel:
Viable moves no longer use the `isViable` flag. Instead, we start by assuming moves with 75+ base power are not useless (and others are useless), and manually list exceptions in the relevant dexsearch function.

In general, the move viability system has been a bit too specific. I don't want to hardcode species because it's nearly always also hardcoding specific metagames. It may help to think of it as "would this move be useful in any metagame ever?" rather than "would it be useful in the current metagame".

We instead use a significantly lower bar of "not useless", although this hasn't been made clear until now. The word "viable" is now banished from the codebase as well as the teambuilder's actual text, which may also help.

Select changes:

- doubles moves now have separate viability lists
- Rock Tomb
  - it's good on Groudon for reasons that don't seem Groudon-specific, and is also good with Technician and in Doubles, so we'll default to yes
- Belly Drum
  - viable with priority, high speed, or Trick Room: too many
    conditions, so we'll default to yes
- Reflect Type
  - viable on Latias and Starmie for reasons that aren't specific to them, we'll default to yes
- Transform
  - required on Ditto, useful in Doubles: default to yes
- Fake Out
  - no longer viable if you have Sheer Force, should hopefully make people choose it less
- Nature's Madness, Super Fang
  - viable in Doubles, has uses in Singles: default to yes
- Dragon Darts, Sparkling Aria, Drill Peck
  - mistakenly categorized as unviable; fixed

### DIFF
--- a/src/battle-dex-data.ts
+++ b/src/battle-dex-data.ts
@@ -1088,7 +1088,6 @@ class Move implements Effect {
 
 	readonly desc: string;
 	readonly shortDesc: string;
-	readonly isViable: boolean;
 	readonly isNonstandard: string | null;
 	readonly isZ: ID;
 	readonly zMove?: {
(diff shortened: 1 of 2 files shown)


Zarel:
Weird, I expected more controversy. Oh well, if no one's against these changes, I'll just push them.

PR #1512, opened 2020-05-09
Fix `canLearn`'s filtering

KrisXV:
This makes it that looking up moves now correctly filters in Battle Spot/Battle Stadium formats, and searching for moves that were dexited in Gen 8 will no longer work outside of National Dex formats.

### DIFF
--- a/src/battle-dex-search.ts
+++ b/src/battle-dex-search.ts
@@ -697,6 +697,9 @@ abstract class BattleTypedSearch<T extends SearchType> {
 		return '' as ID;
 	}
 	protected canLearn(speciesid: ID, moveid: ID) {
+		if (this.dex.gen >= 8 && this.dex.getMove(moveid).isNonstandard === 'Past' && this.formatType !== 'natdex') {
+			return false;
+		}
 		let genChar = `${this.dex.gen}`;
 		if (
 			this.format.startsWith('vgc') ||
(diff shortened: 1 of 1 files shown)


src/battle-dex-search.ts
```diff
@@ -711,6 +715,9 @@ abstract class BattleTypedSearch<T extends SearchType> {
 		while (learnsetid) {
 			let learnset = BattleTeambuilderTable.learnsets[learnsetid];
 			if (learnset && (moveid in learnset) && learnset[moveid].includes(genChar)) {
+				if (this.dex.gen >= 8 && this.dex.getMove(moveid).isNonstandard === 'Past' && this.formatType !== 'natdex') {
+					return false;
+				}
```
Zarel, line 720: Why is a nonstandard move in the Pokémon's learnset? That seems like a bug that should be fixed in the learnset rather than here.

Zarel: Oh! Nevermind, I figured it out.

Zarel: There's no reason this should be this deep into a `while` loop. Put it at the top of `canLearn`, don't repeat it for every prevo and alternate forme.

KrisXV:
done @Zarel

Zarel:
Shit, wait, that was inside the `if` block, it would only have run once either way.

I'm actually not clear on which one would be faster...

PR #1513, opened 2020-05-10
fix css visual bug on mobile

WallyTheBully:
.tablist .button "width" isn't good on mobile 
how it currently looks: http://prntscr.com/sdygem
how it should look: http://prntscr.com/sdygnr


Zarel:
I should probably investigate what this was originally intended to do and why it broke, but this workaround seems to work so I'll just merge now and figure it out later.

PR #1514, opened 2020-05-10
Fix Metronome Battle teambuilder

KrisXV:


KrisXV:
Combined with https://github.com/smogon/pokemon-showdown-client/pull/1515

PR #1516, opened 2020-05-12
Fix Gmax Pokemon learnsets in the teambuilder

KrisXV:


src/battle-dex-data.ts
```diff
@@ -1370,7 +1370,7 @@ class Species implements Effect {
 		this.battleOnly = data.battleOnly || (this.isGigantamax ? this.baseSpecies : undefined);
 		this.isNonstandard = data.isNonstandard || null;
 		this.unreleasedHidden = data.unreleasedHidden || false;
-		this.changesFrom = data.changesFrom || undefined;
+		this.changesFrom = data.changesFrom || (this.isGigantamax ? this.baseSpecies : undefined);
```
Zarel, line 1373: The `battleOnly` check should cover it; it 100% does not need to be repeated here.

KrisXV: It doesn't.

Zarel: Then it should be fixed in the implementation, not here.

Zarel: Like you didn't update the documentation in `FORMES.md`, nor the jsdoc in `dex-data.ts`. Those are things you _must_ do if you want to change what a property means.

But at some point you should probably just realize that you don't actually want to change what the property means, but just how it's used.

PR #1518, opened 2020-05-13
Report replay hash mismatch properly

whalemer:
see also: https://github.com/smogon/pokemon-showdown-client/blob/a76be69b76d1e11c3d503ce1ff60fab5aaa88b3b/js/client.js#L1286-L1287

### DIFF
--- a/replays/replays.lib.php
+++ b/replays/replays.lib.php
@@ -250,8 +250,7 @@ class Replays {
 
 				// Someone else tried to upload a replay of the same battle,
 				// while we were uploading this
-				// ...pretend it was a success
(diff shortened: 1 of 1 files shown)


Zarel:
Why? We changed to a `success` because there's no real need to report it as a failure. The link should work fine no matter who uploads it...

whalemer:
It's occasionally relevant if you're corrupting the log somehow, and we have the message for it, so I figured that not using it was a mistake. Your call if you want to keep the current behaviour, though.

Zarel:
I'll keep the current behavior.

PR #1520, opened 2020-05-20
Add msg button

PartMan7:
(Will be accompanied by a server PR)

PRed server change.


PartMan7:
Client part for https://github.com/smogon/pokemon-showdown/pull/6724

PartMan7:
And this is why I should probably start from scratch when I realize I'm using an old chat.ts

PR #1521, opened 2020-05-21
Increase autojoin limit to 15

tmagicturtle:
With https://github.com/smogon/pokemon-showdown/pull/6728


js/client.js
```diff
@@ -1853,7 +1853,7 @@ function toId() {
 				autojoins.push(room.id.indexOf('-') >= 0 ? room.id : (room.title || room.id));
 				if (room.id === 'staff' || room.id === 'upperstaff' || (Config.server.id !== 'showdown' && room.id === 'lobby')) continue;
 				autojoinCount++;
-				if (autojoinCount >= 10) break;
+				if (autojoinCount >= 15) break;
```
Zarel, line 1856: Oh, this is for all servers? We'd probably need to increase it for only Main, notify protocol updates, and increase it for side servers after a month or so.

Zarel:
"22 days" is totally close enough to "a month or so", and people have been asking about this, so I guess here it is.

PR #1522, opened 2020-05-23
Added css styling to override new default input range styling

StephenArg:


Zarel:
Hi, I haven't forgotten about this and will get around to it soon. Just posting an update so you know.

StephenArg:
Thanks, I appreciate that!

style/client.css
```diff
@@ -3092,7 +3092,38 @@ a.ilink.yours {
 	height: 16px;
 	font-size: 9pt;
 }
-
+.selector:not(*:root), input[type=range]{
```
Zarel, line 3095: I can't tell what `.selector` is doing here. We don't seem to have any element with `class="selector"`?

StephenArg: I noticed my changes were breaking on Firefox (and potentially other browsers). This snippet makes it only work on Chrome. I'd be lying if I said I understand what it's doing exactly, I found it on stackoverflow or something. It works though.

Zarel: Hm, I'll try to address Firefox separately.

Zarel: Thanks for explaining it to me! I ended up going with a more readable workaround for Firefox: df710473067d3856fa4

PR #1527, opened 2020-06-12
Fix display of Pokémon level in details form

urkerab:
`escapeHTML` doesn't work on numbers, so I ported the Happiness code instead.


Zarel:
`escapeHTML` should probably be changed to do it the same way React does.

(`null`/`undefined` to `""`, numbers to strings)

Issue #1526, opened 2020-06-12
Setting a Pokemon's level in generation 1 and hitting enter disconnects the client


davidstone:
Steps to reproduce:

* Create a generation 1 team
* Add a Pokemon
* Select its Species
* Click on "Level"
* Enter a valid number into the level field
* Hit enter

This occurs for me on Firefox on Linux

PartMan7:
Can confirm, though the issue isn't limited to Gen1.
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8466937 A bit more on it.

PR #1528, opened 2020-06-12
Prevent Enter from crashing during Pokémon level entry

urkerab:
This fixes issue #1526 for me.

(You might not need both calls but I just copied it from another place in the code to be sure.)


Zarel:
It's plausible we might want Enter to work like Tab here?

PR #1534, opened 2020-06-23
Fix RBY accuracy/evasion tooltips

Enigami:
In RBY, Accuracy and Evasion use the same boost table as the other stats. For example, -1 Accuracy or Evasion = 0.66 instead of 0.75. This is implemented on Showdown, but it is incorrectly displayed using the Accuracy/Evasion table from Stadium and GSC onwards.

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -370,7 +370,8 @@ class Pokemon implements PokemonDetails, PokemonHealth {
 		}
 		if (this.boosts[boostStat] > 6) this.boosts[boostStat] = 6;
 		if (this.boosts[boostStat] < -6) this.boosts[boostStat] = -6;
-		if ((this.side.battle.tier.includes('Stadium') || this.side.battle.gen > 2) && (boostStat === 'accuracy' || boostStat === 'evasion')) {
+		const isRBY = this.side.battle.gen <= 1 && !this.side.battle.tier.includes('Stadium');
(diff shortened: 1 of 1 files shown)


src/battle.ts
```diff
@@ -370,7 +370,7 @@ class Pokemon implements PokemonDetails, PokemonHealth {
 		}
 		if (this.boosts[boostStat] > 6) this.boosts[boostStat] = 6;
 		if (this.boosts[boostStat] < -6) this.boosts[boostStat] = -6;
-		if (boostStat === 'accuracy' || boostStat === 'evasion') {
+		if ((this.side.battle.tier.includes('Stadium') || this.side.battle.gen > 2) && (boostStat === 'accuracy' || boostStat === 'evasion')) {
```
Zarel, line 373: Did you use `gen > 2` where you meant `gen > 1`?

Either way, this is a situation where it's a lot easier not to mess up if you have an intermediate variable:

```suggestion
		const isRBY = this.side.battle.gen <= 1 && !this.side.battle.tier.includes('Stadium');
		if (!isRBY && (boostStat === 'accuracy' || boostStat === 'evasion')) {
```

Enigami: Oops, I'm not sure how I missed that. Good catch.

Zarel:
Oh, it looks like I didn't notice when you fixed it.

Thanks!

PR #1535, opened 2020-06-25
Don't require desktop notifications for disconnections

urkerab:
Requested [here](https://www.smogon.com/forums/threads/3659047/).

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -509,10 +509,7 @@ function toId() {
 
 				if (showNotification !== false && (self.popups.length || !self.focused) && window.Notification) {
 					self.rooms[''].requestNotifications();
-					var disconnect = new Notification("Disconnected!", {lang: 'en', body: "You have been disconnected from Pokémon Showdown."});
-					disconnect.onclick = function (e) {
-						window.focus();
-					};
+					self.rooms[''].notifyOnce("Disconnected", "You have been disconnected from Pokémon Showdown.", 'disconnected');
 				}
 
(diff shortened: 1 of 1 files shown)


Zarel:
Wouldn't this give two notifications if you have notifications enabled? That seems probably bad...

Zarel:
I thought the `ReconnectPopup` was clear enough, anyway. If a user isn't seeing it, I'm not sure if this will help at all.

Zarel:
The right cure, I think, probably involves writing our own heartbeat system since SockJS's doesn't handle our use-case?

urkerab:
The user isn't seeing it because they don't have the PS! tab selected, and the tab's icon doesn't change for a disconnect.

But you're right about the second notification, so I've removed the duplicate.

Zarel:
The only remaining problem with this is that the Main Menu would glow orange even after you'd dismissed the "disconnected" popup, giving the false impression that you should open the Main Menu and there will be something in there to look at.

I'm not sure if that's something worth fixing, though? I'll let that be your call.

urkerab:
It's useful in case you accidentally dismiss the reconnect popup? It reminds you that you're disconnected and where you can find the Reconnect button.

Zarel:
The main thing is that when you actually click on the big orange "Home" button, you're greeted with nothing. No message about disconnection or anything like that.

I guess the "search battle" button says "Disconnected" with "Reconnect" under it. I suppose that's enough.

PR #1538, opened 2020-06-28
Clean up join/leave messages

AnnikaCodes:
Implements [this suggestion](https://www.smogon.com/forums/threads/add-left-after-joined.3666495/#post-8518430).

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1548,7 +1548,16 @@
 				this.$chat.append('<div class="message"><small>Loading...</small></div>');
 				this.$joinLeave = this.$chat.children().last();
 			}
-			this.joinLeave[action].push(user.group + user.name);
+
+			var formattedUser = user.group + user.name;
+			if (action === 'join' && this.joinLeave['leave'].includes(formattedUser)) {
+				this.joinLeave['leave'].splice(this.joinLeave['leave'].indexOf(formattedUser), 1);
+			} else if (action === 'leave' && this.joinLeave['join'].includes(formattedUser)) {
+				this.joinLeave['join'].splice(this.joinLeave['join'].indexOf(formattedUser), 1);
+			} else {
+				this.joinLeave[action].push(formattedUser);
+			}
+
 			var message = '';
 			if (this.joinLeave['join'].length) {
 				message += this.displayJoinLeaves(this.joinLeave['join'], 'joined');
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1552,8 +1552,6 @@
 			var formattedUser = user.group + user.name;
 			if (action === 'join' && this.joinLeave['leave'].includes(formattedUser)) {
 				this.joinLeave['leave'].splice(this.joinLeave['leave'].indexOf(formattedUser), 1);
-			} else if (action === 'leave' && this.joinLeave['join'].includes(formattedUser)) {
-				this.joinLeave['join'].splice(this.joinLeave['join'].indexOf(formattedUser), 1);
 			} else {
 				this.joinLeave[action].push(formattedUser);
 			}
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -120,13 +120,13 @@ class BattleLog {
 
 			let buf = '';
 			if (this.joinLeave.joins.length) {
-				buf += this.displayJoinLeaves(this.joinLeave.joins, 'joined');
+				buf += `${this.textList(this.joinLeave.joins)} joined`;
 			}
 			if (this.joinLeave.leaves.length) {
-				if (this.joinLeave.joins.length) buf += '; ';
-				buf += this.displayJoinLeaves(this.joinLeave.leaves, 'left') + '<br />';
+				if (this.joinLeave.joins.length) buf += `; `;
+				buf += `${this.textList(this.joinLeave.leaves)} left`;
 			}
-			this.joinLeave.element.innerHTML = `<small>${buf}</small>`;
+			this.joinLeave.element.innerHTML = `<small>${BattleLog.escapeHTML(buf)}</small>`;
 			(preempt ? this.preemptElem : this.innerElem).append(this.joinLeave.element);
 			return;
 		}
@@ -250,29 +250,30 @@ class BattleLog {
 			break;
 		}
 	}
-	displayJoinLeaves(preList: string[], action: string) {
+	textList(list: string[]) {
 		let message = '';
-		const list: string[] = [];
-		for (const user of preList) {
-			if (!list.includes(user)) list.push(user);
+		const listNoDuplicates: string[] = [];
+		for (const user of list) {
+			if (!listNoDuplicates.includes(user)) listNoDuplicates.push(user);
 		}
-		for (let j = 0; j < list.length; j++) {
-			if (j >= 5) {
+		list = listNoDuplicates;
+		for (let i = 0; i < list.length; i++) {
+			if (i >= 5) {
 				message += `, and ${list.length - 5} others`;
 				break;
 			}
-			if (j > 0) {
-				if (j === 1 && list.length === 2) {
+			if (i > 0) {
+				if (i === 1 && list.length === 2) {
 					message += ' and ';
-				} else if (j === list.length - 1) {
+				} else if (i === list.length - 1) {
 					message += ', and ';
 				} else {
 					message += ', ';
 				}
 			}
-			message += BattleLog.escapeHTML(list[j]);
+			message += list[i];
 		}
-		return `${message} ${action}`;
+		return message;
 	}
 	/**
 	 * To avoid trolling with nicknames, we can't just run this through
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -21,6 +21,11 @@ class BattleLog {
 	atBottom = true;
 	className: string;
 	battleParser: BattleTextParser | null = null;
+	joinLeave: {
+		joins: string[],
+		leaves: string[],
+		element: HTMLDivElement,
+	} | null = null;
 	/**
 	 * * -1 = spectator: "Red sent out Pikachu!" "Blue's Eevee used Tackle!"
 	 * * 0 = player 1: "Go! Pikachu!" "The opposing Eevee used Tackle!"
@@ -67,6 +72,7 @@ class BattleLog {
 		let divClass = 'chat';
 		let divHTML = '';
 		let noNotify: boolean | undefined;
+		if (!['join', 'j', 'leave', 'l'].includes(args[0])) this.joinLeave = null;
 		switch (args[0]) {
 		case 'chat': case 'c': case 'c:':
 			let battle = this.scene?.battle;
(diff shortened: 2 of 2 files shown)


js/client-chat.js
```diff
@@ -1548,7 +1548,16 @@
 				this.$chat.append('<div class="message"><small>Loading...</small></div>');
 				this.$joinLeave = this.$chat.children().last();
 			}
-			this.joinLeave[action].push(user.group + user.name);
+
+			var formattedUser = user.group + user.name;
+			if (action === 'join' && this.joinLeave['leave'].includes(formattedUser)) {
+				this.joinLeave['leave'].splice(this.joinLeave['leave'].indexOf(formattedUser), 1);
+			} else if (action === 'leave' && this.joinLeave['join'].includes(formattedUser)) {
+				this.joinLeave['join'].splice(this.joinLeave['join'].indexOf(formattedUser), 1);
```
Zarel, line 1556:
```suggestion
```

In case my outlining wasn't clear:

- If a user leaves and rejoins, this can be removed completely
- If a user joins and leaves, this should be kept, because it's useful to know that someone was in the room briefly

And the message ordering always puts leaves after joins, so you don't have to worry about the order beyond that.

Zarel:
If you have some extra time, do you want to also implement this system in `BattleLog`, for consistency?

js/client-chat.js
AnnikaCodes: Oh, I misunderstood that.

AnnikaCodes:
I couldn't find a method to remove the `<div>` for a leave message in BattleLog, so I just manually `.replace()`d it. Do you know of a better way to remove it?

Zarel:
Oh, I meant the entire batching system. So it matches chatrooms.

Zarel:
I would use `.parent.replaceChild()` to replace nodes.

Zarel:
Although in this particular case, setting `.innerText` is probably simplest.

AnnikaCodes:
> Oh, I meant the entire batching system. So it matches chatrooms.

sorry, what do you mean by "batching system?"

Zarel:
by "batching system", I mean the system where it says:

> Foo and Bar joined, Baz left

rather than

> Foo joined.
> Baz left.
> Bar joined.

AnnikaCodes:
Well, this works, although I still feel like there's a cleaner way to do this.

src/battle-log.ts
```diff
@@ -93,16 +93,54 @@ class BattleLog {
...
+				const joinSplit = rawJoinsLeaves?.split(';')[0].replace(' joined', '').split(' and ');
+				joins = joinSplit[0].split(', ');
+				if (joinSplit.length > 1) joins.push(joinSplit[1]);
+				if (rawJoinsLeaves.includes('; ')) rawLeaves = rawJoinsLeaves?.split('; ')[1];
+			} else {
+				if (rawJoinsLeaves) rawLeaves = rawJoinsLeaves;
+			}
+
+			if (rawLeaves) {
+				const leaveSplit = rawLeaves.replace(' left', '').split(' and ');
```
Zarel, line 117: Okay I definitely don't think the correct way to do this is to text-parse the previous line.

I would recommend a `BattleLog#joinLeave` property; like how `client-check` does it.

Zarel: You can basically just copy/paste the implementation from `client-chat` but rewrite the jQuery to be raw JavaScript.

Zarel: Here's how I'd do it:

```ts
joinLeave: {
  joins: string[],
  leaves: string[],
  element: HTMLDivElement,
} | null;
```

If `this.joinLeave` is `null`, you add a new element and populate it. Otherwise, you set the `element.innerText` to the new message that you generate from `joinLeave.joins` and `joinLeave.leaves`. You set `this.joinLeave = null` whenever you add a non-join/leave message.

AnnikaCodes: Sure.

src/battle-log.ts
```diff
@@ -93,16 +99,37 @@ class BattleLog {
...
+		case 'join': case 'j': case 'leave': case 'l': {
 			const user = BattleTextParser.parseNameParts(args[1]);
-			divHTML = '<small>' + BattleLog.escapeHTML(user.group + user.name) + ' joined.</small>';
-			break;
-		}
-		case 'leave': case 'l': {
-			const user = BattleTextParser.parseNameParts(args[1]);
-			divHTML = '<small>' + BattleLog.escapeHTML(user.group + user.name) + ' left.</small>';
+			const formattedUser = BattleLog.escapeHTML(user.group + user.name);
+			const isJoin = args[0].includes('j');
```
Zarel, line 105:
```suggestion
			const isJoin = (args[0].charAt(0) === 'j');
```

d-don't mind me, I'm just golfing.

src/battle-log.ts
```diff
@@ -93,16 +99,37 @@ class BattleLog {
...
+			if (!this.joinLeave) {
+				this.joinLeave = {
+					joins: [],
+					leaves: [],
+					element: document.createElement('div') as HTMLDivElement,
```
Zarel, line 110: Is the cast necessary? I thought TypeScript was smarter than that. We might also want to just use `HTMLElement`.

src/battle-log.ts
```diff
@@ -93,16 +99,37 @@ class BattleLog {
...
+			if (this.joinLeave.joins.length) {
+				buf += this.displayJoinLeaves(this.joinLeave.joins, 'joined');
+			}
+			if (this.joinLeave.leaves.length) {
+				if (this.joinLeave.joins.length) buf += '; ';
+				buf += this.displayJoinLeaves(this.joinLeave.leaves, 'left') + '<br />';
+			}
+			this.joinLeave.element.innerHTML = `<small style="color: #555555">${buf}</small>`;
+			(preempt ? this.preemptElem : this.innerElem).append(this.joinLeave.element);
 			break;
```
Zarel, line 130:
```suggestion
			return;
```

src/battle-log.ts
```diff
@@ -93,16 +99,37 @@ class BattleLog {
...
-			const user = BattleTextParser.parseNameParts(args[1]);
-			divHTML = '<small>' + BattleLog.escapeHTML(user.group + user.name) + ' left.</small>';
+			const formattedUser = BattleLog.escapeHTML(user.group + user.name);
+			const isJoin = args[0].includes('j');
+			if (!this.joinLeave) {
+				this.joinLeave = {
+					joins: [],
+					leaves: [],
+					element: document.createElement('div') as HTMLDivElement,
+				};
```
Zarel, line 111:
```suggestion
				};
				this.joinLeave.element.className = 'chat';
```

src/battle-log.ts
```diff
@@ -93,16 +99,37 @@ class BattleLog {
...
+
+			let buf = '';
+			if (this.joinLeave.joins.length) {
+				buf += this.displayJoinLeaves(this.joinLeave.joins, 'joined');
+			}
+			if (this.joinLeave.leaves.length) {
+				if (this.joinLeave.joins.length) buf += '; ';
+				buf += this.displayJoinLeaves(this.joinLeave.leaves, 'left') + '<br />';
+			}
+			this.joinLeave.element.innerHTML = `<small style="color: #555555">${buf}</small>`;
```
Zarel, line 128:
```suggestion
			this.joinLeave.element.innerHTML = `<small>${buf}</small>`;
```

(The style isn't necessary if you're setting the class to `chat`.)

src/battle-log.ts
AnnikaCodes: You're right; TypeScript does that for you.

src/battle-log.ts
```diff
@@ -222,6 +249,30 @@ class BattleLog {
...
+				} else if (j === list.length - 1) {
+					message += ', and ';
+				} else {
+					message += ', ';
+				}
+			}
+			message += BattleLog.escapeHTML(list[j]);
+		}
+		return `${message} ${action}`;
+	}
```
Zarel, line 276: I realize I'm nitpicking, but I think this would be better at a slightly different level of abstraction:

```suggestion
	textList(list: string[]) {
		let message = '';
		const listNoDuplicates: string[] = [];
		for (const user of list) {
			if (!listNoDuplicates.includes(user)) listNoDuplicates.push(user);
		}
		list = listNoDuplicates;
		for (let i = 0; i < list.length; i++) {
			if (i >= 5) {
				message += `, and ${list.length - 5} others`;
				break;
			}
			if (i > 0) {
				if (i === 1 && list.length === 2) {
					message += ' and ';
				} else if (i === list.length - 1) {
					message += ', and ';
				} else {
					message += ', ';
				}
			}
			message += list[i];
		}
		return message;
	}
```

src/battle-log.ts
```suggestion
			if (this.joinLeave.joins.length) {
				buf += `${this.textList(this.joinLeave.joins)} joined`;
			}
			if (this.joinLeave.leaves.length) {
				if (this.joinLeave.joins.length) buf += `; `;
				buf += `${this.textList(this.joinLeave.leaves)} left`;
			}
			this.joinLeave.element.innerHTML = `<small>${BattleLog.escapeHTML(buf)}</small>`;
```

src/battle-log.ts
```diff
@@ -93,16 +99,38 @@ class BattleLog {
...
+
+			let buf = '';
+			if (this.joinLeave.joins.length) {
+				buf += this.displayJoinLeaves(this.joinLeave.joins, 'joined');
+			}
+			if (this.joinLeave.leaves.length) {
+				if (this.joinLeave.joins.length) buf += '; ';
+				buf += this.displayJoinLeaves(this.joinLeave.leaves, 'left') + '<br />';
+			}
+			this.joinLeave.element.innerHTML = `<small>${buf}</small>`;
```
Zarel, line 129:
```suggestion
			let buf = '';
			if (this.joinLeave.joins.length) {
				buf += `${this.textList(this.joinLeave.joins)} joined`;
			}
			if (this.joinLeave.leaves.length) {
				if (this.joinLeave.joins.length) buf += `; `;
				buf += `${this.textList(this.joinLeave.leaves)} left`;
			}
			this.joinLeave.element.innerHTML = `<small>${BattleLog.escapeHTML(buf)}</small>`;
```

src/battle-log.ts
```diff
@@ -222,6 +250,31 @@ class BattleLog {
...
+				if (i === 1 && list.length === 2) {
+					message += ' and ';
+				} else if (i === list.length - 1) {
+					message += ', and ';
+				} else {
+					message += ', ';
+				}
+			}
+			message += list[i];
+		}
```
Zarel, line 275:
```suggestion

		if (list.length === 2) {
			 return `${list[0]} and ${list[1]}`;
		}
		for (let i = 0; i < list.length - 1; i++) {
			if (i >= 5) {
				return `${message}and ${list.length - 5} others`;
			}
			message += `${list[i]}, `;
		}
		return `${message}and ${list[list.length - 1]}`;
```

s-still golfing

Zarel:
Looks great, thanks!

PR #1540, opened 2020-06-29
Fix linter warning in battle-tooltips.ts

AnnikaCodes:
This annoyed me while testing other changes, so I fixed it.


Zarel:
Thanks!

PR #1542, opened 2020-07-04
Fix HTML escaping for battle joins/leaves

AnnikaCodes:
Usernames are already escaped on line 104.


Zarel:
I would rather remove the escaping on line 104.

Zarel:
Thanks!

PR #1543, opened 2020-07-05
Use configurable routes

AnnikaCodes:
This is based on pre's work in #1320, and will allow others to run the full client, loginserver, and replay server for testing purposes (obviously we won't publicly document the process for setting up a login server for security purposes).


replays/theme/wrapper.inc.template.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+include_once '../../pokemonshowdown.com/config/config.inc.php';
+
 if ((substr($_SERVER['REMOTE_ADDR'],0,11) === '69.164.163.') ||
```
AnnikaCodes, line 5:
```suggestion

if ((substr($_SERVER['REMOTE_ADDR'],0,11) === '69.164.163.') ||
```
oops, that shouldnt be necessary anymore since we're building it

build-tools/build-indexes
```diff
@@ -9,7 +9,7 @@ const rootDir = path.resolve(__dirname, '..');
 process.chdir(rootDir);
 
 if (!fs.existsSync('data/Pokemon-Showdown')) {
-	child_process.execSync('git clone https://github.com/smogon/pokemon-showdown.git', {
+	child_process.execSync('git clone https://github.com/smogon/pokemon-showdown.git Pokemon-Showdown', {
```
Zarel, line 12: Good catch! I think my Mac OS is case-insensitive and we haven't re-cloned on the web server in a long time.

config/config-example.js
```diff
@@ -117,6 +117,25 @@ Config.whitelist = [
 	'4cdn\\.org'
 ];
 
+Config.routes = {
+	root: 'pokemonshowdown.com',
+	client: 'play.pokemonshowdown.com',
+	dex: 'dex.pokemonshowdown.com',
+	replays: 'replay.pokemonshowdown.com',
+	usercp: 'pokemonshowdown.com/users'
+};
```
Zarel, line 126: I wonder if there's a way to only specify this in one place instead of two.

Maybe have the real config be JSON, and then PHP and the client can both generate config files from that?

js/replay-embed.js
```diff
@@ -12,26 +12,26 @@ function requireScript(url) {
...
-requireScript('https://play.pokemonshowdown.com/data/pokedex-mini-bw.js?a7');
-requireScript('https://play.pokemonshowdown.com/data/graphics.js?a7');
-requireScript('https://play.pokemonshowdown.com/data/pokedex.js?a7');
-requireScript('https://play.pokemonshowdown.com/data/moves.js?a7');
-requireScript('https://play.pokemonshowdown.com/data/abilities.js?a7');
-requireScript('https://play.pokemonshowdown.com/data/items.js?a7');
-requireScript('https://play.pokemonshowdown.com/data/teambuilder-tables.js?a7');
-requireScript('https://play.pokemonshowdown.com/js/battle-tooltips.js?a7');
-requireScript('https://play.pokemonshowdown.com/js/battle.js?a7');
+linkStyle('https://' + Config.routes.client + '/style/font-awesome.css?');
```
Zarel, line 15: `Config` is undefined here.

lib/ntbb-session.lib.php
```diff
@@ -1,12 +1,12 @@
 <?php
 
 require_once dirname(__FILE__) . '/ntbb-database.lib.php';
+include_once '../../pokemonshowdown.com/config.inc.php';
```
Zarel, line 4: What does it need from main config? It should probably be moved into Play config.

replays/battle.template.php
```diff
@@ -174,13 +174,9 @@ function userid($username) {
 
 		<input type="hidden" name="replayid" value="<?php echo htmlspecialchars($replay['id']); ?>" />
 		<!--
-
 You can get this log directly at https://replay.pokemonshowdown.com/<?php echo $replay['id']; ?>.log
-
 Or with metadata at https://replay.pokemonshowdown.com/<?php echo $replay['id']; ?>.json
-
 Most PS pages you'd want to scrape will have a .json version!
-
```
Zarel, line 183: Is this necessary? :( I like the spacing to make it easier to see.

replays/build
```diff
@@ -8,32 +8,50 @@
...
-		} catch (e) {}
-
-		return b + '="/' + c + '?' + hash + '"';
-	});
-	console.log("DONE");
-
-	process.stdout.write("Writing new `wrapper.inc.php` file... ");
-	fs.writeFileSync('theme/wrapper.inc.php', indexContents);
-	console.log("DONE");
+        let indexContents = fs.readFileSync('theme/wrapper.inc.template.php', {encoding: 'utf8'});
```
Zarel, line 19: Your editor default is 8-space indents? @_@

replays/build
```diff
@@ -8,32 +8,50 @@
...
+        process.stdout.write("Writing new `wrapper.inc.php` file... ");
+        fs.writeFileSync('theme/wrapper.inc.php', indexContents);
+        console.log("DONE");
+        process.stdout.write("Writing new `battle.php` file... ");
+        fs.writeFileSync('battle.php', battleContents);
+        console.log("DONE");
+
+}
+
+function runReplace(a, b, c) {
```
Zarel, line 37: I actually think this function is less readable out-of-line rather than in-line where it happens... It also makes the diff unreadable, although the reindentation isn't doing it any favors...

src/battle-dex.ts
```diff
@@ -114,12 +114,12 @@ if (typeof window === 'undefined') {
...
+	soundManager.setup({url: `https://${Config.routes.client}/swf/`});
 	if (window.Replays) soundManager.onready(window.Replays.soundReady);
 	soundManager.onready(() => {
 		soundManager.createSound({
 			id: 'notif',
-			url: 'https://play.pokemonshowdown.com/audio/notification.wav',
+			url: `https://${Config.routes.client}/audio/notification.wav`,
 		});
 	});
 }
```
Zarel, line 125: Wow, this is still here? It should probably be moved to `BattleScene`...

src/battle-searchresults.tsx
```diff
@@ -8,7 +8,7 @@
  */
 
 class PSSearchResults extends preact.Component<{search: DexSearch}> {
-	readonly URL_ROOT = '//dex.pokemonshowdown.com/';
+	readonly URL_ROOT = `//${Config.routes.dex}/`;
```
Zarel, line 11: `Config` is not defined here, either.

js/replay-embed.js
```diff
@@ -12,26 +12,26 @@ function requireScript(url) {
...
+linkStyle('https://' + Config.routes.client + '/style/battle.css?a7');
+linkStyle('https://' + Config.routes.client + '/style/replay.css?a7');
+
+requireScript('https://' + Config.routes.client + '/js/lib/jquery-1.11.0.min.js');
+requireScript('https://' + Config.routes.client + '/js/lib/lodash.compat.js');
+requireScript('https://' + Config.routes.client + '/js/lib/html-sanitizer-minified.js');
+requireScript('https://' + Config.routes.client + '/js/lib/soundmanager2-nodebug-jsmin.js');
+requireScript('https://' + Config.routes.client + '/js/config.js?a7');
```
Zarel, line 23: As you can see, it's not included until this specific line. And it won't actually exist until after this entire file has been run.

Zarel:
Hm. This is a super ambitious change and I think it will be difficult to get this to a mergeable state.

I'm definitely open to having routes be more configurable, but a big problem is that a lot of these files do not depend on Config, and replacing URLs with `Config.whatever` will make them crash because Config doesn't exist there. `replay-embed`, for instance, is standalone.

I'm not sure what I'd recommend here. Maybe doing things a little at a time? I just want to warn you that there's a big risk that you will tire of this PR before it's mergeable.

config/config-example.js
AnnikaCodes: Sure, that's a good idea.

lib/ntbb-session.lib.php
AnnikaCodes: It needs `$psclient['routes']`. What's this "Play config" you mention?

replays/battle.template.php
AnnikaCodes: No, it's not necessary - I'm not actually sure why I changed that.

replays/build
AnnikaCodes: `runReplace` is called twice, both on `indexContents` and `battleContents`, and I don't want to repeat that twice.

AnnikaCodes:
> Hm. This is a super ambitious change and I think it will be difficult to get this to a mergeable state.
> 
> I'm definitely open to having routes be more configurable, but a big problem is that a lot of these files do not depend on Config, and replacing URLs with `Config.whatever` will make them crash because Config doesn't exist there. `replay-embed`, for instance, is standalone.
> 
> I'm not sure what I'd recommend here. Maybe doing things a little at a time? I just want to warn you that there's a big risk that you will tire of this PR before it's mergeable.

Regarding the standalone files, how would you feel about building them from templates like we already do for the HTML files, `replays/battle.php`, and `replays/theme/wrapper.inc.php`? Simply replacing the routes would eliminate the need to load the routes from a config file whose location depends on the routes that must be loaded.

Zarel:
I don't have a problem with building them from templates.

src/battle-dex.ts
AnnikaCodes: OK, I'll do that.

AnnikaCodes:
Regarding Config being undefined in the Preact client's `.ts` and `.tsx` files: I don't know much about how Preact works; would it be possible to generate a `route-config` module in the build script that could then be included? (Since we already generate a config from `routes.json` for PHP and JavaScript, it doesn't seem like a big deal to me to also generate one for TypeScript.)

Zarel:
> Regarding `Config` being undefined in the Preact client's `.ts` and `.tsx` files: I don't know much about how Preact works; would it be possible to generate a `route-config` module in the build script that could then be included? (Since we already generate a config from `routes.json` for PHP and JavaScript, it doesn't seem like a big deal to me to also generate one for TypeScript.)

Unfortunately, the issue is more subtle than that. The Preact client does already use `config.js`. The problem is that a lot of the files it generates are intended to be used in other places as well, such as PSdex, which doesn't use `config.js`.

config/routes.json
```diff
@@ -0,0 +1,7 @@
+{
+    "root": "pokemonshowdown.com",
+	"client": "play.pokemonshowdown.com",
+	"dex": "dex.pokemonshowdown.com",
+	"replays": "replay.pokemonshowdown.com",
+	"usercp": "pokemonshowdown.com/users"
```
Zarel, line 6:
```suggestion
    "root": "pokemonshowdown.com",
    "client": "play.pokemonshowdown.com",
    "dex": "dex.pokemonshowdown.com",
    "replays": "replay.pokemonshowdown.com",
    "usercp": "pokemonshowdown.com/users"
```

(You had a mix of tabs and spaces here.)

config/routes.json
```diff
@@ -0,0 +1,7 @@
...
+}
```
Zarel, line 7:
```suggestion
}

```

config/config-example.inc.php
```diff
@@ -2,6 +2,8 @@
 
 mb_internal_encoding('UTF-8');
 
+$routes = json_decode(file_get_contents('routes.json'));
```
Zarel, line 5: Heh, I was trying to think of a way to get them to only need one place. This is actually a pretty good approach.

lib/ntbb-session.lib.php
Zarel: Do you mean `$psconfig['routes']`? That's defined in `play.pokemonshowdown.com/config.inc.php`, not `pokemonshowdown.com/config.inc.php`, right?

replays/build
Zarel: Oh, I see. Can you fix the reindentation, though? This file should be tab-indented.

src/battle-searchresults.tsx
Zarel: On reflection, I think we can allow this. Including `config` in PSdex is probably better than rewriting this.

AnnikaCodes:
> Unfortunately, the issue is more subtle than that. The Preact client does already use config.js. The problem is that a lot of the files it generates are intended to be used in other places as well, such as PSdex, which doesn't use config.js.

Hmm, that makes it more difficult. You're sure you don't want to prebuild those files or otherwise manually include the routes in each file? If not, and if I don't think of another way, would it be OK to just support configuring routes in the non-Preact client for now, and save that for a future PR?

lib/ntbb-session.lib.php
AnnikaCodes: Oh, right, it's there - the config paths were a bit hard to wrap my head around. I'll fix that.

replays/build
AnnikaCodes: OK.

Zarel:
The non-Preact client also uses files built from `src/`. They're pretty intermixed at this point.

The main thing you'd want to do is move `config` to the top of the list of files to include, to make sure it exists everywhere you use it.

lib/ntbb-session.lib.php
scheibo: TBH, i think there should probably only be one config. Currently its kind of a mess (pokemonshowdown.com sometimes includes both in places and overwrites the $psconfig global which is ick)

config/config-example.inc.php
scheibo: I agree, this is pretty clever! Nice, Annika :)

build-tools/update
```diff
@@ -37,10 +37,19 @@ try {
...
 
 let configBuf = fs.readFileSync('config/config.js', {encoding: 'utf8'});
-configBuf = configBuf.replace(/\/\* version \*\/[^;\n]*;/, `/* version */ Config.version = ${JSON.stringify(version)};`);
+const routes = JSON.parse(fs.readFileSync('config/routes.json'));
+configBuf = configBuf.replace(/\/\* version \*\/[^;\n]*;/, `/* version */ Config.version = ${JSON.stringify(version)};`)
+	.replace(/\/\* root \*\/[^,\n]*,/, `/* root */ '${routes.root}',`)
+	.replace(/\/\* client \*\/[^,\n]*,/, `/* client */ '${routes.client}',`)
+	.replace(/\/\* dex \*\/[^,\n]*,/, `/* dex */ '${routes.dex}',`)
+	.replace(/\/\* replays \*\/[^,\n]*,/, `/* replays */ '${routes.replays}',`)
+	.replace(/\/\* usercp \*\/[^,\n]*,/, `/* root */ '${routes.usercp}',`);
```
AnnikaCodes, line 46:
```suggestion
	.replace(/\/\* users \*\/[^,\n]*,/, `/* users */ '${routes.users}',`);
```

config/config-example.js
```diff
@@ -117,6 +117,25 @@ Config.whitelist = [
 	'4cdn\\.org'
 ];
 
+Config.routes = {
+	root: /* root */ 'pokemonshowdown.com',
+	client: /* client */ 'play.pokemonshowdown.com',
+	dex: /* dex */ 'dex.pokemonshowdown.com',
+	replays: /* replays */ 'replay.pokemonshowdown.com',
+	usercp: /* root */ 'pokemonshowdown.com/users',
```
AnnikaCodes, line 125:
```suggestion
	users: /* users */ 'pokemonshowdown.com/users',
```

lib/ntbb-ladder.lib.php
```diff
@@ -158,7 +158,7 @@ function getRating(&$user, $create=false) {
 				}
 				$rp = $this->getrp();
 				$ladderdb->query(
-					"INSERT INTO `{$ladderdb->prefix}ladder` (`formatid`,`userid`,`username`,`rptime`) VALUES (?,?,?,?)",
```
scheibo, line 161: @Zarel - Annika and I couldn't figure out why the MySQL on the server was OK without a default value in this `INSERT`? From inspection of the DB, `''` and `0` seem correct here, but i just wanted to call attention to this and ask if maybe you know whats going on? See below for what happens if you run the current query on mysql 5.7 

```
mysql> INSERT INTO `ntbb_ladder` (`formatid`,`userid`,`username`,`rptime`) VALUES ('gen8randombattle','johigufyd','johigufyd','1594198800');
ERROR 1364 (HY000): Field 'rpdata' doesn't have a default value
```

config/routes.json
```diff
@@ -0,0 +1,7 @@
+{
+    "root": "pokemonshowdown.com",
+    "client": "play.pokemonshowdown.com",
+    "dex": "dex.pokemonshowdown.com",
+    "replays": "replay.pokemonshowdown.com",
+    "usercp": "pokemonshowdown.com/users"
```
AnnikaCodes, line 6:
```suggestion
    "users": "pokemonshowdown.com/users"
```

js/client-chat.js
```diff
@@ -912,7 +912,7 @@
 					var userid = toID(targets[0]);
 					var registered = app.user.get('registered');
 					if (registered && registered.userid === userid) {
-						buffer += '<tr><td colspan="8" style="text-align:right"><a href="//pokemonshowdown.com/users/' + userid + '">Reset W/L</a></tr></td>';
+						buffer += '<tr><td colspan="8" style="text-align:right"><a href="//' + Config.routes.usercp + '/' + userid + '">Reset W/L</a></tr></td>';
```
AnnikaCodes, line 915:
```suggestion
						buffer += '<tr><td colspan="8" style="text-align:right"><a href="//' + Config.routes.users + '/' + userid + '">Reset W/L</a></tr></td>';
```

js/client-ladder.js
```diff
@@ -112,7 +112,7 @@
 		leave: function () {},
 		update: function () {
 			if (!this.curFormat) {
-				var buf = '<div class="ladder pad"><p>See a user\'s ranking with <a class="button" href="//pokemonshowdown.com/users/" target="_blank">User lookup</a></p>' +
+				var buf = '<div class="ladder pad"><p>See a user\'s ranking with <a class="button" href="//' + Config.routes.usercp + '/" target="_blank">User lookup</a></p>' +
```
AnnikaCodes, line 115:
```suggestion
				var buf = '<div class="ladder pad"><p>See a user\'s ranking with <a class="button" href="//' + Config.routes.users + '/" target="_blank">User lookup</a></p>' +
```

js/client.js
```diff
@@ -2514,7 +2514,7 @@ function toId() {
 
 			var buf = '<div class="userdetails">';
 			if (avatar) buf += '<img class="trainersprite' + (userid === ownUserid ? ' yours' : '') + '" src="' + Dex.resolveAvatar(avatar) + '" />';
-			buf += '<strong><a href="//pokemonshowdown.com/users/' + userid + '" target="_blank">' + BattleLog.escapeHTML(name) + '</a></strong><br />';
+			buf += '<strong><a href="//' + Config.routes.usercp + '/' + userid + '" target="_blank">' + BattleLog.escapeHTML(name) + '</a></strong><br />';
```
AnnikaCodes, line 2517:
```suggestion
			buf += '<strong><a href="//' + Config.routes.users + '/' + userid + '" target="_blank">' + BattleLog.escapeHTML(name) + '</a></strong><br />';
```

replays/build
```diff
@@ -8,32 +8,50 @@
...
+			const filename = c.replace('/' + Config.routes.client + '/', '');
+			filepath = '../' + filename;
+		}
+		const fstr = fs.readFileSync(filepath, {encoding: 'utf8'});
+		hash = crypto.createHash('md5').update(fstr).digest('hex').substr(0, 8);
+	 } catch (e) {}
+	c = c.replace('/replay.pokemonshowdown.com/', '/' + Config.routes.replay + '/');
+	c = c.replace('/dex.pokemonshowdown.com/', '/' + Config.routes.dex + '/');
+	c = c.replace('/play.pokemonshowdown.com/', '/' + Config.routes.client + '/');
+	c = c.replace('/pokemonshowdown.com/users/', '/' + Config.routes.usercp + '/');
```
AnnikaCodes, line 51:
```suggestion
	c = c.replace('/pokemonshowdown.com/users/', '/' + Config.routes.users + '/');
```

src/battle-log.ts
```diff
@@ -886,20 +886,20 @@ class BattleLog {
...
 		buf += '</style>\n';
 		buf += '<div class="wrapper replay-wrapper" style="max-width:1180px;margin:0 auto">\n';
 		buf += '<input type="hidden" name="replayid" value="' + replayid + '" />\n';
 		buf += '<div class="battle"></div><div class="battle-log"></div><div class="replay-controls"></div><div class="replay-controls-2"></div>\n';
-		buf += '<h1 style="font-weight:normal;text-align:center"><strong>' + BattleLog.escapeHTML(battle.tier) + '</strong><br /><a href="http://pokemonshowdown.com/users/' + toID(battle.p1.name) + '" class="subtle" target="_blank">' + BattleLog.escapeHTML(battle.p1.name) + '</a> vs. <a href="http://pokemonshowdown.com/users/' + toID(battle.p2.name) + '" class="subtle" target="_blank">' + BattleLog.escapeHTML(battle.p2.name) + '</a></h1>\n';
+		buf += `<h1 style="font-weight:normal;text-align:center"><strong>${BattleLog.escapeHTML(battle.tier)}</strong><br /><a href="http://${Config.routes.usercp}/${toID(battle.p1.name)}" class="subtle" target="_blank">${BattleLog.escapeHTML(battle.p1.name)}</a> vs. <a href="http://${Config.routes.usercp}/${toID(battle.p2.name)}" class="subtle" target="_blank">${BattleLog.escapeHTML(battle.p2.name)}</a></h1>\n`;
```
AnnikaCodes, line 896:
```suggestion
		buf += `<h1 style="font-weight:normal;text-align:center"><strong>${BattleLog.escapeHTML(battle.tier)}</strong><br /><a href="http://${Config.routes.users}/${toID(battle.p1.name)}" class="subtle" target="_blank">${BattleLog.escapeHTML(battle.p1.name)}</a> vs. <a href="http://${Config.routes.users}/${toID(battle.p2.name)}" class="subtle" target="_blank">${BattleLog.escapeHTML(battle.p2.name)}</a></h1>\n`;
```

src/panel-topbar.tsx
```diff
@@ -227,7 +227,7 @@ class UserPanel extends PSRoomPanel<UserRoom> {
 						src={Dex.resolveAvatar('' + (user.avatar || 'unknown'))}
 					/>
 				}
-				<strong><a href={`//pokemonshowdown.com/users/${user.userid}`} target="_blank" style={away ? {color: '#888888'} : null}>{name}</a></strong><br />
+				<strong><a href={`//${Config.routes.usercp}/${user.userid}`} target="_blank" style={away ? {color: '#888888'} : null}>{name}</a></strong><br />
```
AnnikaCodes, line 230:
```suggestion
				<strong><a href={`//${Config.routes.users}/${user.userid}`} target="_blank" style={away ? {color: '#888888'} : null}>{name}</a></strong><br />
```

lib/ntbb-session.lib.php
Zarel: I agree. I think the two config files should be merged and moved into `play.pokemonshowdown.com/config/config.inc.php`.

config/config-example.inc.php
Zarel: LOL, it turns out I suggested this earlier and forgot:

https://github.com/smogon/pokemon-showdown-client/pull/1543#discussion_r449928418

lib/ntbb-ladder.lib.php
Zarel: I thought it did have a default value. Either way, this seems easy enough to fix.

AnnikaCodes: Trying to create a table with default value for `rpdata`  causes an error in MySQL

config/config-example.inc.php
scheibo: Clever Zarel then!

lib/ntbb-session.lib.php
scheibo: I think maybe we can punt on that and do it in a follow up? Cross repo stuff like this is awkward and might bog down this PR. We definitely all agree I think that it should be done

lib/ntbb-ladder.lib.php
Zarel: Okay, looks like it's a MEDIUMBLOB in an InnoDB table without a default value; but we're using MySQL 5.5.

It's possible MySQL just lets it go through, or only gives a warning rather than an error.

build-tools/build-indexes
Zarel, line 12: It would break existing installs on case-sensitive operating systems, but I think I'd prefer for the directory to be lowercase going forward.

build-tools/update
```diff
@@ -37,10 +37,19 @@ try {
 } catch (e) {}
 
 let configBuf = fs.readFileSync('config/config.js', {encoding: 'utf8'});
-configBuf = configBuf.replace(/\/\* version \*\/[^;\n]*;/, `/* version */ Config.version = ${JSON.stringify(version)};`);
+const routes = JSON.parse(fs.readFileSync('config/routes.json'));
+configBuf = configBuf.replace(/\/\* version \*\/[^;\n]*;/, `/* version */ Config.version = ${JSON.stringify(version)};`)
```
Zarel, line 41: I think we should probably have a clearly marked autogenerated section of Config. This approach might be a bit unclear as to what is going on.

build-tools/update
```diff
@@ -37,10 +37,19 @@ try {
...
+	.replace(/\/\* client \*\/[^,\n]*,/, `/* client */ '${routes.client}',`)
+	.replace(/\/\* dex \*\/[^,\n]*,/, `/* dex */ '${routes.dex}',`)
+	.replace(/\/\* replays \*\/[^,\n]*,/, `/* replays */ '${routes.replays}',`)
+	.replace(/\/\* users \*\/[^,\n]*,/, `/* users */ '${routes.users}',`);
 
 fs.writeFileSync('config/config.js', configBuf);
 console.log("DONE");
+const context = {Config: {}};
+require('vm').runInNewContext(configBuf, context);
+const Config = context.Config;
```
Zarel, line 52: Neat hack. Not sure if I like this, but I suppose there's no better way.

index.template.html
```diff
@@ -40,6 +40,7 @@
 <link rel="manifest" href="/manifest.json" />
 <meta name="google-signin-client_id" content="912270888098-jjnre816lsuhc5clj3vbcn4o2q7p4qvk.apps.googleusercontent.com">
 <script>var Config = {};</script>
+<script src="//play.pokemonshowdown.com/config/config.js?"></script>
```
Zarel, line 43: Maybe not _this_ far up. Under `"Loading libraries..." but before the first library looks like the correct place.

Also is `var Config = {}` still necessary? I think it was previously to guard against Config not existing, but now that we have a hard dependency on `Config.routes`, I don't think it's necessary anymore.

AnnikaCodes: Hardcoding an empty `Config` seems silly when we include `config.js` *and* proceed to depend upon things that aren't hardcoded in it. If it were necessary, I'd want to hardcode default routes there as well (though I'm fine with just removing it).

lib/ntbb-ladder.lib.php
```diff
@@ -158,7 +158,7 @@ function getRating(&$user, $create=false) {
...
+					"INSERT INTO `{$ladderdb->prefix}ladder` (`formatid`,`userid`,`username`,`rptime`,`rpdata`,`col1`) VALUES (?,?,?,?,'',0)",
```
Zarel, line 161: Looks good. Does `col1` also not have a default value?

build-tools/build-indexes
AnnikaCodes: OK.

build-tools/update
AnnikaCodes: Sure, I took a stab at implementing that (my apologies if I missed something, it's 1 am right now).

lib/ntbb-ladder.lib.php
AnnikaCodes: Not with the current schema, no, but it's a `double`, so I could add a default value of 0. I'm still getting the hang of PS database stuff, but if you think adding a default value wouldn't break anything that seems like a good change to me!

Zarel:
Why was `replays/battle.php` renamed to battle-template? I don't see anything justifying the change, either. It's a PHP file so any routes can just be read from `psconfig` instead of needing to be built, anyway.

index.template.html
Zarel: I agree. Let's remove the hardcode.

AnnikaCodes:
> Why was `replays/battle.php` renamed to battle-template? I don't see anything justifying the change, either. It's a PHP file so any routes can just be read from `psconfig` instead of needing to be built, anyway.

Good point - I don't remember what I was thinking there.

preactalpha.template.html
```diff
@@ -51,6 +51,7 @@ <h3><button class="closebutton" tabindex="-1" aria-label="Close"><i class="fa fa
 		linkStyle("style/battle-search.css");
 		linkStyle("/style/font-awesome.css");
 	</script>
+	<script src="/config/config.js?"></script>
```
Zarel, line 54:
```suggestion
	<script defer src="/config/config.js?"></script>
```

src/battle-animations.ts
```diff
@@ -104,6 +104,17 @@ class BattleScene {
 			return '???pokemon:' + pokemonId + '???';
 		};
 
+		if (window.soundManager) {
```
Zarel, line 107: I hope this has been tested? It seems wayyyyy too late. It should nearly definitely be at the top level.

src/battle-log.ts
```diff
@@ -655,7 +655,7 @@ class BattleLog {
 				return false;
 			},
 			getURI(uri: string) {
-				return 'http://pokemonshowdown.com/interstice?uri=' + encodeURIComponent(uri);
+				return `http://${window.Config?.routes.root || 'pokemonshowdown.com'}/interstice?uri=${encodeURIComponent(uri)}`;
```
Zarel, line 658:
```suggestion
				return `http://${Config.routes.root}/interstice?uri=${encodeURIComponent(uri)}`;
```

This file newly depends on config so we don't need stuff like this anymore.

src/client-main.ts
```diff
@@ -298,13 +298,13 @@ interface PSGroup {
...
-	altport = 80;
-	registered = true;
+	id = Config.defaultserver.id ? Config.defaultserver.id : 'showdown';
+	host = Config.defaultserver.host ? Config.defaultserver.host : 'sim3.psim.us';
+	port = Config.defaultserver.port ? Config.defaultserver.port : 8000;
+	altport = Config.defaultserver.altport ? Config.defaultserver.altport : 80;
+	registered = Config.defaultserver.registered ? Config.defaultserver.registered : true;
 	prefix = '/showdown';
-	protocol: 'http' | 'https' = 'https';
+	protocol: 'http' | 'https' = (Config.defaultserver.https! ? 'https' : 'http');
```
Zarel, line 307: Config is newly required, so these fallbacks are no longer necessary.

testclient.html
```diff
@@ -18,7 +18,15 @@
...
-			var Config = {testclient: true};
+			var Config = {
+				testclient: true,
+				routes: {
+					root: 'pokemonshowdown.com',
+					client: 'play.pokemonshowdown.com',
+					dex: 'dex.pokemonshowdown.com',
+					replays: 'replay.pokemonshowdown.com'
+				},
+			};
```
Zarel, line 29: Maybe just call client config, and add `Config.testclient = true` after that line.

src/battle-animations.ts
AnnikaCodes: I re-tested just now and both music and cries seem to work just fine in battles (and replays!). Happy to move it if you would like though.

crossdomain.php
```diff
@@ -1,12 +1,14 @@
 <?php
 $config = array();
 
+include_once '../pokemonshowdown.com/config/config.inc.php';
```
Zarel, line 4: .

src/battle-animations.ts
Zarel: Yeah, even if it works, I'd rather have it top-level (so it would be preloaded, instead of needing to wait until the battle starts to star the load, which could take some time on older systems).

AnnikaCodes: Sure, good point.

build-tools/update
scheibo: Wait, I'm WTFing at this as well. At the very least, some comments would help. AIUI you:

1) read in the routes json
2) create an automatically generated config fragment.
3) try to read in the config file (and remove an old auto generated config comments) and append the config fragment with routes at the end
4) write this config

- this is all to workaround the fact that the routes.json cant be loaded in the browser so you need to embedded it in the config.json (OK)
- this is also somewhat fragile if a config already has routes defined and we trample over them by appending ours at the end to take precedence (probably not super concerning)
- these sorts of manipulations become much easier if the config is a *data* specification encoding like JSON/JSON5 instead of full blown JS (out of scope)

5) We eval the config (this is in fact just eval, no? Why bother with the vm?).

Once again, much less hacky if the config were JSON and not JS, but thats out of scope. **Why do we need to read the config in again though?** We already know what the value for routes has to be (because we append our autoconfig at the end anyway, so we know Config.routes  is going to equal what we read from routes.json, ie the routes constant)?

Possibly missing something (which means this section could benefit from some comments explaining whatever misconception I have) :)

AnnikaCodes: Oh, evaling the config was a leftover hack from when the routes were actually defined in the config file. I can remove that now.

AnnikaCodes: I agree that moving the modifiable configuration to a JSON file and generating a js file from that would be better, but imo that would go in its own PR.

action.php
```diff
@@ -28,6 +28,7 @@
 
 include_once 'lib/ntbb-session.lib.php';
 include_once '../pokemonshowdown.com/config/servers.inc.php';
+include_once '../pokemonshowdown.com/config/config.inc.php';
```
Zarel, line 31: > `pokemonshowdown.com`

This also needs to go above line 14, for obvious reasons.

build-tools/update
```diff
@@ -36,9 +43,22 @@ try {
...
+	root: '${routes.root}',
+	client: '${routes.client}',
+	dex: '${routes.dex}',
+	replays: '${routes.replays}',
+	users: '${routes.users}',
+};
+${AUTOCONFIG_END}`;
+
+// remove old automatically generated configuration and add the new one
+const configBuf = fs.readFileSync('config/config.js', {encoding: 'utf8'}).replace(autoconfigRegex, '') + autoconfig;
```
Zarel, line 61:
```suggestion
const configBuf = fs.readFileSync('config/config.js', {encoding: 'utf8'}).replace(autoconfigRegex, autoconfig);
```

The way the autoconfig is currently done, it's non-obvious that you can't put stuff after it. This will let you put stuff after it.

build-tools/update
```diff
@@ -36,9 +43,22 @@ try {
 	version += ` (${head.slice(0, 8)}${head !== origin ? `/${origin.slice(0, 8)}` : ''})`;
 } catch (e) {}
 
-let configBuf = fs.readFileSync('config/config.js', {encoding: 'utf8'});
-configBuf = configBuf.replace(/\/\* version \*\/[^;\n]*;/, `/* version */ Config.version = ${JSON.stringify(version)};`);
-
+const routes = JSON.parse(fs.readFileSync('config/routes.json'));
+const autoconfigRegex = new RegExp(`${escapeRegex(AUTOCONFIG_START)}[^]+${escapeRegex(AUTOCONFIG_END)}`, 'g');
```
Zarel, line 47:
```suggestion
const autoconfigRegex = new RegExp(`${escapeRegex(AUTOCONFIG_START)}[^]+${escapeRegex(AUTOCONFIG_END)}`);
```

I think you can make arguments for either having or not having the `g` flag, but I think with my suggested switch to replace it with new config, it's better not to have it.

build-tools/update
Zarel: Oh, I see, appending is a special-case in case the regex didn't find anything.

I think it's worth incurring a double-search for this (annoying that there's no way for `string.replace` to detect if the replace succeeded or failed).

Zarel:
```suggestion
let configBuf = fs.readFileSync('config/config.js', {encoding: 'utf8'});
if (autoconfigRegex.test(configBuf)) {
  configBuf = configBuf.replace(autoconfigRegex, autoconfig);
} else {
  configBuf += autoconfig;
}
```

customcss.php
```diff
@@ -98,7 +98,7 @@
 	Done: <?= htmlspecialchars($customcssuri) ?> was reloaded.
 </p>
 <p>
-	<a href="http://pokemonshowdown.com/servers/<?= $server ?>">Back to server management</a>
+	<a href="http://<?= $psconfig['routes']['root'] ?>/servers/<?= $server ?>">Back to server management</a>
```
Zarel, line 101: You need to add config as a dependency here, if you want this to work.

crossdomain.php
```diff
@@ -1,12 +1,14 @@
 <?php
 $config = array();
 
+include_once '../play.pokemonshowdown.com/config.inc.php';
```
Zarel, line 4:
```suggestion
include_once './config.inc.php';
```

lib/ntbb-session.lib.php
```diff
@@ -1,12 +1,12 @@
 <?php
 
 require_once dirname(__FILE__) . '/ntbb-database.lib.php';
+include_once '../../play.pokemonshowdown.com/config.inc.php';
```
Zarel, line 4:
```suggestion
include_once '../config/config.inc.php';
```

The `config/` is important! Also you don't need to go out and back in, you're already in `play.pokemonshowdown.com`.

lib/ntbb-session.lib.php
```diff
@@ -1,12 +1,12 @@
 <?php
 
 require_once dirname(__FILE__) . '/ntbb-database.lib.php';
```
Zarel, line 3: tbh this is all really old code that I wrote because I didn't understand include paths. These can all be rewritten:

```suggestion
require_once './ntbb-database.lib.php';
```

replays/battle.php
```diff
@@ -5,6 +5,7 @@
 ini_set('display_startup_errors', TRUE);
 
 include_once 'theme/panels.lib.php';
+include_once '../config/config.inc.php';
```
Zarel, line 8:
```suggestion
require_once './theme/panels.lib.php';
require_once '../config/config.inc.php';
```

action.php
```diff
@@ -28,6 +28,7 @@
 
 include_once 'lib/ntbb-session.lib.php';
```
Zarel, line 29:
```suggestion
require_once './lib/ntbb-session.lib.php';
```

lib/ntbb-session.lib.php
```diff
@@ -1,12 +1,12 @@
 <?php
 
-require_once dirname(__FILE__) . '/ntbb-database.lib.php';
+require_once './ntbb-database.lib.php';
+require_once '../config/config.inc.php';
```
AnnikaCodes, line 4:
```suggestion
require_once 'ntbb-database.lib.php';
require_once './config/config.inc.php';
```
I'm not very familiar with PHP paths, but I found that this setup fixed several file not found errors.

lib/ntbb-session.lib.php
Zarel: Yeah, it doesn't need to be in this PR.

Zarel: In general if I suggest something unrelated in PR comments, I'm not actually saying it needs to be in this particular PR.

lib/ntbb-session.lib.php
Zarel: Seriously???

Zarel: Okay, it turns out I was wrong, my workaround was actually necessary...

```suggestion
require_once __DIR__ . '/ntbb-database.lib.php';
require_once __DIR__ . '/../config/config.inc.php';
```

action.php
```diff
@@ -9,9 +9,11 @@
 
 error_reporting(E_ALL);
 
+include_once './config/config.inc.php';
```
Zarel, line 12:
```suggestion
include_once __DIR__ . '/config/config.inc.php';
```

action.php
```diff
@@ -26,7 +28,7 @@
 }
 // header("X-Debug: " . @$_SERVER['HTTP_REFERER']);
 
-include_once 'lib/ntbb-session.lib.php';
+require_once './lib/ntbb-session.lib.php';
```
Zarel, line 31:
```suggestion
require_once __DIR__ . '/lib/ntbb-session.lib.php';
```

action.php
```diff
@@ -26,7 +28,7 @@
...
 include_once '../pokemonshowdown.com/config/servers.inc.php';
 include_once 'lib/dispatcher.lib.php';
```
Zarel, line 33:
```suggestion
include_once __DIR__ . '/../pokemonshowdown.com/config/servers.inc.php';
include_once __DIR__ . '/lib/dispatcher.lib.php';
```

crossdomain.php
```diff
@@ -1,12 +1,14 @@
 <?php
 $config = array();
 
+require_once './config/config.inc.php';
```
Zarel, line 4:
```suggestion
require_once __DIR__ . '/config/config.inc.php';
```

customcss.php
```diff
@@ -3,6 +3,7 @@
 ini_set('max_execution_time', 60); // 1 minute
 
 include '../pokemonshowdown.com/config/servers.inc.php';
+include_once './config/config.inc.php';
```
Zarel, line 6:
```suggestion
require_once __DIR__ . '/../pokemonshowdown.com/config/servers.inc.php';
require_once __DIR__ . '/config/config.inc.php';
```

replays/battle.php
```diff
@@ -4,7 +4,8 @@
 ini_set('display_errors', TRUE);
 ini_set('display_startup_errors', TRUE);
 
-include_once 'theme/panels.lib.php';
+require_once './theme/panels.lib.php';
+require_once '../config/config.inc.php';
```
Zarel, line 8:
```suggestion
require_once __DIR__ . '/theme/panels.lib.php';
require_once __DIR__ . '/../config/config.inc.php';
```

src/battle-log.ts
```diff
@@ -454,7 +454,7 @@ class BattleLog {
 	static usernameColor(name: ID) {
 		if (this.colorCache[name]) return this.colorCache[name];
 		let hash;
-		if (window.Config?.customcolors?.[name]) {
+		if (Config.customcolors?.[name]) {
```
Zarel, line 457:
```suggestion
		if (Config.customcolors[name]) {
```

Let's make this required inside the file.

src/battle-log.ts
```diff
@@ -639,7 +639,7 @@ class BattleLog {
 	}
 
 	static interstice = (() => {
-		const whitelist: string[] = window.Config?.whitelist || [];
+		const whitelist: string[] = Config.whitelist || [];
```
Zarel, line 642:
```suggestion
		const whitelist: string[] = Config.whitelist;
```

config/config-example.inc.php
AnnikaCodes, line 5:
```suggestion
$routes = json_decode(file_get_contents(__DIR__ . '/routes.json'));
```
This is one more place where the odd behavior of PHP file paths must be accounted for.

config/config-example.inc.php
```diff
@@ -2,6 +2,8 @@
 
 mb_internal_encoding('UTF-8');
 
+$routes = json_decode(file_get_contents(__DIR__ . '/routes.json'));
```
Zarel, line 5:
```suggestion
$routes = json_decode(file_get_contents(__DIR__ . '/routes.json'), true);
```

This seems better than the array cast later on, imo.

config/config-example.inc.php
```diff
@@ -22,6 +24,9 @@
 	'prefix' => 'ps_',
 	'charset' => 'utf8',
 
+// routes
+	'routes' => (array) $routes,
```
Zarel, line 28:
```suggestion
	'routes' => $routes,
```

Zarel:
This is still approved; I'm just dreading the deploy, gimme another day.

PR #1544, opened 2020-07-06
Add HP Bug/Flying to recommended moves in Gens 2/3, and add Draining Kiss/Iron Defense to recommended moves

Gitrman19:
In Gens 2/3, Hidden Power Bug and Flying had use on many physical attackers, and Iron Defense and Draining Kiss are commonly used moves.


src/battle-dex-search.ts
```diff
@@ -1202,6 +1202,10 @@ class BattleMoveSearch extends BattleTypedSearch<'move'> {
 			return !moves.includes('energyball') && !moves.includes('gigadrain');
 		case 'hiddenpowerice':
 			return !moves.includes('icebeam') && !moves.includes('aurorabeam');
+		case 'hiddenpowerflying':
+			return (species.baseSpecies === 'Gyarados' || species.baseSpecies === 'Aerodactyl' || species.baseSpecies === 'Salamence' || species.baseSpecies === 'Charizard' || species.baseSpecies === 'Crobat' || species.baseSpecies === 'Dragonite' || species.baseSpecies === 'Ninjask' || species.baseSpecies === 'Swellow' || species.baseSpecies === 'Gligar') && dex.gen < 4;
+		case 'hiddenpowerbug':
+			return (species.baseSpecies === 'Forretress' || species.baseSpecies === 'Pinsir' || species.baseSpecies === 'Armaldo' || species.baseSpecies === 'Ninjask' || species.baseSpecies === 'Tyranitar' || species.baseSpecies === 'Scizor' || species.baseSpecies === 'Aerodactyl' || species.baseSpecies === 'Dugtrio' || species.baseSpecies === 'Flygon' || species.baseSpecies === 'Marowak' || species.baseSpecies === 'Machamp' || species.baseSpecies === 'Sandslash') && dex.gen < 4;
```
AnnikaCodes, line 1208:
```suggestion
    case 'hiddenpowerflying':
        return dex.gen < 4 && [
            'Gyarados', 'Aerodactyl', 'Salamence', 'Charizard', 'Crobat', 'Dragonite', 'Ninjask', 'Swellow', 'Gligar'
        ].includes(species.baseSpecies);
    case 'hiddenpowerbug':
        return dex.gen < 4 && [
            'Forretress', 'Pinsir', 'Armaldo', 'Ninjask', 'Tyranitar', 'Scizor', 
            'Aerodactyl', 'Dugtrio', 'Flygon', 'Marowak', 'Machamp', 'Sandslash'
        ].includes(species.baseSpecies);
```
I'm not very familiar with this code, so there might be some other issue I'm missing here, but you probably should just use arrays here.

Zarel: We tend not to hardcode species like this.

Here's how I'd do it:

```suggestion
		case 'hiddenpowerflying':
			return dex.gen < 4;
		case 'hiddenpowerbug':
			return dex.gen < 4 && !moves.includes('megahorn');
```

Zarel:
Cool, thanks!

Although one question about Draining Kiss – I get the impression that it's usually a bad move. In what situations would it get used?

Gitrman19:
> Cool, thanks!
> 
> Although one question about Draining Kiss – I get the impression that it's usually a bad move. In what situations would it get used?

Ayy thanks! Draining Kiss is commonly used on Fairy types in 1825, like Magearna, Hatterne, Primarina, Comfey, and Gardevoir

Zarel:
1825?

Zarel:
Do you have link to analyses or anything?

KrisXV:
He means the 1825 ladder stats. It’s comfeys most used move and is good for setup sweeper magearna. There are no analyses yet since the metas are still unstable from the DLC

Zarel:
Well, I'll accept this for now. But I do think we should probably think about more precise thresholds for when to suggest Draining Kiss or not. Is it viable without boosting?

PR #1546, opened 2020-07-09
Update database schemas

AnnikaCodes:
There are no schemas provided at all for a number of tables needed to run the replay and login servers, and the schema for the replay table is outdated. This PR fixes that.


Zarel:
Looks good.

PR #1547, opened 2020-07-10
Fix issue with getting the category icon for tooltips

HoeenCoder:


HoeenCoder:
Zarel says he has a patch incoming so closing.

Issue #1271, opened 2019-04-30
Block PMs/Block Challenges is not persistent


scheibo:
- the 'Block PMs' is not persistent at all, even if you refresh the setting gets dropped. 'Block Challenges' persists through refreshes, which is odd because they're basically implemented the same way? I wonder if something here is server related. This definitely worked during my initial testing, so I'll need to figure out where the setting is getting dropped.
- more importantly, both are not going to work properly after logging out or changing names. When you log out, values get overwritten because the server sends `{blockChallenges: false, blockPMs: false}` for the guest account you're now associated with. Simply avoiding overwriting these values for guest accounts is an easy fix, but doesn't really solve renames. 

The main issue is that these **server settings are per account**, unlike the other settings. The `serversettings` pref can key its contents by userid, but there's a question of what behavior exactly we want for these toggles - should the preferences actually be for account/username, or should they apply to the client like the other toggles? (the latter being more difficult to accomplish - if I'm on account A and Block PMs and Challenges using the UI toggle, but then switch to account B and do `/unbch`, if I switch to account A should challenges be unblocked there, or just for account B?)?

I think these settings should probably should apply to the client as much as possible, and not the account. I don't think keying `serversettings` by account ID is the best approach, and I'd propose that the client simply ignore the initial state sent by the server for an account (ie/ ignore the `{blockChallenges: true|false, blockPMs: true|false}` from the first`|updateuser` for a given ID, but update the toggles for any updates *after* that. Thus, in my scenario above, account B doing `/unbch` will change the pref and when A logs back in, A will no longer be blocking challenges either).

*Thanks @HoeenCoder and Flare for bringing this to my attention*

scheibo:
The bad `blockPMs`  behavior is due to this line in [`server/users.js`](https://github.com/Zarel/Pokemon-Showdown/blob/df3d9b91ef18bae97c4e5cae8790615d807c81f9/server/users.js#L1153).

```javascript
if (this.blockPMs && this.can('lock') && !this.can('bypassall')) this.blockPMs = false;
```
This didn't come up in testing because my `this.can('lock')` isn't true for my test account (and most users, I'd imagine).

What is the correct solution here? My proposed solution to ignore the first `|updateuser` and set whatever is in the client's prefs seems like it would be going against the intent of the server code, which is to not allow staff to permanently block PMs? Should staff not get a 'Block PMs' toggle in the UI at all?

Zarel:
[...]

PR #1548, opened 2020-07-11
Introduce a "Don't allow spectators" checkbox

scheibo:
Client side portion of https://github.com/smogon/pokemon-showdown/issues/6950

Screen Shot 2020-07-11 at 11 32 46

- CSS for `.checkbox` was changed to make it optimally aligned on the latest Chrome
- Persisting `serversettings` appropriately was fixed by adding `_.clone` calls so Backbone's change detection will work. This is important for #1271 (though that also requires we send `/updatesettings` on setup and can be done in a followup PR)
- `/ionext` is used instead of `/hidenext` intil https://github.com/smogon/pokemon-showdown/issues/6950 is finished (ie. passwords get added to hidden battle rooms)

One 'gotcha' is that toggling the privacy checkbox in a challenge will not cause the ladder checkbox to update (meaning all of the privacy checkboxes in the UI do not always reflect the current state of the `disallowspectators` pref). This is fine because:

1.  you dont always want the UI to reflect the state of the `disallowspectators` pref (for example, in the case of two difference challenge windows open at the same time that desire different settings)
2. the only requirement is that the checkbox state reflects the privacy of the battle it is scoped to. The persistence of `disallowspectators` is purely a convenience feature to avoid users constantly toggling the checkbox back and forth

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -43,6 +43,7 @@
 				buf += '<div class="menugroup"><form class="battleform" data-search="1">';
 				buf += '<p><label class="label">Format:</label>' + this.renderFormats() + '</p>';
 				buf += '<p><label class="label">Team:</label>' + this.renderTeams() + '</p>';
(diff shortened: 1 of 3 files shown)


scheibo:
Note: we’re going to want to restart tours before deploying this, preferably at 3AM EST.

Zarel:
Wow, nice!

PR #1551, opened 2020-07-12
Add persistent ignore specs and auto timer toggles

scheibo:
Requested by @DaWoblefet: https://www.smogon.com/forums/threads/options-for-automatically-doing-ionext-timer-on-ignoring-spectators-etc.3652588/.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -47,6 +47,7 @@
 			this.battle.startCallback = function () { self.updateControls(); };
 			this.battle.stagnateCallback = function () { self.updateControls(); };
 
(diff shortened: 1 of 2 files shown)


scheibo:
It looks like someone has also requested an all battles version of 'Ignore opponent' (https://www.smogon.com/forums/threads/move-ignore-spectators-and-ignore-opponent-into-the-all-battles-category-in-battle-options.3653847/), though I'm less sure about that one.

Zarel:
The "this battle" vs "all battles" distinction seems difficult to handle, but I think this is a good incremental change.

PR #1555, opened 2020-07-13
Handle Gigantamaxing properly

KrisXV:
i need help with sprites support before this or the server-side part of this (https://github.com/smogon/pokemon-showdown/pull/7001) gets merged @Zarel

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -8,8 +8,13 @@ const child_process = require("child_process");
 const rootDir = path.resolve(__dirname, '..');
 process.chdir(rootDir);
 
+<<<<<<< HEAD
 if (!fs.existsSync('data/pokemon-showdown')) {
 	child_process.execSync('git clone https://github.com/smogon/pokemon-showdown.git', {
+=======
+if (!fs.existsSync('data/Pokemon-Showdown')) {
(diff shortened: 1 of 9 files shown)


Zarel:
The best way for sprites support is to keep the dex entry. :p

Zarel:
Like I mentioned in #sim-dev:

> the validator/simulator can convert it to a pikachu with the gmax flag before use
> convenience doesn't affect that

Zarel:
I do think Gigantamaxing should just be `|-start|POKEMON|Dynamax|Gigantamax` though. I think that's possible to support on the client.

KrisXV:
> I do think Gigantamaxing should just be |-start|POKEMON|Dynamax|Gigantamax though. I think that's possible to support on the client.
@Zarel I believe the gigantamax currently gets sent through a `-formechange`

Zarel:
> @Zarel I believe the gigantamax currently gets sent through a `-formechange`

Yes, I'm saying it shouldn't. It's not a real forme change, and changing it to be a volatile flag better reflects how it works in-game, and should simplify our code, too.

KrisXV:
Bump? @Zarel

Zarel:
Looks good!

PR #1559, opened 2020-07-18
Support renaming battles

AnnikaCodes:
Client support for smogon/pokemon-showdown#7015.


Zarel:
Looks good, although I'm somewhat surprised this is enough.

PR #1562, opened 2020-07-20
Replays: Display !details better

### LINKED ISSUE #1560: The "!details" command is bugged in replays, weird behavior
I think this is an error:
bandicam 2018-12-09 14-36-31-399

AnnikaCodes:
(mostly) fixes #1560.
`<psicon>` tags still don't work in replays -- there's some sort of odd bug there where the CSS for the image gets removed from the `<span>` tag, and fixing it is much more complicated (and not something I'm making a priority to work on right now). If you don't want to merge this until that's fixed, feel free to draft it -- I might work on this more some other time.
However, !details looks much nicer, as you can see below or at https://replays.ps.annika.codes/gen8randombattle-223.
image


Zarel:
Neat!

PR #1563, opened 2020-07-22
Remove SoundManager dependency

Zarel:
All sound stuff is now handled directly by BattleSound, using the HTML5 audio API.

The main complicated thing we do with sound is loop music with an intro. This is unfortunately not supported by ANY sound library out there (I had to manually add support for it myself to soundManager!)

https://github.com/scottschiller/SoundManager2/pull/13

In the end, I don't think the existing libraries out there actually give us anything I care about.


Zarel:
If you're wondering: Yes, there's still a slight stutter as the music loops. I don't know how to fix that.

Zarel:
Since I apparently forgot to explain this: The main reason I did this is because SoundManager is really brittle, crashing if you try to use it before it's loaded. A recent refactor changed the load order, which made SoundManager crash, which was really dumb.

I've been wanting to drop the SoundManager dependency for a long time, because it's a huge dependency that doesn't really do much for us anymore now that usercount of pre-HTML5 browsers is negligible; this just motivated me to actually do it.

Sound simply won't play in browsers that don't support HTML5 `<audio>`, but everything else should work.

Zarel:
In testing, the only weirdness is that linking directly to a battle in testclient doesn't play BGM (because of Chrome's autoplaying sound restrictions), but apparently the old sound system had that problem, too.

js/client.js
```diff
@@ -215,7 +215,7 @@ function toId() {
 		getActionPHP: function () {
 			var ret = '/~~' + Config.server.id + '/action.php';
 			if (Config.testclient) {
-				ret = 'https://' + Config.origindomain + ret;
+				ret = 'https://' + Config.routes.client + ret;
```
Zarel, line 218: Heh, I fixed this but probably should have pushed it.

xfix:
Web Audio API does provide `loopStart` and `loopEnd` properties if that's what you are looking for. Audio API is kinda boiler-platy to use, but https://developer.mozilla.org/en-US/docs/Web/API/AudioBufferSourceNode/loopStart should provide a generator idea of how to use loops.

urkerab:
I think the old system had it worse, since SoundManager hadn't even initialised when the battle started, so that it could never have sound, even on an older browser which didn't block autoplay. (I have the reverse problem; I got so tired of YouTube autoplay that I completely disabled autoplay, which means that I have to jump through hoops to get any sound in PS at all.)

Zarel:
I am, of course, open to using the Web Audio API to get around the stutter, if any of you want to PR that.

PR #1564, opened 2020-07-22
Allow for challenging users with multiple rules

KrisXV:
Currently, when you do `/chall user, gen8ou@@@inversemod,scalemonsmod`, it only detects the first additional rule. This allows users to do the aforementioned command and actually challenge the target user to [Gen 8] OU + Inverse Mod + Scalemons Mod. This works correctly with `/chall user, gen8ou@@@customrule1,customrule2, customrule(n)` as well.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -447,7 +447,7 @@
 				var self = this;
 				var challenge = function (targets) {
 					target = toID(targets[0]);
-					self.challengeData = {userid: target, format: targets[1] || '', team: targets[2] || ''};
+					self.challengeData = {userid: target, format: targets.length > 1 ? targets.slice(1).join(',') : '', team: ''};
 					app.on('response:userdetails', self.challengeUserdetails, self);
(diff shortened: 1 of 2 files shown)


PartMan7:
I always thought the intended syntax was Rule1,Rule2.

`/challenge [Gen 8] OU@@@InverseMod,-Azumarill,Blitz`, for example, works as intended.

KrisXV:
> I always thought the intended syntax was Rule1,Rule2.
> 
> `/challenge [Gen 8] OU@@@InverseMod,-Azumarill,Blitz`, for example, works as intended.

@PartMan7 doesnt look like it to me (I used the exact same command)
t

Zarel:
I don't think anyone's using the optional team parameter at the moment. We should probably remove it. I'd definitely prefer to use `,` because that's what `CUSTOM-RULES.md` teaches, and it would match `/tour rules`.

KrisXV:
alright, did @zarel

PR #1565, opened 2020-07-23
Collapse consecutive rename messages

AnnikaCodes:


src/battle-log.ts
```diff
@@ -133,11 +139,23 @@ class BattleLog {
...
+			if (toID(args[2]) === toID(user.name)) return;
+			if (!this.lastRename) {
+				this.lastRename = {
+					from: args[2],
+					to: '',
+					element: document.createElement('div'),
+				};
+				this.lastRename.element.className = 'chat';
+			} else if (toID(this.lastRename.to) !== toID(user.name)) {
+				this.lastRename.from = args[2];
```
Zarel, line 151: ? Was this a mistake?

Zarel, line 151:
```suggestion
			if (!this.lastRename || toID(this.lastRename.to) !== toID(user.name)) {
				this.lastRename = {
					from: args[2],
					to: '',
					element: document.createElement('div'),
				};
				this.lastRename.element.className = 'chat';
```

This seems like what you'd want? Unless you intentionally wanted to hide earlier renames if multiple people renamed in a row.

Zarel: I suppose it wouldn't be horrible if you did, but I don't know if that's necessary.

AnnikaCodes: Mm, I suppose multiple users renaming would be pretty difficult to abuse with a bot.

PR #1567, opened 2020-07-23
Add website to client repository

AnnikaCodes:
This pull request merges the [website repository](https://github.com/smogon/pokemonshowdown.com/) into the client, as discussed in https://github.com/smogon/pokemonshowdown.com/pull/2#issuecomment-662154003. This will eliminate the issues caused by having two separate configurations. 

Of course, when this PR is merged, some actions will need to be taken on the server that hosts PS. Namely, the `pokemonshowdown.com/config/` and `play.pokemonshowdown.com/config/` directories will need to be combined, and the web server configuration will need to be updated to point to the new directory.


Zarel:
Can https://github.com/smogon/pokemonshowdown.com/pull/2#issuecomment-663106416 be closed in favor of this? Or does this not set up routes? (I could see value in both approaches)

AnnikaCodes:
This doesn't set up routes - I felt that it would be cleaner to add that in a separate commit.

Zarel:
Okay, that's fair enough.

PR #1568, opened 2020-07-24
Show locked status in all trainer cards

AnnikaCodes:
Previously, you would only be able to tell that a user was locked through their usercard if you got that usercard from a room's user list. Now, along with the server PR (https://github.com/smogon/pokemon-showdown/pull/7041) to send groups for locked & namelocked users, locked status will show in all usercards.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -2509,13 +2509,13 @@ function toId() {
 			var name = data.name;
 			var avatar = data.avatar || '';
 			var group = ((Config.groups[data.roomGroup] || {}).name || '');
-			var globalgroup = ((Config.groups[(data.group || Config.defaultGroup || ' ')] || {}).name || '');
-			if (globalgroup) {
-				if (!group || group === globalgroup) {
-					group = "Global " + globalgroup;
-					globalgroup = '';
+			var globalGroupObject = (Config.groups[(data.group || Config.defaultGroup || ' ')] || null);
+			var globalgroup = '';
+			if (globalGroupObject && globalGroupObject.name) {
+				if (!group || group === globalGroupObject.name) {
+					group = (globalGroupObject.type === 'punishment' ? "" : "Global ") + globalGroupObject.name;
 				} else {
-					globalgroup = "Global " + globalgroup;
+					globalgroup = (globalGroupObject.type === 'punishment' ? "" : "Global ") + globalGroupObject.name;
 				}
 			}
 			var ownUserid = app.user.get('userid');
--- a/js/client.js
+++ b/js/client.js
@@ -2508,14 +2508,16 @@ function toId() {
 			var userid = data.userid;
 			var name = data.name;
 			var avatar = data.avatar || '';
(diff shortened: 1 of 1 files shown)


js/client.js
```diff
@@ -2509,13 +2509,13 @@ function toId() {
 			var name = data.name;
 			var avatar = data.avatar || '';
 			var group = ((Config.groups[data.roomGroup] || {}).name || '');
-			var globalgroup = ((Config.groups[(data.group || Config.defaultGroup || ' ')] || {}).name || '');
-			if (globalgroup) {
-				if (!group || group === globalgroup) {
-					group = "Global " + globalgroup;
-					globalgroup = '';
+			var globalGroupObject = (Config.groups[(data.group || Config.defaultGroup || ' ')] || null);
```
Zarel, line 2512:
```suggestion
			var globalGroupObject = (Config.groups[data.group || Config.defaultGroup || ' '] || null);
```

Zarel: tbh I don't really like `Object` as part of a variable name. Probably call these `globalGroup` and `globalGroupName` instead of `globalGroupObject` and `globalgroup`?

js/client.js
```diff
@@ -2509,13 +2509,13 @@ function toId() {
 			var name = data.name;
 			var avatar = data.avatar || '';
 			var group = ((Config.groups[data.roomGroup] || {}).name || '');
```
Zarel, line 2511: Sorry to make you do this, but could you also rename `group` to `groupName`?

This is already a lot more readable, though, so thanks for that! It might seem minor to you, but it's a huge difference in readability to me.

js/client.js
```diff
@@ -2509,13 +2509,13 @@ function toId() {
...
-			var globalgroup = ((Config.groups[(data.group || Config.defaultGroup || ' ')] || {}).name || '');
-			if (globalgroup) {
-				if (!group || group === globalgroup) {
-					group = "Global " + globalgroup;
-					globalgroup = '';
+			var globalGroup = (Config.groups[data.group || Config.defaultGroup || ' '] || null);
+			var globalGroupName = '';
+			if (globalGroup && globalGroup.name) {
+				if (!group || group === globalGroup.name) {
+					group = (globalGroup.type === 'punishment' ? "" : "Global ") + globalGroup.name;
```
Zarel, line 2516: This lets me see what's going on clearly enough that I would write it this way:

```js
			if (globalGroup && globalGroup.name) {
				if (globalGroup.type === 'punishment') {
					globalGroupName = globalGroup.name;
				} else if (!group || group === globalGroup.name) {
					groupName = "Global " + globalGroup.name;
				} else {
					globalGroupName = "Global " + globalGroup.name;
				}
			}
```

js/client.js
AnnikaCodes: OK.

Zarel:
Neat, this will be a good improvement!

PR #1569, opened 2020-07-24
Support |forceregister| popup type

AnnikaCodes:
Client-side support for https://github.com/smogon/pokemon-showdown/pull/7053.


scheibo:
I think this can be closed now that we're not going with the forceregister solution anymore. Thanks for you work here!

PR #1570, opened 2020-07-25
Don't show punishments twice

AnnikaCodes:
Previously, punishments would be shown as both a user's global rank and their room rank when getting a usercard from the room list. (An example is below.)

image

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -2513,8 +2513,7 @@ function toId() {
 			var globalGroupName = '';
 			if (globalGroup && globalGroup.name) {
 				if (globalGroup.type === 'punishment') {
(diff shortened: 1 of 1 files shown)


js/client.js
```diff
@@ -2514,6 +2514,7 @@ function toId() {
 			if (globalGroup && globalGroup.name) {
 				if (globalGroup.type === 'punishment') {
 					globalGroupName = globalGroup.name;
+					groupName = '';
```
Zarel, line 2517:
```suggestion
					groupName = globalGroup.name;
```

Zarel:
Thanks!

PR #1572, opened 2020-07-26
Support /cleartext

AnnikaCodes:
Client-side support for https://github.com/smogon/pokemon-showdown/pull/7065.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1451,7 +1451,7 @@
 					var $messages = $('.chatmessage-' + user);
 					if (!$messages.length) break;
 					$messages.find('a').contents().unwrap();
-					if (row[1] === 'hide' || row[1] === 'delete') {
+					if (row[1] !== 'unlink') {
 						$messages = this.$chat.find('.chatmessage-' + user);
 						if (!$messages.length) break;
 						var lineCount = parseInt(row[3], 10) || 0;
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1441,7 +1441,9 @@
 						var lineCount = parseInt(row[3], 10) || 0;
 						if (lineCount) $messages = $messages.slice(-lineCount);
 						$messages.hide().addClass('revealed').find('button').parent().remove();
-						this.$chat.children().last().append(' <button name="toggleMessages" value="' + user + '" class="subtle"><small>(' + $messages.length + ' line' + ($messages.length > 1 ? 's' : '') + ' from ' + user + ' hidden)</small></button>');
+						if (row[1] !== 'clear') {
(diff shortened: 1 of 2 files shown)


Zarel:
We should probably at this point rename `|unlink|`. Impact on bots seems very small.

I'm thinking `|hidelines|` should be easy enough:

`|hidelines|unlink|USERNAME|LINES`

`|hidelines|hide|USERNAME|LINES`

`|hidelines|delete|USERNAME|LINES`

Zarel:
tbh I don't think we'll end up using the `hide` setting much after this, either. I can't think of use-cases where you'd want to hide things from users but don't particularly care if they can still see it. I guess just spam?

js/client-chat.js
```diff
@@ -1444,7 +1445,21 @@
...
 					}
 					break;
-
+				case 'hidelines':
+					if (Dex.prefs('nounlink')) return;
+					var user = toID(row[2]);
+					var $messages = $('.chatmessage-' + user);
+					if (!$messages.length) break;
+					$messages.find('a').contents().unwrap();
+					if (row[1] === 'hide' || row[1] === 'delete') {
```
Zarel, line 1454:
```suggestion
					if (row[1] !== 'unlink') {
```

js/client-chat.js
```diff
@@ -1444,7 +1445,21 @@
...
+						$messages = this.$chat.find('.chatmessage-' + user);
+						if (!$messages.length) break;
+						var lineCount = parseInt(row[3], 10) || 0;
+						if (lineCount) $messages = $messages.slice(-lineCount);
+						$messages.hide().addClass('revealed').find('button').parent().remove();
+						if (row[1] === 'hide') this.$chat.children().last().append(' <button name="toggleMessages" value="' + user + '" class="subtle"><small>(' + $messages.length + ' line' + ($messages.length > 1 ? 's' : '') + ' from ' + user + ' hidden)</small></button>');
```
Zarel, line 1460: I don't suppose you could grab the user group to show auth the unhide button?

AnnikaCodes: I thought that it would be easier to handle this server side, by sending `|delete|` to regular users and `|hide|` to staff. Handling auth on the server seems like a cleaner permission to me.

Zarel: Hm, I suppose that works too. I think client is "better" but I suppose old-client architecture sucks so it's easier to do it server-side, so I support your decision to do it server-side.

src/battle-log.ts
```diff
@@ -192,6 +193,17 @@ class BattleLog {
 			}
 			return;
 		}
+
+		case 'hidelines': {
+			const user = toID(args[2]);
+			this.unlinkChatFrom(user);
+			if (args[1] === 'hide' || args[1] === 'delete') {
```
Zarel, line 200:
```suggestion
			if (args[1] !== 'unlink') {
```

Zarel:
Looks good!

PR #1573, opened 2020-07-27
Don't allow uploading blank teams to PokéPaste

AnnikaCodes:


js/client-teambuilder.js
```diff
@@ -798,7 +798,9 @@
 			this.update();
 		},
 		pokepasteExport: function () {
-			document.getElementById("pasteData").value = Storage.exportTeam(this.curSetList);
+			var team = Storage.exportTeam(this.curSetList);
+			if (!team) return app.addPopupMessage("You cannot upload a blank team!");
```
AnnikaCodes, line 802:
```suggestion
			if (!team) return app.addPopupMessage("Add a Pokémon to your team before uploading it!");
```

PR #1574, opened 2020-07-27
Support register buttons in popups

AnnikaCodes:


Zarel:
Perhaps it should give some sort of error message if the user is already registered? Silently failing is generally considered horrible.

Zarel:
Looks good, thanks!

PR #1575, opened 2020-07-28
Support configurable URLs in the website

AnnikaCodes:
This is migrating the changes proposed in smogon/pokemonshowdown.com#2 to the new location of the website code.

### DIFF
--- a/website/index.php
+++ b/website/index.php
@@ -2,8 +2,8 @@
 
 // error_reporting(0);
 
-include 'style/wrapper.inc.php';
 include_once __DIR__ . '/../config/config.inc.php';
+include 'style/wrapper.inc.php';
 
 function servercmp($a, $b) {
 	global $usercount;
--- a/website/index.php
+++ b/website/index.php
@@ -3,6 +3,7 @@
 // error_reporting(0);
 
 include 'style/wrapper.inc.php';
+include_once __DIR__ . '/../config/config.inc.php';
 
 function servercmp($a, $b) {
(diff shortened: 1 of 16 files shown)


website/index.php
```diff
@@ -3,6 +3,7 @@
 // error_reporting(0);
 
 include 'style/wrapper.inc.php';
+include_once __DIR__ . '/../config/config.inc.php';
```
Zarel, line 6: Hm, while you're doing this, can you go and move the config to be the first include in every file that includes it?

This is mostly useful because I have Sentry set up in config, so putting it first allows Sentry to report errors in other includes.

website/news/embed.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
+include_once __DIR__ . '/../../config/config.inc.php';
 // in theory this file is no longer used
-
 include __DIR__ . '/../../config/news.inc.php';
```
Zarel, line 5: Awkward place to put the comment, I think. I'd do it like this:

```php
// in theory this file is no longer used

include_once __DIR__ . '/../../config/config.inc.php';
include __DIR__ . '/../../config/news.inc.php';
```

website/index.php
AnnikaCodes: Sure thing!

website/news/embed.php
AnnikaCodes: Works for me.

Zarel:
Thanks!

PR #1577, opened 2020-07-28
Fix undefined $psconfig in wrapper.inc.php

AnnikaCodes:


website/style/wrapper.inc.php
```diff
@@ -48,13 +46,13 @@ function includeHeaderBottom() {
 		<div class="header"><div class="header-inner">
 			<ul class="nav">
 				<li><a class="button nav-first<?php curPage('home') ?>" href="/"><img src="/images/pokemonshowdownbeta.png" alt="Pok&eacute;mon Showdown! (beta)" /> Home</a></li>
-				<li><a class="button" href="//<?= $psconfig['routes']['dex'] ?>/">Pok&eacute;dex</a></li>
-				<li><a class="button" href="//<?= $psconfig['routes']['replays'] ?>/">Replays</a></li>
```
Zarel, line 52: This should be done by `global $psconfig` at the function definition.

Like this:

```
function includeHeaderBottom() {
	global $page, $pageTitle, $headerData, $psconfig;
```

AnnikaCodes: Oh, that's a much better solution

PR #1578, opened 2020-07-29
Correctly display reveal button for /cleartext

AnnikaCodes:
Client-side support for https://github.com/smogon/pokemon-showdown/pull/7082.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1247,6 +1247,9 @@
 		receive: function (data) {
 			this.add(data);
 		},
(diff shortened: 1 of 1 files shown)


Zarel:
What was wrong with your original server-side implementation?

AnnikaCodes:
> What was wrong with your original server-side implementation?

It was looking at the rank of the user whose lines were being cleared, not the user receiving the protocol message.

Zarel:
Ah yes that would make sense. You'd need to send it individually for every user for that one to work.

PR #1581, opened 2020-07-31
Update credits

scheibo:
Server: https://github.com/smogon/pokemon-showdown/pull/7094

### DIFF
--- a/website/credits.php
+++ b/website/credits.php
@@ -61,6 +61,7 @@ includeHeader();
 			<ul>
 				<li><p><strong>Alexander B.</strong> [mathfreak231] <small>&ndash; Development</small></p></li>
 				<li><p><strong>Andrew Goodsell</strong> [Zracknel] <small>&ndash; Art (battle weather backdrops)</small></p></li>
+				<li><p><strong>Annika L.</strong> [Annika] <small>&ndash; Development</small></p></li>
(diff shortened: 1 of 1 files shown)


website/credits.php
```diff
@@ -61,6 +61,7 @@
 			<ul>
 				<li><p><strong>Alexander B.</strong> [mathfreak231] <small>&ndash; Development</small></p></li>
 				<li><p><strong>Andrew Goodsell</strong> [Zracknel] <small>&ndash; Art (battle weather backdrops)</small></p></li>
+				<li><p><strong>Annika L.</strong> [Annika] <small>&ndash; Development</small></p></li>
```
KrisXV, line 64: Don't we typically skip placing usernames when the user's actual name is their username (see: me)?

Zarel: I wonder if we should stop doing that. It doesn't come up much and might be confusing to people.

Zarel:
We should also get Mia here.

website/credits.php
scheibo: In this case it also includes the an additional initial, and `[Annika] L.` looked slightly weird?

My preference is "ask people how they would like to be credited" and that's what I've been doing.

KrisXV: I think what Zarel meant was `Annika L. [Annika]`

Zarel: Yeah, I think we should start doing the redundant thing ("Annika L. [Annika]") with everyone, at this point.

PR #1584, opened 2020-08-01
Fix Weather Ball display with Delta Stream

DaWoblefet:
Before, Weather Ball displayed as having 100 BP in Delta Stream, which it doesn't.


Zarel:
Thanks!

PR #1585, opened 2020-08-01
Implement spacing functionality when user list is toggled

JasTuh:
Context: While playing in tournaments with friends I would get annoyed the user list would overlap the chat so I couldn't read parts of the chat while the user list was open, not the biggest issue so understand if this change is unnecessary, but thought I give a quick go at fixing it. 

before: https://streamable.com/8pz2bu
after: https://streamable.com/2qhnar

Kinda a hacky fix: Just get the inner chat div using JQuery selectors and apply a new class thats just translateX(146px) to it if the user list is open, since 146 seems to be the standard size of the chat window. I can see how adding a magic number here would be bad, but not sure how else to get the width of the column. Ran npm run test/lint and everything seems fine there.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -1725,6 +1725,7 @@ a.ilink.yours {
 .inner-chat-userlist-maximized {
 	position: absolute;
 	left: 145px;
+	right: 0px;
 }
 .chat-log.hasuserlist,
 .chat-log-add.hasuserlist,
--- a/style/client.css
+++ b/style/client.css
@@ -1723,7 +1723,8 @@ a.ilink.yours {
 	background: #DEE4EA;
 }
 .inner-chat-userlist-maximized {
-	transform: translateX(146px);
+	position: absolute;
+	left: 145px;
 }
 .chat-log.hasuserlist,
 .chat-log-add.hasuserlist,
--- a/style/client.css
+++ b/style/client.css
@@ -1722,6 +1722,9 @@ a.ilink.yours {
 .pm-buttonbar button:hover {
 	background: #DEE4EA;
 }
+.inner-chat-userlist-maximized {
+	transform: translateX(146px);
+}
 .chat-log.hasuserlist,
 .chat-log-add.hasuserlist,
 .tournament-wrapper.hasuserlist {
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1725,6 +1725,7 @@
 	var UserList = this.UserList = Backbone.View.extend({
 		initialize: function (options) {
 			this.room = options.room;
+			this.innerMessageLog = $('div.inner.message-log');
 		},
 		events: {
 			'click .userlist-count': 'toggleUserlist'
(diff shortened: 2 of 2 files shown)


AnnikaCodes:
You should be able to just click at the top where it says "N users" to minimize it. I don't think this change is necessary?

JasTuh:
Hey, can totally understand that view so feel free to close this but at the same time it would be nice to see who is in the battle and who is sending messages at the same time

Zarel:
I think this is a pretty useful change. Although I do also think that the list should probably be closed by clicking in the chat area in general.

Zarel:
PS uses `left` and not `transform` for arranging elements, though.

Zarel:
I might try to iterate on this myself for a bit.

Zarel:
I think the main thing is definitely wanting to be clear that the user list can be expanded/collapsed. So interface cues like "clicking on the chat area closes the user list" are probably the most important thing for usability here.

JasTuh:
I'm not too familiar with PS  but agreed the usability should definitely be extended to closing on click like you mentioned. I can also look more into it

Zarel:
Yes, please! That would save me a lot of work.

JasTuh:
Ok this seems to work pretty well upon user selection of the chat bar causing the user list to hide. Attempted to make it so any click besides the chat bar would work, but started running into trouble. Will look into again when I have time

JasTuh:
Video: https://streamable.com/ezg454 
Only problem is the right 3rd and all buttons in the user profile from the userlist work and this battle link from the userlist causes the userlist to disappear 
Screen Shot 2020-08-01 at 10 07 26 PM

style/client.css
```diff
@@ -1722,6 +1722,10 @@ a.ilink.yours {
 .pm-buttonbar button:hover {
 	background: #DEE4EA;
 }
+.inner-chat-userlist-maximized {
+	position: absolute;
```
Zarel, line 1726: Wait, `.inner-chat`? I suppose that _does_ make more sense than any of the things I was thinking.

In hindsight, I think unless you also want to update `right:`, I like your transform implementation more than updating `left`.

JasTuh: Yeah definitely open to naming suggestions here, my thinking was this was how the chat behaves when the userlist is maximized so I tried to place this class around the userlist-maximized classes, but can also see how it may belong somewhere else. Will update with a right value, I think just right:0px should work.

JasTuh:
Also wondering if this is some framework I'm not familiar with here as well and also noticed theres some type script stuff going on in this repository in fact I started in panel-chat.tsx, are you in the process of migrating over? Just curious what stuff type script handles

Zarel:
Yeah, we're in the process of migrating from Backbone+JavaScript to Preact+TypeScript.

`src/battle-log.ts` is already being used for battle logs, but all other logs are still running on `js/client-chat.js` right now.

PR #1587, opened 2020-08-02
Merge the ntbb-database libraries

AnnikaCodes:
I don't see a reason to have 2 separate `lib/` directories, and given that we are planning to migrate databases to Postgres eventually, duplicate database connection code seems especially problematic.


Zarel:
Hm. I think it makes sense to merge `database`, but the website-specific ones like `panels` should probably stay.

Zarel:
Thanks!

PR #1588, opened 2020-08-04
Don't ignore player joins/leaves

AnnikaCodes:
Implements [this suggestion](https://www.smogon.com/forums/threads/turning-on-ignore-spectators-shouldnt-prevent-me-from-seeing-my-opponents-leave-join-messages.3667838/).

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -34,8 +34,6 @@ type WeatherState = [string, number, number];
 type EffectTable = {[effectid: string]: EffectState};
 type HPColor = 'r' | 'y' | 'g';
 
-const PLAYER_GROUP = '\u2606';
-
 class Pokemon implements PokemonDetails, PokemonHealth {
 	name = '';
 	speciesForme = '';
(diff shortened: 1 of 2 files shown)


src/battle.ts
```diff
@@ -34,6 +34,8 @@ type WeatherState = [string, number, number];
 type EffectTable = {[effectid: string]: EffectState};
 type HPColor = 'r' | 'y' | 'g';
 
+const PLAYER_GROUP = '\u2606';
```
Zarel, line 37: How does the ignoreSpects code work usually? Presumably you only want to ignore unauthed users and maybe voices.

Zarel:
It looks like it's in `battle-log`. Can you implement this inside `battle-log` as well?

src/battle.ts
AnnikaCodes: Yes, exactly -- but the logic for displaying `|j|` and `|l|` messages didnt respect that.

Zarel:
Your `battle.ts` changes are no longer necessary, right?

AnnikaCodes:
> Your `battle.ts` changes are no longer necessary, right?

I'm not sure -- I'll test it out locally.

AnnikaCodes:
turns out I messed smth up, this is working now

Zarel:
Great, thanks!

PR #1590, opened 2020-08-05
Improve past gen item and move viability

petuuuhhh:
Improve berry search/Hidden Power search display for past gens. I know Zarel you wanted me to edit my original pull req initially, then said you could do it, but here it is done for you.

### DIFF
--- a/src/battle-dex-search.ts
+++ b/src/battle-dex-search.ts
@@ -1194,15 +1194,15 @@ class BattleMoveSearch extends BattleTypedSearch<'move'> {
 		case 'headbutt':
 			return abilityid === 'serenegrace';
 		case 'hiddenpowerelectric':
-			return !moves.includes('thunderbolt');
+			return (dex.gen < 4 && !moves.includes('thunderpunch')) && !moves.includes('thunderbolt');
 		case 'hiddenpowerfighting':
-			return !moves.includes('aurasphere') && !moves.includes('focusblast');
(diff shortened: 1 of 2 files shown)


src/battle-dex-search.ts
```diff
@@ -1177,6 +1177,8 @@ class BattleMoveSearch extends BattleTypedSearch<'move'> {
 			return ['technician', 'serenegrace'].includes(abilityid) || !moves.includes('powergem');
 		case 'aurawheel':
 			return species.baseSpecies === 'Morpeko';
+		case 'aurorabeam':
+			return !moves.includes('icebeam');
```
Zarel, line 1181: > /ds aurorabeam, !icebeam:
> No Pokémon found.

```suggestion
```

I'd rather not.

urkerab: Xerneas only gets Ice Beam in Gen 7.

KrisXV: I think you mean Aurora Beam.

urkerab: It got Aurora Beam in Gen 6... I'm not saying that it's viable though.

Zarel: Does Gen 7 Xerneas actually hit this code path, though? I suppose this is worth it if so.

petuuuhhh: urkerab is saying Gen 6 Xerneas hits the path. That's why it's Xerneas tech in Gen 6.

Zarel: Works for me. Thanks for the changes!

PR #1593, opened 2020-08-09
Adding types to tooltips

GameModerator:


src/battle-tooltips.ts
```diff
@@ -296,8 +298,10 @@ class BattleTooltips {
 						index++;
 					}
 				}
+				tooltipType = 'pokemon';
 			} else {
 				buf = this.showPokemonTooltip(pokemon);
+				tooltipType = 'pokemon';
 			}
```
Zarel, line 305:
```suggestion
			} else {
				buf = this.showPokemonTooltip(pokemon);
			}
			tooltipType = 'pokemon';
```

These can be combined.

src/battle-tooltips.ts
```diff
@@ -327,23 +332,26 @@ class BattleTooltips {
...
 		}
 		default:
 			// "throws" an error without crashing
 			Promise.resolve(new Error(`unrecognized type`));
 			buf = `<p class="message-error" style="white-space: pre-wrap">${new Error(`unrecognized type`).stack}</p>`;
+			tooltipType = 'unknown';
 		}
 
-		this.placeTooltip(buf, elem, ownHeight);
+		this.placeTooltip(buf, elem, ownHeight, tooltipType);
```
Zarel, line 350: I wonder if you'd prefer to just put the `type` itself here, instead of making a separate `tooltipType` variable?

GameModerator: I don't really mind, as long as there's classes in them, I'm good. Though you'd have to take it from here since my JS knowledge is limited.

Zarel:
Cool, I did that here: 6b020327e72bf2d9b

PR #1594, opened 2020-08-13
Add button to copy replay link to clipboard

varun1729:
The way I did it is kind of hacky. Going with Zarel's preference to avoid new dependencies.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -2757,7 +2757,7 @@ function toId() {
 			// This is a hack. You can only "select" an input field. The trick is to create a short lived input element and destroy it after a copy.
 			dummyReplayLink.id = "dummyReplayLink";
 			dummyReplayLink.value = copyText.href;
-			dummyReplayLink.setAttribute("style", "absolute; left: -1000px; top: -1000px");
+			dummyReplayLink.style.position = 'absolute';
 			copyText.appendChild(dummyReplayLink);
 			dummyReplayLink.select();
 			document.execCommand("copy");
--- a/js/client.js
+++ b/js/client.js
@@ -2752,7 +2752,7 @@ function toId() {
 			this.close();
 		},
 		copyReplayLink: function () {
-			var copyText = document.querySelector("#replayLink");
+			var copyText = this.$(".replay-link")[0];
 			var dummyReplayLink = document.createElement("input");
 			// This is a hack. You can only "select" an input field. The trick is to create a short lived input element and destroy it after a copy.
 			dummyReplayLink.id = "dummyReplayLink";
--- a/js/client.js
+++ b/js/client.js
@@ -2751,7 +2751,7 @@ function toId() {
 		submit: function (i) {
 			this.close();
 		},
-		copyReplayLink: function() {
+		copyReplayLink: function () {
 			var copyText = document.querySelector("#replayLink");
 			var dummyReplayLink = document.createElement("input");
 			// This is a hack. You can only "select" an input field. The trick is to create a short lived input element and destroy it after a copy.
@@ -2761,7 +2761,7 @@ function toId() {
 			copyText.appendChild(dummyReplayLink);
 			dummyReplayLink.select();
 			document.execCommand("copy");
-			copyText.removeChild(dummyReplayLink);		  
+			copyText.removeChild(dummyReplayLink);
 		}
 	});
 
--- a/js/client.js
+++ b/js/client.js
@@ -2757,7 +2757,7 @@ function toId() {
 			// This is a hack. You can only "select" an input field. The trick is to create a short lived input element and destroy it after a copy.
 			dummyReplayLink.id = "dummyReplayLink";
 			dummyReplayLink.value = copyText.href;
-			dummyReplayLink.style = "absolute; left: -1000px; top: -1000px";
+			dummyReplayLink.setAttribute("style", "absolute; left: -1000px; top: -1000px");
 			copyText.appendChild(dummyReplayLink);
 			dummyReplayLink.select();
 			document.execCommand("copy");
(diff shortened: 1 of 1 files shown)


DaWoblefet:
Can you provide a screenshot of what this will look like?

Also, I don't think it's necessary to destroy the copied button after click - maybe a person copies the replay, pastes it somewhere, and wants to do it again later. If you want to indicate that it's finished copying, maybe transition to a "[checkmark] Copied!" sort of thing?

Zarel:
I'm reasonably sure you can select the contents of an `<a>` tag.

Zarel:
Huh, after Googling around, it looks like you might be right here.

js/client.js
```diff
@@ -2741,15 +2741,27 @@ function toId() {
...
 		submit: function (i) {
 			this.close();
+		},
+		copyReplayLink: function () {
+			var copyText = document.querySelector("#replayLink");
+			var dummyReplayLink = document.createElement("input");
+			// This is a hack. You can only "select" an input field. The trick is to create a short lived input element and destroy it after a copy.
+			dummyReplayLink.id = "dummyReplayLink";
+			dummyReplayLink.value = copyText.href;
+			dummyReplayLink.style = "absolute; left: -1000px; top: -1000px";
```
Zarel, line 2760:
```suggestion
			dummyReplayLink.style = "position: absolute; left: -1000px; top: -1000px";
```

Does this actually support being set to a string, though?

varun1729:
> Can you provide a screenshot of what this will look like?
Screen Shot 2020-08-13 at 2 49 33 PM

> Also, I don't think it's necessary to destroy the copied button after click - maybe a person copies the replay, pastes it somewhere, and wants to do it again later. If you want to indicate that it's finished copying, maybe transition to a "[checkmark] Copied!" sort of thing?
The copy button is not destroyed. It create a short lived DOM element, an input element, and writes the link to it. The input field's value is copied to clipboard. Then the input element element is deleted. I tested it out, and you can copy multiple times.
I am not sure, if we need to communicated that link has been copied by setting the button to 'Copied'. Then what happens if we the users copies the link, copies something else and wants to copy this link again? Copy is guaranteed to always succeed, so the confirmation doesn't seem necessary. I may be wrong about this.

js/client.js
varun1729: > Does this actually support being set to a string, though?

Yeah it does. See [here](https://gomakethings.com/two-ways-to-set-an-elements-css-with-vanilla-javascript/). You can provide a semicolon delimited list of attributes `.style`.

Zarel: Good to know!

(You forgot the `position:` part.)

Zarel: Also your link doesn't actually say it's possible to provide a semicolon-delimited list like that.

Zarel: MDN says you can't:

https://developer.mozilla.org/en-US/docs/Web/API/ElementCSSInlineStyle/style

> Styles should not be set by assigning a string directly to the `style` property

I presume you probably want to use `setAttribute`, or by setting individual styles.

varun1729: Okay got it. Thanks @Zarel. I have fixed it now.

DaWoblefet:
That makes more sense, as I was misunderstanding what you were trying to do earlier. Looks good to me!

js/client.js
```diff
@@ -2741,15 +2741,27 @@ function toId() {
 		initialize: function (data) {
 			var buf = '';
 			buf = '<p>Your replay has been uploaded! It\'s available at:</p>';
-			buf += '<p><a href="https://' + Config.routes.replays + '/' + data.id + '" target="_blank" class="no-panel-intercept">https://' + Config.routes.replays + '/' + data.id + '</a></p>';
-			buf += '<p><button class="autofocus" name="close">Close</button></p>';
+			buf += '<p> <a id="replayLink" href="https://' + Config.routes.replays + '/' + data.id + '" target="_blank" class="no-panel-intercept">https://' + Config.routes.replays + '/' + data.id + '</a></p>';
+			buf += '<p><button class="autofocus" name="close">Close</button> <button name="copyReplayLink">Copy</button><p>';
```
Zarel, line 2745:
```suggestion
			buf += '<p> <a id="replayLink" href="https://' + Config.routes.replays + '/' + data.id + '" target="_blank" class="no-panel-intercept">https://' + Config.routes.replays + '/' + data.id + '</a> <button name="copyReplayLink">Copy</button></p>';
			buf += '<p><button class="autofocus" name="close">Close</button><p>';
```

I think putting Copy and Close on the same row will screw with muscle memory. This should be safer.

js/client.js
```diff
@@ -2741,15 +2741,27 @@ function toId() {
...
 			this.$el.html(buf).css('max-width', 620);
 		},
 		clickClose: function () {
 			this.close();
 		},
 		submit: function (i) {
 			this.close();
+		},
+		copyReplayLink: function () {
+			var copyText = document.querySelector("#replayLink");
```
Zarel, line 2755:
```suggestion
			var copyText = this.$("#replayLink")[0];
```

js/client.js
```diff
@@ -2741,15 +2741,27 @@ function toId() {
...
+			var dummyReplayLink = document.createElement("input");
+			// This is a hack. You can only "select" an input field. The trick is to create a short lived input element and destroy it after a copy.
+			dummyReplayLink.id = "dummyReplayLink";
+			dummyReplayLink.value = copyText.href;
+			dummyReplayLink.setAttribute("style", "absolute; left: -1000px; top: -1000px");
```
Zarel, line 2760:
```suggestion
			dummyReplayLink.style.position = 'absolute';
```

You still can't set `style="absolute"`, you have to set `style="position: absolute"`.

All this is academic since we remove the child immediately after appending it. There's only a single setting that would have any effect, and that's `position`, so the rest of the layout doesn't need to be deformed in certain rendering engines.

varun1729:
Thanks @Zarel for helping me.

Zarel:
This looks good now. Thank you!

PR #1595, opened 2020-08-16
Fix type/BP display of -ate abilities + Z/Max

DaWoblefet:
Previously, Z-moves and Max Moves were not displaying the correct typing / base power in move tooltips; this PR fixes bugs such as https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8505457 or Pixilate Z-Hyper Voice being displayed as Fairy-type.


Zarel:
Thanks!

PR #1596, opened 2020-08-16
Fix Z-move BP display in tooltips

DaWoblefet:
Currently, Z-move BP is not displayed because of some additional check in the if statement. Removing it fixes the issue. @KrisXV do you know why it was originally added? 

Fixes bugs such as https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8460441


KrisXV:
no

Zarel:
Thanks!

PR #1597, opened 2020-08-16
UI for persistent user language setting

AnnikaCodes:
Client support for smogon/pokemon-showdown#7220.

### DIFF
--- a/js/client-topbar.js
+++ b/js/client-topbar.js
@@ -442,6 +442,7 @@
 			'change input[name=bwgfx]': 'setBwgfx',
 			'change input[name=nopastgens]': 'setNopastgens',
 			'change select[name=tournaments]': 'setTournaments',
+			'change select[name=language]': 'setLanguage',
 			'change input[name=blockchallenges]': 'setBlockchallenges',
 			'change input[name=blockpms]': 'setBlockpms',
 			'change input[name=inchatpm]': 'setInchatpm',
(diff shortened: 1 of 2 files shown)


js/client-topbar.js
```diff
@@ -501,6 +502,17 @@
 				buf += '<p><label class="optlabel"><input type="checkbox" name="temporarynotifications"' + (Dex.prefs('temporarynotifications') ? ' checked' : '') + ' /> Notifications disappear automatically</label></p>';
 			}
 			buf += '<p><label class="optlabel"><input type="checkbox" name="refreshprompt"' + (Dex.prefs('refreshprompt') ? ' checked' : '') + '> Prompt on refresh</label></p>';
+			var curLang = toID(Dex.prefs('serversettings').language) || 'english';
+			var possibleLanguages = [
+				"English", "Dutch", "French", "German", "Hindi", "Italian", "Japanese", "Portuguese", "Simplified Chinese",
+				"Spanish", "Traditional Chinese", "Turkish",
+			];
```
Zarel, line 509: Should we use language codes instead of names? It also seems weird to have all the language names in English rather than in their own language.

AnnikaCodes: hmm, I agree with it being odd that they're in English. Does the average PS user know that, for example, `en` means English, though? or that `es` means Spanish?

Zarel: Oh I'm just talking about the variable for `Dex.prefs('serversettings').language`.

Also can we stop using `Dex.prefs`? It's a minor thing, but `Dex.prefs` is an access function for the replay player (because prefs aren't available in replays etc). In PS itself, we use `Storage.prefs` which supports getting as well as setting.

AnnikaCodes: That'd involve a server-side change, since the `User#language` attribute is what eventually gets stored in `'serversettings'`. I think that having it be a different identifier from the translations data map (or changing the translation map) would be more trouble than it's worth, but I can if you'd like.

Zarel: Hm. I suppose we can just add a map instead of using `toID`, which would let us use the real language names.

Zarel: (Where by "map" I mean "object", not a literal `Map`)

AnnikaCodes: Sure, that'd work.

PR #1598, opened 2020-08-17
Show items revealed by Poltergeist in tooltips

KrisXV:


KrisXV:
This will require a client rebuild with the next server restart or with the next formats hotpatch

src/battle-text-parser.ts
```diff
@@ -136,7 +136,7 @@ class BattleTextParser {
 				return {args: ['-fieldactivate', effect], kwArgs: {}};
 			}
 
-			if (id === 'symbiosis' || id === 'poltergeist') {
+			if (id === 'symbiosis') {
```
urkerab, line 139: You might want to avoid breaking old replays. This would also mean that the client change can be merged first, and then the server change later.

```suggestion
			if (id === 'poltergeist') {
				return {args: ['-item', arg3], kwArgs: {from: effect}};
			}

			if (id === 'symbiosis') {
```

data/text.js
```diff
@@ -732,7 +732,7 @@ exports.BattleText = {
 		removeItem: '#bugbite',
 	},
 	poltergeist: {
-		activate: "  [POKEMON] is about to be attacked by its [ITEM]!",
+		item: "  [POKEMON] is about to be attacked by its [ITEM]!",
```
urkerab, line 735: (Could use `addItem` perhaps, and avoid the special-case?)

src/battle.ts
```diff
@@ -2077,6 +2077,9 @@ class Battle {
 					poke.itemEffect = 'bestowed';
 					this.scene.resultAnim(poke, item.name, 'neutral');
 					break;
+				case 'poltergeist':
+					poke.itemEffect = 'poltergeisted';
```
urkerab, line 2081: (Going by the German origin of the word, the Geist here is the Ghost-type Pokémon, and the Polter actually refers to the sounds of the objects (typically chains) being moved. Maybe "disturbed" would be a better choice?)

data/text.js
Zarel: I think the fact that we're suggesting `addItem` clarifies that we're using the incorrect message type, and we should be using `activate` because no item is being added.

src/battle-text-parser.ts
Zarel: In general, we always want to avoid breaking replays unless there's a good reason otherwise.

src/battle.ts
KrisXV: I went with disturbed. thanks!

Zarel:
Neat! I'm glad you managed to figure htis out.

PR #1601, opened 2020-08-24
Clarify config whitelist settings

DaWoblefet:
Changes will need to be reflected in an update to config.js.

https://victoryroadvgc.com/ - very popular and influential VGC resource hub; manage streams, host team reports, provide samples / Speed tiers etc.

https://vgcstats.com/ - compiles IRL VGC stats. Not as useful with quarantine and all, but has been a good resource for IRL usage stats at a glance.

https://pikalytics.com/ - combines Battle Stadium and Showdown VGC usage stats. Supports a damage calculator and custom teambuilder.

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -5,127 +5,14 @@ var Config = Config || {};
 Config.bannedHosts = ['cool.jit.su', 'pokeball-nixonserver.rhcloud.com'];
 
 Config.whitelist = [
-	// general sites
 	'wikipedia.org',
-	'wikimedia.org',
-	'wiktionary.org',
-	'github.com',
-	'reddit.com',
-	'gamefaqs.com',
-	'facebook.com',
-	'fbcdn.net',
-	'twitter.com',
-	'tumblr.com',
-	'deviantart.com',
-	'youtube.com',
-	'youtu.be',
-	'zombo.com',
-	'strawpoll.me',
-	'twitch.tv',
-	'take-a-screenshot.org',
-	'myanimelist.net',
-	'4chan.org',
-	'tumblr.com',
-	'git.io',
-	'mibbit.com',
-	'codecademy.com',
-	'xkcd.com',
-	'stackoverflow.com',
-	'stackexchange.com',
-	'malwarebytes.org',
-	'animenewsnetwork.com',
-	'animenewsnetwork.cc',
-	'zombo.com',
-	'html5zombo.com',
-	'whatismyipaddress.com',
 
-	// pokemon sites
-	'pokemonshowdown.com',
-	'psim.us',
-	'smogon.com',
-	'upokecenter.com',
-	'veekun.com',
-	'bulbagarden.net',
-	'serebii.net',
-	'nuggetbridge.com',
-	'pokecommunity.com',
-	'pokemon-online.eu',
-	'pokemonlab.com',
-	'shoddybattle.com',
-	'pokemonxy.com',
-	'pokemon.com',
-	'pokemon-gl.com',
-	'pokecheck.org',
-	'projectpokemon.org',
-	'pokemondb.net',
-	'pokemoncentral.it',
-	'poketrade.us',
-	'neverused.net',
-	'pokestrat.com',
-	'pokestrat.io',
-	'spo.ink',
-	'jooas.com',
-	'pokemongodb.net',
-	'pokeassistant.com',
-	'pokemon-sunmoon.com',
-	'gamepress.gg',
-	'trainertower.com',
-	'pokepast.es',
-	'pokepedia.fr',
-	'randbatscalc.github.io',
-	'ruins-of-alph.github.io',
+	// The full list is maintained outside of this repository so changes to it
+	// don't clutter the commit log. Feel free to copy our list for your own
+	// purposes; it's here: https://play.pokemonshowdown.com/config/config.js
 
-	// personal sites
-	'breakdown.forumotion.com',
-	'pokemonmillennium.net',
-	'thebattletower.no-ip.org',
-	'meltsner.com',
-	'guangcongluo.com',
-	'cathyjf.com',
-	'xiaotai.org',
-	'xfix.pw',
-	'pkmn.cc',
-	'bumba.me',
-	'strategydatabase.jimdo.com',
-	'hidden50.github.io',
-	'krisxv.github.io',
-	// personal hosting sites
-	'forumieren.com',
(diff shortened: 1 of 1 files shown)


Zarel:
I can add these myself, but I'd prefer not to have to maintain a config-example that isn't even being used. If we want the whitelist to be versioned, it should probably be moved out of config-example.

Zarel:
Maybe the correct answer is to put config.js in version control, and force everyone to fork it for their config changes?

scheibo:
> Maybe the correct answer is to put config.js in version control, and force everyone to fork it for their config changes?

This seems like its going to run into the exact same problems as the other open PR about formats.ts constantly causing conflicts.

Zarel:
The difference is that no one should be forking the client.

scheibo:
???

You've got tens of developers maintaining client forks currently (though most are unlikely to touch config.js), but @AnnikaCodes  specifically did work to make hosting a standalone client feasible and this would backtrack on that for no reason?

Zarel:
I mean we're still publicly not providing documentation for rehosting the client. As far as I'm aware, client forks are mostly using testclient rather than rehosting it, and rehosts are expected to be sophisticated/rare enough to keep up with stuff like config changes.

AnnikaCodes:
Personally I'm willing to deal with merge conflicts on my client host but I do think that it would make more sense to at least split off the part of the config file we want to check in, especially since we're going to try to have a sandbox setup. We obviously wouldn't publish any documentation about the client or loginserver, and AFAIK the security concerns that led to the lack of documentation were regarding the loginserver, which uses another config file, rather than the JavaScript browser client.

samsamson33:
If config.js was just `module.exports = require("./config-example.js");` by default, then you allow people to make changes to their config without causing conflicts while still allowing changes to the default config with version control. Would something along those lines work?

Zarel:
That exact change wouldn't work, but we could certainly do something _like_ that.

DaWoblefet:
Talking with Annika, I moved the whitelist to a JSON file and it's now just part of the build script. Of note, JSON.stringify() strips out spaces which eslint doesn't like in arrays; should I do anything about that?

AnnikaCodes:
> [12:59:16] @Annika: lol eslint should ignore config.js

Why isn't eslint respecting the gitignore? It should only check config-example, right?

DaWoblefet:
Hmm actually maybe it's VS Code being weird - if I run ``npm run-script lint`` it doesn't throw any issues with the config file (though it does with a line-limit for ``src/battle-dex-search.ts``) so I think it's probably fine.

AnnikaCodes:
> Hmm actually maybe it's VS Code being weird - if I run `npm run-script lint` it doesn't throw any issues with the config file (though it does with a line-limit for `src/battle-dex-search.ts`) so I think it's probably fine.

If you open a `.js` file it'll try to lint it but if you close it the error should go away.

Zarel:
> Why isn't eslint respecting the gitignore? It should only check config-example, right?

I believe `eslint` has never respected `gitignore`. I've always had to add files to both `gitignore` and `eslintignore` for them to be both ignored.

.gitignore
```diff
@@ -1,6 +1,11 @@
 /sprites/
 /audio/
 /config/
+!/config/config-example.js
+!/config/config-example.inc.php
+!/config/routes.json
+!/config/domain-whitelist.json
+!/config/servers-example.inc.json
```
Zarel, line 8: These don't need to be manually added to un-gitignore, by the way. You can just use `git add` to add a gitignored file.

DaWoblefet: Yes, but I think for domain-whitelist specifically if we want it to be versioned we should explicitly allow it to avoid confusion.

Zarel:
I've added your three sites to the config file on Main, by the way. This could take a while to figure out, and I didn't actually intend to delay that change for this.

I also added the three sites (victoryroadvgc.com, pikalytics.com, vgcstats.com) to the "pokemon sites" section, rather than the "personal sites" section. "Personal sites" I always intended as being mostly about intent, if it's a website about a developer that just happens to have PS-relevant tools, rather than a website about pokemon. It's not supposed to be about size or professionalism or anything like that.

.gitignore
DaWoblefet: I also don't really see much reason to gitignore files that are intended to be versioned?

Zarel:
As for this... I really wish there was a better solution than spreading out config over a bunch of files. I'm not even a fan of `routes.json` (which is currently versioned, anyway, which in hindsight entirely defeats the purpose of leaving config.js out of the repo).

As far as I can tell, the only reason to leave `config.js` out of the repo is to keep color changes from cluttering the commit log, and to have a place for `Config.version`.

On the other hand, the only reason to allow `config.js` _in_ the repo is to make it easier for developers without server access to add sites to the whitelist, and maybe to add color changes as well.

I think, on balance, that I'm fine doing the work of adding whitelisted websites and color changes myself, and that I'd prefer it to keep those from cluttering the commit log. So I guess the question is, what do you two feel about that? I would be willing to put the whitelist in the repo if you felt strongly about that.

DaWoblefet:
No big objections here either way. Would you like me to add a comment like "changes to whitelist domains should be PMed to Zarel, not PRed" or something to that effect?

Zarel:
If we chose to do it that way, I would remove the whitelist from `config-example.js` entirely, and I would use this phrasing:

```
Config.whitelist = [
	'wikipedia.org',

	// The full list is maintained outside of this repository so changes to it
	// don't clutter the commit log. Feel free to copy our list for your own
	// purposes; it's here: https://play.pokemonshowdown.com/config/config.js

	// If you would like to change our list, simply message Zarel on Smogon or
	// Discord.
];
```

But yes, that's simple enough for either of us to do if we chose to do it that way.

AnnikaCodes:
Not to butt in but we can un-version routes.json and add a routes-example.json if thats desirable.

DaWoblefet:
Done @Zarel, couldn't think of a great title though.

Thanks for adding those sites!

Zarel:
> Not to butt in but we can un-version routes.json and add a routes-example.json if thats desirable.

@AnnikaCodes As far as I'm aware, it would be entirely for you. I don't use routes.json at all, but I presume you do for your sandbox stuff, and the impression I got earlier was that the reason `routes.json` exists at all was so you could modify it.

So it's entirely your choice in terms of whether or not you think it's worth the effort, whether or not you want to version routes.json.

PR #1602, opened 2020-08-25
Fix broken /users/ link

cyap:
I restored the link's original destination, but I might suggest linking to `https://replay.pokemonshowdown.com/search?user=<user>` instead.


Zarel:
Thanks!

I think maybe the user page itself might want to have a link to replays.

PR #1604, opened 2020-08-26
Translate rng faq to german

xMarcell:
I translated the rng faq to german.


Zarel:
Thanks for this!

I think we should probably wait for #1605 to resolve how we want the original one to be phrased, though.

xMarcell:
Alright Zarel, tell me when #1605 has been resolved, then I can look over this translation again to check, if anything has to be changed

Zarel:
@xMarcell I think there was some miscommunication on #1605. You never responded to the suggested changes. I ended up just making them for you. Can you update this translation with those changes?

Zarel:
ty!

PR #1605, opened 2020-08-26
Improve RNG article phrasing

xMarcell:
Hey, as I translated the this site to german, I noticed that solely the 3Ds is referenced when describing the RNG used in pokemon games. Now that the switch is the main platform for console pokemon games, it would be fitting to reference it aswell; also I fixed a typo provably->probably.

### DIFF
--- a/website/pages/rng.md
+++ b/website/pages/rng.md
@@ -16,7 +16,7 @@ With every game with randomness, people will complain that the Random Number Gen
 How PS's RNG works
 ------------------
 
-Pokémon Showdown uses the exact same RNG as the actual Pokémon games on the Switch and 3DS use. You can see the source code right here:
+Pokémon Showdown uses the exact same type of RNG as the actual Pokémon games on the Switch and 3DS use. You can see the source code right here:
 
 https://github.com/smogon/Pokemon-Showdown/blob/master/sim/prng.ts
 
--- a/website/pages/rng.md
+++ b/website/pages/rng.md
@@ -4,7 +4,7 @@
 
 With every game with randomness, people will complain that the Random Number Generator (RNG) is bad, or giving them bad numbers. And there's a huge list of reasons why, if you play Pokémon on a simulator, it's even worse:
 
-1. Games go way faster on a simulator than on the 3DS, so you play more often and encounter more randomness.
+1. Games go way faster on a simulator than on the Switch or 3DS, so you play more often and encounter more randomness.
 
 2. There are more people playing. We get over 500,000 battles every day. A one-in-a-million event happens every other day! Even if it's incredibly unlikely, it'll still happen all the time, just because of the numbers.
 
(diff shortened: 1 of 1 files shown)


PartMan7:
Doesn't the provably mean `and it can be proved that it generates the exact same numbers`, or have I been reading it wrong the whole time?

DaWoblefet:
The provably wasn't a typo - it's meant to be there. It's meant to mean something like "you can demonstrate it yourself that the source code used in battle is identical to what we have up on github".

That being said, a bit unrelated to this PR, more generally I'm not a fan of advertising that "provably generate the exact same numbers as the 3DS itself does" or that PS "uses the exact same RNG as the actual Pokémon games on the 3DS use". It's not as if we have any serious modern gen battle RNG documentation; I think we should be making more modest claims.

Zarel:
We can swap "provably" with "can be proven" to be clearer, I suppose.

I also wouldn't mind changing up the phrasing to "the RNG used in Pokémon games to the best of our knowledge" – but I think people have datamined modern games enough that we'd know if they were using a different RNG.

xMarcell:
Thanks Partman7, now I see it aswell, of course it derives from prove, completely missed it.

website/pages/rng.md
```diff
@@ -32,7 +32,7 @@ If you think we're using different source code than the code above, you can alwa
 
 After a battle has ended, saving its replay will also save its input log containing its random seed. For random battles, you can get the log directly from the HTML source of the replay page. For battles with non-random teams, seeds are not visible for privacy reasons – you will need to get permission from your opponent and then use the `/exportinputlog` command.
 
-Once you have the log, you can replay it using the publicly available source from GitHub, and notice that it will provably generate the exact same numbers as the 3DS itself does.
+Once you have the log, you can replay it using the publicly available source from GitHub, and notice that it will probably generate the exact same numbers as the Switch and 3DS themselves do.
```
Zarel, line 35:
```suggestion
Once you have the log, you can replay it using the publicly available source code from GitHub, which proves that it's the same RNG as in the source code, which you can read to see that it's the same RNG as the Switch and 3DS.
```

Zarel:
I think it might also be a good idea to just mention the DS, because apparently that's the last one where we're sure of how the RNG works. But it's still surprising to me that we don't know anything about RNGs for later games. Are you _sure_, @DaWoblefet?

DaWoblefet:
Well, at least I don't know of _any_ public information on battle RNG since Gen 5. The best we've got is people like Kaphotics or Anubis hacking the current RNG state to produce funny results on social media, which doesn't really tell you much about how the game actually rolls the numbers. The way it's currently worded, it sounds like it should be possible to take an inputlog and starting seed, load the starting seed into a Battle Video, choose the same moves as the inputlog, and produce the same battle as what Showdown provides, which I doubt is the case. For example, it'd be correct to say that Showdown picks 1 of 16 possible damage rolls at random. But I'm not sure we can assert that given an identical starting seed, Showdown's RNG will produce an identical damage roll to the roll the game picks, and various things like that. We can say that the chance is the same (1/16) but not that we'll produce the same 1/16. We definitely can't say that for things like accuracy/evasion, where our best knowledge of how they're calculated comes from Gen 4.

I think it'd be more accurate to say we're using the "exact same type of RNG", rather than "exact same RNG". Or that we're "simulating the battle RNG of the Pokemon games", things like that.

Zarel:
I'd support "exact same type of RNG".

website/pages/rng.md
```diff
@@ -16,11 +16,11 @@ With every game with randomness, people will complain that the Random Number Gen
 How PS's RNG works
 ------------------
 
-Pokémon Showdown uses the exact same RNG as the actual Pokémon games on the 3DS use. You can see the source code right here:
+Pokémon Showdown uses the exact same RNG as the actual Pokémon games on the Switch and 3DS use. You can see the source code right here:
```
Zarel, line 19:
```suggestion
Pokémon Showdown uses the exact same type of RNG as the actual Pokémon games on the Switch and 3DS use. You can see the source code right here:
```

(Better phrasing, as agreed upon with DaWoblefet.)

Zarel:
This has gone unmaintained, so I've finished it up myself. Thanks for this!

PR #1609, opened 2020-08-30
User settings: Support new blockpms format

mia-pi-git:
Will transform legacy boolean values stored in the client to match the new Object (`{all: AuthLevel, specific: []}`) type of `blockPMs` on the server (when server PR #7191 is merged.)

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -194,7 +194,9 @@ function toId() {
 			if (settings[setting] !== value) {
 				switch (setting) {
 				case 'blockPMs':
-					app.send(value ? '/blockpms' : '/unblockpms');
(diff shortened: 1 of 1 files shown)


js/client.js
```diff
@@ -194,7 +194,14 @@ function toId() {
 			if (settings[setting] !== value) {
 				switch (setting) {
 				case 'blockPMs':
-					app.send(value ? '/blockpms' : '/unblockpms');
+					// handle new blockpms format
+					if (typeof value === 'object') {
```
scheibo, line 198: `value` here comes from the topbar, which can only ever be `boolean`

scheibo (approved):

scheibo:
Thanks!

PR #1613, opened 2020-09-03
Website: Fix lock duration in the Staff FAQ

AnnikaCodes:
As per [this bug report](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-27#post-8586989).


Zarel:
Thanks.

PR #1614, opened 2020-09-03
Website: Fix bug in usersearch

AnnikaCodes:
`$ip` isn't defined here - the correct behavior is to search by the userlist we just fetched.

### DIFF
--- a/website/usersearch.php
+++ b/website/usersearch.php
@@ -173,8 +173,8 @@ if (!$ip && !$entry) {
 		$newStanding = intval($_POST['standing']);
 
 		$psdb->query(
-			"UPDATE {$psdb->prefix}users, {$psdb->prefix}usermodlog INNER JOIN " +
-			"{$psdb->prefix}usermodlog ON {$psdb->prefix}usermodlog.userid = {$psdb->prefix}users.userid " +
+			"UPDATE {$psdb->prefix}users, {$psdb->prefix}usermodlog " +
+			"INNER JOIN {$psdb->prefix}usermodlog ON {$psdb->prefix}usermodlog.userid = {$psdb->prefix}users.userid " +
 			"SET {$psdb->prefix}users.banstate = ? WHERE {$psdb->prefix}usermodlog.entry LIKE ? AND banstate != 100",
 			[$newStanding, $entry]
 		);
--- a/website/usersearch.php
+++ b/website/usersearch.php
@@ -172,7 +172,7 @@ if (!$ip && !$entry) {
 	if ($csrfOk && isset($_POST['standing'])) {
 		$newStanding = intval($_POST['standing']);
 
-		$users = array();
+		$users = [];
 		foreach ($userlist as $row) {
 			$modlogentry = "Standing changed to $newStanding ({$STANDINGS[$newStanding]}): {$_POST['allreason']}";
 			$psdb->query(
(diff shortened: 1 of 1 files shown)


website/usersearch.php
```diff
@@ -171,7 +171,10 @@
 	$userlist = $psdb->query("SELECT `ntbb_users`.`username`, `ntbb_users`.`userid`, `ntbb_users`.`banstate` FROM `ntbb_users` INNER JOIN `ntbb_usermodlog` ON `ntbb_users`.`userid` = `ntbb_usermodlog`.`userid` WHERE `entry` LIKE ?", [$entry]);
 	if ($csrfOk && isset($_POST['standing'])) {
 		$newStanding = intval($_POST['standing']);
-		$psdb->query("UPDATE ntbb_users SET banstate = ? WHERE ip = ? AND banstate != 100", [$newStanding, $ip]);
+		$psdb->query(
+			"UPDATE ntbb_users SET banstate = ? WHERE userid IN (" . implode(',', $userlist) . ") AND banstate != 100",
```
Zarel, line 175: Err, can `$userlist` be used this way? Isn't it a database resource, not an array? I also don't think it's properly escaped...

AnnikaCodes: Oh, yeah, I meant to convert it to an array; I'll do that now.

website/usersearch.php
```diff
@@ -171,16 +171,22 @@
 	$userlist = $psdb->query("SELECT `ntbb_users`.`username`, `ntbb_users`.`userid`, `ntbb_users`.`banstate` FROM `ntbb_users` INNER JOIN `ntbb_usermodlog` ON `ntbb_users`.`userid` = `ntbb_usermodlog`.`userid` WHERE `entry` LIKE ?", [$entry]);
 	if ($csrfOk && isset($_POST['standing'])) {
 		$newStanding = intval($_POST['standing']);
-		$psdb->query("UPDATE ntbb_users SET banstate = ? WHERE ip = ? AND banstate != 100", [$newStanding, $ip]);
 
+		$users = array();
```
Zarel, line 175:
```suggestion
		$users = [];
```

website/usersearch.php
```diff
@@ -171,16 +171,22 @@
...
 		foreach ($userlist as $row) {
 			$modlogentry = "Standing changed to $newStanding ({$STANDINGS[$newStanding]}): {$_POST['allreason']}";
 			$psdb->query(
 				"INSERT INTO `{$psdb->prefix}usermodlog` (`userid`,`actorid`,`date`,`ip`,`entry`) VALUES (?, ?, ?, ?, ?)",
 				[$row['userid'], $curuser['userid'], time(), $users->getIp(), $modlogentry]
 			);
+			array_push($users, $psdb->escape($row['userid']));
```
Zarel, line 182:
```suggestion
			$users []= $psdb->escape($row['userid']);
```

Zarel:
Is there any reason why we can't define `$ip` here? I'm not a big fan of manually constructing an unbounded string to be executed from an array, even if there are no longer any security issues.

Zarel:
(Sorry, last time I reviewed I was still confused about exactly what this was doing any why; I understand what this is doing now.)

AnnikaCodes:
> Is there any reason why we can't define `$ip` here? I'm not a big fan of manually constructing an unbounded string to be executed from an array, even if there are no longer any security issues.

Why would we define `$ip`? As far as I know, this code is searching by modlog note, not by IP address?

Zarel:
This code is designed to change the standing of all users with the same registration IP at once; to permalock a user with a lot of alts.

Zarel:
(It's also been working fine this entire time, so I'm also still confused about what bug you're seeing.)

AnnikaCodes:
I haven't seen a bug personally - it was reported by pre and @asgdf.
image

Zarel:
Oh wait I did get it mixed up with the one I wrote.

I wonder if you should just write a new query for that one.

Zarel:
I think you can do this with a JOIN.

AnnikaCodes:
Alright.

Zarel:
Thanks! I like this way a lot better.

PR #1616, opened 2020-09-11
Rename mia-pi in Credits

mia-pi-git:


Zarel:
You still use `mia-pi-git`, and it already lists your first name as "Mia". I think it's less confusing if you leave it alone? Not that it's a big deal to me, though.

mia-pi-git:
> You still use `mia-pi-git`, and it already lists your first name as "Mia". I think it's less confusing if you leave it alone? Not that it's a big deal to me, though.

I don't know if i'm going to _keep_ mia-pi-git, i'm only not changing it in server because i don't wanna jump the gun. I feel like it's a better interrim here to just use my PS name.

Zarel:
Sure.

PR #1619, opened 2020-09-19
Fix format display for hidden battles

AnnikaCodes:
This was meant to be done with my original battle privacy PR but I forgot :/


Zarel:
I hope this doesn't break something...

Zarel:
Oh! I remember why it was like this before! Replays from side servers!

Zarel:
`battle-smogtours-ou-589172398528`

AnnikaCodes:
Doesn't it already handle uploaded replays in a special case?

Zarel:
Yes, it does, this is the special case...

PR #1620, opened 2020-09-20
Ensure that the BGM restarts when starting a new battle

urkerab:
I don't claim to understand the old BGM code, but somehow the BGM would always restart when starting a new battle. This doesn't happen with the new BGM code from PR #1563.
- When starting a new battle, we don't have a sound yet. So although we ask to reset it, nothing happens.
- When stopping a sound instance, if it's never played, it won't get rewound.

This changes fixes the second part, and indirectly the first part by making new sound objects always rewind, but if you don't want that, I can fix the first by part instead by changing `src/battle-animations.ts` something like this:
```diff
	resetBgm() {
-		if (this.bgm) this.bgm.stop();
+		if (!this.bgm) this.rollBgm();
+		this.bgm!.stop();
	}
```

### DIFF
--- a/src/battle-sound.ts
+++ b/src/battle-sound.ts
@@ -17,13 +17,17 @@ class BattleBGM {
 	 */
 	isPlaying = false;
 	isActuallyPlaying = false;
+	/**
+	 * The sound should be rewound when it next plays.
+	 */
+	willRewind = true;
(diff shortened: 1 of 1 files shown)


Zarel:
Hm. I never like adding a new variable for this sort of thing, but I can't think of any better way to do it. :s

Zarel:
Thanks!

PR #1622, opened 2020-09-23
Add some weather tool tips

AustinXII:


Zarel:
I believe the reason these aren't already here is because they're not implemented as base power modifiers.

I guess it's an open question whether our mouseovers should be "accurate" and "base power modifiers only". It might be worth dropping that in favor of being more informative. But I think if it does happen, it should be as a "power rating" thing.

Zarel:
I've been thinking about displaying pokemon/attacks in terms of "Power rating" and "Bulk rating".

"Power rating" would be the percentage damage against a neutral enemy with 100/100 defenses, shown for every attack.

And then "bulk rating" would be how bulky you are, as a percentage relative to a pokemon with 100/100 defenses, shown for your own pokemon and given a range for opponents.

So if your power rating was higher than your opponent's bulk rating, it would be a 1HKO. Otherwise it wouldn't. Simple and informative.

I never got around to actually _implementing_ that, but I think it could be a pretty cool project for someone.

Zarel:
Actually, I'm going to post that in Suggestions. Until then, this will... be in limbo unless someone has strong opinions about whether or not we should "lie" about base power. I assume @Marty-D would object.

KrisXV:
We already lie about base power with burns because there’s no better way to represent damage modifiers

DaWoblefet:
I personally don't like representing modifiers in ways that they're not; e.g. I don't think we should represent super-effectiveness by anything more than a "super-effective" or "not very effective" indicator, and the same would apply for weather, or Life Orb, or spread moves, etc. The problem is that there's just no way to _accurately_ represent "this is an increase in damage", being technically precise, while not representing it as a multiplier of BP or Atk.

Marty-D:
I object to showing halved Attack for burn, too, but I don't usually deal with handholdy tooltips so follow your heart here.

The problem is, yes, it spreads misinformation. We sometimes get bug reports from people who are actually informed, explaining how wrong some tooltips are. You'll remember there was huge pushback for implementing STAB misleadingly, for example. The users who aren't informed just take these things as fact because they aren't going to read any relevant text explaining it.

Zarel:
As expected, this ended up being controversial. Should we punt until damage rating / bulk rating is implemented?

PR #1624, opened 2020-10-04
Correctly pluralize "users" in the userlist

AnnikaCodes:


js/client-chat.js
```diff
@@ -1730,8 +1730,12 @@
 			'click .userlist-count': 'toggleUserlist'
 		},
 		construct: function () {
+			var plural = this.room.userCount.users === 1 ? ' user' : ' users';
 			var buf = '';
-			buf += '<li class="userlist-count" id="' + this.room.id + '-userlist-users" style="text-align:center;padding:2px 0"><small><span id="' + this.room.id + '-usercount-users">' + (this.room.userCount.users || '0') + '</span> users</small></li>';
+			buf += '<li class="userlist-count" id="' + this.room.id + '-userlist-users" style="text-align:center;padding:2px 0"><small>';
+			buf += '<span id="' + this.room.id + '-usercount-users">' + (this.room.userCount.users || '0') + '</span>';
+			buf += '<span id="' + this.room.id + '-usercount-plural">' + plural + '</span></small></li>';
```
Zarel, line 1737:
```suggestion
			var usersString = "" + (this.room.userCount.users || '0') + plural;
			buf += '<li class="userlist-count" id="' + this.room.id + '-userlist-users" style="text-align:center;padding:2px 0">';
			buf += '<small id="' + this.room.id + '-usercount-users">' + usersString + '</small></li>';
```

I think I'd rather just have this all be updated as one thing, rather than two things. RIP to userscripts depending on the previous layout but perf is important!

Zarel:
Thanks!

PR #1625, opened 2020-10-06
Improve errors when using room commands in PMs

AnnikaCodes:
Ideally we can refactor commands to specify whether they apply in PMs, rooms, or both. Is that something that Preact is doing?

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -407,16 +407,29 @@
 					app.addPopup(LoginPopup);
 				}
 				return false;
+			case 'chal':
+			case 'chall':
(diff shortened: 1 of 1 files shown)


Zarel:
It's not currently being done in the Preact client. I'll let you decide if it should be.

Zarel:
(Also thanks)

AnnikaCodes:
I think it should be; a lot of these commands could work just as well in PMs (e.g. `/user` and `/highlight`). It's not a huge deal though since people can join a room.

Zarel:
What I mean is, feel free to implement a system like this into the Preact client.

PR #1626, opened 2020-10-07
Support temporarily turning off notifications with a command

mia-pi-git:
This seems to be a thing people want, and it's not overly difficult so here we are.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -575,18 +575,18 @@
 			case 'disablenotifications':
 				var command = cmd.toLowerCase();
 				if (command.includes('disable')) {
-					if (Dex.prefs('nonotifs')) {
+					if (Storage.prefs('nonotifs')) {
 						this.add("|error|You already have notifications disabled.");
 						return false;
 					}
-					Dex.prefs('nonotifs', true);
+					Storage.prefs('nonotifs', true);
 					this.add("You will now not receive notifications.");
 				} else if (command.includes('enable')) {
-					if (!Dex.prefs('nonotifs')) {
+					if (!Storage.prefs('nonotifs')) {
 						this.add("|error|You are already receiving notifications.");
 						return false;
 					}
-					Dex.prefs('nonotifs', false);
+					Storage.prefs('nonotifs', false);
 					this.add("You will now receive notifications.");
 				}
 				return false;
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -572,14 +572,14 @@
 			case 'notifications':
 			case 'notifs':
 				target = toID(target);
-				if (target === 'off') {
+				if (target === 'off' || target === 'false') {
 					if (Storage.prefs('nonotifs')) {
 						this.add("|error|You already have notifications disabled.");
 						return false;
 					}
 					Storage.prefs('nonotifs', true);
 					this.add("You will now not receive notifications.");
-				} else if (target === 'on') {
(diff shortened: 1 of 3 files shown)


js/client-chat.js
```diff
@@ -569,7 +569,27 @@
...
+						return false;
+					}
+					Dex.prefs('nonotifs', true);
+					this.add("You will now not receive notifications.");
+				} else if (command.includes('enable')) {
+					if (!Dex.prefs('nonotifs')) {
+						this.add("|error|You are already receiving notifications.");
+						return false;
+					}
+					Dex.prefs('nonotifs', false);
```
Zarel, line 589: This doesn't do anything, does it?

Zarel: Huh, when did everywhere start using Dex.prefs?

mia-pi-git: > This doesn't do anything, does it?

It does, in fact, set the value to `false` in storage. Dunno about the second part.

mia-pi-git: Saw your commit. Changing now.

Zarel:
I wonder if this should be `/notifs on` and `/notifs off`. A lot of our commands would probably be better implemented this way rather than as a pair... :/ Maybe a bit late for that, though.

mia-pi-git:
> I wonder if this should be `/notifs on` and `/notifs off`. A lot of our commands would probably be better implemented this way rather than as a pair... :/ Maybe a bit late for that, though.

Done.

js/client-chat.js
```diff
@@ -569,7 +569,27 @@
 			case 'query':
 				this.add('This is a PS system command; do not use it.');
 				return false;
-
+			case 'notifications':
+			case 'notifs':
+				target = toID(target);
+				if (target === 'off') {
```
Zarel, line 575:
```suggestion
				if (target === 'off' || target === 'false') {
```

js/client-chat.js
```diff
@@ -569,7 +569,27 @@
...
+					if (Storage.prefs('nonotifs')) {
+						this.add("|error|You already have notifications disabled.");
+						return false;
+					}
+					Storage.prefs('nonotifs', true);
+					this.add("You will now not receive notifications.");
+				} else if (target === 'on') {
```
Zarel, line 582:
```suggestion
				} else if (target === 'on' || target === 'true') {
```

Zarel:
This is no notifications at all, not even the glowy panels or anything, btw. Are you sure this is what you want?

mia-pi-git:
> This is no notifications at all, not even the glowy panels or anything, btw. Are you sure this is what you want?

That was the intent. I can certainly change it to have the glowy panels, though.

Zarel:
Was there a suggestion thread? Maybe we should make one. I'd feel better about these sorts of smaller decisions if I could understand why people want this.

Zarel:
Turns out it was here: https://www.smogon.com/forums/threads/ability-to-turn-hls-on-and-off.3665267/

Zarel:
As discussed in Discord DMs, I'd ideally like some way to notify the user that they have notifications disabled and provide a button to turn it back on.

Showing a notification that a comment was ignored has been moderately unpopular, so I suggested instead just a window below News. You mentioned it would probably be difficult but this is a lot of my frustration with adding more and more features which will need to be rewritten for Preact, instead of just working on Preact. :/

PR #1627, opened 2020-10-08
Support halved toxic damage when in Dynamax form

majorlyps:


src/battle-tooltips.ts
```diff
@@ -814,7 +814,8 @@ class BattleTooltips {
 				if (pokemon.status === 'tox') {
 					if (pokemon.ability === 'Poison Heal' || pokemon.ability === 'Magic Guard') {
 						text += ' <small>Would take if ability removed: ' + Math.floor(100 / 16 * Math.min(clientPokemon.statusData.toxicTurns + 1, 15)) + '%</small>';
-					} else {
+				    } if (pokemon.dynamax) {text += ' Next damage: ' + Math.floor(100 / 8 * Math.min(clientpokemon.statusData.toxicTurns + 1, 15))
+                   else {
```
mia-pi-git, line 818:
```suggestion
				  }  else if (pokemon.dynamax) {text += ' Next damage: ' + Math.floor(100 / 8 * Math.min(clientpokemon.statusData.toxicTurns + 1, 15))
                   else {
```

AnnikaCodes (changes requested):
Can you edit the title of this PR to explain what it does, please?
Also, it's not clear what the `if` statements do — you should work on formatting those better. A newline (and indented block) should follow the `{`.

mia-pi-git:
> Can you edit the title of this PR to explain what it does, please?
> Also, it's not clear what the `if` statements do — you should work on formatting those better. A newline (and indented block) should follow the `{`.

Adding onto this - see [this](https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md) for title guidelines.

AnnikaCodes:
Also, you should probably implement this by creating an intermediary variable for the damage that Toxic would do rather than using another `if` block. That way, it'll account for the increase in HP from dynamaxing regardless of whether the Pokémon has an ability preventing Toxic damage.

src/battle-tooltips.ts
```diff
@@ -814,7 +814,8 @@ class BattleTooltips {
 				if (pokemon.status === 'tox') {
 					if (pokemon.ability === 'Poison Heal' || pokemon.ability === 'Magic Guard') {
 						text += ' <small>Would take if ability removed: ' + Math.floor(100 / 16 * Math.min(clientPokemon.statusData.toxicTurns + 1, 15)) + '%</small>';
-					} else {
+				  } else if (pokemon.dynamax) {text += ' Next damage: ' + Math.floor(100 / 8 * Math.min(clientpokemon.statusData.toxicTurns + 1, 15))
```
AnnikaCodes, line 817:
```suggestion
					} else if (pokemon.dynamax) {
						text += ' Next damage: ' + Math.floor(100 / 8 * Math.min(clientpokemon.statusData.toxicTurns + 1, 15))
```

src/battle-tooltips.ts
```diff
@@ -814,7 +814,8 @@ class BattleTooltips {
 				if (pokemon.status === 'tox') {
 					if (pokemon.ability === 'Poison Heal' || pokemon.ability === 'Magic Guard') {
```
AnnikaCodes, line 815:
```suggestion
					const damageIncreaseStep = pokemon.dynamax ? 100 / 8 : 100 / 16;
					const toxicDamage = Math.floor(damageIncreaseStep * Math.min(clientpokemon.statusData.toxicTurns + 1, 15);
					if (pokemon.ability === 'Poison Heal' || pokemon.ability === 'Magic Guard') {
```

AnnikaCodes: This is what I mean by using another variable (you need to use this in the two remaining if branches for Poison Heal/Magic Guard and regular damage).
```suggestion
					let damageIncreaseStep = 100 / 16;
					if (pokemon.dynamax) damageIncreaseStep /= 2;
					const toxicDamage = Math.floor(damageIncreaseStep * Math.min(clientpokemon.statusData.toxicTurns + 1, 15));
					if (pokemon.ability === 'Poison Heal' || pokemon.ability === 'Magic Guard') {
```

src/battle-tooltips.ts
```diff
@@ -814,7 +814,9 @@ class BattleTooltips {
...
 						text += ' <small>Would take if ability removed: ' + Math.floor(100 / 16 * Math.min(clientPokemon.statusData.toxicTurns + 1, 15)) + '%</small>';
-					} else {
+					} else if (pokemon.dynamax) {
+						text += ' Next damage: ' + Math.floor(100 / 8 * Math.min(clientpokemon.statusData.toxicTurns + 1, 15))
+                   else {
```
mia-pi-git, line 819:
```suggestion
                   else {
```
```suggestion
						text += ' Next damage: ' + Math.floor(100 / 8 * Math.min(clientpokemon.statusData.toxicTurns + 1, 15))
				} else {
```

Marty-D:
You're actually doubling Toxic damage instead of halving it. If you halve the divisor you make the number bigger, not smaller.

AnnikaCodes:
@majorlyps are you still working on this?

PR #1630, opened 2020-10-10
Teambuilder: Fix ability display for Pokemon with Special Abilities and no HA

KrisXV:


Zarel:
I honestly originally did it that way so it would be possible to tell apart special abilities and hidden abilities.

Zarel:
(It also happened to make the code more concise.)

KrisXV:
We can do what we do with special abilities in /dt and surround it in parentheses if that helps with differentiating them

Zarel:
I like that!

KrisXV:
did

PR #1632, opened 2020-10-10
Fix |tempnotify| in HTML rooms

sirDonovan:
This fixes `|tempnotify|` by parsing rows the same way as in `client-chat`.


Zarel:
Thanks!

PR #1636, opened 2020-10-12
Add a custom html tag for namecolors

mia-pi-git:
Pretty sure i did this right? Works in testing. Also not sure if the source should be `src` or just `name` - went with `src` for now.

### DIFF
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -719,7 +719,7 @@ class BattleLog {
 			'psicon::item': 0,
 			'psicon::type': 0,
 			'psicon::category': 0,
-			'username::src': 0,
+			'username::name': 0,
 			'*::aria-label': 0,
 			'*::aria-hidden': 0,
 		});
@@ -759,7 +759,7 @@ class BattleLog {
 				let name = '';
 				tagName = 'strong';
 				for (let i = 0; i < attribs.length - 1; i += 2) {
-					if (attribs[i] === 'src') {
+					if (attribs[i] === 'name') {
 						name = toID(attribs[i + 1]);
 					}
 				}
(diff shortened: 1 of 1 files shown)


Zarel:
I think ideally this should just be `<username>Zarel</username>`.

Zarel:
You could manually specify a separate `name` attribute but it should be optional.

mia-pi-git:
> I think ideally this should just be `<username>Zarel</username>`.

Is this possible in Caja, the way it's set up currently? Looking at how `BattleLog.tagPolicy` works, only the attribs (i.e `src`) are available; not the data in the tag. I may be missing something, though.

Zarel:
After a lot of googling:

1. apparently this is just entirely undocumented. feels great after all the time I was yelled at for having supposedly horrible documentation. -_-

2. looking at the actual source code, yes, it's impossible: https://npmdoc.github.io/node-npmdoc-sanitizer/build/apidoc.html#apidoc.element.sanitizer.makeHtmlSanitizer

I would suggest `<namecolor name="Zarel">Zarel</namecolor>`. It's important not to use `src` because that already has a specific meaning pointing to a URL, and usernames are obviously not URLs.

Zarel:
Well, one thing that's possible is that we could preparse with a regex

```js
str.replace(/<namecolor>([^<]*)</namecolor>/gi, '<namecolor name="$1">$1</namecolor>')
```

But probably more complicated.

Zarel:
```js
str.replace(/<namecolor([^>]*)>([^<]*)</namecolor>/gi, (match, attrs, username) => {
  if (/\bname\s*=\s*"/.test(attrs)) return match;
  const escapedUsername = username.replace(/"/g, '&quot;').replace(/>/g, '&gt;');
  return `<namecolor${attrs} name="${escapedUsername}">${username}</namecolor>`;
})
```

(This will additionally support having other attributes, and also deal with some improperly-escaped usernames.)

Zarel:
Looks good, thanks!

Zarel:
We should probably document PS custom HTML somewhere.

PR #1637, opened 2020-10-16
Support a custom html tag for youtube videos

mia-pi-git:
Not sure if whitelisting the `src` attribute is the best approach here, but since <iframe> is still banned serverside, it might be nice to have. It was certainly cleaner than anything else i could do.

### DIFF
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -804,8 +804,12 @@ class BattleLog {
 				const color = this.usernameColor(toID(getAttrib('name')));
 				const style = getAttrib('style');
 				setAttrib('style', `${style};color:${color}`);
+			} else if (tagName === 'youtube') {
+				tagName = 'iframe';
+				const id = getAttrib('video') || "";
+				setAttrib('src', id);
 			} else if (tagName === 'psicon') {
-				// <psicon> is a custom element which supports a set of mutually incompatible attributes:
+				// <psicon> is a csustom element which supports a set of mutually incompatible attributes:
 				// <psicon pokemon> and <psicon item>
 				let iconType = null;
 				let iconValue = null;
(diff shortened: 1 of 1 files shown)


src/battle-log.ts
```diff
@@ -720,6 +721,8 @@ class BattleLog {
 			'psicon::type': 0,
 			'psicon::category': 0,
 			'username::name': 0,
+			'youtube::video': 0,
+			'iframe::src': 0,
```
Zarel, line 725: Why is this necessary? "`<iframe>` is still banned serverside" is meaningless because side servers can change that code.

mia-pi-git: It's necessary because otherwise the `src` element was removed. I didn't know of another way to handle it, though i asked on Discord :p

Zarel: Your problem is that you're running `sanitizeAttribs`. Look at where `sanitizeAttribs` is executed and you should be able to figure out the correct way to do this.

Zarel (changes requested):
security vulnerability

src/battle-log.ts
mia-pi-git: Doesn't Caja do that internally?

mia-pi-git: Oh, no, i'm wrong. I'll sort this out.

src/battle-log.ts
```diff
@@ -742,6 +744,10 @@ class BattleLog {
 				}
 			}
 			let dataUri = '';
+			attribs = html.sanitizeAttribs(tagName, attribs, (urlData: any) => {
+				if (urlData.scheme_ === 'geo' || urlData.scheme_ === 'sms' || urlData.scheme_ === 'tel') return null;
+				return urlData;
+			});
```
Zarel, line 750: Moving this here ruins all the other behavior of this function (the `img` data-URI sanitization no longer works, for instance)

mia-pi-git:
I'll PM you on Discord, got a few questions that will be easier to discuss there.

PR #1639, opened 2020-10-20
Few more multi battle changes

Spandamn:


Spandamn:
idk how macs commits show up everytime lol

Zarel:
Rebasing might fix it.

Or just hard reset to the latest master.

src/battle.ts
```diff
@@ -1051,6 +1052,8 @@ class Battle {
 	yourSide: Side = null!;
 	p1: Side = null!;
 	p2: Side = null!;
+	p3?: Side = null!;
+	p4?: Side = null!;
```
Zarel, line 1056:
```suggestion
	p3: Side | null = null;
	p4: Side | null = null;
```

src/battle.ts
```diff
@@ -579,6 +579,7 @@ class Side {
 	id = '';
 	n: number;
 	foe: Side = null!;
+	ally: Side = null!;
```
Zarel, line 582:
```suggestion
	ally: Side | null = null;
```

src/battle.ts
```diff
@@ -1135,6 +1138,15 @@ class Battle {
...
 		this.p1 = this.mySide;
 		this.p2 = this.yourSide;
+		if (this.gameType === 'multi') {
+			this.p3 = new Side(this, 2);
+			this.p3.foe = this.p2;
+			this.p3.ally = this.p1;
+			this.p4 = new Side(this, 3);
+			this.p4.ally = this.p2;
+			this.p4.foe = this.p1;
+			this.sides.concat([this.p3, this.p4]);
```
Zarel, line 1148:
```suggestion
			this.sides.push(this.p3, this.p4);
```

src/battle.ts
```diff
@@ -2984,7 +2996,7 @@ class Battle {
 			let pokemon = side.pokemon[i];
 			if (pokemon.fainted) continue;
 			// already active, can't be switching in
-			if (side.active.includes(pokemon)) continue;
+			if (side.active.includes(pokemon) && side.ally) continue;
```
Zarel, line 2999: What is this intended to do? I don't think it does what it's intended to.

src/battle.ts
Zarel: Although what's this for? It doesn't appear to be used anywhere.

Zarel: Oh, nvm, I see it's set in the constructor.

src/battle.ts
Spandamn: so currently, it you cant click on your second pokemon which is not active, because it thinks its active, because your teammate has an active pokemon. adding with willl make sure you can switch to your second pokemon, since his way the client knows its an ally and wont make the second pokemon unclickable.
should probably be ```!side.ally```

Zarel: That's not what this code does...

Spandamn: cause what i believe, side.active includes your allys pokemon rn notsure

Spandamn: i study this deeper

Spandamn:
Putting everyting iin one place

PR #1641, opened 2020-10-20
(Preact) Partial drag/drop panel tab rearranging

Zarel:
This is stalled at partially done so I guess I publish what I have so far.

Of note is that mini-windows (News, PMs) can now be dragged into the tabbar, where they'll become regular tabs.

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -26,10 +26,10 @@ Dex.includeData();
 console.log("DONE");
 
 function es3stringify(obj) {
-	let buf = JSON.stringify(obj);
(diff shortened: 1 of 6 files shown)


Zarel:
The old version definitely had a lot of interesting functionality and a lot of experiences that it kind of sucks to relearn.

The HTML5 drag/drop API continues to be horrible.

But it's worth it for the new features this time around.

Zarel:
Supported:

- Rearranging the topbar
- Dragging PMs from the Main Menu to the topbar

Not supported:

- Dragging things from the topbar to the Main Menu

PR #1643, opened 2020-10-25
Add animations for Crown Tundra moves

KrisXV:
I also added an (amazing) animation for Jungle Healing and improved Life Dew's animation


Zarel:
Thanks!

(In general, we should do our best not to add new effects, because they slow down loading times.)

PR #1645, opened 2020-10-25
Support a |pagescroll| protocol message

AnnikaCodes:


js/client-ladder.js
```diff
@@ -52,6 +52,17 @@
 				this.subtleNotifyOnce();
 				break;
 
+			case 'pagescroll':
+				var page = this.$el.get()[0];
+				if (!page) break;
+
+				if (row[1] === 'top') {
+					page.scrollBy(0, -1 * page.scrollHeight);
```
Zarel, line 60: This looks like a bleeding-edge API. I would recommend using `scrollTop` instead, which should have better compatibility.

Zarel:
```suggestion
					page.scrollTop = 0;
```

js/client-ladder.js
```diff
@@ -52,6 +52,17 @@
 				this.subtleNotifyOnce();
 				break;
 
+			case 'pagescroll':
+				var page = this.$el.get()[0];
```
Zarel, line 56:
```suggestion
				var page = this.$el[0];
```

Typo?

js/client-ladder.js
```diff
@@ -52,6 +52,17 @@
...
+				if (!page) break;
+
+				if (row[1] === 'top') {
+					page.scrollBy(0, -1 * page.scrollHeight);
+				} else if (row[1] === 'bottom') {
+					page.scrollBy(0, page.scrollHeight);
```
Zarel, line 62:
```suggestion
					page.scrollTop = page.scrollHeight;
```

js/client-ladder.js
Zarel: https://github.com/smogon/pokemon-showdown-client/blob/9bf683df7be1ffb92a242410052256d49f353049/src/battle-log.ts#L328-L332

Yeah, this is how we do it in `battle-log`. It's best to be consistent.

Zarel:
I wonder if we should also support `selectorscroll` or something. Maybe also scroll to selector?

Zarel:
I'm just thinking about in the future, where we might have `scroll` equivalents of all the `html` tags. `|controlsscroll|`, `|pagescroll|`, `|fieldscroll|`, `|selectorscroll|`...

That "HTML tag" idea would certainly simplify the API...

Zarel:
A lot of the time when I say "I wonder", you usually don't reply and just implement the thing I'm wondering about, but to be clear, I'm always asking for your opinions on the matter when I say "I wonder".

AnnikaCodes:
Oh! I didn't know you were asking for my opinion. Does `|selectorscroll|` have to be its own protocol message? I feel like `|pagescroll|selector|...` would work just as well, and be more consistent (since we don't have `|topscroll|` or `|bottomscroll|`).

Zarel:
I mean, `|pagehtml|` and `|selectorhtml|` are separate messages, currently.

To be clear, I think we'd want both `|pagescroll|selector|...` (for "scroll to selector"), and `|selectorscroll|...` (for "scroll the element matched by this selector to...")

AnnikaCodes:
Oh, you meant scrolling within a selector. Yeah, I can do that.

Zarel:
Annika! I'm still trying to have a technical discussion! The question is: Do you think having `|whateverscroll|` versions of every `|whateverhtml|` message is a good idea, or should we consider maybe making it something generalizable like a fake HTML element?

I'm actually really starting to like the fake HTML element idea. Or, to be exact, a fake _attribute_. We already have `autofocus` as an attribute to mean "focus this after updating the HTML". We could add an `autoscroll` as an attribute meaning "scroll to this after updating the HTML".

AnnikaCodes:
I didn't like the idea you brought up in private of a tag that scrolled to the bottom; it seems more suited to a protocol message that can be sent whenever. An attribute makes a lot more sense to me, and it's more flexible (currently, selectors 
are only used by Mafia).

I think there might still be some utility in implementing `|...scroll|top` messages in case we want to scroll to the top/bottom of a page without resending all of the HTML, but a tag would work fine for PM logs and most of the other usecases in PS.

Zarel:
Can you think of a situation where you'd want to scroll without resending all the HTML?

I've been wanting to implement linking to specific lines in PM logs, but I'd want to highlight the line in yellow (like how `/roomlog` does it) which would require resending the HTML anyway.

AnnikaCodes:
Jumping to a particular line in chat/PM logs is the main one yeah.

Zarel:
I'm pretty set on the attribute option, just because it's a lot more flexible. Sorry for not using your work here.

monsanto:
Apparently this was superceeded

PR #1647, opened 2020-10-29
Don't allow ! for client commands

mia-pi-git:
None of these commands support this, so instead of erroring or anything similar, the best solution seems to be this: let the text through to the server and let it handle the "this can't be broadcasted" error.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -423,7 +423,7 @@
 			var cmd = '';
 			var target = '';
 			var noSpace = false;
-			if (text.substr(0, 2) !== '//' && text.charAt(0) === '/' || text.charAt(0) === '!') {
+			if (text.substr(0, 2) !== '//' && text.charAt(0) === '/') {
 				var spaceIndex = text.indexOf(' ');
 				if (spaceIndex > 0) {
 					cmd = text.substr(1, spaceIndex - 1);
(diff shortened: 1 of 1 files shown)


Zarel:
As discussed in Discord, this gives the generic error message "command not found". We'd probably prefer to give the same error message as on the server ("command cannot be broadcast"), even though this means _significantly_ more code.

js/client-chat.js
```diff
@@ -418,7 +418,14 @@
 		},
 
 		// command parsing
-
+		checkBroadcast: function (cmd, text) {
+			if (text.charAt(0) === '!') {
+				this.add('The command "/' + cmd + '" cannot be broadcast.');
+				this.add('Use /' + cmd + ' to use it normally.');
```
Zarel, line 424:
```suggestion
				this.add('|error|The command "/' + cmd + '" cannot be broadcast.');
				this.add('|error|Use /' + cmd + ' to use it normally.');
```

js/client-chat.js
```diff
@@ -418,7 +418,14 @@
 		},
 
 		// command parsing
-
+		checkBroadcast: function (cmd, text) {
+			if (text.charAt(0) === '!') {
+				this.add('|error|The command "/' + cmd + '" cannot be broadcast.');
```
Zarel, line 423:
```suggestion
				this.add('|error|The command "!' + cmd + '" cannot be broadcast.');
```

Zarel:
ty!

PR #1649, opened 2020-10-30
Client Side for Multi Battles

Spandamn:
idk whats wrong with the indentation, it seems good on VS Code.
Obviously needs a lot of work. 
Needs ```node build full```  to work properly.
image


Spandamn:
jesus checks

Spandamn:
Pulling removed a lot of changes i made. fixing.

KrisXV:
You don't really need "SUPER WIP" in the title if this is a draft repository

Spandamn:
battle*

src/battle.ts
```diff
@@ -3244,6 +3292,15 @@ class Battle {
 			this.scene.updateSidebar(side);
 			if (this.joinButtons) this.scene.hideJoinButtons();
 			this.log(args);
+			let userid = app!.user.get('userid');
```
Spandamn, line 3295: Can someone explain this error?

Spandamn: @Zarel

Spandamn:
image
Tooltips for allies are visible now too!

src/battle-text-parser.ts
```diff
@@ -529,6 +545,18 @@ class BattleTextParser {
...
 
+		case 'canDynamax': {
+			let [, side] = args;
+			const own = this.own(side);
+			let template = '';
+			if (this.turn === 1) {
+				if (own) template = this.template('canDynamax', own);
+			} else {
+				template = this.template('canDynamax', own);
+			}
```
Zarel, line 556: Can you explain what the turn tracker is about?

Zarel: Looks like it displays a message unless it's turn 1 and it's not yourself? What's the message and why?

MacChaeger: In gen 8 multi battles, there are two different dynamax messages:
```
canDynamax: "  [TRAINER] can dynamax now!",
canDynamaxOwn: "Dynamax Energy gathered around [TRAINER]!",
```
The "Energy" message is only used on the first turn a player is allowed to dynamax, and only for that player. The other players just see the "can" message, but on the first turn, only the players who can dynamax (the ones on their respective lefts) see any message at all.

MacChaeger: Server-side, this is done by sending the canDynamax messages as secret messages to their respective sides, but the server and client don't support a super-secret message protocol that only gets sent to one player on a team, so the client has to do a bit of work on its own.

Spandamn: Just saying, request.active[0].canDynamax is true if that side can dynamax, otherwise its not defined.

Spandamn:
Fuck i accidentally undrafted it

src/battle.ts
```diff
@@ -1187,10 +1196,12 @@ class Battle {
 	}
 	switchSides() {
 		this.setSidesSwitched(!this.sidesSwitched);
+		this.scene.updateSidebars();
```
Zarel, line 1199: This is the wrong place for this. `resetToCurrentTurn` already redraws everything, including sidebars – if it doesn't, that's a bug and should be fixed, probably in `animationOn`.

Zarel:
```suggestion
```

src/battle.ts
```diff
@@ -1187,10 +1196,12 @@ class Battle {
...
 		this.resetToCurrentTurn();
 	}
 	setSidesSwitched(sidesSwitched: boolean) {
 		this.sidesSwitched = sidesSwitched;
+		if (this.mySide === this.p3 || this.mySide === this.p4) return;
```
Zarel, line 1204: I think switching sides should still be supported if the current side is `p3` or `p4`. What's the meaning of disabling it?

src/battle.ts
```diff
@@ -2941,22 +2954,16 @@ class Battle {
...
+		if (name.match(/^p[0-9]$|p[0-9]: |p[0-9][a-f]: /)) {
+			const serverSideN = parseInt(name.charAt(1), 10) - 1;
+			siden = serverSideN;
+			if (name.match(/^p[0-9]$|p[0-9]: /)) {
+				name = name.substr(4);
+			} else {
+				slot = slotChart[name.charAt(2)];
+				name = name.substr(5);
+				pokemonid = 'p' + (serverSideN + 1) + ': ' + name;
+			}
```
Zarel, line 2966: This is a bit _too_ unreadable in the name of preciseness.

```suggestion
		if (/^p[1-9]($|: )/.test(name)) {
			siden = parseInt(name.charAt(1), 10) - 1;
			name = name.slice(4);
		} else if (/^p[1-9][a-f]: /.test(name))
			siden = parseInt(name.charAt(1), 10) - 1;
			slot = slotChart[name.charAt(2));
			name = name.slice(5);
			pokemonid = `p${siden + 1}: ${name}`;
```

src/battle.ts
```diff
@@ -2966,14 +2973,14 @@ class Battle {
 		pokemonid = parsedPokemonid;
 
 		const searchid = `${pokemonid}|${details}`;
-		const side = this.sides[siden];
+		const side = this.gameType === 'multi' ? this.getSide(`p${siden + 1}`) : this.sides[siden];
```
Zarel, line 2976: Why is `getSide` required here? I can't think of a reason why `this.sides[siden]` wouldn't work for everyone, unless something's seriously wrong.

src/battle.ts
```diff
@@ -3037,6 +3044,8 @@ class Battle {
 	getSide(sidename: string): Side {
 		if (sidename === 'p1' || sidename.substr(0, 3) === 'p1:') return this.p1;
 		if (sidename === 'p2' || sidename.substr(0, 3) === 'p2:') return this.p2;
+		if ((sidename === 'p3' || sidename.substr(0, 3) === 'p3:') && this.p3) return this.p3;
+		if ((sidename === 'p4' || sidename.substr(0, 3) === 'p4:') && this.p4) return this.p4;
```
Zarel, line 3048:
```suggestion
		if (sidename === 'p1' || sidename.startsWith('p1:')) return this.p1;
		if (sidename === 'p2' || sidename.startsWith('p2:')) return this.p2;
		if ((sidename === 'p3' || sidename.startsWith('p3:')) && this.p3) return this.p3;
		if ((sidename === 'p4' || sidename.startsWith('p4:')) && this.p4) return this.p4;
```

src/battle.ts
```diff
@@ -3106,6 +3115,20 @@ class Battle {
...
+				this.pokemonControlled = 1;
+				this.p3 = new Side(this, 2);
+				this.p4 = new Side(this, 3);
+				this.p3.foe = this.p4.ally = this.p2;
+				this.p3.ally = this.p4.foe = this.p1;
+				this.p1.ally = this.p3;
+				this.p2.ally = this.p4;
+				this.sides = [this.p1, this.p2, this.p3, this.p4];
+				this.mySide.isFar = this.mySide.ally.isFar = false;
+				this.farSide.isFar = this.farSide.ally.isFar = true;
```
Zarel, line 3128: This seems wrong, `isFar` should be set by `switchSides` and never changed here.

src/battle.ts
```diff
@@ -3106,6 +3115,20 @@ class Battle {
...
+				this.nearSide.ally.active = this.nearSide.active = [null, null];
+				this.farSide.ally.active = this.farSide.active = [null, null];
```
Zarel, line 3130:
```suggestion
				this.pokemonControlled = 1;
				this.p3 = new Side(this, 2);
				this.p4 = new Side(this, 3);
				this.p3.foe = this.p4.ally = this.p2;
				this.p3.ally = this.p4.foe = this.p1;
				this.p1.ally = this.p3;
				this.p2.ally = this.p4;
				this.sides = [this.p1, this.p2, this.p3, this.p4];
				for (const side of this.sides) side.active = [null, null];
```

src/battle.ts
```diff
@@ -3106,6 +3115,20 @@ class Battle {
 				this.nearSide.active = [null];
 				this.farSide.active = [null];
```
Zarel, line 3116:
```suggestion
				for (const side of this.sides) side.active = [null];
```

src/battle.ts
```diff
@@ -1187,10 +1196,12 @@ class Battle {
...
 	switchSides() {
 		this.setSidesSwitched(!this.sidesSwitched);
+		this.scene.updateSidebars();
 		this.resetToCurrentTurn();
 	}
 	setSidesSwitched(sidesSwitched: boolean) {
 		this.sidesSwitched = sidesSwitched;
+		if (this.mySide === this.p3 || this.mySide === this.p4) return;
 		if (this.sidesSwitched) {
 			this.nearSide = this.mySide = this.p2;
```
Zarel, line 1206: Wait, shouldn't `this.mySide` never change? I thought that was the point of separating `nearSide` and `mySide`?

src/battle.ts
```diff
@@ -1153,7 +1160,7 @@ class Battle {
 		for (const side of this.sides) {
 			if (side) side.reset();
 		}
-		this.myPokemon = null;
+		this.mySide.myPokemon = null;
```
Zarel, line 1163: Wait, what's the reason `myPokemon` was moved? I think it's fine where it was.

Zarel: Also I don't think there's a need to reset `myPokemon`.

src/battle.ts
```diff
@@ -2966,14 +2973,14 @@ class Battle {
...
-		const side = this.sides[siden];
+		const side = this.gameType === 'multi' ? this.getSide(`p${siden + 1}`) : this.sides[siden];
 
 		// search inactive revealed pokemon
 		for (let i = 0; i < side.pokemon.length; i++) {
 			let pokemon = side.pokemon[i];
 			if (pokemon.fainted) continue;
 			// already active, can't be switching in
-			if (side.active.includes(pokemon)) continue;
+			if (side.active.includes(pokemon) && side.ally) continue;
```
Zarel, line 2983: This shouldn't matter. Active Pokémon shouldn't be able to switch in.

style/battle.css
```diff
@@ -253,93 +281,126 @@ License: GPLv2
...
 	right: 0;
-	bottom: 102px;
+	bottom: 75px;
 }
+
 .rightbar .trainer {
 	bottom: 232px;
 }
+
+#p1 {
```
Zarel, line 364: You can't! People might have multiple battles! This should be something like `.trainer.p1`.

src/battle-text-parser.ts
```diff
@@ -226,20 +230,16 @@ class BattleTextParser {
 
 	pokemonName = (pokemon: string) => {
 		if (!pokemon) return '';
-		if (!pokemon.startsWith('p1') && !pokemon.startsWith('p2')) return `???pokemon:${pokemon}???`;
+		if (!pokemon.startsWith('p1') && !pokemon.startsWith('p2') && !pokemon.startsWith('p3') && !pokemon.startsWith('p4')) return `???pokemon:${pokemon}???`;
```
Zarel, line 233:
```suggestion
		if (!pokemon.startsWith('p')) return `???pokemon:${pokemon}???`;
```

src/battle-text-parser.ts
```diff
@@ -226,20 +230,16 @@ class BattleTextParser {
...
-		let side;
-		switch (pokemon.slice(0, 2)) {
-		case 'p1': side = 0; break;
-		case 'p2': side = 1; break;
-		default: return `???pokemon:${pokemon}???`;
-		}
+		let side = pokemon.slice(0, 2);
+		if (!['p1', 'p2', 'p3', 'p4'].includes(side)) return `???pokemon:${pokemon}???`;
 		const name = this.pokemonName(pokemon);
 		const template = BattleText.default[side === this.perspective ? 'pokemon' : 'opposingPokemon'];
```
Zarel, line 244: Shouldn't allies get different text than foes? I suppose that can be left for a future PR.

src/battle-text-parser.ts
```diff
@@ -257,28 +257,39 @@ class BattleTextParser {
...
+		return '';
+	}
+
+	team(side: string, isFar: boolean = false) {
 		side = side.slice(0, 2);
-		if (side === (this.perspective === isFar ? 'p1' : 'p2')) {
-			return BattleText.default.team;
+		let team = [BattleText.default.team, BattleText.default.opposingTeam];
+		if (side === this.perspective || side === BattleTextParser.allyID(side as SideID)) {
+			return team[+isFar];
```
Zarel, line 277:
```suggestion
			return team[isFar ? 1 : 0];
```

src/battle-text-parser.ts
```diff
@@ -257,28 +257,39 @@ class BattleTextParser {
...
 		}
-		return BattleText.default.opposingTeam;
+		return team[+isFar ^ 1];
```
Zarel, line 279:
```suggestion
		return team[isFar ? 0 : 1];
```

Bitwise tricks are really unreadable...

src/battle-text-parser.ts
Zarel, line 277:
```suggestion
		if (side === this.perspective || side === BattleTextParser.allyID(side as SideID)) {
			return isFar ? BattleText.default.team : BattleText.default.opposingTeam;
```

src/battle-text-parser.ts
Zarel: Okay, but protocol messages have never been mixed-case.

I also don't think `canDynamax` is a major message. I would go with `-candynamax`.

src/battle-tooltips.ts
```diff
@@ -319,13 +336,27 @@ class BattleTooltips {
...
+			// mouse over ally's pokemon
+			// serverPokemon definitely exists, sidePokemon maybe
+			// let side = this.battle.mySide.ally;
+			let activeIndex = parseInt(args[1], 10);
+			let pokemon = null;
+			/*if (activeIndex < side.pokemon.length) {
+				pokemon = side.pokemon[activeIndex] || side.ally ? side.ally.pokemon[activeIndex] : null;
+			}*/
+			let serverPokemon = this.battle.mySide.ally.myPokemon ? this.battle.mySide.ally.myPokemon[activeIndex] : null;
 			buf = this.showPokemonTooltip(pokemon, serverPokemon);
```
Zarel, line 360: Why is this different from `activepokemon`?

src/battle-animations.ts
```diff
@@ -693,17 +691,33 @@ class BattleScene {
...
+			$sidebar.html(sidebarhtml);
 			$sidebar.find('.trainer').css('opacity', 1);
 		} else {
 			$sidebar.find('.trainer').css('opacity', 0.4);
 		}
 	}
 	updateSidebars() {
-		for (const side of this.battle.sides) this.updateSidebar(side);
+		this.$leftbar.html(this.getSidebarHTML(this.battle.nearSide) + (this.battle.nearSide.ally ? this.getSidebarHTML(this.battle.nearSide.ally, true) : ''));
+		this.$rightbar.html(this.getSidebarHTML(this.battle.farSide) + (this.battle.farSide.ally ? this.getSidebarHTML(this.battle.farSide.ally, true) : ''));
```
Zarel, line 720:
```suggestion
		this.updateSidebar(this.battle.nearSide);
		this.updateSidebar(this.battle.farSide);
```

src/battle-animations.ts
```diff
@@ -666,7 +664,7 @@ class BattleScene {
 		for (let i = 0; i < sidebarIcons.length; i++) {
 			const [iconType, pokeIndex] = sidebarIcons[i];
 			const poke = pokeIndex !== null ? side.pokemon[pokeIndex] : null;
-			const tooltipCode = ` class="picon has-tooltip" data-tooltip="pokemon|${side.n}|${pokeIndex}${iconType === 'pokemon-illusion' ? '|illusion' : ''}"`;
+			const tooltipCode = ` class="picon has-tooltip" data-tooltip="pokemon|${this.battle.sides.indexOf(side)}|${pokeIndex}${iconType === 'pokemon-illusion' ? '|illusion' : ''}"`;
```
Zarel, line 667:
```suggestion
			const tooltipCode = ` class="picon has-tooltip" data-tooltip="pokemon|${side.n}|${pokeIndex}${iconType === 'pokemon-illusion' ? '|illusion' : ''}"`;
```
This shouldn't be necessary, right? This should just be `side.n`.

src/battle-animations.ts
```diff
@@ -693,17 +691,33 @@ class BattleScene {
...
+		if (isAlly) {
+			if (!side.isFar) {
+				posStr += "p3";
+			} else {
+				posStr += "p4";
+			}
+		} else {
+			posStr = !side.isFar ? "p1" : "p2";
+		}
+		return `<div class="trainer" id="${posStr}"><strong>${BattleLog.escapeHTML(side.name)}</strong><div class="trainersprite"${ratinghtml} style="background-image:url(${Dex.resolveAvatar(side.avatar)})"></div>${pokemonhtml}</div>`;
```
Zarel, line 705:
```suggestion
		return `<div class="trainer trainer-${posStr}"><strong>${BattleLog.escapeHTML(side.name)}</strong><div class="trainersprite"${ratinghtml} style="background-image:url(${Dex.resolveAvatar(side.avatar)})"></div>${pokemonhtml}</div>`;
```

style/battle.css
```diff
@@ -315,11 +315,25 @@ License: GPLv2
...
+}
+#p2 {
+	bottom: 232px;
+}
+#p3 {
+	bottom: 232px;
+}
+#p4 {
+	bottom: 75px;
+}
```
Zarel, line 335:
```suggestion
.trainer-p1 {
	bottom: 75px;
}
.trainer-p2 {
	bottom: 232px;
}
.trainer-p3 {
	bottom: 232px;
}
.trainer-p4 {
	bottom: 75px;
}
```

src/battle-animations.ts
Zarel: See what I said in `battle.css` - IDs should straight-up never be used for any reason in modern code.

src/battle-animations.ts
```diff
@@ -693,17 +691,33 @@ class BattleScene {
...
+		let posStr = "";
+		if (isAlly) {
+			if (!side.isFar) {
+				posStr += "p3";
+			} else {
+				posStr += "p4";
+			}
+		} else {
+			posStr = !side.isFar ? "p1" : "p2";
+		}
```
Zarel, line 704: It sounds like we shouldn't be using `p1/p2/p3/p4` for these at all, considering they look like side IDs even though they aren't. Maybe `near`, `near2`, `far`, and `far2`?

src/battle.ts
Zarel: Oh, I see! Because your ally also has Pokémon. I would actually put it in `battle.myAllyPokemon`.

Zarel: The problem with putting it in `.mySide.myPokemon` is that it's very brittle. It gets cleared on `resetStep` and a lot of other stuff.

src/battle-tooltips.ts
Zarel: Oh, I see! It's for the switch menu.

js/client-battle.js
```diff
@@ -1007,14 +1040,21 @@
...
-			if (this.battle.sidesSwitched !== !!(this.side === 'p2')) {
+			if (this.battle.sidesSwitched !== !!(this.side === 'p2' || this.side === 'p4')) {
 				this.battle.switchSides();
 				this.$chat = this.$chatFrame.find('.inner');
 			}
 		},
 		updateSide: function () {
 			var sideData = this.request.side;
 			this.battle.myPokemon = sideData.pokemon;
+			this.battle.mySide = this.battle.nearSide = this.battle[sideData.id];
```
Zarel, line 1051: Err, I don't think `nearSide` should be set to your own side. It should be set to `mySide` or `mySide.ally` depending on which one is `siden <= 1`.

js/client-battle.js
```diff
@@ -715,7 +727,14 @@
...
 				);
-
+				/*if (this.battle.mySide.ally) {
+					this.$allyPokemon.html(
+						'<div class="ally-pokemon">' +
+						'<div class="allyTeam"><button name="allyteam">Ally Party</button></div>' +
+						'<div class="allyparty">' + this.displayAllyParty() + '</div>' +
+						'</div>'
+					);
+				}*/
```
Zarel, line 737:
```suggestion

```

js/client-battle.js
```diff
@@ -555,6 +562,10 @@
 			if (type === 'movetarget') {
 				requestTitle += 'At who? ';
 
+				if (this.request && this.request.side) {
+					pos += Math.floor((parseInt(this.side.charAt(1), 10) - 1) / 2);
+				}
```
Zarel, line 567: ?

Zarel: It looks like this is doing the same thing as `activePos` but different?

Zarel:
Okay, I think my review covered everything concerning in the diff.

MacChaeger:
Remind me, and I’ll try and respond to stuff tomorrow; I’m too sleepy to remember/figure out what this code does right now.

Zarel:
Neat! Sorry for taking so long, but my review does cover everything, so once we work through everything I brought up, this should be mergeable.

js/client-battle.js
MacChaeger: Yeah. They both do the same thing in a normal Multi Battle, but whereas this supports a theoretical Multi-Triple Battle with a p5 and p6, `activePos` is written to support a theoretical 2v2v2v2 kind of battle. I couldn't tell you why, though.

js/client-battle.js
MacChaeger: So something like this?
```
this.battle.mySide = this.battle[sideData.id];
this.battle.nearSide = this.sides[this.battle.mySide.n % 2];
```

src/battle-text-parser.ts
MacChaeger: No, that's an easy enough fix for this one. Something like
```
const teamOf = {'p1': 0, 'p2': 1, 'p3': 0, 'p4': 1};
const isOpposing = teamOf[side] === teamOf[this.perspective];
const template = BattleText.default[isOpposing ? 'pokemon' : 'opposingPokemon'];
```
When supporting other game types with different team setups like Battle Royal we can just use different values for `teamOf`;

src/battle.ts
MacChaeger: If the current side is `p3` or `p4` then we're not a spectator, and active players should never be switching sides except when they first join as a player. Spectator perspectives are always `p1` or `p2`.

src/battle.ts
MacChaeger: I'm pretty sure anything that uses `mySide` is exclusive to actual players and that all the code used for spectators only ever uses `nearSide`, so assuming the code that sets `mySide` when a player joins actually does work for p2, then this doesn't need to set `mySide`, but only because it doesn't matter.

src/battle.ts
MacChaeger: I can't tell what either version does, so I'll take your word for it that it's unreadable. 😅

src/battle.ts
MacChaeger: I honestly have no idea why that's there. Git blame isn't any help, either, so I guess I'll try taking it out and see what happens.

src/battle.ts
```diff
@@ -2941,22 +2953,14 @@ class Battle {
...
+		if (/^p[1-9]($|: )/.test(name)) {
+			siden = parseInt(name.charAt(1)) - 1;
+			name = name.slice(4);
+		} else if (/^p[1-9][a-f]: /.test(name))
+			siden = parseInt(name.charAt(1)) - 1;
+			slot = slotChart[name.charAt(2));
+			name = name.slice(5);
+			pokemonid = `p${siden + 1}: ${name}`;
 		}
 		return {name, siden, slot, pokemonid};
```
MacChaeger, line 2965: All of a sudden this isn't valid syntax anymore and is causing build errors. What gives?

style/battle.css
```diff
@@ -315,11 +315,25 @@ License: GPLv2
...
+.trainer-p2 {
+	bottom: 232px;
+}
+.trainer-p3 {
+	bottom: 232px;
+}
+.trainer-p4 {
+	bottom: 75px;
```
MacChaeger, line 334: This doesn't work. Seems like the `.rightbar .trainer` above is overriding this one.

Zarel: You can just remove the `.rightbar .trainer`; it's redundant with your changes here, right?

js/client-battle.js
Zarel: I don't really care which one you go with, but probably they should match for readability reasons.

js/client-battle.js
Zarel: Yes! That's what I was thinking.

src/battle-text-parser.ts
Zarel: We already have an `ally` functiom somewhere in the text parser, don't we?

src/battle.ts
Zarel: I'm saying that it's bad for this to set `mySide`, because it would cause this function to break when a player joins. Because of that, this should be taken out.

src/battle.ts
Zarel: There's no reason why we shouldn't support active players switching sides. I think it's bad to couple them like this.

Zarel: There's a lot of code that assumes `mySide` will always be `nearSide` and that shouldn't be the assumption; we made them separate variables to support cases where they're not.

src/battle.ts
Zarel: In fact, I think it would make a lot of sense for `.mySide` to be `null` if the player isn't in a battle. I think that would make it easier to reason about.

style/battle.css
Zarel: Did it work? It looks like it worked.

MacChaeger:
That should be everything from the review addressed.

Zarel:
Haven't forgotten.

src/battle-animations.ts
```diff
@@ -714,16 +720,16 @@ class BattleScene {
 	}
 
 	teamPreviewEnd() {
-		for (let siden = 0; siden < 2; siden++) {
-			this.$sprites[siden].empty();
+		for (let siden = 0; siden < this.battle.sides.length; siden++) {
+			this.$sprites[siden % 2].empty();
 			this.battle.sides[siden].updateSprites();
 		}
```
Zarel, line 726: `siden % 2` isn't `spriteIndex`, right? It might be easier if we did these in separate loops.

```suggestion
		this.$sprites[0].empty();
		this.$sprites[1].empty();
		for (let siden = 0; siden < this.battle.sides.length; siden++) {
			this.battle.sides[siden].updateSprites();
		}
```

Zarel: (I'll handle this)

Zarel:
Thanks!

js/client-battle.js
```diff
@@ -877,19 +898,20 @@
...
 			} else if (this.choice.choices && this.request) {
-				var nearActive = this.battle.nearSide.active;
-				for (var i = 0; i < this.choice.choices.length; i++) {
+				var myActive = this.battle.myPokemon;
+				for (var i = 0; i < (this.battle.myPokemon || this.choice.choices.length); i++) {
 					var parts = this.choice.choices[i].split(' ');
 					switch (parts[0]) {
 					case 'move':
 						var move;
+						myActive = this.battle.nearSide.active;
```
Zarel, line 907: Was this a typo? I don't think there's a point to this.

js/client-battle.js
```diff
@@ -877,19 +898,20 @@
 				}
 				buf += leads.join(', ') + ' will be sent out first.<br />';
 			} else if (this.choice.choices && this.request) {
-				var nearActive = this.battle.nearSide.active;
-				for (var i = 0; i < this.choice.choices.length; i++) {
+				var myActive = this.battle.myPokemon;
+				for (var i = 0; i < (this.battle.myPokemon || this.choice.choices.length); i++) {
```
Zarel, line 902: I also don't think there's a point to this change, either...

js/client-battle.js
```diff
@@ -909,7 +931,7 @@
 								var targetActive = this.battle.farSide.active;
 								// Targeting your own side in doubles / triples
 								if (targetPos < 0) {
-									targetActive = nearActive;
+									targetActive = myActive;
```
Zarel, line 934: Ah, it's for this!

js/client-battle.js
Zarel: Figured it out, it's for the `targetActive` stuff later.

PR #1650, opened 2020-11-01
Add support for Gen 3 Doubles formats

KrisXV:
This will be necessary for one of the November RoA spotlight ladders


Zarel:
tyty

PR #1652, opened 2020-11-02
Support text formatting in |error|

mia-pi-git:


js/client-chat.js
```diff
@@ -1404,7 +1404,8 @@
 					break;
 
 				case 'error':
-					this.$chat.append('<div class="notice message-error">' + BattleLog.escapeHTML(row.slice(1).join('|')) + '</div>');
+					// eslint-disable-next-line no-undef
+					this.$chat.append('<div class="notice message-error">' + formatText(row.slice(1).join('|'), true) + '</div>');
```
Zarel, line 1408: is `formatText` not defined here?

mia-pi-git: It works locally. Not sure what's with ESLint.

Zarel: isn't ``formatText`` already used elsewhere in client-chat?

mia-pi-git: Not on a cursory search. The only usage i see is in `battle-log.ts`.

Zarel: aha, looks like you should be using `BattleLog.parseMessage`. You should be adding a `trusted` param _there_.

mia-pi-git: Done.

Zarel:
Thanks!

PR #1653, opened 2020-11-02
Mainmenu: Show out-of-date app users an error

mia-pi-git:
Not sure if this is the right way to do this, but doing it in mainmenu gave access to tools to make the popup easier, and made the most sense since similar initialization things are done there.

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -104,7 +104,7 @@
 				}
 				app.roomsFirstOpen = 2;
 			}
-			if ('nw' in window && !nw.process.version.startsWith('v13')) {
+			if ('nw' in window && !nw.process.version.startsWith('v13.')) {
 				app.addPopupMessage(
 					"Your version of the app is out of date.\n" +
 					"Go to pokemonshowdown.com to update it."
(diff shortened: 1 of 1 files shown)


js/client-mainmenu.js
```diff
@@ -104,6 +104,12 @@
 				}
 				app.roomsFirstOpen = 2;
 			}
+			if ('nw' in window && !nw.process.version.startsWith('v13')) {
```
Zarel, line 107:
```suggestion
			if ('nw' in window && !nw.process.version.startsWith('v13.')) {
```

js/client-mainmenu.js
```diff
@@ -104,6 +104,12 @@
 				}
 				app.roomsFirstOpen = 2;
 			}
+			if ('nw' in window && !nw.process.version.startsWith('v13.')) {
+				app.addPopupMessage(
+					"Your version of the app is out of date.\n" +
+					"Go to pokemonshowdown.com to update it."
```
Zarel, line 110:
```suggestion
					"Please go to pokemonshowdown.com to update it."
```

Zarel:
Can you make this an unclosable modal?

Zarel:
Hm, maybe we should use this version for now and make sure there aren't any bugs, before making it an unclosable modal. Thanks!

PR #1655, opened 2020-11-03
Fix National Dex teambuilder

KrisXV:


monsanto:
I don't know what this does, but Kris says it is an emergency

PR #1657, opened 2020-11-06
Add support for Bots to receive input from pages

PartMan7:
### Code and description has drastically changed, please read https://github.com/smogon/pokemon-showdown-client/pull/1657#issuecomment-723853783

Basically, this does this:
a) When the user clicks a button with the name `sendToBot`, the client will
b) Look for all input elements that have the name `botInput` with type text (any amount) and type radio (only one group, since the name is already fixed)
c) Send the JSON of the array of the values of all these elements in order
d) Prepended by the value of the sendToBot button.

As an example:

```<input type="text" name="botInput"><input type="text" name="botInput"><input type="text" name="botInput"><br/><input type="radio" name="botInput" value="A"><input type="radio" name="botInput" value="B"><input type="radio" name="botInput" value="C"><br/><button name="sendToBot" value=",demo ">Go!</button>```

When the button is clicked directly, the client will send `,demo ["","","",null]` to the author of the Bot page.
If the user fills in some information in the textboxes, the respective strings will be sent when the user clicks the button.
The value of the sole clicked radio button will be sent instead of null if a radio option is checked. The index of the radio option is always the index of the first valid radio element in the page (with the right input name attribute) relative to all other valid input elements from the start of the HTML.

The primary use of this is to allow users to use HTML input in the page itself, without having to resort to tedious workarounds, like manually programming in an entire keyboard just for typing numbers.

If used in any HTML source except a sent HTML page, the client ignores it.

Note: These probably shouldn't be merged until the corresponding server changes (mainly for restricting permissions and ensuring that people don't have the button value as something like /logout)

Oh, and I'm fairly inexperienced with the client, so hopefully I haven't broken anything else by adding the e attribute to buttonclick functions.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -2002,7 +2002,7 @@ function toId() {
 				app.dispatchingButton = target;
 				e.preventDefault();
 				e.stopImmediatePropagation();
-				this[target.name](target.value, target);
+				this[target.name](target.value, target, e);
 				delete app.dismissingSource;
 				delete app.dispatchingButton;
 			}
@@ -2028,6 +2028,46 @@ function toId() {
 		 */
 		receive: function (data) {
 			//
+		},
+		sendToBot: function (data, target, e) {
+			var sentPage = e.currentTarget.offsetParent;
+			var pageTitle = sentPage.attributes.getNamedItem('id');
+			if (!pageTitle) return;
+			else pageTitle = pageTitle.nodeValue;
+			if (pageTitle.substr(0, 14) !== 'room-view-bot-') return; // Not a sent HTML page
+			var author = pageTitle.substr(14).split('-')[0];
+			var toSend = [];
+			var radioIndex = null;
+			for (var i = 0; i < sentPage.children.length; i++) {
+				var tag = sentPage.children[i];
+				if (tag.localName !== 'input') continue;
+				var tagName = tag.attributes.getNamedItem('name');
+				if (!tagName || tagName.nodeValue !== 'botInput') continue;
+				var tagType = tag.attributes.getNamedItem('type');
+				if (!tagType) tagType = 'text';
+				else tagType = tagType.nodeValue;
+				switch (tagType) {
+				case 'text': {
+					toSend.push(tag.value);
+					break;
+				}
+				case 'radio': {
+					/**
+					 * Since radio buttons use names for grouping,
+					 * pages can only support one radio group per page.
+					 * Because of this, we only store the one checked value.
+					 */
+					if (typeof radioIndex !== 'number') {
+						radioIndex = toSend.length;
+						toSend.push(null);
+					}
+					if (tag.checked) toSend[radioIndex] = tag.value;
+					break;
+				}
+				}
+			}
+			var stringToSend = "/pm " + author + "," + data + " " + JSON.stringify(toSend);
+			this.send(stringToSend);
 		},
 
 		// layout
--- a/js/client.js
+++ b/js/client.js
@@ -2029,44 +2029,90 @@ function toId() {
 		receive: function (data) {
 			//
 		},
-		sendToBot: function (data, target, e) {
+		submit: function (data, target, e) {
 			var sentPage = e.currentTarget.offsetParent;
 			var pageTitle = sentPage.attributes.getNamedItem('id');
 			if (!pageTitle) return;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Zarel** [comment]
We already support `name="send" value="/msg [BOT], [MESSAGE]"`, right? What does this add over that?

**Comment by PartMan7** [comment]
That doesn't take user input, though. The quickest example I can give is this:

https://cdn.discordapp.com/attachments/576487985709580309/751500689590583336/Screenshot_2020-09-04_PartBot_partbot-mastermindboardgamespartman_-_Showdown_.png

In Mastermind at least, there's only 10 needed buttons. For a game like Scrabble, or if there's any need for something like a name, there's no way to do it without manually PMing the Bot - which requires you to switch to the Home screen and then reopen the page. Apart from that, this also lets users simply type and click a button to answer, as opposed to needing to type something like `,scrabble boardgames 1604677864702 play a 4 down pizza`.

**Comment by Zarel** [comment]
Oh, I see! This is for form submission!

I think form submission should be handled by an attribute on the `<form>` itself. This is important to support submitting forms by pressing `Enter` in a text box.

**Comment by Zarel** [comment]
Something like this:

```html
<form data-sendonsubmit="/msg botname, {name}, {guess}">
  <label>Name: <input type="text" name="name" autofocus /></label>
  <label>Guess: <input type="text" name="guess" /></label>
  <button type="submit">Submit</submit>
</form>
```

PartMan7:
That looks pretty good; I'll take a shot at trying something like that ~but will probably mess up~

PartMan7:
Description:

When a button with the name ``submit`` and a value is clicked, the client scans upwards for the first matching form element. If the client is unable to find one before reaching the parent of the HTMLpage, it stops execution there.
Next, it looks at all named input elements within the form. Out of these, it currently only considers elements with type as `text`, `radio`, and `checkbox`.
Finally, the client replaces all instances of {var_name} with the corresponding value - text returns the text in the box, radio returns the value of the selected element or ``'null'`` if no option is selected, and checkbox returns the values of all selected elements delimited by a `|`. Please note that names are case-sensitive and must be written exclusively using alphanumeric characters in order to be recognized.

```
<form>
	<label>Name: <input type="text" name="name" autofocus /></label>
	<label>Guess: <input type="text" name="guess" /></label>
	<input type="radio" name="radio" value="A">A<br />
	<input type="radio" name="radio" value="B">B<br />
	<input type="radio" name="radio" value="C">C<br />
	<br /><br />
	<input type="checkbox" name="check" value="A">A<br />
	<input type="checkbox" name="check" value="B">B<br />
	<input type="checkbox" name="check" value="C">C<br />
	<input type="checkbox" name="check" value="D">D<br />
	<button name="submit" type="submit" value="{toID-name}, {guess}, {adkb}, {efvkj}, {radio}, {constructor}, {check}">Submit</button>
</form>
```

Also included a couple useful features, notably the ability to prepend a ``toID-`` to the variable name in the button value in order to replace the regular value with its ID. (For checkboxes, this first converts the values to their IDs and then joins them with a `|`)

Zarel:
Can you describe why you decided not to go with the `<form>` attribute approach?

PartMan7:
PS trimmed `data-actionsend` out.

Zarel:
Oh yeah you need to manually whitelist it in Caja, which you can find here:

https://github.com/smogon/pokemon-showdown-client/blob/ab4f06316fa07ebadac9c13afc5191caf449dda9/src/battle-log.ts#L693-L857

Zarel:
My main worry is – like mentioned before – there are ways to submit forms without using buttons, like by pressing Enter in a text box. That's why I think the code makes more sense on the `<form>` than on the `<button>`. That's also where HTML usually puts it.

Like the analogous HTML constructs are `<button onclick="...">` vs `<form onsubmit="...">`.

Zarel:
Caja is not documented at all, but the comments in `battle-log.ts` should be enough. If you still need help, just ask!

sparkychildcharlie:
Was just wondering is this still going to be a thing? :o It would be realllly useful for implementing a lot of features

PartMan7:
@mia-pi-git had taken over this, since there were additional usecases they wished to adapt this for.

Zarel:
I ended up implementing the sanitizer part of this, so the only thing left to do is this part.

PR #1658, opened 2020-11-08
Refactor sides to use .isFar over .n for graphics

MacChaeger:
This is a step in the right direction for 4 player battle support. `Side.n` is still equivalent to its index in `Battle.sides`, but now that array is never rearranged, and instead we use `Side.isFar` to determine which side graphics should appear on. The unary `+` operator is used in this code to convert boolean values into numbers so `Side.isFar` can function exactly the same as the old `Side.n` when used a boolean or a number, and thus we don't have to keep track of two separate values that are essentially equivalent. Accordingly, `Sprite.isBackSprite` has been changed to `Sprite.isFrontSprite` for consistency.


Zarel:
For the record, this is my highest priority to review, it's just a big refactor so it's taking me a while. Please be patient and I promise this won't take long!

MacChaeger:
I've tried to be patient by working on other PRs, but that seems to be just distracting from this one, so I guess I'll have to try actually just patiently waiting.

Zarel:
I will merge this now; I don't think I'll be able to code review. We can revert if it causes problems later.

src/battle-animations.ts
```diff
@@ -1822,16 +1821,16 @@ class PokemonSprite extends Sprite {
 	}
 
 	behindx(offset: number) {
-		return this.x + (this.isBackSprite ? -1 : 1) * offset;
+		return this.x + (this.isFrontSprite ? 1 : -1) * offset;
 	}
 	behindy(offset: number) {
-		return this.y + (this.isBackSprite ? 1 : -1) * offset;
+		return this.y + (this.isFrontSprite ? 1 : -1) * offset;
```
MacChaeger, line 1827: Whoops, looks like I missed something. I'll fix this real quick and PR. Looking over the rest, I think this is the only mistake.

PR #1661, opened 2020-11-09
Fix teambuilder top bar for boxes

MacChaeger:


MacChaeger:
Side note: I had originally intended to add a button that could automatically convert a team into a box, but I forgot. Where would be a good place for such a button, and what would it say? I was going to call it "Boxify".

js/client-teambuilder.js
```diff
@@ -1711,7 +1711,14 @@
 			if (this.curSetList.length && !this.curSetList[this.curSetList.length - 1].species && this.curSetLoc !== this.curSetList.length - 1) {
 				this.curSetList.splice(this.curSetList.length - 1, 1);
 			}
-			for (var i = 0; i < this.curSetList.length; i++) {
+			// if in a box, try to show at least 2 and up to 4 other pokemon in each direction
+			// but don't step outside the array bounds (obviously)
+			var scrolling = this.curTeam.capacity > 6 || this.curSetList.length > 6;
```
Zarel, line 1716: It seems to me that you only need `this.curSetList.length > 6`?

js/client-teambuilder.js
```diff
@@ -1711,7 +1711,14 @@
...
+			var endRoom = Math.max(this.curSetLoc + 3, 5);
+			var end = scrolling ? Math.min(this.curSetList.length, endRoom) : this.curSetList.length;
+			var startRoom = Math.min(this.curSetLoc - 2, end - 5);
+			var start = scrolling ? Math.max(0, startRoom) : 0;
```
Zarel, line 1720:
```suggestion
			var start = 0;
			var end = this.curSetList.length;
			if (end > 6) {
				start = this.curSetLoc - 2;
				if (start < 0) start = 0;
				if (start + 5 > end) start = end - 5;
				end = start + 5;
			}
```

MacChaeger: This breaks for a box with exactly six Pokemon in it, which is why it's necessary to check `capacity`.

Zarel: Oh!

Zarel:
```suggestion
			var start = 0;
			var end = this.curSetList.length;
			if (end > 6 || (end === 6 && this.curTeam.capacity > 6)) {
				start = this.curSetLoc - 2;
				if (start < 0) start = 0;
				if (start + 5 > end) start = end - 5;
				end = start + 5;
			}
```

Zarel:
Thanks!

PR #1663, opened 2020-11-10
Simplify protocol for the Unnerve ability

MacChaeger:
I tested using [this old replay](https://replay.pokemonshowdown.com/nu-593386897), and this appears to be fully backwards compatible. I'll also make a server PR to remove the extraneous side data from Unnerve activation messages, but that one won't be totally compatible with an outdated client, so this for sure has to be merged and public before that is.

### DIFF
--- a/src/battle-text-parser.ts
+++ b/src/battle-text-parser.ts
@@ -260,9 +260,9 @@ class BattleTextParser {
 		return `???side:${side}???`;
 	}
 
-	team(side: string) {
+	team(side: string, isFar: 0 | 1 = 0) {
 		side = side.slice(0, 2);
-		if (side === (this.perspective === 0 ? 'p1' : 'p2')) {
+		if (side === (this.perspective === isFar ? 'p1' : 'p2')) {
 			return BattleText.default.team;
 		}
 		return BattleText.default.opposingTeam;
@@ -609,7 +609,7 @@ class BattleTextParser {
 			const id = BattleTextParser.effectId(ability);
 			if (id === 'unnerve') {
 				const template = this.template('start', ability);
-				return line1 + template.replace('[TEAM]', this.team(arg4));
+				return line1 + template.replace('[TEAM]', this.team(pokemon.slice(0, 2), 1));
 			}
 			let templateId = 'start';
 			if (id === 'anticipation' || id === 'sturdy') templateId = 'activate';


MacChaeger:
Actually, on second thought, it's probably fine to leave the server as-is so it's also compatible with other clients. I'll make the other PR if you want,, though, or you could do it yourself in a jiffy; it's just removing the fourth param from the three `'-ability'` messages for Unnerve in `data/abilities.ts`.

Zarel:
Why though?

MacChaeger:
Because there’s absolutely no reason the client needs to be sent a side id to figure out which team is being affected by a given Pokémon’s Unnerve. Like the Unnerve ability itself, this change isn’t terribly useful, but it should be strictly beneficial if the client doesn’t have to receive quite as much data.

Zarel:
The difference in data is unnoticeable, and the previous behavior allows the team to be changed in mods (I can imagine it being relevant in free-for-all).

Zarel:
I don't think it's worth losing flexibility.

MacChaeger:
In mods? What mod would ever need Unnerve specifically to say that its own team is too afraid to eat berries? FFA isn’t a good reason to keep this behavior either; in fact, it’s part of the problem. The ``Battle.add`` function explicitly doesn’t support sending multiple sides at once, so the client would have to check which side it originated from anyway.

MacChaeger:
…should I have mentioned the FFA thing earlier?

Zarel:
We could change it to support multiple sides at once.

I think mainly I don't want to lose the consistency of "things sent are the information in the textual message".

MacChaeger:
It’s already being sent, though, which is why this change is backwards-compatible. The text parser can only make a distinction between the user’s (perspective’s) team and not that team. I can’t even imagine a use case for sending multiple sides where the client wouldn’t be able to infer that the message should apply to an effect’s source and its allies or all of the effect’s opponents because in any situation like that, you would want the client to just receive multiple messages instead (imagine something like screen cleaner, which sends a message for each individual side condition).

MacChaeger:
If you really want flexible, verbose protocol, I suggest sending some kind of flag to say “this message should refer to all the source’s opponents instead of its allies”.

Zarel:
...I mean, I can see how this would be the _easiest_ way to support FFA. There's probably no need to argue about this; I suppose this isn't all that big of a deal to me.

MacChaeger:
Well, we might better settle this with some research. @Marty-D or @DaWoblefet do either of you know how Unnerve’s text actually works in a Battle Royal if an opponent’s Pokémon with the ability switches in? It’s pretty much the only case I can think of where an effect would apply to multiple teams as a whole including the player’s team but not every team, and while I imagine it would say “Your team is too nervous”, it would be nice to know for sure.

Marty-D:
unnerve1
unnerve2

Zarel:
OK, knowing this, I think this change is the correct move.

PR #1664, opened 2020-11-11
Adjust support for building text files

KrisXV:
client side part of https://github.com/smogon/pokemon-showdown/pull/7680

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -1156,17 +1156,15 @@ process.stdout.write("Building `data/moves,items,abilities,typechart,learnsets.j
 
 	function assignData(id, entry) {
 		for (const key in entry) {
-			if (key === 'name' || /^(modded)?(short)?desc/i.test(key)) continue;
-			if (key === 'moddedPlines') continue;
-			const plines = entry.moddedPlines;
+			if (key === 'name' || /(short)?desc/i.test(key)) continue;
 			const textEntry = Text[id] || (Text[id] = {});
-			textEntry[key] = entry[key];
-			if (plines) {
-				for (const pline in plines) {
-					for (const gen in plines[pline]) {
-						textEntry[pline + 'Gen' + gen[3]] = plines[pline][gen];
-					}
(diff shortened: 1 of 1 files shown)


Zarel:
Do you want to just support the new syntax instead of rebuilding it into the old syntax?

build-tools/build-indexes
```diff
@@ -1156,11 +1156,16 @@ process.stdout.write("Building `data/moves,items,abilities,typechart,learnsets.j
 
 	function assignData(id, entry) {
 		for (const key in entry) {
-			if (key === 'name' || key.startsWith('desc') || key.startsWith('shortDesc')) {
-				continue;
-			}
+			if (key === 'name' || /(short)?desc/i.test(key)) continue;
```
Zarel, line 1159:
```suggestion
			if (['name', 'desc', shortDesc'].includes(key)) continue;
```

build-tools/build-indexes
```diff
@@ -1156,11 +1156,16 @@ process.stdout.write("Building `data/moves,items,abilities,typechart,learnsets.j
...
 			const textEntry = Text[id] || (Text[id] = {});
-			textEntry[key] = entry[key];
+			if (key.startsWith('gen')) {
+				for (const key2 in entry[key]) {
+					if (/(short)?desc/i.test(key2)) continue;
+					textEntry[key2 + 'Gen' + key[3]] = entry[key][key2];
+				}
+			} else {
+				textEntry[key] = entry[key];
+			}
```
Zarel, line 1168:
```suggestion
			if (key.startsWith('gen')) {
				for (const modKey in entry[key]) {
					if (['desc', shortDesc'].includes(key)) continue;
					textEntry[modKey + 'Gen' + key.charAt(3)] = entry[key][modKey];
				}
			} else {
				textEntry[key] = entry[key];
			}
```
readability pass

Zarel:
You mentioned that it would be hard to convert the client to the new style, but with `?.` it should be exactly the same as before, right?

PR #1665, opened 2020-11-11
Standardize volatile tag formatting

MacChaeger:
I also snuck Tar Shot into `statusTable` because somehow it's been missing all this time.


src/battle-animations.ts
```diff
@@ -2554,117 +2554,120 @@ class PokemonSprite extends Sprite {
...
-			else status += statusTable[i];
+			status += statusTable[i] !== null ? PokemonSprite.buildEffectTag(statusTable[i] || i) : '';
 		}
 		let statusbar = this.$statbar.find('.status');
 		statusbar.html(status);
 	}
 
+	private static buildEffectTag(effect: [string, 'good' | 'bad' | 'neutral'] | string) {
+		if (typeof effect === 'string') effect = [`[[${effect}]]`, 'neutral'];
+		return `<span class="${effect[1]}">${effect[0].replace(' ', '&nbsp;')}</span> `;
```
Zarel, line 2668:
```suggestion
		return `<span class="${effect[1]}">${effect[0].replace(/ /g, '&nbsp;')}</span> `;
```

src/battle-animations.ts
```diff
@@ -2554,117 +2554,120 @@ class PokemonSprite extends Sprite {
...
 		for (let i in pokemon.turnstatuses) {
 			if (i === 'roost' && !pokemon.getTypeList().includes('Flying')) continue;
-			if (typeof statusTable[i] === 'undefined') status += '<span class="neutral">[[' + i + ']]</span>';
-			else status += statusTable[i];
+			status += statusTable[i] !== null ? PokemonSprite.buildEffectTag(statusTable[i] || i) : '';
 		}
 		for (let i in pokemon.movestatuses) {
-			if (typeof statusTable[i] === 'undefined') status += '<span class="neutral">[[' + i + ']]</span>';
-			else status += statusTable[i];
+			status += statusTable[i] !== null ? PokemonSprite.buildEffectTag(statusTable[i] || i) : '';
```
Zarel, line 2660: It seems like you could move more of this logic into `buildEffectTag`.

Like

```suggestion
			status += PokemonSprite.buildEffectTag(i);
```

MacChaeger: Where would I put the `statusTable` declaration? I wouldn’t want to redeclare it every time the method is called, so should it be a property on `PokemonSprite`?

MacChaeger: actually, what if I make it a static property of the `PokmeonSprite` prototype, and after each status is built I store the result in the prototype property to cache it? seems like a good blend of readability and performance.

Zarel: Yeah, it can be a static property. Storing the result in the prototype property to cache it, though, is I think an unnecessary hyper-optimization not worth the loss in readability.

(But if you really want to do it, I won't stop you.)

Zarel:
Thanks!

PR #1668, opened 2020-11-14
Close uploaded replay when opening a new one

MacChaeger:
Previously, if an uploaded replay was already open, trying to open a new one wouldn't replace it automatically.


Zarel:
Thanks!

PR #1670, opened 2020-11-17
Add text for G-Max and Pledge effects; modify Focus Energy / Attract volatile text

DaWoblefet:
The games use the following text for these effects, which I have added and / or updated. These are based on the strings used in SwSh.

- G-Max Wildfire / G-Max Volcalith / G-Max Vine Lash / G-Max Cannonade - as they are
- Pledge rainbow = Rainbow
- Pledge sea of fire = Sea of Fire
- Pledge swamp = Swamp
- Focus Energy / Lansat Berry / whatever else can do that = Critical Hit Boost
- Attract / Cute Charm = Infatuation


Zarel:
Looks good, thanks!

PR #1673, opened 2020-11-19
Fix challenging "boolean" usernames

DaWoblefet:
This particularly affected challenging user false, who would properly receive the challenge, but it would be canceled on the challenger's end later due to 'false' !== false in the later check on client-mainmenu:729. I'm not sure if there's a smarter way to do this.

We have more calls to data.('userid') - should I just go ahead and sanitize all of them?

Before:
image

After:
image

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -697,7 +697,7 @@
 			var self = this;
 			this.$('.pm-window').each(function (i, el) {
 				var $pmWindow = $(el);
-				var userid = $pmWindow.getAttribute('data-userid');
+				var userid = el.getAttribute('data-userid');
 				var name = $pmWindow.data('name');
 				if (data.challengesFrom[userid]) {
 					var format = data.challengesFrom[userid];
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -697,7 +697,7 @@
 			var self = this;
 			this.$('.pm-window').each(function (i, el) {
 				var $pmWindow = $(el);
-				var userid = '' + $pmWindow.data('userid');
+				var userid = $pmWindow.getAttribute('data-userid');
 				var name = $pmWindow.data('name');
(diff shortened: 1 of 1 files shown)


js/client-mainmenu.js
```diff
@@ -727,6 +727,9 @@
 								$challenge.html('<form class="battleform"><p>The challenge was cancelled.</p><p class="buttonbar"><button name="dismissChallenge">OK</button></p></form>');
 							}
 						} else if ($challenge.find('button[name=cancelChallenge]').length && challengeToUserid !== userid) {
+							//console.log($challenge.find('button[name=cancelChallenge]').length);
+							//console.log(challengeToUserid);
+							//console.log(userid);
```
AnnikaCodes, line 732:
```suggestion
```

Zarel:
Ugh, that's annoying. I think it'd be best if we stopped using jQuery's `data` entirely. I hate APIs that try to be "magic" because it leads to bugs like this.

js/client-mainmenu.js
```diff
@@ -697,7 +697,7 @@
 			var self = this;
 			this.$('.pm-window').each(function (i, el) {
 				var $pmWindow = $(el);
-				var userid = $pmWindow.data('userid');
+				var userid = '' + $pmWindow.data('userid');
```
Zarel, line 700:
```suggestion
				var userid = el.dataset && el.dataset.userid;
```

Zarel:
I'd suggest refactoring all the other `.data` calls into `getAttribute`, which should solve this problem.

DaWoblefet:
I attempted to do a proper refactor and ended up breaking the client's functionality pretty badly. I've done a good deal of playtesting with the 'false' account and haven't noticed any other unusual behavior, so if a refactor should be done, that's beyond the scope of what I want to accomplish in this PR.

Zarel:
Sure, we can leave it like this.

PR #1674, opened 2020-11-20
Fix the teambuilders of various OMs

KrisXV:
- Fixes a bug where every time you re-click on the ability box for a set in AAA/Hackmons/Metronome Battle, it continues to load the list of abilities until eventually none remain.
- Fixed moves from Alola/Galar formes showing up on the base species of a Pokemon (Dark moves showing on Kanto Moltres when they shouldn't)
- Properly shows moves for both Sketch and Balanced Hackmons


Zarel:
Thanks!

PR #1675, opened 2020-11-20
Usermodlog: Hide emails from lower staff

AnnikaCodes:


website/usermodlog.php
```diff
@@ -153,6 +156,12 @@
 		$entry = $row['entry'];
 		$fromindex = strpos($entry, " from: ");
 		if ($fromindex !== false) $entry = substr($entry, 0, $fromindex);
+
+		if (!$upperstaff) {
+			// Hide email addresses from lowerstaff
+			$entry = preg_replace($EMAIL_REGEX, "", $entry);
```
Zarel, line 162:
```suggestion
			$entry = preg_replace($EMAIL_REGEX, "[email address hidden]", $entry);
```

Should make it easier to tell what happened and also to find bugs if they appear.

Zarel:
Thanks!

PR #1677, opened 2020-11-22
Actually remove healing berries after they're eaten

MacChaeger:
~I think this should also fix a rare case where a Pokemon A's Sticky Barb is stolen by an opponent through an effect like Magician, then Pokemon A gets the Sticky Barb back via Sticky Barb's effect (which has no message), because it looks like the old code would not have set Pokemon A's item to be a Sticky Barb again after the damage effect since it's previously held item was also a Sticky Barb.~ Edit: nah, that wasn't a problem to begin with.

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -1519,6 +1519,7 @@ class Battle {
 		if (kwArgs.then) this.waitForAnimations = false;
 		if (kwArgs.simult) this.waitForAnimations = 'simult';
 
+		const CONSUMED = ['eaten', 'popped', 'consumed', 'held up'];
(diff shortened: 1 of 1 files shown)


Zarel:
Interesting, I believe this fixes the problem with no negative consequences, but it _does_ make it harder for other clients to keep up. I wonder if we should add a `|[from]item: Sitrus Berry|[consumed]` to mean "this isn't a currently-held item"?

Zarel:
lol shit I merged the commit summary with the grammar error, oh well

Zarel:
Thanks!

PR #1678, opened 2020-11-22
Tooltips: fix phantom Sticky Barbs for players

MacChaeger:
Should fix [this bug](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8619891), if only from the perspective of the player controlling the affected Pokemon.


Zarel:
Thanks!

PR #1682, opened 2020-11-25
Tooltips: Always show previous item if it exists

MacChaeger:


Zarel:
Thanks! We might want to make this DRYer at some point, but this is a good fix for now.

PR #1683, opened 2020-11-26
Fix tooltips for switchables after rejoining

MacChaeger:
I checked through again and didn't find any more hard-coded indexing of `battle.sides`, so hopefully this is the last holdover.


Zarel:
ty!

PR #1684, opened 2020-11-26
Add <spotify> tag

sirKiraUzumaki:
Adding a <spotify> tag to embed song from Spotify to ps.

Example: 
umtrBapwb6ll

Just like <youtube> tag, it will play songs on ps.

### DIFF
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -810,8 +810,8 @@ class BattleLog {
 				const src = getAttrib('src');
 				const songId = /(?:\?v=|\/track\/)([A-Za-z0-9]+)/.exec(src)[1];
 						
-            return {
-				   tagName: 'iframe',
+				return {
+					tagName: 'iframe',
 					attribs: ['src', `https://open.spotify.com/embed/track/${songId}`,'width', '300', 'height', '380', 'frameborder', '0', 'allowtransparency', 'true', 'allow', 'encrypted-media'],
 				};
 		   } else if (tagName === 'youtube') {
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -805,7 +805,16 @@ class BattleLog {
 				const color = this.usernameColor(toID(getAttrib('name')));
 				const style = getAttrib('style');
 				setAttrib('style', `${style};color:${color}`);
-			} else if (tagName === 'youtube') {
+			} else if (tagName === 'spotify') {
+				// <iframe src="https://open.spotify.com/embed/track/6aSYnCIwcLpnDXngGKAEzZ" width="300" height="380" frameborder="0" allowtransparency="true" allow="encrypted-media"></iframe>
+				const src = getAttrib('src');
+				const songId = /(?:\?v=|\/track\/)([A-Za-z0-9]+)/.exec(src)[1];
+						
+            return {
+				   tagName: 'iframe',
+					attribs: ['src', `https://open.spotify.com/embed/track/${songId}`,'width', '300', 'height', '380', 'frameborder', '0', 'allowtransparency', 'true', 'allow', 'encrypted-media'],
+				};
+		   } else if (tagName === 'youtube') {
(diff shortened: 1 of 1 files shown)


src/battle-log.ts
```diff
@@ -805,7 +805,16 @@ class BattleLog {
...
-			} else if (tagName === 'youtube') {
+			} else if (tagName === 'spotify') {
+				// <iframe src="https://open.spotify.com/embed/track/6aSYnCIwcLpnDXngGKAEzZ" width="300" height="380" frameborder="0" allowtransparency="true" allow="encrypted-media"></iframe>
+				const src = getAttrib('src');
+				const songId = /(?:\?v=|\/track\/)([A-Za-z0-9]+)/.exec(src)[1];
+						
+            return {
+				   tagName: 'iframe',
+					attribs: ['src', `https://open.spotify.com/embed/track/${songId}`,'width', '300', 'height', '380', 'frameborder', '0', 'allowtransparency', 'true', 'allow', 'encrypted-media'],
+				};
```
KrisXV, line 816:
```suggestion
				return {
					tagName: 'iframe',
					attribs: ['src', `https://open.spotify.com/embed/track/${songId}`,'width', '300', 'height', '380', 'frameborder', '0', 'allowtransparency', 'true', 'allow', 'encrypted-media'],
				};
```

src/battle-log.ts
```diff
@@ -805,7 +805,16 @@ class BattleLog {
 				const color = this.usernameColor(toID(getAttrib('name')));
 				const style = getAttrib('style');
 				setAttrib('style', `${style};color:${color}`);
-			} else if (tagName === 'youtube') {
+			} else if (tagName === 'spotify') {
+				// <iframe src="https://open.spotify.com/embed/track/6aSYnCIwcLpnDXngGKAEzZ" width="300" height="380" frameborder="0" allowtransparency="true" allow="encrypted-media"></iframe>
+				const src = getAttrib('src');
+				const songId = /(?:\?v=|\/track\/)([A-Za-z0-9]+)/.exec(src)[1];
```
KrisXV, line 811:
```suggestion
				const songId = /(?:\?v=|\/track\/)([A-Za-z0-9]+)/.exec(src)?.[1];
```

src/battle-log.ts
```diff
@@ -805,7 +805,16 @@ class BattleLog {
...
+						
+            return {
+				   tagName: 'iframe',
+					attribs: ['src', `https://open.spotify.com/embed/track/${songId}`,'width', '300', 'height', '380', 'frameborder', '0', 'allowtransparency', 'true', 'allow', 'encrypted-media'],
+				};
+		   } else if (tagName === 'youtube') {
```
Zarel, line 817: The indentation on this line is wrong, too, which is what's leading to it being part of the diff:

```suggestion
			} else if (tagName === 'youtube') {
```

src/battle-log.ts
AnnikaCodes: There should probably be a check to not return a tag if `songId` is undefined. I'm not familar with the tag parser but https://open.spotify.com/embed/track/undefined doesn't seem like a very valid URL :p

src/battle-log.ts
```diff
@@ -805,6 +805,15 @@ class BattleLog {
 				const color = this.usernameColor(toID(getAttrib('name')));
 				const style = getAttrib('style');
 				setAttrib('style', `${style};color:${color}`);
+			} else if (tagName === 'spotify') {
+				// <iframe src="https://open.spotify.com/embed/track/6aSYnCIwcLpnDXngGKAEzZ" width="300" height="380" frameborder="0" allowtransparency="true" allow="encrypted-media"></iframe>
+				const src = getAttrib('src');
```
sirKiraUzumaki, line 810:
```suggestion
				const src = getAttrib('src') || '';
```

Zarel:
Looks good; thanks!

PR #1686, opened 2020-11-29
Update Battle class API

Zarel:
The client Battle class API has been pretty old and crusty, so this updates it to be saner.

The constructor now takes an options object. Any setting you'd want to initialize with is now a constructor option, instead of needing to call methods after the constructor.

(Deprecated settings `roomid` and `joinButtons` still need to be set separately.)

The old callback system is removed. It's replaced with a subscription system vaguely resembling `PSStreamModel`. Any callbacks only intended to be used by the warstory generator are removed (anyone who wants to write their own warstory generator should extend `BattleSceneStub` instead).

Battles no longer start paused. You can still start them paused by passing `paused: true` as an option.

Playback state tracking had a bunch of rearrangement:

- `playbackState` no longer exists; state should be directly read from `paused`, `atQueueEnd`, `turn`, and `seeking`.

- `turn` is now initialized to `-1`. `-1` now means "we haven't reached `|teampreview|` or `|start|` yet". Reaching those sets turn to `0`.

- "Fast forwarding" and "seeking" are now consistently named "seeking".
  - `seeking` tracks seek state; changes from `fastForward`:
    - `null` means not seeking (replaces `0`)
    - `0` means seeking the start (replaces `0.5`)
    - `Infinity` means seeking the end (replaces `-1`)
  - `fastForward` deprecated and replaced with `seeking`
  - `fastForwardTo()` deprecated and replaced with `seekTurn()`

- `resultWaiting` is removed (it's unused)

- The "activity queue" has been renamed the "step queue", which means some renamed properties:
  - `activityQueue` to `stepQueue`
  - `activityStep` to `currentStep`
  - `nextActivity()` to `nextStep()`

- new property: `atQueueEnd` to track if animation has caught up to the end of the step queue (replaces checking `playbackState`)

- new property/option: `isReplay` - will automatically set `ended` when reaching the end of a replay (stopping music and showing a message), if the replay was saved before the end of the battle

- both replay players (`replay.pokemonshowdown.com` and downloaded files) have been rewritten to use an observer system, instead of the previous manual updating

- `reset(true)` has been renamed `resetStep()`

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -8,6 +8,9 @@
 		maxWidth: 1180,
 		initialize: function (data) {
 			this.me = {};
+			this.choice = undefined;
+			/** are move/switch/team-preview controls currently being shown? */
+			this.controlsShown = false;
 
 			this.battlePaused = false;
 			this.autoTimerActivated = false;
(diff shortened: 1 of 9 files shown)


Zarel:
This apparently has a delayed start on randbats which needs to be investigated at some point.

Zarel:
Okay, I think I fixed everything this time around.

PR #1687, opened 2020-11-29
Properly interpret Max Moves from base move type

MacChaeger:
This covers the rest of the missing cases where a tooltip should show the true max move. For example:
- Pixilate Sylveon's Max Strike from Hyper Voice will show Max Starfall
- Galvanize Toxtricity-Gmax's Max Strike from Hyper Voice will show G-Max Stun Shock
- Refrigerate Genesect's Max Strike from Techno Blast will show Max Hailstorm
- Normalize Bug Memory Silvally's Max Strike from Multi Attack will show Max Strike.

This also changes the type shown by the max move buttons themselves to match the true max move.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -333,7 +333,7 @@
 			var switchables = [];
 			if (this.request) {
 				// TODO: investigate when to do this
-				this.updateSide(this.request.side);
+				this.updateSide();
 
 				act = this.request.requestType;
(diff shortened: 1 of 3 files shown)


Zarel:
Looks great, thanks!

TheImmortal:
This changes Max Guard to whatever type the base Status move is. In-game, Max Guard is always Normal.

PR #1688, opened 2020-11-30
Fix teambuilder for the downloadable client

MacChaeger:


MacChaeger:
It seems a little weird to have a function that specifically does this and also have three other functions that use almost-but-not-quite identical code to do the same thing.

js/storage.js
```diff
@@ -1532,10 +1532,16 @@ Storage.nwLoadTeamFile = function (filename, localApp) {
...
 	if (bracketIndex >= 0) {
 		format = line.slice(1, bracketIndex);
+		if (format && format.slice(0, 3) !== 'gen') format = 'gen6' + format;
+		if (format && format.endsWith('-box')) {
+			format = format.slice(0, -4);
+			capacity = 24;
+		}
 		line = $.trim(line.slice(bracketIndex + 1));
 	}
 	if (format && format.slice(0, 3) !== 'gen') format = 'gen6' + format;
```
urkerab, line 1547: Even weirder is this line which appears to have no effect (but obviously your copy is significant).

MacChaeger: gen 6 formats didn't have a gen prefix in gen 6, so somebody with _really_ old teams might have some that this line hasn't fixed yet.

urkerab: Oh, I was confused by the patch. The problem with this line is actually that it's obsoleted by your copy, since there's no other way to have a format by this point.

MacChaeger: Oh, I see.

Zarel:
```suggestion
```

js/storage.js
```diff
@@ -1532,10 +1532,16 @@ Storage.nwLoadTeamFile = function (filename, localApp) {
 		return;
 	}
 
-	var format = '';
+	var format = 'gen8';
+	var capacity = 6;
 	var bracketIndex = line.indexOf(']');
 	if (bracketIndex >= 0) {
 		format = line.slice(1, bracketIndex);
+		if (format && format.slice(0, 3) !== 'gen') format = 'gen6' + format;
```
Zarel, line 1540:
```suggestion
		if (format && format.startsWith('gen')) format = 'gen6' + format;
```

Zarel:
Yeah I agree that's messed up.

Zarel:
It's the sort of thing I plan to fix after Preact. :s

alaserdolphin:
The bug where it was not possible to directly add Pokemon to older teams is fixed, but the "Box Bug" is not.

That issue still remains. In addition (and I can submit a new issue if that's preferred), all of my teams became "gen6<format>. Everything turned into gen6gen8ou, gen6gen3ou, etc. I was able to fix the ones I attempted by manually going through each team and reselecting their formats. That said, any formats that are past OMs, etc. could only be fixed through manually editing the titles of the text files for those teams. 
Obviously of course I'd also prefer to not have to redo every team, but I understand if that's unavoidable.
[
AllTeamsGen6Bug
](url)

Zarel:
I pushed a fix for the `gen6gen` issue.

Zarel:
@alaserdolphin Can you clarify which bug the "Box bug" is? Is it one of the two issues I closed reffing this PR? If so, could you comment in that issue so I can reopen it?

alaserdolphin:
@Zarel 

The bug (and it's visible in the included image with two boxes (121 and 122) that are both doing it) where the visual frame that normally surrounds a team doesn't "stay sized up" to hold teams larger than 6 Pokémon. The pokemon after the first six eventually overflow and will display on top of/below the other teams.

Zarel:
Thanks.

PR #1693, opened 2020-12-08
BattleRoom: Fix bug with autotimer

AnnikaCodes:
When a `BattleRoom` is initialized, `this.users` is empty, so we can't check if we're the player. The `p1` trick may be a little hacky, but I don't know what a better place for the timer activation is.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -10,6 +10,7 @@
 			this.me = {};
 
 			this.battlePaused = false;
+			this.autoTimerActivated = false;
 
 			this.isSideRoom = Dex.prefs('rightpanelbattles');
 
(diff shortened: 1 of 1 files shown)


js/client-battle.js
```diff
@@ -218,6 +213,12 @@
 				} else if (logLine.substr(0, 6) === '|chat|' || logLine.substr(0, 3) === '|c|' || logLine.substr(0, 4) === '|c:|' || logLine.substr(0, 9) === '|chatmsg|' || logLine.substr(0, 10) === '|inactive|') {
 					this.battle.instantAdd(logLine);
 				} else {
+					if (Dex.prefs('autotimer') && logLine.startsWith('|player|p1')) {
+						// Autostart the timer when we get the first |player| message,
+						// so that we can check if we're a player or not — the users will have joined by now
+						var user = this.users[app.user.get('userid')];
+						if (user && user.group === '\u2606') this.setTimer('on');
+					}
```
Zarel, line 221: You could do it on the `receiveRequest` – only players receive requests. And then maybe set some state so the timer doesn't get turned on again?

Zarel:
ty!

PR #1696, opened 2020-12-19
Update forme cries

DaWoblefet:
I also rearranged them alphabetically to organize them a bit. Here are the cries themselves (credit to dataminer mattyoukhana (@sora10pls), I just converted to mp3 and added the names of the mons): https://drive.google.com/drive/folders/1DWp3k2jsViSH64ivH-2y1EYQ5sqoPGwG


DaWoblefet:
Marty pointed out this is a good uptick in bandwidth (~9.17 MB -> ~15.9 MB). I'm going to do some compression and update with a different Google drive folder.

PR #1697, opened 2020-12-20
Teambuilder/Battle/Sprite support for mods

scoopapa:
EDIT: This is now mostly ready to go. I'm still willing to refactor anything that needs it. There is one issue that I wanted to ask a dev about: where should custom sprites go? Right now it looks for sprites on the server at data/mods/[modName]/sprites but I want to double check before I finalize that aspect.


build-tools/build-indexes
```diff
@@ -9,7 +9,7 @@ const rootDir = path.resolve(__dirname, '..');
 process.chdir(rootDir);
 
 if (!fs.existsSync('data/pokemon-showdown')) {
-	child_process.execSync('git clone https://github.com/smogon/pokemon-showdown.git', {
+	child_process.execSync('git clone https://github.com/scoopapa/DH.git pokemon-showdown', {
```
AnnikaCodes, line 12:
```suggestion
	child_process.execSync('git clone https://github.com/smogon/pokemon-showdown.git pokemon-showdown', {
```

build-tools/build-indexes
```diff
@@ -1017,7 +1016,216 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
+			const modData = modDex.data;
+			let modGen = 0;
+			BattleTeambuilderTable.ClientMods[modid].formats = [];
+			for (let formatid in Dex.data.Formats) { //Find formats using this mod
+				let format = Dex.getFormat(formatid);
+				if (format.mod === modid) {
+					if (formatid.slice(0, 3) === 'gen' && formatid.charAt(3) !== '8') modGen = (formatid.slice(0, 4));
+					BattleTeambuilderTable.ClientMods[modid].formats.push(format.fullname);
+				}
+			}
```
AnnikaCodes, line 1036:
```suggestion
			// Find formats using this mod
			for (const formatid in Dex.data.Formats) {
				const format = Dex.getFormat(formatid);
				if (format.mod === modid) {
					if (formatid.startsWith('gen') && formatid.charAt(3) !== '8') modGen = formatid.slice(0, 4);
					BattleTeambuilderTable.ClientMods[modid].formats.push(format.fullname);
				}
			}
```

build-tools/build-indexes
```diff
@@ -1017,7 +1016,216 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
+			const standardTiers = ['uber', 'ou', 'uubl', 'uu', 'rubl', 'ru', 'nubl', 'nu', 'publ',
+				'pu', 'zu', 'nfe', 'lcuber', 'lc', 'cap', 'caplc', 'capnfe', 'ag', 'duber', 'dou',
+				'dbl', 'duu', 'dnu', 'illegal', 'unreleased'];
+			if (ClientMods[modid].customTiers) BattleTeambuilderTable.ClientMods[modid].customTiers = ClientMods[modid].customTiers;
+			else BattleTeambuilderTable.ClientMods[modid].customTiers = [];
```
AnnikaCodes, line 1041:
```suggestion
			if (ClientMods[modid].customTiers) {
				BattleTeambuilderTable.ClientMods[modid].customTiers = ClientMods[modid].customTiers;
			} else {
				BattleTeambuilderTable.ClientMods[modid].customTiers = [];
			}
```

build-tools/build-indexes
```diff
@@ -1017,7 +1016,216 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
+			for (const speciesid in modData.FormatsData) { //Find any nonstandard tiers used by this mod
+				const tier = modData.FormatsData[speciesid].tier;
+				if (typeof tier === 'undefined') continue;
+				if (!standardTiers.includes(toID(tier)) &&
+					!BattleTeambuilderTable.ClientMods[modid].customTiers.includes(tier)) {
+					BattleTeambuilderTable.ClientMods[modid].customTiers.push(tier);
+				}
+			}
```
AnnikaCodes, line 1049:
```suggestion
			// Find any nonstandard tiers used by this mod
			for (const speciesid in modData.FormatsData) {
				const tier = modData.FormatsData[speciesid].tier;
				if (tier === undefined) continue;
				if (!standardTiers.includes(toID(tier)) &&
					!BattleTeambuilderTable.ClientMods[modid].customTiers.includes(tier)) {
					BattleTeambuilderTable.ClientMods[modid].customTiers.push(tier);
				}
			}
```

AnnikaCodes: Although tbh it's probably easiest to just check `tier &&` in the second `if` statement

build-tools/build-indexes
scoopapa: will do, was only keeping it in case someone wanted to download the files to verify that it works

build-tools/build-indexes
```diff
@@ -1017,7 +1016,216 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
+			BattleTeambuilderTable[modid].tiers = tiers;
+			BattleTeambuilderTable[modid].formatSlices = formatSlices;
+			const modTiers = BattleTeambuilderTable.ClientMods[modid].customTiers;
+			let tierOrder = ["OU", "AG", "Uber", "(Uber)", "OU", "(OU)", "UUBL", "UU", "RUBL", "RU", "NUBL", "NU", "PUBL", "PU", "(PU)", "New", "NFE", "LC Uber", "LC", "Unreleased"];
+			if (typeof modTiers !== "undefined") {
+				if (ClientMods[modid].excludeStandardTiers) tierOrder = modTiers;
+				else tierOrder = modTiers.concat(tierOrder);
+			}
+			for (const tier of tierOrder) {
+				if (tier in {OU:1, AG:1, Uber:1, UU:1, "(UU)":1, RU:1, NU:1, "(NU)":1, PU:1, "(PU)":1, NFE:1, LC:1, DOU:1, DUU:1, "(DUU)":1, New:1, Legal:1, Regular:1, "Restricted Legendary":1, "CAP LC":1}) {
```
AnnikaCodes, line 1075: This should be 1) an array or Set, and 2) a constant variable

build-tools/build-indexes
```diff
@@ -1017,7 +1016,216 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
+					}
+					if (modEntry.abilities['0'] !== baseEntry.abilities['0']) {
+						overrideAbility[id] = modEntry.abilities['0'];
+					}
+					if (modEntry.abilities['1'] !== baseEntry.abilities['1']) {
+						overrideSecondAbility[id] = modEntry.abilities['1'];
+					}
+					if (modEntry.abilities['H'] !== baseEntry.abilities['H']) {
+						overrideHiddenAbility[id] = modEntry.abilities['H'];
+					}
```
AnnikaCodes, line 1131:
```suggestion
					for (const ability of ['0', '1', 'H']) {
						if (modEntry.abilities[ability] !== baseEntry.abilities[ability]) {
							overrideAbility[id] = modEntry.abilities[ability];
						}
					}

File: build-tools/build-indexes [hunk 7]
```diff
@@ -1017,7 +1016,216 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
+			}
+			//learnsets
+			const learnsets = {};
+			BattleTeambuilderTable[modid].learnsets = learnsets;
+			for (const id in modDex.data.Learnsets) {
+				const learnset = modDex.data.Learnsets[id].learnset;
+				if (!learnset) continue;
+				learnsets[id] = {};
+				for (const moveid in learnset) {
+					learnsets[id][moveid] = '012345678qgp';
```
AnnikaCodes, line 1145: What is this about?

build-tools/build-indexes
```diff
@@ -1017,7 +1016,216 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
+					items.push(id);
+					fullItemName[id] = modData.Items[id].name;
+				}
+				if (fakeItem || (modEntry.shortDesc || modEntry.desc) !== (baseEntry.shortDesc || baseEntry.desc)) {
+					overrideItemDesc[id] = (modEntry.shortDesc || modEntry.desc);
+				}
+			}
+			let itemTable = BattleTeambuilderTable.items;
+			if (modGen) itemTable = BattleTeambuilderTable[modGen].items;
+			for (let i = 1; i < itemTable.length; i++) items.push(itemTable[i]);
```
AnnikaCodes, line 1170:
```suggestion
			items.push(...itemTable.slice(1));
```

AnnikaCodes (changes requested):
I recommend that you read over the [contributing and style guide](https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md).

scoopapa:
Yikes sorry, i'm not very good with github

build-tools/build-indexes
scoopapa: A catchall to make sure moves would legal... possibly something I tried when trying to fix an issue and never changed back. If the proper learnsets string is available we could pass that. Move compatibility isn't normally a thing in mods but this would mess that up if it ever was.

build-tools/build-indexes
scoopapa: you want me to make tierOrder a const?

scoopapa:
I'll read over the style guide, thanks for all the help

build-tools/build-indexes
scoopapa: Just to clarify, another part of the code checks this string for characters related to different generations or offshoots that a move might be from. 

I mentioned a potential problem with move compatibility but that is not an issue; the teambuilder already does not take it into account (the validator handles that). 

The fact that it is loading the list of moves from the mod in question also means that we can assume any moves being added to learnsets[id] should be showing up in the teambuilder no matter what. I verified that it works for past gen mods, gen8 OU-based mods, and natdex-based mods. All of them get the right list of moves. Nothing special had to be done, so I'm pretty sure this will load movesets properly for all mods, provided the mod itself functions properly.

build-tools/build-indexes
scoopapa: This doesn't work because overrideAbility only overrides the first ability. The pre-existing code only checks for overrideAbility and overrideHiddenAbility in BattleTeambuilderTable. overrideSecondAbility is something I added. If you don't think it should be done that way, I'll have to change the way past gen abilities are done as well (willing to, just let me know).

Zarel:
Hm. Unfortunately I've been trying to decrease reliance on teambuilder-tables, so this will be complicated to integrate.

scoopapa:
> Hm. Unfortunately I've been trying to decrease reliance on teambuilder-tables, so this will be complicated to integrate.

Ah, I thought I might have made too much use of that. Just to get an idea of what I might try to change, are you trying to reduce the amount of data in these tables, or are you trying to clean up the code? I happy to refactor things to make this easier.

As I mentioned above, I think the ModConfig table in build-indexes and the BattleTeambuilderTable.ModConfig table could probably be done away with, given the main server has fewer mods and only has working mods. If we did that, we could probably remove references to BattleTeambuilderTable (at least the ones that I added) outside of build-indexes and src/battle-dex.ts. Not sure if that helps.

scoopapa:
This (or a version very similar to it) has been used and tested for many weeks here: https://dragonheaven.herokuapp.com/
Some people have come from the main server to use its support for Megas For All, the current PMOTS.

This now has:
- Teambuilder support for changed and custom Items / Abilities / Moves / Pokemon / Learnsets / Tiers
- Search support, with custom features only showing up in their respective mods
- Battle UI and tooltip support to display everything properly for the mod during battle
- Sprite support for Pokemon, Pokemon Icons, Type Icons, and Item Icons

Please let me know if you need me to refactor anything. I have generally minimized the size of teambuilder-tables, as mentioned before, and moved some data into other build tables with more appropriate names. I'm willing to do a lot if you'll point me in the right direction, but I also understand if this is too much to change with Zarel's current overhaul of the client.

PR #1699, opened 2020-12-22
Add a horizontal scroll bar for "infobox-limited" class.

sirKiraUzumaki:
Ig infobox-limited class is in room intros and in the event box. On mobile phones, it is hard to check the whole room intro or event and you can't slide because overflow for the x axis is hidden. So just removing overflow-x: hidden.

### DIFF
--- a/style/battle-log.css
+++ b/style/battle-log.css
@@ -555,7 +555,6 @@ details.readmore[open] summary:hover:before {
 .infobox-limited {
 	max-height: 200px;
 	overflow: auto;
-	overflow-x: hidden;
 }
 .broadcast-red {
 	background-color: #AA5544;


DaWoblefet:
Mobile room intros are one of the worst things on the site, since you can't use any custom CSS to make them responsive. I would love this change.

Zarel:
Can't you? In theory we have Caja configured to allow CSS. Does it not work for `<style>` tags?

Zarel:
(btw, thanks!)

DaWoblefet:
It keeps it in the source, @Zarel, but won't actually show up in the roomintro. For example,

``/roomintro <style>
.test-color {
color: blue;
}
</style>
<div class = "test-color">this text should be blue</div>``

image


image


It doesn't include the new style tags in the DOM at all:

image

Zarel:
Unfortunate. :(

PR #1703, opened 2021-01-03
Muting Features

Distrib-ps:
Pr for https://www.smogon.com/forums/threads/muting-ps-chatrooms.3675505/

### DIFF
--- a/js/client-topbar.js
+++ b/js/client-topbar.js
@@ -447,6 +447,7 @@
 			'change input[name=blockpms]': 'setBlockpms',
 			'change input[name=inchatpm]': 'setInchatpm',
 			'change input[name=dark]': 'setDark',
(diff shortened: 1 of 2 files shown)


Zarel:
I think the request was asking for per-room notification settings, not global notification settings.

Distrib-ps:
Oh yes mb, i watch during the week.

sirKiraUzumaki:
The best we can do is that a command which will mute the room in the room in which the command is called.

tps48:
With permission from @Distrib-ps , I'm taking over on this and working on a per-room notification setting. When that's finished up, should I make a separate PR?

Zarel:
Looks like this is being superseded by #1706.

We can always reopen this if that falls through.

PR #1706, opened 2021-01-07
Chatroom-specific mute options

tps48:
Pull request for [this suggestion](https://www.smogon.com/forums/threads/muting-ps-chatrooms.3675505/). Was originally attempted [here](https://github.com/smogon/pokemon-showdown-client/pull/1703), but that ended up implementing a global notification setting instead of a chatroom-specific one.

The original suggestion proposed using either a right-click on the room tab or a chat command to do this. I opted for the former, mostly because I couldn't think of a succinct command name for this particular behavior, though if my current solution is deemed too clunky we should probably use a chat command instead. Right now a right-click on a chatroom tab brings up a popup that has a "Hide new message indicator" option.

Right-click submenu

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1738,7 +1738,7 @@
 			} else if (mayNotify && this.id.substr(0, 5) === 'help-') {
 				this.notifyOnce("Help message from " + name, "\"" + message + "\"", 'pm');
 			} else if (mayNotify && name !== '~') { // |c:|~| prefixes a system message
-				if(!Dex.prefs('chatmute' + this.title)) {
+				if (!Dex.prefs('chatmute' + this.title)) {
 					this.subtleNotifyOnce();
 				}
 			}
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1727,7 +1727,9 @@
 			} else if (mayNotify && this.id.substr(0, 5) === 'help-') {
 				this.notifyOnce("Help message from " + name, "\"" + message + "\"", 'pm');
 			} else if (mayNotify && name !== '~') { // |c:|~| prefixes a system message
-				this.subtleNotifyOnce();
+				if(!Dex.prefs('chatmute' + this.title)) {
+					this.subtleNotifyOnce();
+				}
 			}
 
 			if (message.slice(0, 4) === '/me ' || message.slice(0, 5) === '/mee') {
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1738,7 +1738,8 @@
 			} else if (mayNotify && this.id.substr(0, 5) === 'help-') {
 				this.notifyOnce("Help message from " + name, "\"" + message + "\"", 'pm');
 			} else if (mayNotify && name !== '~') { // |c:|~| prefixes a system message
-				if (!Dex.prefs('chatmute' + this.title)) {
+				var mutedChats = Dex.prefs('mutedchats') || {};
+				if (!mutedChats[this.id]) {
 					this.subtleNotifyOnce();
 				}
 			}
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1738,8 +1738,8 @@
 			} else if (mayNotify && this.id.substr(0, 5) === 'help-') {
 				this.notifyOnce("Help message from " + name, "\"" + message + "\"", 'pm');
 			} else if (mayNotify && name !== '~') { // |c:|~| prefixes a system message
-				var mutedChats = Dex.prefs('mutedchats') || {};
-				if (!mutedChats[this.id]) {
+				var mutedRooms = Dex.prefs('mutedrooms') || {};
+				if (!mutedRooms[this.id]) {
 					this.subtleNotifyOnce();
 				}
 			}
--- a/js/client-topbar.js
+++ b/js/client-topbar.js
@@ -1080,7 +1080,7 @@
 			var buf = '';
 			buf += '<p><strong>' + (sourceTab.context.innerText) + ' chat options</strong></p>';
 			buf += '<p><label class="optlabel"><input type="checkbox" name="chatmuted"' + (chatMuted ? ' checked' : '') + '/>Hide new message indicator</label></p>';
-			buf += '<p><button name="close">Close</button></p>';
+			buf += '<p><button name="closeRoom" value="' + roomId + '" aria-label="Leave Room">Leave Room</button></p>';
 			this.$el.html(buf).css('max-width', 200);
 		},
 		events: {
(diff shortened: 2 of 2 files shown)


Zarel:
I don't have anything against right-click functionality, but it's definitely weird for this to be the only one. It should probably at least also have a "Leave & Close" button.

(I would also really really prefer if you all worked on the Preact client instead of piling features onto an outdated client...)

tps48:
Gotcha. I'm brand new to this project, so I wasn't sure which client to add it to. Tomorrow, I will begin working on adding the close button and porting this feature to the Preact client.

Zarel:
Oh, focus on the client everyone's using first. I'm just whining. The Preact client can come later.

tps48:
I added a close button:

Screen Shot 2021-01-07 at 12 04 35 PM

Once this feature is finalized and merged into the old client, I'll open a PR to port it to the Preact client (just so we don't have loose ends to tie up later).

Zarel:
I'm worried this looks like "Close the popup" rather than "Leave room".

This should probably be called "Leave room"? With separate "Forfeit" and "Forfeit & Leave" buttons in battles.

tps48:
Ah, sorry, I misunderstood. That is a "close popup" button. I figured since there is already a close button on the tab itself, it would be redundant to have one in the popup, though I can change that.

Also, currently this menu is only implemented for chat rooms (not battle rooms), as that was what the original request proposed. Should I extend it to battle rooms?

I will also change the implementation to use a table instead of messing with the prefs.

js/client-chat.js
```diff
@@ -1738,7 +1738,9 @@
 			} else if (mayNotify && this.id.substr(0, 5) === 'help-') {
 				this.notifyOnce("Help message from " + name, "\"" + message + "\"", 'pm');
 			} else if (mayNotify && name !== '~') { // |c:|~| prefixes a system message
-				this.subtleNotifyOnce();
+				if (!Dex.prefs('chatmute' + this.title)) {
```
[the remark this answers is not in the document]
tps48: Thanks for the feedback; I've updated my implementation to use a table instead of cluttering the top level with arbitrary text (oops):
Screen Shot 2021-01-07 at 7 22 40 PM

 Also changed the parameter to roomid rather than title.

tps48:
Screen Shot 2021-01-07 at 8 33 12 PM
 
I've changed the "close popup" button to a "leave room" button. I have not implemented this popup for battle rooms, because those rooms don't have subtle notifications when there has been a new chat message. I can implement a separate popup without the "hide new message indicator" and with forfeit buttons for battle rooms, but I feel that may be redundant with current functionality.

Zarel:
Oh, nah, it's because browsers also have "Close Tab" in their right-click menus, even though their tabs have close buttons, too. I think it's just a normal thing to have.

js/client-chat.js
```diff
@@ -1738,7 +1738,10 @@
 			} else if (mayNotify && this.id.substr(0, 5) === 'help-') {
 				this.notifyOnce("Help message from " + name, "\"" + message + "\"", 'pm');
 			} else if (mayNotify && name !== '~') { // |c:|~| prefixes a system message
-				this.subtleNotifyOnce();
+				var mutedChats = Dex.prefs('mutedchats') || {};
```
Zarel, line 1741: I would call this "muted rooms" or "muted panels". I think that would be much less confusing than "muted chats", considering what the rest of the codebase calls them.

js/client-topbar.js
```diff
@@ -1060,4 +1069,34 @@
 		}
 	});
 
+	var MutePopup = this.MutePopup = Popup.extend({
```
Zarel, line 1072: `TabRightClickPopup` I think would be clearer.

js/client-topbar.js
```diff
@@ -1060,4 +1069,34 @@
...
+			buf += '<p><button name="closeRoom" value="' + roomId + '" aria-label="Leave Room">Leave Room</button></p>';
+			this.$el.html(buf).css('max-width', 200);
+		},
+		events: {
+			'change input[name=chatmuted]': 'setChatMute'
+		},
+		setChatMute: function (e) {
+			var chatMuted = !!e.currentTarget.checked;
+			var mutedChats = Dex.prefs('mutedchats') || {};
+			mutedChats[this.roomId] = chatMuted;
```
Zarel, line 1092:
```suggestion
			if (chatMuted) {
				mutedChats[this.roomId] = 1;
			} else {
				delete mutedChats[this.roomId];
			}
```

Zarel: (Just some minor optimization.)

tps48:
Thanks for the feedback! I believe that's everything, unless we want to implement a battle-tab right-click menu as well.

Zarel:
Sorry, Github makes it really easy to accidentally leave review comments pending instead of submitting them. :(

js/client-topbar.js
```diff
@@ -112,7 +114,7 @@
 				}
 				return buf + ' draggable="true"><i class="text">' + BattleLog.escapeFormat(formatid) + '</i><span>' + name + '</span></a><button class="closebutton" name="closeRoom" value="' + id + '" aria-label="Close"><i class="fa fa-times-circle"></i></a></li>';
 			case 'chat':
-				return buf + ' draggable="true"><i class="fa fa-comment-o"></i> <span>' + (BattleLog.escapeHTML(room.title) || (id === 'lobby' ? 'Lobby' : id)) + '</span></a><button class="closebutton" name="closeRoom" value="' + id + '" aria-label="Close"><i class="fa fa-times-circle"></i></a></li>';
+				return buf + 'oncontextmenu="return false;" draggable="true" data-chat="true"><i class="fa fa-comment-o"></i> <span>' + (BattleLog.escapeHTML(room.title) || (id === 'lobby' ? 'Lobby' : id)) + '</span></a><button class="closebutton" name="closeRoom" value="' + id + '" aria-label="Close"><i class="fa fa-times-circle"></i></a></li>';
```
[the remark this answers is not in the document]
tps48: Oh, yes that is a much better solution, thanks!

tps48:
No worries, thanks for the feedback!

js/client-topbar.js
```diff
@@ -242,6 +244,14 @@
 		tablist: function () {
 			app.addPopup(TabListPopup);
 		},
+		showRoomTabRightClickMenu: function (e) {
+			e.preventDefault();
+			if ($(e.currentTarget).data('chat')) {
```
Zarel, line 249:
```suggestion
			if ($(e.currentTarget).data('chat')) {
				e.preventDefault();
```
Right-clicking should never do nothing.

tps48: Ah right, that slipped my mind completely.

PR #1708, opened 2021-01-15
Update Preact Lib for Fragments

aviettran:
To support, for example:

(in loop):
`<preact.Fragment key={key}><h3>...</h3><ul>...</ul></preact.Fragment>`

This change also removes "\<undefined\>" tags from the DOM when Fragments are used.

I grabbed the latest from the CDN (2021/01/15):

https://cdn.jsdelivr.net/npm/preact/dist/preact.min.js


Zarel:
Thanks!

PR #1709, opened 2021-01-16
Preact Client: Ladder

aviettran:
Please review for my approach, code quality, consistency, etc.

Outstanding:

Now Complete:

_- I have a note with the comment `// TO REVIEW` that should be looked at_
_- There is some local ladder stuff I'm not sure how to setup and test and it'll take me a while to code dive it. If someone can at least get me started, it'd save me a significant amount of time._
_- I need to implement and test an error case that occurs when local ladders don't come back with a response_


aviettran:
@Zarel this has reached functional completeness, so please review. Moving out of draft.

src/panel-ladder.tsx
```diff
@@ -0,0 +1,182 @@
+/**
+ * Ladder Panel
+ *
+ * Panel for ladder formats and associated ladder tables.
+ *
+ * @author Guangcong Luo <guangcongluo@gmail.com>
```
Zarel, line 6: Set this to yourself! Or just remove this if you don't want ownership of it.

src/panel-ladder.tsx
```diff
@@ -0,0 +1,182 @@
...
+ * @license AGPLv3
```
Zarel, line 7: Let's license this MIT.

```suggestion
 * @license MIT
```

I wrote a lot of stuff about how I want licensing to work, but the short answer is that I prefer to be the only one with AGPL content, so I can relicense it to MIT if someone requests it.

src/panel-ladder.tsx
```diff
@@ -0,0 +1,182 @@
...
+				<div dangerouslySetInnerHTML={{__html: room.ladderData}}></div>
+			</div>;
+		}
+		return <div class="ladder pad">
+			<p><this.FormatListButton/></p><p><button class="button" name="refresh" onClick={this.handleRequestLadderData}><i class="fa fa-refresh"></i> Refresh</button>
+			<form class="search" onSubmit={this.submitSearch}><input type="text" name="searchValue" class="textbox searchinput" value={BattleLog.escapeHTML(searchValue)} placeholder="username prefix" onChange={this.setSearchValue} /><button type="submit"> Search</button></form></p>
+			<h3>{BattleLog.escapeFormat(selectedFormat)} Top {BattleLog.escapeHTML(lastSearch ? `- '${lastSearch}'` : '500')}</h3>
+			<div dangerouslySetInnerHTML={{__html: room.ladderData}}></div>
+		</div>; // That's right, danger!
+	};
```
Zarel, line 164: Should these all be static?

src/panel-ladder.tsx
```diff
@@ -0,0 +1,182 @@
...
+		const { teams } = PS;
+		const { room } = this.props;
+		const { searchValue, lastSearch } = this.state;
+		const selectedFormat = room.selectedFormat as string;
+		if (room.ladderData === undefined) {
+			return <div class="ladder pad"><p><this.FormatListButton/></p><p><em>Loading...</em></p></div>;
+		} else if (teams.usesLocalLadder) {
+			return <div class="ladder pad">
+				<p><this.FormatListButton/></p>
+				<div dangerouslySetInnerHTML={{__html: room.ladderData}}></div>
```
Zarel, line 155: Isn't this normally sanitized with `sanitizeHTML`?

testclient-beta.html
```diff
@@ -93,6 +93,7 @@ <h3><button class="closebutton" tabindex="-1" aria-label="Close"><i class="fa fa
 	<script src="js/battle-dex-search.js?"></script>
 	<script src="js/battle-searchresults.js?"></script>
 	<script src="js/panel-teambuilder-team.js?"></script>
+	<script src="js/panel-ladder.js?"></script>
```
Zarel, line 96: (This also goes in `preactalpha.template.html`)

src/panel-ladder.tsx
```diff
@@ -0,0 +1,182 @@
...
+		</div>; // That's right, danger!
+	};
+	render() {
+		const room = this.props.room;
+		const { showHelp } = this.state;
+		return <PSPanelWrapper room={room} scrollable>
+			<div class="ladder pad">
+				{showHelp && <this.Help/>}
+				{!showHelp && room.selectedFormat === undefined && <this.ShowFormatList/>}
+				{!showHelp && room.selectedFormat !== undefined && <this.ShowFormat/>}
```
Zarel, line 172: I wonder if this should be separate rooms with separate room IDs.

Like:

`ladder`, `ladder-gen8ou`, `ladderhelp`.

That's how the new teambuilder is set up and I think it'd be better.

src/panels.tsx
```diff
@@ -90,7 +90,7 @@ class PSRouter {
 }
 PS.router = new PSRouter();
 
-class PSRoomPanel<T extends PSRoom = PSRoom> extends preact.Component<{room: T}> {
+class PSRoomPanel<T extends PSRoom = PSRoom, S = {}> extends preact.Component<{room: T}, S> {
```
Zarel, line 93: Room panel state should ideally go into the room model (the `PSRoom` instance). This is necessary because Preact doesn't support retaining state when reparenting a component, which we plan to do (partial implementation in #1641).

src/panel-ladder.tsx
Zarel: We might even want to make a component sanitizing HTML, if it doesn't already exist.

src/panel-ladder.tsx
```diff
@@ -0,0 +1,182 @@
...
+		this.subscriptions.push(PS.teams.subscribe(() => {
+			this.forceUpdate();
+		}));
+	};
+	componentWillUnmount() {
+		for (const subscription of this.subscriptions) {
+			subscription.unsubscribe();
+		}
+		this.subscriptions = [];
+	}
```
Zarel, line 81: Shouldn't `PSRoomPanel` do this automatically?

src/panel-ladder.tsx
```diff
@@ -0,0 +1,182 @@
...
+	lastSearch: string;
+}
+
+class LadderPanel extends PSRoomPanel<LadderRoom, LadderPanelState> {
+	constructor() {
+		super();
+		this.state = { showHelp: false, searchValue: '', lastSearch: '' };
+	}
+	subscriptions: PSSubscription[] = [];
+	componentDidMount = () => {
```
Zarel, line 61: I don't believe this needs to be bound.

src/panel-ladder.tsx
```diff
@@ -0,0 +1,182 @@
...
+	searchValue: string;
+	lastSearch: string;
+}
+
+class LadderPanel extends PSRoomPanel<LadderRoom, LadderPanelState> {
+	constructor() {
+		super();
+		this.state = { showHelp: false, searchValue: '', lastSearch: '' };
+	}
+	subscriptions: PSSubscription[] = [];
```
Zarel, line 60: This should be inherited automatically from `PSRoomPanel`, right?

src/panel-ladder.tsx
```diff
@@ -0,0 +1,182 @@
...
+		}));
+	};
+	componentWillUnmount() {
+		for (const subscription of this.subscriptions) {
+			subscription.unsubscribe();
+		}
+		this.subscriptions = [];
+	}
+	setShowHelp = (showHelp: boolean) => () => this.setState({ showHelp });
+	setSearchValue = (e: Event) => this.setState({ searchValue: (e.currentTarget as HTMLInputElement).value });
```
Zarel, line 83: My plan for the Preact client (which I guess I forgot to document) is for button dispatch to work through commands, so the button would look like `<button name="cmd" value="/searchvalue [number]">` and then the command would be dispatched in `handleMessage`.

Not really a big deal for ladder, but probably a useful thing to remember in general. My plan is that mousing over buttons would display the command in the closest text box, which would be a nice hint for the user in some cases.

.eslintignore
```diff
@@ -29,3 +29,4 @@ node_modules/
 /js/panel-teambuilder-team.js
 /js/panel-teamdropdown.js
 /js/panel-battle.js
+/js/panel-ladder.js
```
Zarel, line 32:
```suggestion
/js/panel-ladder.js

```

src/panels.tsx
```diff
@@ -478,3 +478,9 @@ class PSMain extends preact.Component {
 }
 
 type PanelPosition = {top?: number, bottom?: number, left?: number, right?: number} | null;
+
+class SanitizedHTML extends preact.Component<{ innerHTML: string }> {
+	render() {
+		return <div dangerouslySetInnerHTML={{__html: BattleLog.sanitizeHTML(this.props.innerHTML)}}/>;
+	}
+}
```
Zarel, line 486: Is there a reason this isn't a pure function component?

```suggestion
function SanitizedHTML(props: {children: string}) {
	return <div dangerouslySetInnerHTML={{__html: BattleLog.sanitizeHTML(props.children)}}/>;
}
```

aviettran: Learned something new

src/panel-ladder.tsx
```diff
@@ -0,0 +1,186 @@
...
+			} else {
+				this.requestLadderData();
+			}
+		} else {
+			this.update(null);
+		}
+	};
+	requestLadderData = (searchValue?: string) => {
+		this.setLadderData(undefined); // "Loading..."
+		$.get('/ladder.php', { // TO REVIEW: I imagine this may need to be changed to use PSLoginServer, but that does not have support for prepping GETs with query parameters
```
aviettran, line 51: this comment is to remind myself to update this to use the new net lib

src/panel-ladder.tsx
```diff
@@ -0,0 +1,186 @@
...
+			format: this.selectedFormat,
+			server: Config.server.id.split(':')[0],
+			output: 'html',
+			prefix: toID(searchValue),
+		},
+		this.setLadderData);
```
Zarel, line 57:
```suggestion
		Net('/ladder.php').get({query: {
			format: this.selectedFormat,
			server: Config.server.id.split(':')[0],
			output: 'html',
			prefix: toID(searchValue),
		}}).then(this.setLadderData);
```

aviettran: Thanks!

src/panel-ladder.tsx
```diff
@@ -48,13 +48,12 @@ class LadderRoom extends PSRoom {
 	};
 	requestLadderData = (searchValue?: string) => {
 		this.setLadderData(undefined); // "Loading..."
-		$.get('/ladder.php', { // TO REVIEW: I imagine this may need to be changed to use PSLoginServer, but that does not have support for prepping GETs with query parameters
+		Net('/ladder.php').get({query: {
 			format: this.selectedFormat,
```
Zarel, line 53: tbh maybe we should support `undefined` to mean "leave this off", considering how hard it is to have a property that's only conditionally there.

aviettran: I'm going to have an entirely separate room for the ladder table, so it's also entirely possible that the user joins a room corresponding to a format that does not even exist. I'll have to run through a few scenarios.

src/panel-ladder.tsx
aviettran: Complete

src/panel-ladder.tsx
aviettran: Marking resolved (discussed in Discord)

aviettran:
@Zarel just a reminder that this is ready for review

Zarel:
Thanks! I still want individual ladder pages to be separate pages, but that can wait for a different PR.

PR #1715, opened 2021-01-28
Fix broadcasting /help

mia-pi-git:
:s
(currently attempting to broadcast !help mostly breaks.)


Zarel:
hrm

Zarel:
I'm not sure what this does

Zarel:
But this does seem more consistent

mia-pi-git:
> I'm not sure what this does

This makes it so you can broadcast !help again - before you couldn't, since this would catch it.

PR #1716, opened 2021-01-29
Preact: Page Panel

aviettran:


src/client-main.ts
```diff
@@ -786,7 +786,6 @@ const PS = new class extends PSModel {
 					}
 				}
 				this.update();
-				continue;
```
Zarel, line 789: ?

aviettran: Don't remember why I removed this. I thought I had a good reason, but I don't think I did.

src/panel-page.tsx
```diff
@@ -52,15 +68,64 @@ function PagerLadderHelp(props: { room: PageRoom }) {
...
+				room.setHTMLData(args[2]);
+			}
+			return true;
+		case 'pagehtml':
+			room.setHTMLData(args[1]);
+			return true;
+		}
+	}
 	render() {
 		const { room } = this.props;
```
Zarel, line 123:
```suggestion
		const {room} = this.props;
```

Zarel: We don't have tslint but you know the rules and so do I!

src/panel-page.tsx
```diff
@@ -52,15 +68,64 @@ function PagerLadderHelp(props: { room: PageRoom }) {
...
 		const RenderPage = () => {
```
Zarel, line 124: You definitely seem to define components like `RenderPage` and `PageServerHTML` that I would just have let be functions. Like I would just use a variable for `RenderPage`, and `PageServerHTML` I would probably just implement inline.

src/panel-page.tsx
```diff
@@ -8,17 +8,33 @@
...
+			PS.send(`|/join ${this.id}`);
+			this.connected = true;
+			this.connectWhenLoggedIn = false;
+		}
+	}
 }
 
-function PagerLadderHelp(props: { room: PageRoom }) {
+function PageLadderHelp(props: { room: PageRoom }) {
 	const { room } = props;
```
Zarel, line 38:
```suggestion
	const {room} = props;
```

src/panel-page.tsx
```diff
@@ -8,17 +8,33 @@
...
+		if (!this.connected) {
+			PS.send(`|/join ${this.id}`);
+			this.connected = true;
+			this.connectWhenLoggedIn = false;
+		}
+	}
 }
 
-function PagerLadderHelp(props: { room: PageRoom }) {
+function PageLadderHelp(props: { room: PageRoom }) {
```
Zarel, line 37:
```suggestion
function PageLadderHelp(props: {room: PageRoom}) {
```

src/panel-page.tsx
```diff
@@ -52,15 +68,64 @@ function PagerLadderHelp(props: { room: PageRoom }) {
...
+}
+
 class PagePanel extends PSRoomPanel<PageRoom> {
+	clientRooms: { [key: string]: JSX.Element } = { 'ladderhelp': <PageLadderHelp room={this.props.room}/> };
+
+	/**
+	 * @return true to prevent line from being sent to server
+	 */
+	receiveLine(args: Args) {
+		const { room } = this.props;
```
Zarel, line 89:
```suggestion
		const {room} = this.props;
```

src/panel-page.tsx
aviettran: I have made these changes, but what are the rules exactly?

src/panel-page.tsx
aviettran: Revised.

Zarel:
Thanks!

Issue #1717, opened 2021-01-30
Show Password Button Conflicts with Password Text Field


mashirochan:
Sadly, the new show password button that was just added has a couple issues - it appears to be overlapping the password text field, so it can only be clicked on the very far right of the button. Also, I've noticed that pressing enter now interacts with the show password button, instead of the log in button. This seems to be an inferior interaction, because I'd imagine most users would rather login quickly than show their password quickly.

https://i.imgur.com/eNnuZco.gif

urkerab:
Looks like this is because the button doesn't have the right `type` attribute, so it's defaulting to a submit button. I checked in Developer Tools and adding `type="button"` seems to fix the problem.

Zarel:
Out of curiosity, what browser is this? Firefox?

mashirochan:
Yeah, Firefox 85.0 (64-bit) on Windows 10

Zarel:
Fixed.

PR #1718, opened 2021-01-30
Explicitly mark more non-submit buttons with their type

urkerab:
Anywhere where there's a form with more than one button, I added `type="button"` to all the other buttons. (Unless I missed some...) In particular this should fix issue #1717.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1381,9 +1381,9 @@
 				buf += '<p><button type="submit"><strong>Forfeit</strong></button> ';
 			}
 			if (this.gameType === 'battle' && this.room.battle && !this.room.battle.rated) {
-				buf += '<button name="replacePlayer">Replace player</button> ';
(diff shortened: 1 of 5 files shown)


Zarel:
This is a lot of work for something that usually doesn't make a difference. Are you sure it makes a difference here?

Zarel:
I'm pretty sure PS's dispatcher intentionally doesn't require `type="button"`. Issue #1717 probably has to do more with browser compatibility than with `type="button"`. Have you tested it to make sure the difference exists?

urkerab:
It makes a difference for #1717, because we don't want the reveal password button to be the default button, and I thought I should mark other buttons for consistency, otherwise it might cause confusion in future.

urkerab:
It makes a difference for #1717, because we don't want the reveal password button to be the default button, and I thought I should mark other buttons for consistency, otherwise it might cause confusion in future.

Zarel:
Oh, I see! That's valid.

PR #1719, opened 2021-02-01
Improve banning servers

mia-pi-git:
This makes it so that if a server's name includes the name of a banned server, it's also treated as banned.
(reason for this explained elsewhere)

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -709,7 +709,7 @@ function toId() {
 		connect: function () {
 			if (this.down) return;
 
-			var serverIsBanned = Config.bannedHosts && Config.bannedHosts.includes(Config.server.host);
+			var serverIsBanned = false;
 			if (Config.bannedHosts) {
 				for (var i = 0; i < Config.bannedHosts.length; i++) {
 					var host = Config.bannedHosts[i];
--- a/js/client.js
+++ b/js/client.js
@@ -712,7 +712,7 @@ function toId() {
 			if (Config.bannedHosts) {
 				for (var i = 0; i < Config.bannedHosts.length; i++) {
 					var host = Config.bannedHosts[i];
-					if (Config.server.host.includes(host)) {
+					if (typeof host === 'string' ? Config.server.host === host : host.test(Config.server.host)) {
 						Config.server.banned = true;
 						break;
 					}
--- a/js/client.js
+++ b/js/client.js
@@ -709,7 +709,7 @@ function toId() {
 		connect: function () {
 			if (this.down) return;
 
-			var serverIsBanned = Config.bannedHosts && Config.bannedHosts.indexOf(Config.server.host) >= 0;
+			var serverIsBanned = Config.bannedHosts && Config.bannedHosts.includes(Config.server.host);
 			if (Config.bannedHosts) {
 				for (var i = 0; i < Config.bannedHosts.length; i++) {
 					var host = Config.bannedHosts[i];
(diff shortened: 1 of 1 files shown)


js/client.js
```diff
@@ -709,7 +709,17 @@ function toId() {
 		connect: function () {
 			if (this.down) return;
 
-			if (Config.server.banned || (Config.bannedHosts && Config.bannedHosts.indexOf(Config.server.host) >= 0)) {
+			var serverIsBanned = Config.bannedHosts && Config.bannedHosts.indexOf(Config.server.host) >= 0;
+			if (Config.bannedHosts) {
+				for (let i = 0; i < Config.bannedHosts.length; i++) {
```
Zarel, line 714: Build error is because you're using `let`, which isn't supported by IE7.

Zarel:
```suggestion
				for (var i = 0; i < Config.bannedHosts.length; i++) {
```

js/client.js
```diff
@@ -709,7 +709,17 @@ function toId() {
 		connect: function () {
 			if (this.down) return;
 
-			if (Config.server.banned || (Config.bannedHosts && Config.bannedHosts.indexOf(Config.server.host) >= 0)) {
+			var serverIsBanned = Config.bannedHosts && Config.bannedHosts.indexOf(Config.server.host) >= 0;
```
Zarel, line 712:
```suggestion
			var serverIsBanned = Config.bannedHosts && Config.bannedHosts.includes(Config.server.host);
```

We have a polyfill, though!

js/client.js
```diff
@@ -709,7 +709,17 @@ function toId() {
 		connect: function () {
 			if (this.down) return;
 
-			if (Config.server.banned || (Config.bannedHosts && Config.bannedHosts.indexOf(Config.server.host) >= 0)) {
+			var serverIsBanned = Config.bannedHosts && Config.bannedHosts.includes(Config.server.host);
```
Zarel, line 712:
```suggestion
			var serverIsBanned = false;
```

This is redundant with the for-loop.

js/client.js
```diff
@@ -709,7 +709,17 @@ function toId() {
...
+			if (Config.bannedHosts) {
+				for (var i = 0; i < Config.bannedHosts.length; i++) {
+					var host = Config.bannedHosts[i];
+					if (Config.server.host.includes(host)) {
+						serverIsBanned = true;
+					}
+				}
+			}
+
+			if (Config.server.banned || serverIsBanned) {
```
Zarel, line 722:
```suggestion
			if (Config.bannedHosts) {
				for (var i = 0; i < Config.bannedHosts.length; i++) {
					var host = Config.bannedHosts[i];
					if (Config.server.host.includes(host)) {
						Config.server.banned = true;
						break;
					}
				}
			}

			if (Config.server.banned) {
```

There's no real need for the intermediate variable, either.

js/client.js
```diff
@@ -709,7 +709,17 @@ function toId() {
 		connect: function () {
 			if (this.down) return;
 
-			if (Config.server.banned || (Config.bannedHosts && Config.bannedHosts.indexOf(Config.server.host) >= 0)) {
+			if (Config.bannedHosts) {
+				for (var i = 0; i < Config.bannedHosts.length; i++) {
+					var host = Config.bannedHosts[i];
+					if (Config.server.host.includes(host)) {
```
Zarel, line 715:
```suggestion
					if (typeof host === 'string' ? Config.server.host === host : host.test(Config.server.host)) {
```

Zarel:
Thanks!

PR #1720, opened 2021-02-05
Support a twitch embed

mia-pi-git:
I don't like the replace hack but i'm not sure how else to ensure it doesn't get escaped.

### DIFF
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -809,7 +809,7 @@ class BattleLog {
 					tagName: 'iframe',
 					attribs: [
 						'src', `https://player.twitch.tv/?channel=${channelId}&parent=${location.hostname}`,
-						'allowfullscreen', 'true', 'height', "189", 'width', "310"
+						'allowfullscreen', 'true', 'height', "400", 'width', "340"
 					],
 				};
 			} else if (tagName === 'username') {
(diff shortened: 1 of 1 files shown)


src/battle-log.ts
```diff
@@ -938,7 +950,7 @@ class BattleLog {
 		// allows T, however it's more practical to also allow spaces.
 		return sanitized.replace(
 			/<time>\s*([+-]?\d{4,}-\d{2}-\d{2})[T ](\d{2}:\d{2}(?::\d{2}(?:\.\d{3})?)?)(Z|[+-]\d{2}:\d{2})?\s*<\/time>/ig,
-		this.localizeTime);
+		this.localizeTime).replace(/<(.*)&amp;(.*)>/ig, '<$1&$2>');
```
Zarel, line 953:
```suggestion
		this.localizeTime);
```

:[ This will break the HTML sanitizer and create invalid output. You'll need a different solution to this problem.

Zarel: I'm pretty sure you don't need this, anyway. `<iframe src="&amp;">` does in fact create an iframe with a source of `&`, if you were worried about that.

src/battle-log.ts
```diff
@@ -800,6 +801,17 @@ class BattleLog {
...
 				}
+			} else if (tagName === 'twitch') {
+				// <iframe src="https://player.twitch.tv/?channel=ninja&parent=www.example.com" allowfullscreen="true" height="378" width="620"></iframe>
+				const src = getAttrib('src') || "";
+				const channelId = /(https?:\/\/)?twitch.tv\/([A-Za-z0-9]+)/i.exec(src)?.[2];
+				return {
+					tagName: 'iframe',
+					attribs: [
+						'src', `https://player.twitch.tv/?channel=${channelId}&parent=${location.hostname}`,
+						'allowfullscreen', 'true', 'height', "400", 'width', "340"
```
Zarel, line 812:
```suggestion
						'allowfullscreen', 'true', 'height', "400", 'width', "340",
```

src/battle-log.ts
mia-pi-git: hm, alright

Zarel:
ty!

PR #1727, opened 2021-03-04
Remove objectionable replays from the homepage

AnnikaCodes:
Removes replays with slurs, inappropriate usernames, permabanned users, and/or runMinor errors from the homepage


Zarel:
The errors should be fixed, and the slurs and usernames can be removed. I don't think this should be done in this way.

PR #1728, opened 2021-03-09
Add a button to copy/paste password reset urls

mia-pi-git:
Requested by aeonic.
Dumb hack, but there isn't really a better way to do this, is there?

### DIFF
--- a/website/js/ladder.js
+++ b/website/js/ladder.js
@@ -34,13 +34,12 @@ var LadderPanel = Panels.StaticPanel.extend({
 		// this works since this is in the root as well
 		var url = 'https://' + document.location.hostname + '/resetpassword/' + token;
 		var dummy = document.createElement("input");
-		var placeholder = $('.pfx-topbar')[0];
 		dummy.value = url;
-		dummy.style.position = 'absolute';
-		placeholder.appendChild(dummy);
+		dummy.style.display = 'none';
+		document.body.appendChild(dummy);
 		dummy.select();
 		document.execCommand("copy");
-		placeholder.removeChild(dummy);
+		document.body.removeChild(dummy);
 		e.currentTarget.textContent = 'Copied!';
 	},
 	openReset: function (e) {
--- a/website/js/ladder.js
+++ b/website/js/ladder.js
@@ -26,7 +26,21 @@ var LadderPanel = Panels.StaticPanel.extend({
 	events: {
 		'change select[name=standing]': 'changeStanding',
 		'click button[name=openReset]': 'openReset',
-		'click button[name=cancelReset]': 'cancelReset'
+		'click button[name=cancelReset]': 'cancelReset',
+		'click button[name=copyUrl]': 'copyUrl',
+	},
+	copyUrl: function (e) {
+		var token = e.currentTarget.value;
+		var url = 'https://' + Config.routes.root + '/resetpassword/' + token;
+		var dummy = document.createElement('input');
+		dummy.value = url;
+		dummy.style.display = 'none';
+		var placeholder = $('.pfx-topbar')[0];
+		placeholder.appendChild(dummy);
+		dummy.select();
+		document.execCommand('copy');
+		e.currentTarget.text('Copied!');
+		placeholder.removeChild(dummy);
 	},
 	openReset: function (e) {
 		e.preventDefault();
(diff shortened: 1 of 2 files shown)


website/users.php
```diff
@@ -285,11 +285,30 @@
...
+			dummyLink.id = "dummyLink";
+			dummyLink.value = $("#token").text();
+			dummyLink.style.position = 'absolute';
+			appended.appendChild(dummyLink);
+			dummyLink.select();
+			document.execCommand("copy");
+			appended.removeChild(dummyLink);
+			$('#copy').text('Copied!');
+		}
+		</script>
```
Zarel, line 303: This doesn't work in searches. JS needs to go in `website/js/ladder.js`.

mia-pi-git: Done.

website/users.php
```diff
@@ -289,7 +289,8 @@
 				Use this link:
 			</p>
 			<p style="margin: 1em -13px">
-				<small><code>https://<?= $psconfig['routes']['root'] ?>/resetpassword/<?php echo $token; ?></code></small>
+				<small><code id="token">https://<?= $psconfig['routes']['root'] ?>/resetpassword/<?php echo $token; ?></code></small>
```
Zarel, line 292: IDs are required to be unique by the spec; a static ID can't be unique if it's in a panel.

The best way to do this is to use a class and use a local search. You also want to replace the `onclick` handler with a Backbone click handler in `ladder.js`

Zarel: So find this:

```js
	events: {
		'change select[name=standing]': 'changeStanding',
		'click button[name=openReset]': 'openReset',
		'click button[name=cancelReset]': 'cancelReset'
	},
```

and add your copy button:

```js
	events: {
		'change select[name=standing]': 'changeStanding',
		'click button[name=copyLink]': 'copyLink',
		'click button[name=openReset]': 'openReset',
		'click button[name=cancelReset]': 'cancelReset'
	},
```

Zarel: Then you can follow along how `openReset` works.

mia-pi-git: Okay. Done.

website/js/ladder.js
```diff
@@ -26,7 +26,23 @@ var LadderPanel = Panels.StaticPanel.extend({
 	events: {
 		'change select[name=standing]': 'changeStanding',
 		'click button[name=openReset]': 'openReset',
-		'click button[name=cancelReset]': 'cancelReset'
+		'click button[name=cancelReset]': 'cancelReset',
+		'click button[name=copyUrl]': 'copyUrl',
```
Zarel, line 30:
```suggestion
		'click button[name=copyUrl]': 'copyUrl'
```

We don't have eslint in `website/` but that's no reason to break ES3 compatibility!

mia-pi-git: this does what now
is it really _that_ frail?

mia-pi-git: smh ES3

website/js/ladder.js
```diff
@@ -26,7 +26,23 @@ var LadderPanel = Panels.StaticPanel.extend({
...
+	},
+	copyUrl: function (e) {
+		var token = e.currentTarget.value;
+		// this works since this is in the root as well
+		var url = 'https://' + document.location.hostname + '/resetpassword/' + token;
+		var dummy = document.createElement("input");
+		var placeholder = $('.pfx-topbar')[0];
+		dummy.id = "dummy";
```
Zarel, line 38:
```suggestion
```
:|

While this ID is unlikely to cause harm, it's not impossible, and it's also pointless.

website/js/ladder.js
```diff
@@ -26,7 +26,23 @@ var LadderPanel = Panels.StaticPanel.extend({
...
+		dummy.value = url;
+		dummy.style.position = 'absolute';
```
Zarel, line 39: does `display:none` not work here?

website/js/ladder.js
```diff
@@ -26,7 +26,23 @@ var LadderPanel = Panels.StaticPanel.extend({
...
-		'click button[name=cancelReset]': 'cancelReset'
+		'click button[name=cancelReset]': 'cancelReset',
+		'click button[name=copyUrl]': 'copyUrl',
+	},
+	copyUrl: function (e) {
+		var token = e.currentTarget.value;
+		// this works since this is in the root as well
+		var url = 'https://' + document.location.hostname + '/resetpassword/' + token;
+		var dummy = document.createElement("input");
+		var placeholder = $('.pfx-topbar')[0];
```
Zarel, line 37: Why not just `document.body`?

website/js/ladder.js
mia-pi-git: Probably, i just stole this from the other place we do copying that's not in the main site.

website/js/ladder.js
```diff
@@ -67,7 +81,7 @@ var App = Panels.App.extend({
 		'usersearch/': LadderPanel,
 		'usersearch/:q': LadderPanel,
 		'users/:user/modlog': LadderPanel
-	}
+	},
```
Zarel, line 84:
```suggestion
	}
```

...ES3!

Zarel:
Should be good now. Thanks!

Zarel:
Still surprised people care, though. Triple-click + Cmd+C is what I do. You need your keyboard for Cmd+V anyway.

PR #1729, opened 2021-03-13
Website: Support letting non-admins change specific standings

mia-pi-git:


lib/ntbb-session.lib.php
```diff
@@ -331,6 +331,24 @@ function csrfData() {
...
 	}
 
+	function canChangeStanding($targetid, $newstanding) {
+		global $psconfig, $curuser;
+		if (!isset($psconfig['authconfig'])) return false;
+		if (!$this->csrfCheck()) return false;
+		if ($curuser['group'] === 6) return true; // leaders / admins can change to anything
+		if (!isset($psconfig['authconfig'][$curuser['group']])) return false;
+		$tarUser = $this->getUser($this->userid($targetid));
+		if ($tarUser['group'] >= $curUser['group']) {
```
AnnikaCodes, line 341: I think group 2 is also used for admins?

PR #1731, opened 2021-03-15
Support using form elements to send commands

mia-pi-git:
Might not be the best way to do this, but for some reason the global event listener on `app` didn't work, so i had to trickle it down into ConsoleRoom and HTMLRoom.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -115,7 +115,7 @@
 		submit: function (e) {
 			e.preventDefault();
 			e.stopPropagation();
-			if (e.currentTarget.attributes['data-send']) {
+			if (e.currentTarget.getAttribute('data-submitsend')) {
 				return app.submitSend(e);
 			}
 			var text = this.$chatbox.val();
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -51,7 +51,7 @@
 
 		focus: function (e, focusTextbox) {
 			var target = e && e.target;
-			if (target && target.tagName === 'TEXTAREA') {
+			if (target && ['TEXTAREA', 'INPUT'].includes(target.tagName)) {
 				// this workaround works for iOS 12 but not iOS 13
 				/* if (window.isiOS) {
 					// iOS will not bring up a keyboard unless you manually blur and refocus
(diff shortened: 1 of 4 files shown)


Zarel:
I guess you don't know how event bubbling/capture works. It's okay, the way we used Backbone's event system led to a huge mess, I tried to clean it up but gave up. Your solution works and that's the important part, I'm going to leave it alone.

PartMan7:
Oh, this :D

Would this also allow Bots to add form submission in HTML pages?

mia-pi-git:
> Oh, this :D
> 
> Would this also allow Bots to add form submission in HTML pages?

It would allow bots to have users send commands to PS, yes.

PartMan7:
Including ``/msg``?

mia-pi-git:
> Including `/msg`?

Yes, any command, there are no client-side limits.

Zarel:
Anyway, my main actual problem with this is in the syntax. The way I specified this was something like:

```html
<form data-submitsend="/nick {username}">
  <input name="username" />
  <button type="submit">Change name</button>
</form>
```

You seem to just loop through every element directly in the form and add its value with a comma. That would break utterly if you have anything but input elements in the form, such as paragraphs, or buttons...

Zarel:
...ok we agreed on `data-submitsend` and this clearly uses `data-send` :s

Zarel:
Also, an important thing to do is to clear the form after submitting (design standard D6). Otherwise people can accidentally send messages multiple times if they click twice.

Famous UX problem! https://thedailywtf.com/articles/Push-the-Button

mia-pi-git:
> Also, an important thing to do is to something like close the form after submitting (design standard D6). Otherwise people can accidentally send messages multiple times if they click twice.
> 
> Famous UX problem! https://thedailywtf.com/articles/Push-the-Button

Huh. Thought I did that. Don't know if you're making more changes, do you want to make these two or shall I?

Zarel:
I did them. Oh, also, general note: you should be using `getAttribute` rather than `attributes`.

mia-pi-git:
> you should be using `getAttribute` rather than `attributes`.

Ah. Wish I knew about that earlier.

PR #1732, opened 2021-03-23
Fix links not opening in a new tab

HoeenCoder:
@mia-pi-git can you check and make sure this doesn't break your changes?


mia-pi-git:
Checked on testclient, all seems good.

Zarel:
I'm not sure why you needed to search for the attribute here at all...

PR #1734, opened 2021-03-30
FIx BattleTextParser#team

KrisXV:
Compared to what it was like before:

The return cases should be returning the opposite of what they actually are.


AnnikaCodes:
This seems like something that would be simple to unit test; do you want to do that?

KrisXV:
> This seems like something that would be simple to unit test; do you want to do that?

i can try it in a bit

PR #1741, opened 2021-04-05
Fixed Negative pp Tooltip Issue

test123:
More details, on this issue:
https://github.com/smogon/pokemon-showdown-client/issues/1740

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -1228,7 +1228,7 @@ class BattleTooltips {
 			return `${bullet} ${move.name} <small>(0/${maxpp})</small>`;
 		}
 		if (ppUsed || moveName.charAt(0) === '*') {
-			return `${bullet} ${move.name} <small>(${maxpp - ppUsed}/${maxpp})</small>`;
+			return `${bullet} ${move.name} <small>(${Math.max(maxpp - ppUsed, 0)}/${maxpp})</small>`;
 		}
 		return `${bullet} ${move.name} ${showKnown ? ' <small>(revealed)</small>' : ''}`;
 	}


Zarel:
I don't think this is a good change. I intentionally didn't do this, because in modded formats, moves often have different PP from what you think they have. I think it's more important to know exactly how much PP has been deducted, than to show `0` rather than `-1` in some obscure situations.

HoeenCoder:
Would it be feasible to make this only apply to standard formats (base, gen1-7)?

Zarel:
We'd need some way to detect a format as a standard format, but yes, I would be fine with that.

Zarel:
My main thing is that I don't think anyone would be _confused_ about what -1 PP means, so this seems like a pretty unnecessary improvement in general.

DaWoblefet:
I disagree personally Zarel; I think it looks bad to have -1 PP displayed when you can never actually go below 0 PP in the games. In what mods would it ever make sense to have -1 PP displayed anyway?

HoeenCoder:
@DaWoblefet we also have to consider mods on side servers that don't get client support, and even some formats on main that don't get that (iirc unless something changed, only some formats on main won't receive this update). Still regardless I'm unsure how I feel about negative PP in tooltips.

Zarel:
On the other hand, this is relevant for Leppa Berry interactions... but yes, I'm worried about mods on side servers.

Zarel:
> On the other hand, this is relevant for Leppa Berry interactions... but yes, I'm worried about mods on side servers.

On second thought: This is really easy to handle inside Leppa Berry without breaking anything else.

test123:
On the note of Leppa Berries, the counter on the Tooltip just disappears, in some cases. Check around 7:00 minutes, in [this video](https://www.youtube.com/watch?v=uGwLzQuNTOo&).

Zarel:
> On the note of Leppa Berries, the counter on the Tooltip just disappears, in some cases. Check around 7:00 minutes, in [this video](https://www.youtube.com/watch?v=uGwLzQuNTOo&).

Do you have a link to a replay? That'd help a lot for fixing this sort of problem.

test123:
> > On the note of Leppa Berries, the counter on the Tooltip just disappears, in some cases. Check around 7:00 minutes, in [this video](https://www.youtube.com/watch?v=uGwLzQuNTOo&).
> 
> Do you have a link to a replay? That'd help a lot for fixing this sort of problem.

Hey sorry, the replay is not mine, and it isn't uploaded to the PS server. It's just something that I found on a popular YouTube Video.

Zarel:
Unfortunately we can't accept this, as has been discussed extensively. The Leppa Berry issue has been fixed, but negative PP is still displayed, and is most clearly explained here:

https://www.smogon.com/forums/threads/pp-shows-as-1-due-to-facing-pressure-pokemon.3717565/

PR #1742, opened 2021-04-05
Battle-Log: Support custom start times in youtube tags

mia-pi-git:
Requested, fairly easy, and useful.

### DIFF
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -843,9 +843,15 @@ class BattleLog {
 				const videoId = /(?:\?v=|\/embed\/)([A-Za-z0-9_\-]+)/.exec(src)?.[1];
 				if (!videoId) return {tagName: 'img', attribs: ['alt', `invalid src for <youtube>`]};
 
+				const time = getAttrib('start') || /(?:\?|&)t=([0-9]+)/.exec(src)?.[1] || false;
+
 				return {
 					tagName: 'iframe',
-					attribs: ['width', width, 'height', height, 'src', `https://www.youtube.com/embed/${videoId}`, 'frameborder', '0', 'allow', 'accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture', 'allowfullscreen', 'allowfullscreen'],
+					attribs: [
+						'width', width, 'height', height,
+						'src', `https://www.youtube.com/embed/${videoId}${time ? `?start=${time}` : ''}`,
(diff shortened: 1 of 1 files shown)


src/battle-log.ts
```diff
@@ -843,9 +843,15 @@ class BattleLog {
 				const videoId = /(?:\?v=|\/embed\/)([A-Za-z0-9_\-]+)/.exec(src)?.[1];
 				if (!videoId) return {tagName: 'img', attribs: ['alt', `invalid src for <youtube>`]};
 
+				const time = getAttrib('start') || /(?:\?|&)t=([0-9]+)/.exec(src)?.[1] || false;
```
Zarel, line 846:
```suggestion
				const time = getAttrib('start') || /(?:\?|&)t=([0-9a-z]+)/.exec(src)?.[1] || false;
```

We should probably support hms notation.

src/battle-log.ts
```diff
@@ -843,9 +843,15 @@ class BattleLog {
...
+
 				return {
 					tagName: 'iframe',
-					attribs: ['width', width, 'height', height, 'src', `https://www.youtube.com/embed/${videoId}`, 'frameborder', '0', 'allow', 'accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture', 'allowfullscreen', 'allowfullscreen'],
+					attribs: [
+						'width', width, 'height', height,
+						'src', `https://www.youtube.com/embed/${videoId}${time ? `?start=${time}` : ''}`,
```
Zarel, line 852: Does `t=` not work in embeds?

mia-pi-git: What do you mean? `url?t=` unto itself, with nothing else after it?

Zarel: No, I mean, you only accept `t=` in the URL, and you only use `start=` in the embed URL. Should we all be searching for `start=` in the input URL?

Zarel: Also, are attributes sanitized in this? What happens if I try to use an attribute of `start="<i>test"`?

src/battle-log.ts
```diff
@@ -843,9 +843,15 @@ class BattleLog {
 				const videoId = /(?:\?v=|\/embed\/)([A-Za-z0-9_\-]+)/.exec(src)?.[1];
 				if (!videoId) return {tagName: 'img', attribs: ['alt', `invalid src for <youtube>`]};
 
+				const time = /(?:\?|&)t=([0-9]+)/.exec(src)?.[1];
```
Zarel, line 846:
```suggestion
				const time = /(?:\?|&)(?:t|start)=([0-9]+)/.exec(src)?.[1];
```

Zarel:
ty!

PR #1743, opened 2021-04-06
Fix display of move choices

DaWoblefet:
I honestly do not know what this was doing before, but it won't enter the for loop because this.battle.myPokemon is an array. This restores the decision-making text, as seen below.

With changes:
image

image


Zarel:
This could crash inmultibattle. Let me review it more.

Zarel:
Bug introduced in 1fe0b7eb852d97ca904c, more recent than I thought

Zarel:
You accidentally used your master branch for this! Remember not to do that in the future.

Zarel:
Thanks!

PR #1748, opened 2021-04-14
Improve /ignore

mia-pi-git:
https://www.smogon.com/forums/threads/ignore-improvements.3682273/

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -599,7 +599,7 @@
 				} else if (app.ignore[toUserid(target)]) {
 					this.add("User '" + toName(target) + "' is already on your ignore list. (Moderator messages will not be ignored.)");
 				} else {
-					app.ignore[toUserid(target)] = {notified: true};
+					app.ignore[toUserid(target)] = 1;
 					this.add("User '" + toName(target) + "' ignored. (Moderator messages will not be ignored.)");
 					app.saveIgnore();
 				}
@@ -1693,11 +1693,11 @@
 			var user = (this.users && this.users[app.user.get('userid')]) || {};
 			var readerHasAuth = !" +\u2606\u203D!".includes(user.group || ' ');
 			if (app.ignore[userid] && !speakerHasAuth && !readerHasAuth) {
-				if (!app.ignore[userid].notified) {
+				if (!app.ignoreNotified) {
 					this.$chat.append(
 						'<div class="chat">A message from ' + BattleLog.escapeHTML(name) + ' was ignored. (to unignore use /unignore)</div>'
 					);
-					app.ignore[userid].notified = true;
+					app.ignoreNotified = true;
 				}
 				return;
 			}
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -599,7 +599,7 @@
 				} else if (app.ignore[toUserid(target)]) {
 					this.add("User '" + toName(target) + "' is already on your ignore list. (Moderator messages will not be ignored.)");
 				} else {
-					app.ignore[toUserid(target)] = {notified: false};
+					app.ignore[toUserid(target)] = {notified: true};
 					this.add("User '" + toName(target) + "' ignored. (Moderator messages will not be ignored.)");
 					app.saveIgnore();
 				}
@@ -1694,7 +1694,9 @@
 			var readerHasAuth = !" +\u2606\u203D!".includes(user.group || ' ');
 			if (app.ignore[userid] && !speakerHasAuth && !readerHasAuth) {
 				if (!app.ignore[userid].notified) {
-					this.$chat.append('<div class="chat">1 message from user ' + name + ' ignored. (to unignore use /unignore)</div>');
+					this.$chat.append(
+						'<div class="chat">A message from ' + BattleLog.escapeHTML(name) + ' was ignored. (to unignore use /unignore)</div>'
+					);
 					app.ignore[userid].notified = true;
 				}
 				return;
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -613,7 +613,7 @@
 				}
 				if (toID(target) !== 'confirm') {
 					this.add("Are you sure you want to clear your ignore list?");
-					this.add("Type /clearignore confirm if you're sure.");
+					this.add('|html|If you\'re sure, use <code>/clearignore confirm</code>');
 					return false;
 				}
 				if (!Object.keys(app.ignore).length) {
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -599,11 +599,31 @@
 				} else if (app.ignore[toUserid(target)]) {
 					this.add("User '" + toName(target) + "' is already on your ignore list. (Moderator messages will not be ignored.)");
 				} else {
-					app.ignore[toUserid(target)] = 1;
+					app.ignore[toUserid(target)] = {notified: false};
 					this.add("User '" + toName(target) + "' ignored. (Moderator messages will not be ignored.)");
+					app.saveIgnore();
 				}
 				return false;
 
+			case 'clearignore':
+				if (this.checkBroadcast(cmd, text)) return false;
+				if (!target) {
+					this.parseCommand('/help ignore');
+					return false;
+				}
+				if (toID(target) !== 'confirm') {
+					this.add("Are you sure you want to clear your ignore list?");
+					this.add("Type /clearignore confirm if you're sure.");
+					return false;
+				}
+				if (!Object.keys(app.ignore).length) {
+					this.add("You have no ignored users.");
+					return false;
+				}
+				app.ignore = {};
+				app.saveIgnore();
+				this.add("Your ignore list was cleared.");
+				return false;
 			case 'unignore':
 				if (this.checkBroadcast(cmd, text)) return false;
 				if (!target) {
@@ -615,6 +635,7 @@
 				} else {
 					delete app.ignore[toUserid(target)];
 					this.add("User '" + toName(target) + "' no longer ignored.");
+					app.saveIgnore();
 				}
 				return false;
 
@@ -1082,6 +1103,7 @@
 					this.add('/ignore [user] - Ignore all messages from the user [user].');
 					this.add('/unignore [user] - Remove the user [user] from your ignore list.');
 					this.add('/ignorelist - List all the users that you currently ignore.');
+					this.add('/clearignore - Remove all users on your ignore list.');
 					this.add('Note that staff messages cannot be ignored.');
 					return false;
 				case 'nick':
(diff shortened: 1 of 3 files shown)


js/client-chat.js
```diff
@@ -1670,7 +1692,13 @@
 			var speakerHasAuth = " +\u2606".indexOf(name.charAt(0)) < 0;
 			var user = (this.users && this.users[app.user.get('userid')]) || {};
 			var readerHasAuth = !" +\u2606\u203D!".includes(user.group || ' ');
-			if (app.ignore[userid] && !speakerHasAuth && !readerHasAuth) return;
+			if (app.ignore[userid] && !speakerHasAuth && !readerHasAuth) {
+				if (!app.ignore[userid].notified) {
+					this.$chat.append('<div class="chat">1 message from user ' + name + ' ignored. (to unignore use /unignore)</div>');
```
Zarel, line 1697: The number of messages is kind of irrelevant if you stop notifying after the first.

```suggestion
					this.$chat.append('<div class="chat">A message from ' + BattleLog.escapeHTML(name) + ' was ignored. (to unignore use /unignore)</div>');
```

js/client-mainmenu.js
```diff
@@ -394,8 +399,9 @@
 				if (app.ignore[userid]) {
 					$chat.append('<div class="chat">User ' + userid + ' is already on your ignore list. (Moderator messages will not be ignored.)</div>');
 				} else {
-					app.ignore[userid] = 1;
+					app.ignore[userid] = {notified: false};
```
Zarel, line 402:
```suggestion
					app.ignore[userid] = {notified: true};
```

No need to remind them if they literally just used the command.

js/client-mainmenu.js
```diff
@@ -168,8 +168,13 @@
...
 			var userid = toUserid(name);
-			if (app.ignore[userid] && name.substr(0, 1) in {' ': 1, '+': 1, '!': 1, '✖': 1, '‽': 1}) return;
-
+			if (app.ignore[userid] && name.substr(0, 1) in {' ': 1, '+': 1, '!': 1, '✖': 1, '‽': 1}) {
+				if (!app.ignore[userid].notified) {
+					message = '/nonotify 1 message from ' + name + ' ignored.';
+					name = '&';
+					app.ignore[userid].notified = true;
+				} else return;
+			}
```
Zarel, line 177:
```suggestion
			if (app.ignore[userid] && name.substr(0, 1) in {' ': 1, '+': 1, '!': 1, '✖': 1, '‽': 1}) {
				if (!app.ignore[userid].notified) {
					message = '/nonotify A PM from ' + name + ' was ignored. (To unignore, use /unignore.)';
					name = '&';
					app.ignore[userid].notified = true;
				} else {
					return;
				}
			}
```

js/client.js
```diff
@@ -967,7 +967,9 @@ function toId() {
 			} else if (data.substr(0, 3) === '|N|') {
 				var names = data.substr(1).split('|');
 				if (app.ignore[toUserid(names[2])]) {
-					app.ignore[toUserid(names[1])] = 1;
+					app.ignore[toUserid(names[1])] = {
+						notified: false,
```
Zarel, line 971:
```suggestion
						notified: app.ignore[toUserid(names[2])].notified,
```

js/client.js
```diff
@@ -2711,9 +2723,10 @@ function toId() {
 				delete app.ignore[this.userid];
 				buf += " no longer ignored.";
 			} else {
-				app.ignore[this.userid] = 1;
+				app.ignore[this.userid] = {notified: false};
```
Zarel, line 2726:
```suggestion
				app.ignore[this.userid] = {notified: true};
```

js/client.js
```diff
@@ -1158,6 +1161,15 @@ function toId() {
 				break;
 			}
 		},
+		saveIgnore: function () {
+			var buf = {};
+			if (!this.ignore) return; // ??
+			for (var k in this.ignore) {
+				// we set this to false so they're notified once per session of it
+				buf[k] = {notified: false};
```
Zarel, line 1169: This seems like it would be notify more than once per session, but rather every time it was saved?

mia-pi-git: No, because buf isn't app.ignore. 
it just ensures that next time they start a new session, they'll be notified again.

js/client.js
mia-pi-git: Yeah, I was thinking about that originally when I wrote that. But I felt it would be better to err on the side of "make sure they're aware of this."

Zarel: Definitely no. It's better to err on the side of "don't annoy them with notifications". The only reason we're notifying at all is so people don't accidentally ignore the wrong person and never realize, or if a friend messes with their settings they still know.

mia-pi-git: Okay.

js/client.js
Zarel: Please don't resolve conversations after you answer the question, that makes it so I can't see the answer. I'll resolve it once I see the answer. :[

Zarel: It seems like it's better to just set it to `1` and then handle it in the loader?

js/client.js
```diff
@@ -1158,6 +1162,15 @@ function toId() {
...
+			}
+			app.saveIgnore();
+		},
```
Zarel, line 1173: image

You forgot to actually write `saveIgnore`

js/client.js
mia-pi-git: Okay. Done.

js/client.js
mia-pi-git: Oops. Fixed.

js/client.js
```diff
@@ -1158,6 +1162,24 @@ function toId() {
...
 			}
 		},
+		saveIgnore: function () {
+			var buf = {};
+			if (!this.ignore) return; // ??
+			for (var k in this.ignore) {
+				// we set this to false so they're notified once per session of it
+				buf[k] = 1;
+			}
+			Storage.prefs('ignorelist', buf);
```
Zarel, line 1172: This should just be an array, right? It makes more sense as an array.

```suggestion
			if (!this.ignore) return; // ??
			Storage.prefs('ignorelist', Object.keys(this.ignore));
```

js/client.js
```diff
@@ -1158,6 +1162,24 @@ function toId() {
...
+		},
+		loadIgnore: function () {
+			var data = Storage.prefs('ignorelist');
+			if (!data) return {};
+			var buf = {};
+			for (var k in data) {
+				buf[k] = {notified: false};
+			}
+			return buf;
```
Zarel, line 1181:
```suggestion
			var ignoreList = Storage.prefs('ignorelist');
			if (!ignoreList) return {};
			var ignore = {};
			for (var i = 0; i < ignoreList.length; i++) {
				ignore[ignoreList[i]] = {notified: false};
			}
			return ignore;
```

js/client.js
```diff
@@ -1158,6 +1162,18 @@ function toId() {
...
 		},
+		saveIgnore: function () {
+			Storage.prefs('ignorelist', Object.keys(this.ignore));
+		},
+		loadIgnore: function () {
+			var data = Storage.prefs('ignorelist');
+			if (!data) return {};
+			var buf = {};
+			for (var i = 0; i < data.length; i++) {
+				buf[data[i]] = {notified: false};
```
Zarel: I think we'd be better off with one overall `ignoreNotified` flag, rather than a per-user flag? I think this would make it easier on people with a long ignore list.

js/client-mainmenu.js
```diff
@@ -168,8 +168,13 @@
...
 		addPM: function (name, message, target) {
 			var userid = toUserid(name);
-			if (app.ignore[userid] && name.substr(0, 1) in {' ': 1, '+': 1, '!': 1, '✖': 1, '‽': 1}) return;
-
+			if (app.ignore[userid] && name.substr(0, 1) in {' ': 1, '+': 1, '!': 1, '✖': 1, '‽': 1}) {
+				if (!app.ignoreNotified) {
+					message = '/nonotify A message from ' + BattleLog.escapeHTML(name) + ' was ignored.';
+					name = '&';
+					app.ignoreNotified = true;
+				} else return;
```
Zarel, line 176:
```suggestion
				}
				return;
```

js/client-mainmenu.js
```diff
@@ -168,8 +168,13 @@
 
 		addPM: function (name, message, target) {
 			var userid = toUserid(name);
-			if (app.ignore[userid] && name.substr(0, 1) in {' ': 1, '+': 1, '!': 1, '✖': 1, '‽': 1}) return;
-
+			if (app.ignore[userid] && name.substr(0, 1) in {' ': 1, '+': 1, '!': 1, '✖': 1, '‽': 1}) {
```
Zarel, line 171:
```suggestion
			if (app.ignore[userid] && ' +!✖‽'.includes(name.charAt(0))) {
```

js/client-mainmenu.js
```diff
@@ -168,8 +168,13 @@
...
+				if (!app.ignoreNotified) {
+					message = '/nonotify A message from ' + BattleLog.escapeHTML(name) + ' was ignored.';
```
Zarel, line 173: This doesn't need an HTML escape, right?

Zarel: Turns out it does, `/nonotify` is in fact HTML. What a bad name for it. :/

mia-pi-git: Shouldn't it? it gets appended into the PM.

mia-pi-git: Oh, didn't see your comment. MB.

js/client-mainmenu.js
```diff
@@ -168,8 +168,13 @@
...
+					name = '&';
```
Zarel, line 174: Should this just be from the user? Would make it easier to reply probably.

Although the app should also prevent you from replying to users you've ignored.

mia-pi-git: It's mostly consistency. The blockpms notified thing on the server is from &, so I figured it should be the same here.

Zarel: Is it possible to PM users you've ignored? we should probably fix that

Zarel: The blockpms notification should probably be updated to also be from the user I think

mia-pi-git: Fixed both.

js/client-chat.js
```diff
@@ -601,9 +601,29 @@
...
 
+			case 'clearignore':
+				if (this.checkBroadcast(cmd, text)) return false;
+				if (!target) {
+					this.parseCommand('/help ignore');
+					return false;
+				}
+				if (toID(target) !== 'confirm') {
+					this.add("Are you sure you want to clear your ignore list?");
+					this.add("Type /clearignore confirm if you're sure.");
```
Zarel, line 616: The boundary between `confirm` and `if` is very unclear.

Ideally we should use code blocks, but I don't actually know how to put code blocks into `.add`

Zarel:
```suggestion
					this.add('|html|If you\'re sure, use <code>/clearignore confirm</code>');
```

js/client-mainmenu.js
Zarel: MIA YOU DIDN'T MERGE THIS

mia-pi-git: ...but i clicked on it! :c

js/client-mainmenu.js
Zarel: MIA THIS IS A MAJOR BUG THAT MAKES IGNORING NOT WORK WITHOUT MY FIX

PR #1749, opened 2021-04-14
Teambuilder: Fix shiny problem

Distrib-ps:
Pr for https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8821350


Zarel:
`Dex.types.get` always returns a truthy value and this code could set an invalid `hpType` which could cause other crashes.

PR #1750, opened 2021-04-14
Add support for searching the logindb via dispatcher

mia-pi-git:


lib/dispatcher.lib.php
```diff
@@ -304,6 +304,23 @@ public function upkeep($dispatcher, &$reqData, &$out) {
...
+		$server = $dispatcher->findServer();
+		if (!$server || $server['id'] !== 'showdown') {
+			$out['actionerror'] = 'access denied';
+		} else if (!isset($_REQUEST['search'])) {
+			$out['actionerror'] = 'Specify a valid search.';
+		} else {
+			$search = $psdb->escape($_REQUEST['search']);
+			$result = $psdb->query(
+				"SELECT `userid`, `ip` FROM `ntbb_users` WHERE `ip` LIKE '" . $search . "%' " .
+				"OR `userid` LIKE '%" . $search . "%';"
```
Zarel, line 318: We do not have fulltext indexes... the database can't handle this kind of query. :[

PR #1751, opened 2021-04-16
Tooltips: Fix hover for opponent's right slot

DaWoblefet:
In doubles, the tooltip for p2b (opponent's right) was mistakenly given the same values as p2c in https://github.com/smogon/pokemon-showdown-client/pull/1738. I assume this was a typo, but just to be sure, @pyuk-bot do you remember if this was necessary for Multi Battles? I tested in FFA and Doubles and now it works correctly.


Zarel:
ty!

PR #1752, opened 2021-04-16
Tooltips: Fix Victory Star and Battery

DaWoblefet:
Minor regression from Multi Battles. Currently, when checking the ally's Ability, if it wasn't announced (i.e. an entrance ability like Fairy Aura or after Skill Swap or something), it falls back on the multi battle ally's ability. But this is null in doubles, so I readded logic to bring it back. Thanks to @pyuk-bot for her help!


Zarel:
tyty

PR #1753, opened 2021-04-16
Tooltips: Fix undefined Hidden Power description

DaWoblefet:
This has been around a while; no idea when it was introduced, but it only affects Hidden Power so far as I know. There's likely a "correct place" to fix this in the build or something but idk how to do that.

Before:
image

After:
image


Zarel:
ty!

PR #1754, opened 2021-04-17
Tooltips: Fix Max Move typing / name

DaWoblefet:
The change in client-battle.js fixes an issue where a move like Weather Ball would display as its effective type but be called base name.

Before:
image

After:
image

The change in battle-tooltips.js removes the restriction on -ate Abilities displaying relative typing. _Technically_ the games don't display Pixilate Max Hyper Voice as Max Starfall, but that's what it becomes when used during the turn. The games don't display regular Pixilate Hyper Voice as Fairy-type either though, and I think it's better to be consistent and display the effective typing for convenience.

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -667,6 +667,9 @@
 								// might not exist, such as for Z status moves - fall back on base move to determine type then
 								var specialMove = gigantamax || this.battle.dex.moves.get(specialMoves[i].move);
 								var moveType = this.tooltips.getMoveType(specialMove.exists && !specialMove.isMax ? specialMove : baseMove, typeValueTracker, specialMove.isMax ? gigantamax || switchables[pos].gigantamax || true : undefined)[0];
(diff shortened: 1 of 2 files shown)


Zarel:
ty!

js/client-battle.js
```diff
@@ -667,6 +667,9 @@
 								// might not exist, such as for Z status moves - fall back on base move to determine type then
 								var specialMove = gigantamax || this.battle.dex.moves.get(specialMoves[i].move);
 								var moveType = this.tooltips.getMoveType(specialMove.exists && !specialMove.isMax ? specialMove : baseMove, typeValueTracker, specialMove.isMax ? gigantamax || switchables[pos].gigantamax || true : undefined)[0];
+								if (specialMove.isMax && specialMove.name !== 'Max Guard') {
+									specialMove = this.tooltips.getMaxMoveFromType(moveType);
+								}
```
urkerab, line 672: What about gigantamax moves? Would only doing this if `getMoveType` actually changed the type work?

PR #1755, opened 2021-04-17
Tooltips: Fix Hidden Power and -ate Abilities

DaWoblefet:
Previously, effects like Normalize would override the typing of Hidden Power and incorrectly indicate it would receive a BP boost. It does not.


Zarel:
ty!

PR #1757, opened 2021-04-17
Tooltips: Fix unique Z-move BP

DaWoblefet:
Certain moves like Gear Grind or Grass Knot have custom BP that doesn't follow the standard Z-move BP chart. This fixes it so their proper Z-move BP is used.

I think this solution is dumb though - I don't get why we'd have to do it this way for Z-moves and not for Max Moves. I'm surely missing something.


urkerab:
Possibly because status moves are handled differently?

PR #1760, opened 2021-04-17
Fix double escape in battle and browser tab

DaWoblefet:
Hopefully this is the correct place to remove the escape (sample replay: https://replay.pokemonshowdown.com/gen81v1-1148003675)

Before:
image

After:
image

### DIFF
--- a/js/client-topbar.js
+++ b/js/client-topbar.js
@@ -91,7 +91,7 @@
 			case 'rooms':
 				return buf + ' aria-label="Join chatroom"><i class="fa fa-plus" style="margin:7px auto -6px auto"></i> <span>&nbsp;</span></a></li>';
 			case 'battle':
-				var name = BattleLog.escapeHTML(room.title);
+				var name = room.title;
 				var idChunks = id.substr(7).split('-');
 				var formatid;
 				if (idChunks.length <= 1) {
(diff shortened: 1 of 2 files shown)


Zarel:
No, that is unfortunately a dangerous place to remove an escape. The correct place is elsewhere.

js/client-topbar.js
```diff
@@ -91,7 +91,7 @@
 			case 'rooms':
 				return buf + ' aria-label="Join chatroom"><i class="fa fa-plus" style="margin:7px auto -6px auto"></i> <span>&nbsp;</span></a></li>';
 			case 'battle':
-				var name = BattleLog.escapeHTML(room.title);
+				var name = room.title;
```
Zarel, line 94: Specifically, this is bad and will need to be reverted.

js/client.js
```diff
@@ -937,7 +937,7 @@ function toId() {
 					var replayLink = 'https://' + Config.routes.replays + '/' + replayid;
 					$.ajax(replayLink + '.json', {dataType: 'json'}).done(function (replay) {
 						if (replay) {
-							var title = BattleLog.escapeHTML(replay.p1) + ' vs. ' + BattleLog.escapeHTML(replay.p2);
+							var title = replay.p1 + ' vs. ' + replay.p2;
```
Zarel, line 940: This one is OK.

js/client-topbar.js
Zarel:
```suggestion
				var name = BattleLog.escapeHTML(room.title);
```

DaWoblefet: Turns out it was also unnecessary! The other was all that was needed to get the correct results.

PR #1763, opened 2021-04-22
Allow commands in battle when ignoring spectators

mia-pi-git:
Problematic for some people.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -129,8 +129,7 @@
 			text = this.parseCommand(text);
 			if (
 				this.battle && this.battle.ignoreSpects &&
-				app.user.get('userid') !== this.battle.p1.id &&
-				app.user.get('userid') !== this.battle.p2.id &&
+				app.user.get('userid') !== this.battle.p1.id && app.user.get('userid') !== this.battle.p2.id &&
 				!(text.startsWith('/') && !text.startsWith('/me'))
 			) {
 				this.add("You can't chat in this battle as you're currently ignoring spectators");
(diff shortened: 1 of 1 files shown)


js/client-chat.js
```diff
@@ -127,7 +127,12 @@
 			this.tabComplete.reset();
 			this.chatHistory.push(text);
 			text = this.parseCommand(text);
-			if (this.battle && this.battle.ignoreSpects && app.user.get('userid') !== this.battle.p1.id && app.user.get('userid') !== this.battle.p2.id) {
+			if (
+				this.battle && this.battle.ignoreSpects &&
+				app.user.get('userid') !== this.battle.p1.id &&
+				app.user.get('userid') !== this.battle.p2.id &&
+				!(text.startsWith('/') || text.startsWith('!'))
```
Zarel, line 134: Broadcast commands should still be unusable, as should `/me`.

```suggestion
				!(text.startsWith('/') && !text.startsWith('/me'))
```

js/client-chat.js
```diff
@@ -127,7 +127,12 @@
 			this.tabComplete.reset();
 			this.chatHistory.push(text);
 			text = this.parseCommand(text);
-			if (this.battle && this.battle.ignoreSpects && app.user.get('userid') !== this.battle.p1.id && app.user.get('userid') !== this.battle.p2.id) {
+			if (
+				this.battle && this.battle.ignoreSpects &&
+				app.user.get('userid') !== this.battle.p1.id &&
+				app.user.get('userid') !== this.battle.p2.id &&
```
Zarel, line 133:
```suggestion
				app.user.get('userid') !== this.battle.p1.id && app.user.get('userid') !== this.battle.p2.id &&
```

PR #1765, opened 2021-04-30
Tooltips: Halve Earthquake, Bulldoze BP in Grassy Terrain

DaWoblefet:
https://www.smogon.com/forums/threads/show-grassy-terrains-effect-on-the-3-ground-moves.3683343/

No point in displaying for Magnitude since it doesn't a base power anyway, since it can't be known before the turn and all.

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -1548,6 +1548,9 @@ class BattleTooltips {
 			else if (ppLeft === 4) basePower = 50;
 			value.set(basePower);
 		}
+		if (move.id === 'magnitude') {
+			value.setRange(10, 150);
+		}
 		if (move.id === 'venoshock' && target) {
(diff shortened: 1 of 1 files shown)


Zarel:
> No point in displaying for Magnitude since it doesn't a base power anyway, since it can't be known before the turn and all.

Why not? Magnitude should display a range, and `value.modify` will modify both ends of the range.

Zarel:
Like if Magnitude _doesn't_ display "10-150 BP", it should, and `value.modify` is all you need to do to make it display "5-75 BP" instead.

DaWoblefet:
> Like if Magnitude _doesn't_ display "10-150 BP", it should, and `value.modify` is all you need to do to make it display "5-75 BP" instead.

Excellent point!

image

Zarel:
Thanks!

PR #1766, opened 2021-04-30
Tooltips: account halved BP from certain ground moves in grassy terrain

xInstruct:
https://www.smogon.com/forums/threads/show-grassy-terrains-effect-on-the-3-ground-moves.3683343/

before

image

after

image


xInstruct:
Well I got sniped so gg.

PR #1767, opened 2021-04-30
Tooltips: Display Quick Claw after activation

DaWoblefet:
Sample replay: https://replay.pokemonshowdown.com/gen8anythinggoes-1323722693


Zarel:
ty!

PR #1769, opened 2021-04-30
Add radio/checkbox serialization to forms

PartMan7:
These changes allow for submitting data from radio/checkbox elements in form's data-submitsend.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -835,25 +835,21 @@ function toId() {
 			// querySelector dates back to IE8 so we can use it
 			// fortunate, because form serialization is a HUGE MESS in older browsers
 			var elements = form.querySelectorAll('input[name], select[name], textarea[name], keygen[name]');
-			var out = [], checkboxes = {}, radios = {};
+			var out = [], radios = {};
 			for (var i = 0; i < elements.length; i++) {
 				var element = elements[i];
 				if (element.type === 'radio') {
 					var name = ' ' + element.name;
 					if (element.checked) radios[name] = element.value;
 					else if (!radios[name]) radios[name] = false;
-				} else if (element.type === 'checkbox') {
-					if (!checkboxes[' ' + element.name]) checkboxes[' ' + element.name] = [];
-					if (element.checked) checkboxes[' ' + element.name].push(element.value);
-				} else out.push([element.name, element.value]);
-				// For now, checkboxes are handled as a comma-delimited string with all the checked values,
-				// while radios are handled by only returning the checked value / empty string if unchecked
-				// Dunno what others should be added
+				} else if (element.type === 'checkbox') out.push([element.name, element.checked ? element.value : '']);
+				else out.push([element.name, element.value]);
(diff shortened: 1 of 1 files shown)


Zarel:
I don't understand what the `checkboxes` table is for. It seems completely unnecessary. The `radios` table also seems overcomplicated.

js/client.js
```diff
@@ -835,12 +835,25 @@ function toId() {
 			// querySelector dates back to IE8 so we can use it
 			// fortunate, because form serialization is a HUGE MESS in older browsers
 			var elements = form.querySelectorAll('input[name], select[name], textarea[name], keygen[name]');
-			var out = [];
+			var out = [], checkboxes = {}, radios = {};
 			for (var i = 0; i < elements.length; i++) {
 				var element = elements[i];
-				// TODO: values are a mess in the DOM; checkboxes/select probably need special handling
```
Zarel, line 841: You deleted this comment but you don't seem to have handled the `select` element.

js/client.js
```diff
@@ -835,12 +835,25 @@ function toId() {
...
-				out.push([element.name, element.value]);
-			}
+				if (element.type === 'radio') {
+					var name = ' ' + element.name;
+					if (element.checked) radios[name] = element.value;
+					else if (!radios[name]) radios[name] = false;
+				} else if (element.type === 'checkbox') {
+					if (!checkboxes[' ' + element.name]) checkboxes[' ' + element.name] = [];
+					if (element.checked) checkboxes[' ' + element.name].push(element.value);
```
Zarel, line 847: Oh, I see, you're trying to handle the case where multiple checkboxes have the same name.

Is that actually how the spec handles it?

PartMan7: Okay I have no idea what I was thinking when I wrote the checkboxes, sorry
Lemme rewrite this, gimme a minute

PartMan7:
Wait, doesn't select already return the selected value as element.value?

Zarel:
I didn't know. Anyway, while I'm sure the stuff you wrote is _useful_, none of it belongs in this file.

js/client.js
```diff
@@ -835,12 +835,21 @@ function toId() {
 			// querySelector dates back to IE8 so we can use it
 			// fortunate, because form serialization is a HUGE MESS in older browsers
 			var elements = form.querySelectorAll('input[name], select[name], textarea[name], keygen[name]');
-			var out = [];
+			var out = [], radios = {};
```
Zarel, line 838:
```suggestion
			var out = [];
```

Zarel:
No hard feelings, but 6383b4adc4bb4fabd is just a much simpler and more spec-compliant implementation.

I do recognize what you were trying to do with checkboxes now: support having multiple with the same name. But that's not how forms are usually serialized, so you should be doing that inside `submitSend` rather than inside `serializeForm`.

(Although I don't think it's really necessary in the first place. It's easy enough to write your commands not to need such a niche feature.)

PR #1771, opened 2021-05-02
Make ranges appear again in Firefox

urkerab:
Firefox removed support for `-moz-appearance: range;` last year. I've replaced it with `revert` but you could also make a case for `auto`.

While I was checking the console I noticed an illegal linear gradient so I fixed that while I was there.

### DIFF
--- a/style/battle-log.css
+++ b/style/battle-log.css
@@ -282,7 +282,7 @@ select {
 
 input[type=range] {
 	-webkit-appearance: none;
-	-moz-appearance: range;
+	-moz-appearance: revert;
 	cursor: pointer;
 	height: 18px;
 	background: transparent;
--- a/style/client.css
+++ b/style/client.css
@@ -2881,7 +2881,7 @@ a.ilink.yours {
 
 	border: 1px solid #999999;
 	border-radius: 5px;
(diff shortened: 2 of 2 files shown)


Zarel:
Would there be anything wrong with simply removing that line completely?

Zarel:
Presumably leaving the line off would make it work correctly on most modern Firefox versions.

So the question is: what value would add support for even older Firefox versions, without breaking compatibility with the newest Firefox versions? If none, there's no reason not to remove it completely.

Zarel:
It also worries me that `revert` is entirely undocumented.

urkerab:
Sufficiently old versions of Firefox ignore `-webkit-appearance` so for them it doesn't matter whether they recognise the `-moz-appearance` value.

I don't know if there are versions of Firefox that recognise `-webkit-appearance` but not `revert` but if you wanted to support them you would need both lines. I did check that `revert` was added some time before `range` was removed.

`revert` is documented on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/revert).

Zarel:
Oh, I understand now.

Why does it say `-webkit-appearance: none`? Is that to work around a different WebKit bug?

urkerab:
> Why does it say `-webkit-appearance: none`?

Seems like you didn't like Chrome's range style and replaced it with one of your own?

PR #1772, opened 2021-05-02
Implement Zacian/Zamazenta/Xerneas forme leak

Karthik99999:
Corresponding server changes here: https://github.com/smogon/pokemon-showdown/pull/8261

More info on the mechanic is in this twitter thread: https://twitter.com/DaWoblefet/status/1353213137785380864

This does a couple things:

- Point Neutral Xerneas to the correct icon
- Makes Xerneas show its Neutral forme on Team Preview (only the icon for now as Neutral Xerneas currently uses Active forme sprites for everything else)
- Adds support for a new protocol `updatepokemon` to change the icons of Zacian/Zamazenta/Xerneas. Normally I would've used `findCorrespondingPokemon` to get the pokemon to update it, however `searchid` isn't resolved until the pokemon is sent into battle, so I couldn't use that to search for a pokemon and instead had to use only the details to search.

Let me know if the code can be cleaned up anywhere.

### DIFF
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -794,6 +794,10 @@ class BattleScene {
 			let lombreCount = 0;
 			for (let i = 0; i < side.pokemon.length; i++) {
 				let pokemon = side.pokemon[i];
+				if (pokemon.speciesForme === 'Xerneas-*') {
+					side.pokemon[i].name = 'Xerneas-Neutral';
+					side.pokemon[i].speciesForme = 'Xerneas-Neutral';
+				}
 				if (pokemon.speciesForme === 'Ludicolo') ludicoloCount++;
 				if (pokemon.speciesForme === 'Lombre') lombreCount++;
 
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -795,8 +795,8 @@ class BattleScene {
 			for (let i = 0; i < side.pokemon.length; i++) {
 				let pokemon = side.pokemon[i];
 				if (pokemon.speciesForme === 'Xerneas-*') {
-					side.pokemon[i].name = 'Xerneas-Neutral';
-					side.pokemon[i].speciesForme = 'Xerneas-Neutral';
+					pokemon.name = 'Xerneas-Neutral';
+					pokemon.speciesForme = 'Xerneas-Neutral';
 				}
 				if (pokemon.speciesForme === 'Ludicolo') ludicoloCount++;
 				if (pokemon.speciesForme === 'Lombre') lombreCount++;
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -795,7 +795,6 @@ class BattleScene {
 			for (let i = 0; i < side.pokemon.length; i++) {
 				let pokemon = side.pokemon[i];
 				if (pokemon.speciesForme === 'Xerneas-*') {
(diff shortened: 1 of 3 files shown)


Zarel:
Apparently this is `|updatepokemon|OLDDETAILS|NEWDETAILS|`?

Karthik99999:
> Apparently this is `|updatepokemon|OLDDETAILS|NEWDETAILS|`?

It's `|updatepokemon|pokemonid|OLDDETAILS|NEWDETAILS|`

Zarel:
Okay. In theory, you only need `NEWDETAILS`.

Zarel:
`|updatepokemon|POKEMONID|NEWDETAILS`

src/battle.ts
```diff
@@ -3292,6 +3292,18 @@ class Battle {
 			}
 			break;
 		}
+		case 'updatepokemon': {
+			const {siden} = this.parsePokemonId(args[1]);
+			const side = this.sides[siden];
+			for (let i = 0; i < side.pokemon.length; i++) {
+				const pokemon = side.pokemon[i];
+				if (pokemon.details === args[2]) {
+					side.addPokemon('', '', args[3], i);
```
Zarel, line 3301:
```suggestion
				if (pokemon.checkDetails(args[2])) {
					side.addPokemon('', '', args[2], i);
```

Zarel: In theory, this should be all you need to make it support `|updatepokemon|POKEMONID|NEWDETAILS`

Karthik99999: this worked for the dogs but not xerneas, presumably because it's getting "Neutral" added on rather than "-*". If xerneas was changed to have neutral as the base form this could probably work (but that would be a big hassle).

Zarel: Hm. There are a few ways around that.

Zarel: The simplest way is to send `-*`, and then hardcode Neutral form in Team Preview.

src/battle-animations.ts
```diff
@@ -794,6 +794,10 @@ class BattleScene {
 			let lombreCount = 0;
 			for (let i = 0; i < side.pokemon.length; i++) {
 				let pokemon = side.pokemon[i];
+				if (pokemon.speciesForme === 'Xerneas-*') {
+					side.pokemon[i].name = 'Xerneas-Neutral';
```
Zarel: I believe nicknames conventionally never include the forme.

```suggestion
					side.pokemon[i].name = 'Xerneas';
```

src/battle-animations.ts
```diff
@@ -794,6 +794,10 @@ class BattleScene {
...
+					side.pokemon[i].speciesForme = 'Xerneas-Neutral';
```
Zarel: Hrm, if this works with no negative consequences, I'd be pretty happy about that. Otherwise it might be better to implement this inside `getSpriteData`.

Zarel: Hm. Yeah, this seems safe enough. We just need that edit above and this should be good.

Zarel:
ty!

Zarel:
Actually, I forgot: I think this should be `-updatepokemon`.

Zarel:
Hm... On further reflection, no dash is probably better. But I think `updatepoke` would be more consistent with the other messages `poke` and `clearpoke`.

PR #1773, opened 2021-05-02
Fix NatDex Teambuilder Table for UUBL

Karthik99999:
NatDex UUBL got changed to be placed in tags, and the client didn't update to reflect those changes. Fixes https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-54#post-8842754


Zarel:
Thanks!

PR #1774, opened 2021-05-03
Fix missing all BPs in FFA tooltips

Timbuktu22:
This fix will add the adjacent pokemon (if active) to a check for multiple possible BP calculations in a Free-For-All battle.

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -612,8 +612,7 @@ class BattleTooltips {
 			let basePower: string = '';
 			let difference = false;
 			let basePowers = [];
-			let targetsActive = this.battle.gameType === 'freeforall' ? foeActive.concat(allyActive) : foeActive;
-			for (const active of targetsActive) {
+			for (const active of foeActive) {
 				if (!active) continue;
 				value = this.getMoveBasePower(move, moveType, value, active);
 				basePower = '' + value;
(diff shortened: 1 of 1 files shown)


Timbuktu22:
On second thought, I'm unsure if the array returned by `pokemon.side.active` will include the user's own Pokemon. If that is the case, a check could be placed on new line 615 to allow that such member of `targetsActive` to envoke `continue`. Thoughts?

src/battle-tooltips.ts
```diff
@@ -530,6 +530,7 @@ class BattleTooltips {
 
 		let zEffect = '';
 		let foeActive = pokemon.side.foe.active;
+		let allyActive = pokemon.side.active;
```
Zarel, line 533:
```suggestion
		if (this.battle.gameType === 'freeforall') {
			foeActive = [...foeActive, ...pokemon.side.active].filter(active => active !== pokemon);
		}
```

I believe this is all you need, and then the other change can be reverted.

Timbuktu22:
Thank you for pointing that out. I've implemented your fix and reverted my previous change, so I think it's good to go.

Zarel:
You've tested this and made sure it works? Can I get a screenshot?

Timbuktu22:
BPtest
Here is Gyro Ball showing 3 different BPs on testclient.html

Zarel:
Yeah, those error messages are a different bug which I'll need to fix; they're not related to this change. Thanks!

PR #1775, opened 2021-05-04
Fix FFA battles referring to nearSide as 'your' Pokemon

Timbuktu22:
Simple check to see if gameType is FFA.


Timbuktu22:
From issue https://github.com/smogon/pokemon-showdown-client/issues/1747

Screenshot:
ffa_your

Zarel:
Thanks!

PR #1776, opened 2021-05-05
Teambuilder: Add more LC level support

KrisXV:
Adds level support for Doubles LC (and CAP LC if it starts existing)

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -2884,7 +2884,7 @@
 				if (baseFormat.substr(0, 8) === 'pokebank') baseFormat = baseFormat.substr(8);
 				if (this.curTeam && this.curTeam.format) {
 					if (baseFormat === 'battlespotsingles' || baseFormat === 'battlespotdoubles' || baseFormat.substr(0, 3) === 'vgc') set.level = 50;
-					if (baseFormat.substr(0, 2) === 'lc') set.level = 5;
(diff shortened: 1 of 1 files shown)


js/client-teambuilder.js
```diff
@@ -2884,7 +2884,7 @@
 				if (baseFormat.substr(0, 8) === 'pokebank') baseFormat = baseFormat.substr(8);
 				if (this.curTeam && this.curTeam.format) {
 					if (baseFormat === 'battlespotsingles' || baseFormat === 'battlespotdoubles' || baseFormat.substr(0, 3) === 'vgc') set.level = 50;
-					if (baseFormat.substr(0, 2) === 'lc') set.level = 5;
+					if (baseFormat.substr(0, 2) === 'lc' || baseFormat.substr(-2) === 'lc') set.level = 5;
```
Zarel, line 2887:
```suggestion
					if (baseFormat.startsWIth('lc') || baseFormat.endsWitH('lc')) set.level = 5;
```

Zarel: We have a modernizing polyfill:

https://github.com/smogon/pokemon-showdown-client/blob/master/js/lib/ps-polyfill.js

So at least modern functions can be used anywhere in the codebase, it's just modern syntax that requires a TS file.

Zarel:
ty!

PR #1778, opened 2021-05-05
Fix Fury Cutter not showing modified BP in tooltip

Timbuktu22:
From https://github.com/smogon/pokemon-showdown/projects/3#card-23557332

``pokemon.volatiles['furycutter'].multiplier`` was my best guess at retrieving the true Fury Cutter multiplier, found in L5801 of https://github.com/smogon/pokemon-showdown/blob/master/data/moves.ts, but this value does not exist and produces a runtime error because the ``volatiles`` object in ``pokemon`` is blank when I check the console. I suppose there must be _some_ way to retrieve the multiplier from the server since the Fury Cutter mechanics do work correctly.  Any suggestions how to find the multiplier are greatly appreciated :)

Screenshot when I forced ``furyCutterMod = 2``:
furycutter


DaWoblefet:
How is Metronome (the item) handled? They use the same style of counter.

Timbuktu22:
> How is Metronome (the item) handled? They use the same style of counter.

AFAIK the Metronome item does not affect any tooltip, I just tried it out and the move I used shows the same tooltip each time.

In ``items.js`` Metronome seems to add an entry to the ``volatiles`` object like Fury Cutter would, but I still can't access this data. I think the ``onStart()`` function might be relevant but I'm not sure.

src/battle-tooltips.ts
```diff
@@ -1526,6 +1526,13 @@ class BattleTooltips {
 			else basePower = 20;
 			value.set(basePower);
 		}
+		if (move.id === 'furycutter'){
+			// what will actually retrieve the multiplier?
+      	let furyCutterMod = pokemon.volatiles['furycutter'].multiplier;
```
Timbuktu22, line 1531: produces runtime error

DaWoblefet:
Sounds like it's not being tracked then at all client-side. Whatever fix you come up with for Fury Cutter could also be used for Metronome!

src/battle-tooltips.ts
```diff
@@ -1526,6 +1526,13 @@ class BattleTooltips {
 			else basePower = 20;
 			value.set(basePower);
 		}
+		if (move.id === 'furycutter'){
+			// what will actually retrieve the multiplier?
```
Zarel, line 1530: You have to keep track of the multiplier yourself. The server data isn't available to clients

DaWoblefet:
This "Fury Cutter ID" approach isn't great; in past gens, Fury Cutter has a lower starting BP, so you need more of them than 2. I would Make it a "FuryCutterUsedSuccessfully" volatile or something and increment it.

src/battle-animations.ts
```diff
@@ -1804,6 +1804,7 @@ class PokemonSprite extends Sprite {
 		laserfocus: ['Laser Focus', 'good'],
 		spotlight: ['Spotlight', 'neutral'],
 		itemremoved: null,
+		furycutter: null,
```
Timbuktu22, line 1807: This assignment doesn't seem to do anything, the furycutter volatile still appears as some default effect.

Interestingly, I tried removing L1806 (since itemremoved is the volatile I am trying to mimic), but that did not produce a default effect for itemremoved
furycutter
. It again did nothing.

Timbuktu22:
I have implemented an incremental solution to the Fury Cutter mod, so everything is now working except for the [[furycutter]] box displaying by default.

Timbuktu22:
I have thoroughly tested these changes, and everything is in working order. The only thing is that battle-animations.ts will not cooperate with me, and my very simple change to that file is not something I can verify 100% will take effect, but I am confident it will work.

src/battle-animations.ts
Zarel: Yes, `battle-animations` is not built by default because it bundles `battle-animations-moves` which makes the build extremely slow.

Simply `./build full` to opt into building it, and it should work after that.

Zarel: I think Fury Cutter _should_ show up on the bar, though. As should Echoed Voice.

Zarel: (With its current multiplier, correct for the generation. Although feel free to leave that for a future PR.)

src/battle-animations.ts
```diff
@@ -2716,24 +2717,28 @@ class PokemonSprite extends Sprite {
...
+	private static getEffectTag(id: string, pokemon: Pokemon) {
 		let effect = PokemonSprite.statusTable[id];
 		if (typeof effect === 'string') return effect;
 		if (effect === null) return PokemonSprite.statusTable[id] = '';
 		if (effect === undefined) effect = [`[[${id}]]`, 'neutral'];
+		if (id === 'furycutter') {
+			let bp = pokemon.volatiles['furycutter'][2];
+			effect[0] = `Fury Cutter BP: ${bp * pokemon.volatiles['furycutter'][1]}`;
+		}
 		return PokemonSprite.statusTable[id] = `<span class="${effect[1]}">${effect[0].replace(/ /g, '&nbsp;')}</span> `;
```
Timbuktu22, line 2742: Bug: the HP bar effect will only update once, showing the BP of Fury Cutter doubled only once. Upon further usage of Fury Cutter, the tooltip will update as normal, but the bar effect will remain stuck.

src/battle.ts
```diff
@@ -3355,6 +3359,27 @@ class Battle {
...
+			} else if (this.gen === 5) {
+				furyCutterCap = 8;
+				furyCutterBP = 20;
+			}
+
+			if (move.id === 'furycutter') {
+				if (!poke.hasVolatile('furycutter' as ID)) {
+					poke.addVolatile('furycutter' as ID, 2, furyCutterBP);
+				} else if (poke.volatiles['furycutter'][1] < furyCutterCap) {
+					poke.volatiles['furycutter'][1] *= 2;
```
Zarel, line 3377:
```suggestion
					poke.volatiles['furycutter'][1] *= 2;
					poke.sprite.updateStatbar(poke);
```

Zarel: (This is why the statbar isn't updating.)

Timbuktu22: This did not work unfortunately

Timbuktu22: My plan right now is to possibly add and remove the fury cutter volatile when appropriate, in the way that the 'perish' volatiles refresh their status bar message by doing so. Except that it would be copies of the same volatile and not 4 different types of one.

Timbuktu22:
Fury Cutter status effect and Echoed Voice implementation will be shelved for future PRs.

But, the original bug has been fixed: Fury Cutter will now accurately show the modified BP for all generations, and it works incrementally. The BP will be reset under the normal Fury Cutter conditions: the user selects a different move, the user fails to move, the user misses, the user switches, or if the user is prevented from damaging the opponent (from Protect variants or immunity).

Example screenshots:
furycutter0
furycutter1
furycutter2
furycutter3
furycutter4

src/battle.ts
```diff
@@ -1868,6 +1893,7 @@ class Battle {
 				}
 				break;
 			}
+			poke.removeVolatile('furycutter' as ID);
```
Zarel, line 1896: Probably just add it as a movestatus instead of having fifty `removeVolatile`s?

Zarel: The client splits `volatile`s into volatile/turnstatus/movestatus.

It'd probably be better to modify movestatus to work for this than to manually reimplement every movestatus feature.

Timbuktu22:
Looks like I accidentally merged my branch when I didn't want to, how do I restore to my last valid commit? ("Fix Protect and Immunity bug")

Zarel:
I would:

- `git reset --hard afb24cb` (to reset to that commit)
- `git cherry-pick 3bda7c5` (to grab your change from after the commit)

Note that you'd need to forcepush after that, because you'd be rewriting history

src/battle.ts
```diff
@@ -306,25 +306,25 @@ class Pokemon implements PokemonDetails, PokemonHealth {
...
 		if (this.hasMovestatus(volatile)) return;
-		this.movestatuses[volatile] = [volatile];
+		this.movestatuses[volatile] = [volatile, ...args] as EffectState;
 		this.side.battle.scene.addEffect(this, volatile);
 	}
 	hasMovestatus(volatile: ID) {
 		return !!this.movestatuses[volatile];
 	}
-	clearMovestatuses() {
+	clearMovestatusesExcept(exceptions: ID[] = []) {
```
Zarel, line 318: I wouldn't rename this; `clearMovestatuses(except: ID[] = []) {` is clear enough.

Zarel: I also don't think this needs to be an array.

On second thought, I would probably do this as `clearMovestatuses(move: Move)`, and then check for Fury Cutter inside the function.

Timbuktu22: I chose array because I'm implementing Metronome (item) much the same way in https://github.com/smogon/pokemon-showdown-client/pull/1792. Would I be able to keep the array for that purpose?

PR #1779, opened 2021-05-06
Automatically filter types for monotype formats

Karthik99999:
https://www.smogon.com/forums/threads/automatically-filter-by-type-when-teambuilding-for-monotype.3657897/

It will only filter if the 1st pokemon is a single type or when there is only 1 shared type between the team. (so the part in the suggestion I linked above about filtering dark and steel types at the same time for bisharp would be unfulfilled, but I didn't feel like changing how filters worked). I am also somewhat unsure if where I put the functionality is the best place to put it, but it seems to work pretty well in my testing.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1330,6 +1330,24 @@
 			this.curChartName = '';
 			this.update();
 			this.$('input[name=pokemon]').select();
+			if (this.curTeam.format.includes('monotype')) {
+				let typeTable = [];
+				for (const [i, pokemon] of this.curSetList.entries()) {
+					const species = Dex.forGen(this.curTeam.gen).species.get(pokemon.species);
+					if (!species.exists) continue;
+					if (i === 0) {
+						typeTable = species.types;
+					} else {
+						typeTable = typeTable.filter(function(type) {
+							return species.types.includes(type);
+						});
+					}
+				}
+				if (typeTable.length !== 1) return;
+				this.search.engine.addFilter(['type', typeTable[0]]);
+				this.search.filters = this.search.engine.filters;
+				this.search.find('');
+			}
 		},
 		pastePokemon: function (i, btn) {
 			if (!this.curTeam) return;
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1332,10 +1332,12 @@
 			this.$('input[name=pokemon]').select();
 			if (this.curTeam.format.includes('monotype')) {
 				var typeTable = [];
+				var dex = Dex.forGen(this.curTeam.gen);
 				for (var i = 0; i < this.curSetList.length; i++) {
-					var species = Dex.forGen(this.curTeam.gen).species.get(this.curSetList[i].species);
+					var set = this.curSetList[i];
+					var species = dex.species.get(set.species);
 					if (species.isMega) {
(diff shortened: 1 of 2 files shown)


js/client-teambuilder.js
```diff
@@ -1330,6 +1330,24 @@
 			this.curChartName = '';
 			this.update();
 			this.$('input[name=pokemon]').select();
+			if (this.curTeam.format.includes('monotype')) {
+				var typeTable = [];
+				for (var i = 0; i < this.curSetList.length; i++) {
+					const species = Dex.forGen(this.curTeam.gen).species.get(this.curSetList[i].species);
```
KrisXV, line 1336: I believe the client is still using getSpecies etc outside of build files currently

Karthik99999: > I believe the client is still using getSpecies etc outside of build files currently

it was matched to update the server api here https://github.com/smogon/pokemon-showdown-client/commit/31fcd418ecc905a4b46ba1f1dda9d19cb7755b4e

KrisXV: oh its because ur using `const` then mb

js/client-teambuilder.js
```diff
@@ -1330,6 +1330,24 @@
...
+				var typeTable = [];
+				for (var i = 0; i < this.curSetList.length; i++) {
+					var species = Dex.forGen(this.curTeam.gen).species.get(this.curSetList[i].species);
+					if (!species.exists) continue;
+					if (i === 0) {
+						typeTable = species.types;
+					} else {
+						typeTable = typeTable.filter(function (type) {
+							return species.types.includes(type);
+						});
```
Zarel, line 1343:
```suggestion
						});
						if (!typetable.length) break;
```

js/client-teambuilder.js
```diff
@@ -1330,6 +1330,24 @@
...
+					}
+				}
+				if (typeTable.length !== 1) return;
+				this.search.engine.addFilter(['type', typeTable[0]]);
+				this.search.filters = this.search.engine.filters;
+				this.search.find('');
```
Zarel, line 1349:
```suggestion
				if (typeTable.length === 1) {
					this.search.engine.addFilter(['type', typeTable[0]]);
					this.search.filters = this.search.engine.filters;
					this.search.find('');
				}
```

Early return can make things nicer but I think this is the wrong place for it.

Karthik99999:
There seems to be a bug where `isMega` is always false for older gens, so the mega implementation doesn't work properly on non gen 8 formats (which makes it useless lol). Don't know how I would go about fixing that, since the code that sets `isMega` to true seems like it should work.

js/client-teambuilder.js
```diff
@@ -1330,6 +1330,41 @@
...
+								return species.types.includes(type);
+							});
+							if (!typeTable.length) break;
+						}
+					}
+				}
+				if (typeTable.length === 1) {
+					this.search.engine.addFilter(['type', typeTable[0]]);
+					this.search.filters = this.search.engine.filters;
+					this.search.find('');
```
Zarel, line 1365: Not the ideal place for this (I'd definitely prefer to put it closer to where `engine` was cleared), but this will do for now.

Zarel:
Thanks!

PR #1780, opened 2021-05-06
Keep yawn volatile on -status

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8845946

Removes an unnecessary removal of the yawn volatile during the `-status` event, as a pokemon can still be drowsy with a status condition if they received the status after they were drowsy, and the yawn volatile gets removed on its own after its duration is up anyway.

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -1935,7 +1935,6 @@ class Battle {
 			let effect = Dex.getEffect(kwArgs.from);
 			let ofpoke = this.getPokemon(kwArgs.of) || poke;
 			poke.status = args[2] as StatusName;
-			poke.removeVolatile('yawn' as ID);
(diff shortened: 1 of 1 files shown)


DaWoblefet:
Just to make absolutely sure, as an example, if a Yawned Pokemon is paralyzed in the same turn, the Yawn volatile _shouldn't_ go away until next turn.

Zarel:
Does this apply whether or not the status is Sleep?

(Also, thanks!)

DaWoblefet:
> Does this apply whether or not the status is Sleep?
> 
> (Also, thanks!)

Yeah, you can Rest -> Lum Berry -> fall asleep from Yawn.

PR #1782, opened 2021-05-07
Fix D O G A R S

MrSableye:
Pokémon Showdown! Beta is absolutely unplayable without Roxie's JP gym theme playing in the background when I switch in a Koffing named Dogars. On a more serious note, it also does not trigger Miror B's theme when I use my extremely viable multi-Ludicolo team in AG.

This fixes the `setBgm` function to call `updateBgm` after changing the BGM. It also changes the `loadBgm` function to stop the current BGM from playing before removing it

### DIFF
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -1592,6 +1592,8 @@ class BattleScene {
 			this.bgm = BattleSound.loadBgm('audio/sm-rival.mp3', 11389, 62158, this.bgm);
 			break;
 		}
+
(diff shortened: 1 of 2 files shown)


MrSableye:
Please let me know if any changes or additional details are needed for this to move forward.

Zarel:
Sorry, I think this got lost in some other shuffle. I'm still confused what broke this, but I guess I'll figure that out later.

Thanks for the fix!

PR #1783, opened 2021-05-09
Add -swapsideconditions for court change

Karthik99999:
Corresponding server change: smogon/pokemon-showdown#8287

Does 2 things:
- adds an optional `info` argument to `side.addSideCondition`, which will allow preexisting side condition info to be used when adding a side condition (similar to `replaceSlot` for `side.addPokemon`).
- adds a new `-swapsideconditions` action, which swaps side conditions between sides

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -648,8 +648,13 @@ class Side {
 			if (this.foe && this.avatar === this.foe.avatar) this.rollTrainerSprites();
 		}
 	}
-	addSideCondition(effect: Effect) {
+	addSideCondition(effect: Effect, info?: [string, number, number, number]) {
 		let condition = effect.id;
+		if (info) {
+			this.sideConditions[condition] = info;
+			this.battle.scene.addSideCondition(this.n, condition);
+			return;
+		}
 		if (this.sideConditions[condition]) {
 			if (condition === 'spikes' || condition === 'toxicspikes') {
 				this.sideConditions[condition][1]++;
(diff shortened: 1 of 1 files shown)


src/battle.ts
```diff
@@ -2778,6 +2783,49 @@ class Battle {
 			this.log(args, kwArgs);
 			break;
 		}
+		case '-sideswitch': {
+			let effect = Dex.getEffect(args[1]);
+			if (this.sides.length > 2) {
```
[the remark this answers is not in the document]
Karthik99999: don't know why I didn't think of that

src/battle.ts
```diff
@@ -648,8 +648,13 @@ class Side {
...
 		}
 	}
-	addSideCondition(effect: Effect) {
+	addSideCondition(effect: Effect, info?: [string, number, number, number]) {
 		let condition = effect.id;
+		if (info) {
+			this.sideConditions[condition] = info;
+			this.battle.scene.addSideCondition(this.n, condition);
+			return;
+		}
```
Zarel, line 657: Considering this shares zero code with the other `addSideCondition` code, you should probably implement this as a separate `switchSides` function.

src/battle.ts
```diff
@@ -2778,6 +2802,30 @@ class Battle {
 			this.log(args, kwArgs);
 			break;
 		}
+		case '-sideswitch': {
+			let effect = Dex.getEffect(args[1]);
```
Zarel, line 2806: Are we sending this individually for each condition? It might make sense to switch them all at once?

src/battle.ts
```diff
@@ -2778,6 +2807,12 @@ class Battle {
 			this.log(args, kwArgs);
 			break;
 		}
+		case '-sideswitch': {
```
Zarel, line 2810:
```suggestion
		case '-swapsideconditions': {
```

On reflection, I think `sideswitch` is at least a little ambiguous.

Zarel: Sadly this doesn't match `sidestart` or `sideend` very well, but I think clarity is probably more important?

Zarel:
Thanks!

Zarel:
The behavior in FFA will have to wait until we can get consensus, but fortunately this syntax is backwards compatible so I don't mind.

PR #1784, opened 2021-05-09
Fix |updatepoke

Karthik99999:
While the use of `checkDetails` did allow for only the new details to be sent, I overlooked the fact that it didn't work if there is more than one zacian/zamazenta/xerneas (as shown [here](https://replay.pokemonshowdown.com/gen8anythinggoes-1337388488), where the battle log shows 2 `updatepoke` for 2 zacian-crowned but only 1 zacian gets updated). I've added an extra check to see if the details are already an exact match, which should fix the issue.


Zarel:
Thanks! Smart fix.

PR #1785, opened 2021-05-10
Fix Echoed Voice BP display and field display

Timbuktu22:


Timbuktu22:
This Fix will correct the BP shown in the move tooltip as well as display the BP in the move/scene tooltip.

Scene tooltip:
ev1_scene

Example screenshots for move tooltip:
[examples.zip](https://github.com/smogon/pokemon-showdown-client/files/6448884/examples.zip)

Zarel:
Shouldn't this only apply to consecutive turns? I don't see a turn counter.

Timbuktu22:
> Shouldn't this only apply to consecutive turns? I don't see a turn counter.
`echoedVoiceState` is activated per turn by Pokemon who use Echoed Voice, and if nobody activates it, the BP will be reset. I've added clarifying comments to that and other places where the logic might not be clear.

Timbuktu22:
Rebasing this branch to fix that failed check is currently not going well... idk how to fix it.

PR #1787, opened 2021-05-10
Fix some buttons shown after battle/replay not conforming to dark mode

Joehuu:
| Before | After |
| --- | --- |
| play pokemonshowdown com_battle-gen8randombattle-1338161221 (2) | _Users_joseph_Documents_GitHub_pokemon-showdown-client_testclient html (1) |
| play pokemonshowdown com_battle-gen8randombattle-1338161221 (1) | _Users_joseph_Documents_GitHub_pokemon-showdown-client_testclient html |

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -283,7 +283,7 @@
 
 			if (this.battle.ended) {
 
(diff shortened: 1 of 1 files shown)


Zarel:
Putting the switch sides button into its own variable is certainly a choice. This isn't the Preact client so I'm just going to not bother doing a more in-depth modernization; I suppose this is fine as-is.

Zarel:
Preact client uses `class="button"` for all buttons. Probably we should also port that to the old client?

Off the top of my head, there's a lot of buttons in the teambuilder that could be converted. Also the usercard, and the challenge UI.

Joehuu:
Yeah, just did the change on a specific area. Will see all other button usages and make another PR.

PR #1789, opened 2021-05-11
Fix nduubl again

Karthik99999:
Just had to move uubl above ou in the return order. Probably should've done this the first time


Zarel:
Thanks!

PR #1793, opened 2021-05-13
Challenge : Fix notification sound

Distrib-ps:
image


Zarel:
LOL, right as we refactor challenges. Thanks!

PR #1797, opened 2021-05-14
Teambuilder: Fix Gen VII analysis links

DaWoblefet:
Fixes https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8849061

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1994,9 +1994,9 @@
 
 			var generationNumber = 8;
 			if (format.substr(0, 3) === 'gen') {
-				var number = format.charAt(3);
-				if ('1' <= number && number <= '7') {
-					generationNumber = +number;
+				var number = parseInt(format.charAt(3));
+				if (1 <= number && number <= 7) {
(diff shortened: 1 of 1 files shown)


js/client-teambuilder.js
```diff
@@ -1995,7 +1995,7 @@
 			var generationNumber = 8;
 			if (format.substr(0, 3) === 'gen') {
 				var number = format.charAt(3);
```
Zarel, line 1997:
```suggestion
				var number = parseInt(format.charAt(3));
```

js/client-teambuilder.js
```diff
@@ -1995,7 +1995,7 @@
...
-				if ('1' <= number && number <= '6') {
+				if ('1' <= number && number <= '7') {
 					generationNumber = +number;
```
Zarel, line 1999:
```suggestion
				if (1 <= number && number <= 7) {
					generationNumber = number;
```

Zarel:
ty!

PR #1799, opened 2021-05-14
Support new challenge protocol

Zarel:
Previously, we'd get information for challenges in a single `|updatechallenges|` JSON blob.

That was designed for an even older UX, before challenges were integrated inside PM boxes.

The new format sends `/challenge` inside the PM, allowing the server to more directly control the challenge box inside the PM. This also makes it easier for things that aren't challenges to use the same UI, so e.g. FFA battles can ask for teams, or people can challenge each other to Rock Paper Scissors.

The new format looks like this: `|pm|SENDER|RECEIVER|/challenge GAME|TEAM|MESSAGE|ACCEPTBUTTON|REJECTBUTTON`.

Everything after `/challenge` is optional:

- `GAME` is the format name or game name - if left blank: the format/game name box won't show

- `TEAM` is the teambuilder format name - if left blank: the team dropdown won't show

- `MESSAGE` is the message up top - if left blank: "[user] wants to battle!"

- `ACCEPTBUTTON` is the label on the Accept button - if left blank: "Accept"

- `REJECTBUTTON` is the label on the Reject button - if left blank: "Reject"

1. A message containing `GAME` and `TEAM` is a regular challenge. If you are `RECEIVER`, it will display the format name, a team selector, and "Accept" / "Reject" buttons. Otherwise, you are `SENDER` and it will display the format and a "Cancel" button.

    `/challenge gen8ou|gen8ou`

   image

2. A message containing `GAME` (with `TEAM` blank) will work as above, but with no team selector.

   `/challenge Rock Paper Scissors`

   image

3. A message containing `MESSAGE` (with both `GAME` and `TEAM` blank) will work as above, but with no format or team box.

   `/challenge ||Do you like cheese?|Yes|No`

   image

4. `MESSAGE`, `ACCEPTBUTTON`, and `REJECTBUTTON` are optional for all of the above (unless `GAME` and `TEAM` are both blank, in which case `MESSAGE` is required), and can be mixed and matched:

   `/challenge gen8freeforall|gen8freeforall|Please pre-register your team for the upcoming tournament.||No thanks`

   image

5. If you are the sender, the message will always be replaced with "Waiting for [username]...", the Accept/Reject buttons will always be replaced with "Cancel", and there will never be a team selector.

   (as the sender) `/challenge gen8freeforall|gen8freeforall|Please pre-register your team for the upcoming tournament.||No thanks`
   image

6. A blank message (`/challenge` by itself) will close any open challenge box

   `/challenge`

    image

7. Any other message is invalid (currently treated as a blank message, but don't rely on this)

No matter what they're labeled, the buttons send `/accept`, `/reject`, and `/cancelchallenge`. If a team dropdown exists, the team will be sent with `/utm` right before `/accept`.

### DIFF
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1746,7 +1746,7 @@
 
 			var isHighlighted = userid !== app.user.get('userid') && this.getHighlight(message);
 			var parsedMessage = MainMenuRoom.parseChatMessage(message, name, ChatRoom.getTimestamp('chat', msgTime), isHighlighted, this.$chat, true);
-			if (parsedMessage.challenge) {
+			if (typeof parsedMessage.challenge === 'string') {
 				this.$chat.append('<div class="chat message-error">The server sent a challenge but this isn\'t a PM window!</div>');
 				return;
 			}
--- a/js/client-chat.js
+++ b/js/client-chat.js
@@ -1746,6 +1746,10 @@
 
 			var isHighlighted = userid !== app.user.get('userid') && this.getHighlight(message);
 			var parsedMessage = MainMenuRoom.parseChatMessage(message, name, ChatRoom.getTimestamp('chat', msgTime), isHighlighted, this.$chat, true);
+			if (parsedMessage.challenge) {
+				this.$chat.append('<div class="chat message-error">The server sent a challenge but this isn\'t a PM window!</div>');
+				return;
+			}
 			if (typeof parsedMessage === 'object' && 'noNotify' in parsedMessage) {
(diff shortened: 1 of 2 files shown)


Zarel:
Sending only `MESSAGE` with no `GAME` can also be used for a generic confirmation prompt! Like possibly a friend request!

Zarel:
When this gets implemented into Preact, it will be supported in rooms as well as PMs... allowing you to use this UI for tournaments and roomgames! But that's extremely not a feature I have any intention of adding to the old client. :(

PR #1800, opened 2021-05-14
Explicitly notate messages sent to lobby

mia-pi-git:
Explanation from Zarel:

PS's client-to-server messages are in the format `ROOMID|MESSAGE`, but with a special exception: Lobby uses a blank roomid. This exception saves a small amount of bandwidth, but causes a few weird bugs, and the bandwidth savings are no longer worth it.

-----

Necessary before server changes to make `|MSG` send to a console, instead of lobby.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -816,7 +816,7 @@ function toId() {
 		 * Send to sim server
 		 */
 		send: function (data, room) {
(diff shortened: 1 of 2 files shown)


Zarel:
Does the Preact client need any changes?

Zarel:
ty!

PR #1805, opened 2021-05-15
Implement Dynamax Levels

Karthik99999:
Server Side: smogon/pokemon-showdown#8308

Luckily this was pretty easy to add on the preact code (for now anyway since there's not much for the teambuilder :P).


js/client-teambuilder.js
```diff
@@ -2647,6 +2651,14 @@
 				delete set.shiny;
 			}
 
+			// dynamax level
+			var dynamaxLevel = parseInt(this.$chart.find('input[name=dynamaxlevel]').val(), 10);
+			if (isNaN(dynamaxLevel)) dynamaxLevel = 10;
+			if (dynamaxLevel > 10) dynamaxLevel = 10;
+			if (dynamaxLevel < 0) dynamaxLevel = 10;
```
KrisXV, line 2658:
```suggestion
			if (isNaN(dynamaxLevel) || dynamaxLevel > 10 || dynamaxLevel < 0) dynamaxLevel = 10;
```

js/client-teambuilder.js
```diff
@@ -1200,6 +1200,7 @@
 					}
 				}
 				if (this.curTeam.gen === 8) {
+					if (!species.cannotDynamax) buf += '<span class="detailcell"><label>Dmax Level</label>' + (typeof set.dynamaxLevel === 'number' ? set.dynamaxLevel : 10) + '</span>';
```
KrisXV, line 1203: we probably shouldnt display this and should arguably remove more things because it causes the UI to become extremely ugly for people with smaller screens

Zarel: I definitely think it should be left off if it's 10.

Zarel: I think it makes sense to leave it in if it's some other number, though.

AnnikaCodes:
Thanks!

PR #1806, opened 2021-05-19
Default Doubles LC sets to level 5

AnnikaCodes:
Requested by Coconut.

Should this also be refactored to use `String#startsWith`/`String#endsWith`?

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -3154,7 +3154,7 @@
 				if (baseFormat.substr(0, 8) === 'pokebank') baseFormat = baseFormat.substr(8);
 				if (this.curTeam && this.curTeam.format) {
 					if (baseFormat.substr(0, 10) === 'battlespot' && baseFormat.substr(0, 19) != 'battlespotspecial13' || baseFormat.substr(0, 3) === 'vgc') set.level = 50;
-					if (baseFormat.substr(0, 2) === 'lc' || baseFormat.substr(0, 5) === 'caplc') set.level = 5;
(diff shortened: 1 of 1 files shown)


js/client-teambuilder.js
```diff
@@ -3154,7 +3154,7 @@
 				if (baseFormat.substr(0, 8) === 'pokebank') baseFormat = baseFormat.substr(8);
 				if (this.curTeam && this.curTeam.format) {
 					if (baseFormat.substr(0, 10) === 'battlespot' && baseFormat.substr(0, 19) != 'battlespotspecial13' || baseFormat.substr(0, 3) === 'vgc') set.level = 50;
-					if (baseFormat.substr(0, 2) === 'lc' || baseFormat.substr(0, 5) === 'caplc') set.level = 5;
+					if (baseFormat.substr(0, 2) === 'lc' || baseFormat.substr(0, 5) === 'caplc' || baseFormat.substr(-2) === 'lc') set.level = 5;
```
xInstruct, line 3157: > Should this also be refactored to use String#startsWith/String#endsWith?

Probably? ` baseFormat.substr(-2) === 'lc'` makes `baseFormat.substr(0, 5) === 'caplc'` reduandant if you look at it.

AnnikaCodes: I assumed that check should be kept in case we introduce CAP LC UU or something.

KrisXV:
Didn’t I already do this?

AnnikaCodes:
The client doesn't default to level 5 for Doubles LC currently, so no, it doesn't look like you did.

Marty-D (approved):
Thanks!

PR #1807, opened 2021-05-20
Tooltips: Don't show sandstorm Sp. Def boost after type change

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8521042

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -1412,7 +1412,7 @@ class BattleTooltips {
 		value.reset(move.accuracy === true ? 0 : move.accuracy, true);
 
 		let pokemon = value.pokemon!;
-		if (move.id === 'toxic' && this.battle.gen >= 6 && this.pokemonHasType(pokemon, 'Poison', pokemon.getTypeList())) {
+		if (move.id === 'toxic' && this.battle.gen >= 6 && this.pokemonHasType(pokemon, 'Poison')) {
 			value.set(0, "Poison type");
 			return value;
 		}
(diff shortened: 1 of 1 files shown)


src/battle-tooltips.ts
```diff
@@ -1076,7 +1076,7 @@ class BattleTooltips {
 			stats.atk = Math.floor(stats.atk * 1.5);
 		}
 		if (weather) {
-			if (this.battle.gen >= 4 && this.pokemonHasType(serverPokemon, 'Rock') && weather === 'sandstorm') {
+			if (this.battle.gen >= 4 && this.pokemonHasType(serverPokemon, 'Rock', clientPokemon?.getTypeList(serverPokemon)) && weather === 'sandstorm') {
```
Zarel, line 1079:
```suggestion
			if (this.battle.gen >= 4 && this.pokemonHasType(pokemon, 'Rock') && weather === 'sandstorm') {
```

src/battle-tooltips.ts
```diff
@@ -1412,7 +1412,7 @@ class BattleTooltips {
 		value.reset(move.accuracy === true ? 0 : move.accuracy, true);
 
 		let pokemon = value.pokemon!;
-		if (move.id === 'toxic' && this.battle.gen >= 6 && this.pokemonHasType(pokemon, 'Poison')) {
+		if (move.id === 'toxic' && this.battle.gen >= 6 && this.pokemonHasType(pokemon, 'Poison', pokemon.getTypeList())) {
```
Zarel, line 1415: in theory this shouldn't do anything because it should happen automatically

src/battle-tooltips.ts
```diff
@@ -1433,7 +1433,7 @@ class BattleTooltips {
 				return value;
 			}
 			if (move.id === 'sheercold' && this.battle.gen >= 7) {
-				if (!this.pokemonHasType(pokemon, 'Ice')) value.set(20, 'not Ice-type');
+				if (!this.pokemonHasType(pokemon, 'Ice', pokemon.getTypeList())) value.set(20, 'not Ice-type');
```
Zarel, line 1436: ditto

Zarel:
> Should the `pokemonHasType` function just do this automatically instead of taking a parameter?

It already does it automatically...

Karthik99999:
> It already does it automatically...

oop, was not aware of this (probably should've tested it)

Marty-D (approved):
Thanks!

PR #1808, opened 2021-05-21
Gen 1: Account for 1/256 accuracy glitch in move tooltip

Karthik99999:
https://www.smogon.com/forums/threads/consider-the-x-256-uncertainty-in-rby-move-accuracy.3668508/

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -1471,7 +1471,7 @@ class BattleTooltips {
 			value.modify(5 / 3, "Gravity");
 		}
 		// 1/256 glitch
-		if (this.battle.gen === 1) {
+		if (this.battle.gen === 1 && !toID(this.battle.tier).includes('stadium')) {
 			value.set((Math.floor(value.value * 255 / 100) / 256) * 100);
 		}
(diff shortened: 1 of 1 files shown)


src/battle-tooltips.ts
```diff
@@ -1470,6 +1470,10 @@ class BattleTooltips {
 		if (this.battle.hasPseudoWeather('Gravity')) {
 			value.modify(5 / 3, "Gravity");
 		}
+		// 1/256 glitch
+		if (this.battle.gen === 1 && !toID(this.battle.tier).includes('stadium')) {
+			value.set((Math.floor(value.value * 255 / 100) / 256) * 100);
```
Zarel, line 1475: Should this be rounded? Is it already rounded?

Karthik99999: It gets rounded whenever accuracy is shown (Tested with 100% accurate move, actual number is 99.609375 and it showed 99.61% in the tooltip).

Zarel: Would prefer fewer digits, but I suppose that's good enough.

PR #1809, opened 2021-05-22
Tooltips: Don't show stat boosts from items if Magic Room is active or if the Pokemon is affected by Embargo

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-40#post-8716218


Marty-D (approved):
Thanks!

PR #1810, opened 2021-05-24
Tooltips: Fix various Pressure related bugs

Karthik99999:
- Makes sure neutralizing gas isn't active and that the target isn't affected by gastro acid
- Makes sure the pokemon with pressure isn't fainted
- Fixes pressure in FFA
- Adds a hardcode to singles for moves without a target (such as 2 turn moves and moves that fail).

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -1365,10 +1365,10 @@ class Battle {
 				}
 			}
 			if (!ngasActive) {
-				if (move.target === 'all') {
+				if (['foeSide', 'all', 'allAdjacent', 'allAdjacentFoes'].includes(move.target)) {
 					// Looping through all sides for FFA
 					for (const side of this.sides) {
-						if (side === pokemon.side) continue;
+						if (side === pokemon.side || side === pokemon.side.ally) continue;
 						for (const active of side.active) {
 							if (active && !active.fainted && toID(active.ability) === 'pressure') {
 								pp += 1;
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -1364,8 +1364,17 @@ class Battle {
 					}
 				}
 			}
-			if (!ngasActive) {
-				if (['foeSide', 'all', 'allAdjacent', 'allAdjacentFoes'].includes(move.target)) {
+			// Sticky Web is never affected by pressure
+			if ((!target?.volatiles['gastroacid'] || !ngasActive) && move.id !== 'stickyweb') {
+				// Hardcode for moves without a target in singles
+				const activeFoe = pokemon.side.foe.active[0];
+				if (
+					!target && this.gameType === 'singles' &&
+					!['self', 'allies', 'allySide', 'adjacentAlly', 'adjacentAllyOrSelf'].includes(move.target) &&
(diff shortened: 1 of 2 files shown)


DaWoblefet:
To clarify the last bullet, this is necessary because no target is sent anymore for things like, say, Toxic into a burned target or the first turn of Dig, because those things leak target information in formats with multiple Pokemon. It would be an information leak to send that kind of info in doubles. But in singles, there's no such potential ambiguity, so it's OK to hardcode it there.

Zarel (approved):
Not the simplest code, but it looks like it works.

If I were doing this I would probably split up the "get list of targets" and "check if targets have Pressure" code.

src/battle.ts
```diff
@@ -1355,14 +1355,38 @@ class Battle {
...
 			}
 			let pp = 1;
-			if (move.target === "all") {
-				for (const active of pokemon.side.foe.active) {
-					if (active && toID(active.ability) === 'pressure') {
-						pp += 1;
+			let ngasActive = false;
+			for (const side of this.sides) {
+				for (const active of side.active) {
+					if (active && toID(active.ability) === 'neutralizinggas') {
```
Zarel, line 1361:
```suggestion
					if (active && !active.fainted && toID(active.ability) === 'neutralizinggas') {
```

src/battle.ts
```diff
@@ -1355,14 +1355,38 @@ class Battle {
...
+						if (side === pokemon.side || side === pokemon.side.ally) continue;
+						for (const active of side.active) {
+							if (active && !active.fainted && toID(active.ability) === 'pressure') {
+								pp += 1;
+							}
+						}
 					}
+				} else if (target && target.side !== pokemon.side && !target.fainted && toID(target.ability) === 'pressure') {
+					pp += 1;
 				}
```
Zarel, line 1389:
```suggestion
			if (!ngasActive && move.id !== 'stickyweb') {
				const foeTargets = [];

				if (
					!target && this.gameType === 'singles' &&
					!['self', 'allies', 'allySide', 'adjacentAlly', 'adjacentAllyOrSelf'].includes(move.target)
				) {
					// Hardcode for moves without a target in singles
					foeTargets.push(pokemon.side.foe.active[0]);
				} else if (['all', 'allAdjacent', 'allAdjacentFoes', 'foeSide'].includes(move.target)) {
					// Looping through all sides for FFA
					for (const side of this.sides) {
						if (side === pokemon.side || side === pokemon.side.ally) continue;
						for (const active of side.active) {
							foeTargets.push(active);
						}
					}
				} else if (target && target.side !== pokemon.side) {
					foeTargets.push(target);
				}

				for (const foe of foeTargets) {
					if (foe && !foe.fainted && toID(foe.ability) === 'pressure' && !foe.volatiles['gastroacid']) {
						pp += 1;
					}
				}
```

Zarel: This is what I meant by

> split up the "get list of targets" and "check if targets have Pressure" code

Zarel:
(Looks good now, just leaving open a bit for Marty)

Marty-D (approved):
Nice, thanks!

PR #1811, opened 2021-05-25
Don't display 100% HP lost if HP still exists

DaWoblefet:
If an attack deals more than 99.5% damage, the display will round up on the animation that shows up mid-screen above the Pokemon's sprite in red to indicate the Pokemon took "100%" damage even when it survived. Now, it only displays "99%" damage if the Pokemon were to take near-100% damage.

Note that this only affects player perspective mid-battle; you wouldn't have this issue as a spectator or from the opponent's perspective because it's truncated to 99% earlier.

I originally tried modifying this in getFormattedRange() but encountered problems getting it to return the more precise percentages in the battle log itself.

Replay: https://replay.pokemonshowdown.com/gen8ubers-1346819616
image
image

### DIFF
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -1408,6 +1408,7 @@ class BattleScene {
 			callback = () => { $hp.addClass('hp-yellow hp-red'); };
 		}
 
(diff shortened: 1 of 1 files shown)


Zarel:
Thanks!

PR #1812, opened 2021-05-25
Tooltips: Display sound/powder only after Gen 5

DaWoblefet:
Sound moves only bypass Substitute in Gen 6 and later. Same for Powder moves into Grass, etc.

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -709,7 +709,7 @@ class BattleTooltips {
 			if (!move.flags.protect && !['self', 'allySide'].includes(move.target)) {
 				text += `<p class="movetag">Not blocked by Protect <small>(and Detect, King's Shield, Spiky Shield)</small></p>`;
 			}
-			if (move.flags.authentic) {
+			if (move.flags.authentic && !(move.flags.sound && this.battle.gen < 6)) {
 				text += `<p class="movetag">Bypasses Substitute <small>(but does not break it)</small></p>`;
 			}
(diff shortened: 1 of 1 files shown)


Zarel:
This is why we used to use short descriptions in old gens. There are probably more things like this... :/

Zarel:
@Karthik99999 the "bypass Substitute" line is no longer necessary with your past-gen move data change, right?

Karthik99999:
> @Karthik99999 the "bypass Substitute" line is no longer necessary with your past-gen move data change, right?

Pretty sure it's still needed, as the gen 5 mod doesn't explicitly remove these flags, rather the sim just doesn't use them for gen 5 and below.

Zarel:
That doesn't seem possible... let me look into that.

Zarel:
It looks like yes, the Gen 5 mod does, in fact, explicitly remove these flags:

```ts
snore: {
	inherit: true,
	basePower: 40,
	flags: {protect: 1, mirror: 1, sound: 1},
},
```

Zarel:
I'm definitely in favor of automating this change to make the `moves` page smaller, but this is what it is now.

Karthik99999:
> It looks like yes, the Gen 5 mod does, in fact, explicitly remove these flags:
> 
> ```ts
> snore: {
> 	inherit: true,
> 	basePower: 40,
> 	flags: {protect: 1, mirror: 1, sound: 1},
> },
> ```

oops, I didn't read your first comment properly. i checked for powder lol

PR #1813, opened 2021-05-25
Override more Pokedex and Move data

Karthik99999:
Rather than individually picking and choosing which data to override, I think it's best to just have one object containing all the data to override, at least for the Pokedex and Moves since those have a lot of data to override. The new way I implemented the override allows for more data to be overridden without having to manually add each attribute.

For example, the Arceus formes will now automatically get their plate in gen 6 and below. It also doesn't show unreleased hidden abilities from gen 5 (right now Zapdos, for example, shows static in gen 5 despite it not being valid).

As for moves, the target of a move can be overridden now, as well as its flags (one example of flags needing to be overridden is rock blast, which wasn't considered a bullet move in gen 6 but is gen 7, so right now it incorrectly shows rock blast being a bullet move in gen 6 games).

I also went ahead and added a `dex` attribute to `curTeam`, so that `Dex.forGen` doesn't have to be called everytime, and so that `overrideType` doesn't have to be used there.

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -933,80 +933,38 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 		const gen = 'gen' + genNum;
 		const genData = Dex.mod(gen).data;
 		const nextGenData = Dex.mod('gen' + (genNum + 1)).data;
-		const overrideStats = {};
-		BattleTeambuilderTable[gen].overrideStats = overrideStats;
-		const overrideType = {};
-		BattleTeambuilderTable[gen].overrideType = overrideType;
-		const overrideAbility = {};
-		BattleTeambuilderTable[gen].overrideAbility = overrideAbility;
-		const overrideHiddenAbility = {};
-		BattleTeambuilderTable[gen].overrideHiddenAbility = overrideHiddenAbility;
-		const removeSecondAbility = {};
-		BattleTeambuilderTable[gen].removeSecondAbility = removeSecondAbility;
+		const overrideDexData = {};
+		BattleTeambuilderTable[gen].overrideDexData = overrideDexData;
 		for (const id in genData.Pokedex) {
 			const pastEntry = genData.Pokedex[id];
 			const nowEntry = Dex.data.Pokedex[id];
-			const nowType = nowEntry.types.join('/');
-			for (const stat in pastEntry.baseStats) {
-				if (stat === 'spd' && genNum === 1) continue;
-				if (pastEntry.baseStats[stat] !== nowEntry.baseStats[stat]) {
-					if (!overrideStats[id]) overrideStats[id] = {};
-					overrideStats[id][stat] = pastEntry.baseStats[stat];
+			for (const key in pastEntry) {
+				const pastString = JSON.stringify(pastEntry[key]);
+				const nowString = JSON.stringify(nowEntry[key]);
+				if (pastString !== nowString) {
+					if (!overrideDexData[id]) overrideDexData[id] = {};
+					overrideDexData[id][key] = pastEntry[key];
 				}
-			}
-			if (pastEntry.types.join('/') !== nowType) {
-				overrideType[id] = pastEntry.types.join('/');
-			}
-			if (pastEntry.abilities['0'] !== nowEntry.abilities['0']) {
-				overrideAbility[id] = pastEntry.abilities['0'];
-			}
-			if (pastEntry.abilities['H'] !== nowEntry.abilities['H']) {
-				overrideHiddenAbility[id] = pastEntry.abilities['H'];
-			}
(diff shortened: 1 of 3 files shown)


Zarel:
It looks like at some point, we stopped caching intermediate move desc changes.

(It used to be that `overrideMoveDesc` would only exist for changes between current gen and next gen, rather than for changes between current gen and latest gen, for less memory usage.)

Zarel:
While I agree this new format could lead to simpler code, I'm not sure we want to include every property ever. I think our past strategy of only including properties relevant to the client would be much better bandwidth-wise.

Karthik99999:
> It looks like at some point, we stopped caching intermediate move desc changes.
> 
> (It used to be that `overrideMoveDesc` would only exist for changes between current gen and next gen, rather than for changes between current gen and latest gen, for lower bandwidth usage.)

I know when this changed, because it was in one my other PR's lol (the one about old gen teambuilder bugs). I can go back to that system if you want.

Karthik99999:
> While I agree this new format could lead to simpler code, I'm not sure we want to include every property ever. I think our past strategy of only including properties relevant to the client would be much better bandwidth-wise.

I guess. Maybe I could make an array of relevant properties and only modify those?

Zarel:
> I know when this changed, because it was in one my other PR's lol (the one about old gen teambuilder bugs). I can go back to that system if you want.

It might be worth just going to that system for everything. So all diffs were between current gen and next gen. That would simplify code over the hybrid system, while providing even more bandwidth savings.

Zarel:
> I guess. Maybe I could make an array of relevant properties and only modify those?

Sounds great!

build-tools/build-indexes
```diff
@@ -933,80 +933,40 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
-		BattleTeambuilderTable[gen].overrideStats = overrideStats;
-		const overrideType = {};
-		BattleTeambuilderTable[gen].overrideType = overrideType;
-		const overrideAbility = {};
-		BattleTeambuilderTable[gen].overrideAbility = overrideAbility;
-		const overrideHiddenAbility = {};
-		BattleTeambuilderTable[gen].overrideHiddenAbility = overrideHiddenAbility;
-		const removeSecondAbility = {};
-		BattleTeambuilderTable[gen].removeSecondAbility = removeSecondAbility;
+		const overrideDexData = {};
```
Zarel, line 936: I think `overrideSpeciesData` would be clearer than `overrideDexData`.

src/battle-dex.ts
```diff
@@ -851,15 +851,16 @@ class ModdedDex {
...
-			if (id in table.overrideMoveCategory) data.category = table.overrideMoveCategory[id];
-			if (id in table.overrideMoveDesc) data.desc = table.overrideMoveDesc[id];
-			if (id in table.overrideMoveShortDesc) data.shortDesc = table.overrideMoveShortDesc[id];
-
+			const overrideMoveKeys = ['accuracy', 'basePower', 'category', 'desc', 'flags', 'pp', 'shortDesc', 'target', 'type'];
+			for (const key of overrideMoveKeys) {
+				for (var i = this.gen; i < 8; i++) {
+					const table = window.BattleTeambuilderTable['gen' + i];
+					if (id in table.overrideMoveData && key in table.overrideMoveData[id]) {
+						data[key] = table.overrideMoveData[key];
```
Zarel, line 859: You missed `id` here. `table.overrideMoveData[id][key]` not ``table.overrideMoveData[key]`. Of course, that's moot because I'm rewriting this anyway.

src/battle-dex.ts
```diff
@@ -851,15 +851,16 @@ class ModdedDex {
...
+						break;
+					}
+				}
+			}
```
Zarel, line 863:
```suggestion
			for (let i = Dex.gen - 1; i >= this.gen; i--) {
				const table = window.BattleTeambuilderTable[`gen${i}`];
				if (id in table.overrideMoveData) {
					Object.assign(data, table.overrideMoveData[id]);
				}
			}
```

Surprised the linter isn't complaining about the `var`.

Anyway, here, this code passes TypeScript.

Zarel: It's also more generic (doesn't require updating this if we change what properties we add to `overrideMoveData`). I don't even think the perf difference is noticeable.

Zarel: Remember to do the same thing for Species.

Zarel:
Looks like `teambuilder-tables.js` is down to 3.8MB from 4.0MB. Nice!

Zarel:
Thanks a lot!

Some comments on future plans:

- We might want to migrate items and abilities to also use this system
- We might want to move descriptions outside of this system (and into `text.js`, with full descriptions); so that `.desc` and `.shortDesc` aren't properties on `Move` etc but separated into the text parser or something.

PR #1814, opened 2021-05-26
Add OU by technicality to RBY

PlagueVonKarma:
so this works, read the proposal there for details: https://github.com/smogon/pokemon-showdown/pull/8336


Zarel:
tyty

PR #1815, opened 2021-05-26
Show Air Lock/Cloud Nine activation in battle scene

Karthik99999:
https://www.smogon.com/forums/threads/alter-don%E2%80%99t-show-weather-backgrounds-if-air-lock-cloud-nine-is-in-effect.3684932/

Works for the most part, except for the fact that it doesn't immediately update the weather once air lock leaves, and I don't know the right place to run `updateWeather` to fix that.

### DIFF
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -921,28 +921,8 @@ class BattleScene {
 			} else if (this.battle.weatherTimeLeft !== 0) {
 				weatherhtml += ` <small>(${this.battle.weatherTimeLeft} turn${this.battle.weatherTimeLeft === 1 ? '' : 's'})</small>`;
 			}
-			let nullWeather = false;
-			let ngasActive = false;
-			for (const side of this.battle.sides) {
-				for (const active of side.active) {
-					if (active && !active.fainted && toID(active.ability) === 'neutralizinggas' && !active.volatiles['gastroacid']) {
-						ngasActive = true;
-						break;
-					}
-				}
-			}
-			for (const side of this.battle.sides) {
-				for (const active of side.active) {
-					if (
-						active && !active.fainted &&
-						!ngasActive && !active.volatiles['gastroacid'] &&
-						['Air Lock', 'Cloud Nine'].includes(active.ability)
-					) {
-						nullWeather = true;
-					}
-				}
-			}
-			weatherhtml = `${nullWeather ? '<s>' : ''}${weatherhtml}${nullWeather ? '</s>' : ''}`;
+			const nullifyWeather = this.battle.anyHasAbility('Air Lock', 'Cloud Nine');
+			weatherhtml = `${nullifyWeather ? '<s>' : ''}${weatherhtml}${nullifyWeather ? '</s>' : ''}`;
 		}
 
 		for (const pseudoWeather of this.battle.pseudoWeather) {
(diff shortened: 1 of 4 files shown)


Karthik99999:
Don't know what the build error I got is... didn't get it on my end.

DaWoblefet:
What's the current behavior if one weather overrides another while Air Lock is active? It updates the strikethough text, right?

Make sure to account for Ability negation (Gastro Acid / NGas). We probably ought to make a general function for that with how often we need to call it.

Also, I wonder if there's a way to indicate the weather is suppressed without dropping the weather entirely - maybe reducing the opacity so it looks like it's faded out?

DaWoblefet:
Also, Air Lock/Cloud Nine don't always have an entrance message. For example, if Neutralizing Gas switches out, Air Lock will actually be active again, but it won't trigger the messages again. A similar thing happens for Skill Swap. This behavior doesn't seem dependent on that but would lead to information leaks in a situation like this:

1) Neutralizing Gas switches in
2) Cloud Nine switches in
3) Neutralizing Gas switches out

Now Cloud Nine is active again, but it was never announced to have been active. If it's ambiguous whether or not the Pokemon has Cloud Nine (say, a Golduck) then this would be an information leak. I'm not sure how we'd get around this.

Zarel:
This isn't a problem because everything the client knows is from announcement messages. So the client will be wrong in those cases, but it won't leak information.

Karthik99999:
I tested, and yeah the thing with Air Lock not announcing itself if it was brought it when Neutralizing Gas was active is not a problem, the weather showed just fine despite air lock being active, since it wasn't announced.

Karthik99999:
Also I'm not too sure how I can change the opacity of just the weather image, since it seems like changing the opacity with how its setup now also affects the weather text.

DaWoblefet:
Makes sense re: client leaks - just want to be careful. Sounds good to me!

Zarel:
> Works for the most part, except for the fact that it doesn't immediately update the weather once air lock leaves, and I don't know the right place to run `updateWeather` to fix that.

Considering you're checking for `active.ability`, presumably whenever `active` changes, such as at the end of the  `|switch|` handler.

src/battle-animations.ts
```diff
@@ -948,6 +970,13 @@ class BattleScene {
 		if (!this.animating) return;
 		let isIntense = false;
 		let weather = this.battle.weather;
+		for (const side of this.battle.sides) {
+			for (const active of side.active) {
+				if (active && !active.fainted && ['Air Lock', 'Cloud Nine'].includes(active.ability)) {
```
Zarel, line 975: Considering all the places we need to check for Neutralizing Gas and Gastro Acid, we probably want a `pokemon.effectiveAbility()` function.

Zarel: Which can also check `.fainted`, saving us another thing here.

src/battle.ts
```diff
@@ -512,6 +512,24 @@ class Pokemon implements PokemonDetails, PokemonHealth {
 		}
 		return !this.getTypeList(serverPokemon).includes('Flying');
 	}
+	effectiveAbility(abilities: string | string[], serverPokemon?: ServerPokemon) {
```
Zarel, line 515: I would rather this not take an array, but instead return a string.

src/battle.ts
```diff
@@ -1138,6 +1156,26 @@ class Battle {
...
+		for (const side of this.sides) {
+			for (const active of side.active) {
+				if (active && !active.fainted && toID(active.ability) === 'neutralizinggas' && !active.volatiles['gastroacid']) {
+					return true;
+				}
+			}
+		}
+		return false;
+	}
+	anyHasAbility(...abilities: string[]) {
```
Zarel, line 1169: Possibly call this `abilityActive`.

Zarel: Also if you wanted to optimize this you could, instead of using `effectiveAbility`, call `ngasActive` once and then manually check for effective abilities, skipping that loop.

Zarel:
ty!

PR #1816, opened 2021-05-27
Fix Mummy showing duplicate messages on failure

Karthik99999:
https://github.com/smogon/pokemon-showdown/projects/3#card-61195884

### DIFF
--- a/src/battle-text-parser.ts
+++ b/src/battle-text-parser.ts
@@ -837,7 +837,7 @@ class BattleTextParser {
 				return line1 + template.replace('[POKEMON]', this.pokemon(kwArgs.of)).replace('[SOURCE]', this.pokemon(pokemon));
 			}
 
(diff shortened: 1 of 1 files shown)


Zarel:
What _should_ happen in this case?

Zarel:
Should Mummy not activate at all? If so, the server shouldn't be sending the Mummy activation.

Karthik99999:
If mummy fails due to something like Protective Pads (which is what the bug report mentions), it should still show the ability activating but nothing else (which requires the server sending the `-activate` message). `kwArgs.ability` is not defined when Mummy should fail, which is why it should be checked here to determine if Mummy’s `changeAbility` text should be shown.

Zarel:
OK, I see. I think this is the correct solution, then.

Zarel:
Thanks!

PR #1817, opened 2021-05-27
Fix transformed Pokemon showing wrong speed range

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8747482


Zarel:
Thanks!

I'm surprised `getSpecies` doesn't follow the Transform volatile. Maybe it should?

PR #1818, opened 2021-05-28
Use getGen3Category to override move category for gen 3 and below

Karthik99999:
~~I probably shouldn't have removed this in the first place~~
It turns out new gen moves make it on to the `overrideMoveData` for gen 3 because of me not using `getGen3Category`, which probably shouldn't be happening.


Zarel:
Thanks!

PR #1819, opened 2021-05-28
Tooltips: Improve accuracy

DaWoblefet:
This fixes a number of display issues with accuracy:

- Accuracy values are more precise in their display, while keeping relative values for multipliers for regular users to understand. For example, the difference between Victory Star Fire Blast and Wide Lens Fire Blast:
image
image

- Related to the above, accuracy values are now properly rounded. For example, Compound Eyes Sleep Powder:
image

Note that this chaining process is not _technically 100% correct_ because of the potential of multiple Victory Star. The order of Abilities being chained would depend on raw Speed (like LightningRod). However, in order to get it 100% correct, you'd have to start tracking host stuff, when a Pokemon last acquired an ability relative to other mons, etc. We don't even have a way to sort by Speed on client afaik, and I don't feel like doing the extra work for such miniscule gains. If someone uses Role Play on their Victini in doubles, they may have to deal with rare cases of the accuracy tooltip being 1% off (it may not even matter mathematically speaking, but I have not crunched the numbers).

One point of discussion might be whether or not we want to cap accuracy calculations at 100% or not. Currently, we do not cap accuracy calculations at 100%, presumably because we don't account for the opponent at all in our calculations. I left this as-is.

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -1484,12 +1484,12 @@ class BattleTooltips {
 			accuracyModifiers.push(5325);
 			value.abilityModify(1.3, "Compound Eyes");
 		}
-		
+
 		if (value.tryItem('Wide Lens')) {
 			accuracyModifiers.push(4505);
 			value.itemModify(1.1, "Wide Lens");
 		}
-		
+
 		// Chaining modifiers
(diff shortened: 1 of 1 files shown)


Zarel:
ty!

PR #1820, opened 2021-05-29
Fix AltFormPopup and setForm

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8871200

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -3310,7 +3310,7 @@
 			this.room = data.room;
 			this.curSet = data.curSet;
 			this.chartIndex = data.index;
-			var species = this.curTeam.dex.species.get(this.curSet.species);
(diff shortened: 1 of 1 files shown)


Zarel:
Thanks!

Zarel:
I wonder which change caused this problem. I know it worked in the past.

Karthik99999:
it was during my Species/Move override refactor :P. I made `curTeam.dex` a thing cause the old overrides were being used in there, so this just made it easier to not have to use overrides in `client-teambuilder.js` + it's much more efficient than doing `Dex.forGen` every time. I just forgot to add the `room` property for `AltFormPopup` and `setForm` when I implemented it.

Zarel:
Another bug that wouldn't have happened in TypeScript...

PR #1821, opened 2021-05-31
Eval enables /showdebug

Slayer95:
See https://github.com/smogon/pokemon-showdown/pull/8346

### DIFF
--- a/src/battle-log.ts
+++ b/src/battle-log.ts
@@ -557,6 +557,18 @@ class BattleLog {
 		return undefined;
 	}
 
+	static applyPrefs(options: any) {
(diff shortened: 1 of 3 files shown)


src/battle-log.ts
```diff
@@ -557,6 +557,18 @@ class BattleLog {
 		return undefined;
 	}
 
+	static applyPrefs(options: object) {
+		if (options.showDebug !== undefined) {
```
Slayer95, line 561: > error TS2339: Property 'showDebug' does not exist on type 'object'.
Pain in the neck :)

PR #1822, opened 2021-06-04
Update sidebar for all forme changes

Karthik99999:
Currently it runs only when the forme change is permanent, which means changes like aegislash or minior don't get updated icons when they should unless you skip the turn.


Zarel:
Thanks!

PR #1823, opened 2021-06-05
Show correct typechange content for Pokemon with 3+ types

WeWuzNidokangz:
(First client pull request) I agree to make this code available under the MIT license.

This behaviour is exposed in Bonus Type. Currently, if a Pokemon's bonus type is its second type, it is correctly displayed by the typechange in the main battle UI. However, if the bonus type is its third type, it is not displayed except in the tooltip, and if a Pokemon increases from having two to three typing due to an in-battle forme change, the bonus type will disappear. With this change, all of the types are displayed consistently.

### DIFF
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -2707,7 +2707,7 @@ class PokemonSprite extends Sprite {
 			status += '<span class="frz">FRZ</span> ';
 		}
 		if (pokemon.volatiles.typechange && pokemon.volatiles.typechange[1]) {
-			let types = pokemon.volatiles.typechange[1].split('/');
+			const types = pokemon.volatiles.typechange[1].split('/');
 			for (const type of types) {
 				status += '<img src="' + Dex.resourcePrefix + 'sprites/types/' + encodeURIComponent(type) + '.png" alt="' + type + '" class="pixelated" /> ';
 			}
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -2708,8 +2708,7 @@ class PokemonSprite extends Sprite {
 		}
 		if (pokemon.volatiles.typechange && pokemon.volatiles.typechange[1]) {
 			let types = pokemon.volatiles.typechange[1].split('/');
-			for (let i in types) {
-				const type = types[i];
+			for (const type of types) {
 				status += '<img src="' + Dex.resourcePrefix + 'sprites/types/' + encodeURIComponent(type) + '.png" alt="' + type + '" class="pixelated" /> ';
 			}
 		}
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -2708,9 +2708,9 @@ class PokemonSprite extends Sprite {
 		}
 		if (pokemon.volatiles.typechange && pokemon.volatiles.typechange[1]) {
 			let types = pokemon.volatiles.typechange[1].split('/');
-			status += '<img src="' + Dex.resourcePrefix + 'sprites/types/' + encodeURIComponent(types[0]) + '.png" alt="' + types[0] + '" class="pixelated" /> ';
-			if (types[1]) {
-				status += '<img src="' + Dex.resourcePrefix + 'sprites/types/' + encodeURIComponent(types[1]) + '.png" alt="' + types[1] + '" class="pixelated" /> ';
+			for (let i in types) {
+				const type = types[i];
+				status += '<img src="' + Dex.resourcePrefix + 'sprites/types/' + encodeURIComponent(type) + '.png" alt="' + type + '" class="pixelated" /> ';
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/battle-animations.ts
```diff
@@ -2708,9 +2708,9 @@ class PokemonSprite extends Sprite {
 		}
 		if (pokemon.volatiles.typechange && pokemon.volatiles.typechange[1]) {
 			let types = pokemon.volatiles.typechange[1].split('/');
-			status += '<img src="' + Dex.resourcePrefix + 'sprites/types/' + encodeURIComponent(types[0]) + '.png" alt="' + types[0] + '" class="pixelated" /> ';
-			if (types[1]) {
-				status += '<img src="' + Dex.resourcePrefix + 'sprites/types/' + encodeURIComponent(types[1]) + '.png" alt="' + types[1] + '" class="pixelated" /> ';
+			for (let i in types) {
+				const type = types[i];
```
Zarel, line 2712: I'm surprised eslint isn't complaining about this. Oh, we're using an old tslint in the client, maybe that's why. Anyway, using `for...in` on arrays is considered a JS anti-pattern _and_ we're doing polyfills in a way that break it.

```suggestion
			for (const type of types) {
```

Zarel:
Oh the validator isn't even running yet.

Anyway, PS doesn't support 3 types in the `types` array. Usually the third type must be considered an "added" type like `typeadd` rather than `typechange`; relevant for interactions with Trick-or-Treat and Forest's Curse.

Zarel:
Looking at your thread: https://www.smogon.com/forums/threads/bonus-type-om-of-the-month.3683173/

> Q: How do Forest's Curse and Trick Or Treat work?
> A: They replace the Pokémon's third type, or act as the third type if the Pokemon only has 2 types.

You should be implementing this as an added type, by stuffing 3 types into an engine made for 2 types.

src/battle-animations.ts
WeWuzNidokangz: Oh...I wondered why that felt so convoluted...

WeWuzNidokangz:
> Looking at your thread: https://www.smogon.com/forums/threads/bonus-type-om-of-the-month.3683173/
> 
> > Q: How do Forest's Curse and Trick Or Treat work?
> > A: They replace the Pokémon's third type, or act as the third type if the Pokemon only has 2 types.
> 
> You should be implementing this as an added type, not by stuffing 3 types into an engine made for 2 types.

(I should say that I'm not on the council for Bonus Type, nor did I write the server implementation for Bonus Type... This PR was based on the assumption that the server implementation was best practice due already being accepted. Coincidentally I did do something like that for another OM but it felt janky to me at the time and I only did it to force it to be displayed without having to try create a custom client https://github.com/WeWuzNidokangz/TrashChannel-Limited/commit/679da8be9e480e46ba4d58a8511626811187a128#diff-c7b144af9efea6f0eb31f259dc22de01d5f5bffa03a19ecaa1b232a8e2584dd3)

I am a bit surprised that the server implementation of BT is "wrong"... The server simulation _seems_ to support an arbitrary number of types (damage calculations, etc are correct in BT under typechange) and so does the tooltip in the client. The battle UI was the only part of the battle system I could notice that doesn't already support 3+ types, so I thought it was the odd one out...

Zarel:
The simulator doesn't currently break for arbitrary numbers of types, but they're "unsupported" in that we make no guarantees that it won't break in the future. I'm also referring to the thread, where Forest's Curse and Trick-or-Treat are described as a "third type", but they're a special added type with different mechanics, not a regular type in the type list.

WeWuzNidokangz:
I see... Should I just close this PR (i.e. this approach is unacceptable under any circumstances since it might randomly stop working in future) or is it acceptable to discuss the current situation in the thread and offer this approach as a possibility (in case the council wasn't aware it was possible for the bonus type to behave consistently with type-adding moves)?

Zarel:
I think this PR should be merged. The code is better regardless. How you want to handle the bonus type varies, but the simplest fix is to edit the thread to say that Forest's Curse and Trick-or-Treat don't actually overwrite the third type.

WeWuzNidokangz:
(Asking the Bonus Type council's opinion here: https://www.smogon.com/forums/threads/bonus-type-om-of-the-month.3683173/page-4#post-8878303 )

src/battle-animations.ts
```diff
@@ -2708,9 +2708,8 @@ class PokemonSprite extends Sprite {
 		}
 		if (pokemon.volatiles.typechange && pokemon.volatiles.typechange[1]) {
 			let types = pokemon.volatiles.typechange[1].split('/');
```
Zarel, line 2710:
```suggestion
			const types = pokemon.volatiles.typechange[1].split('/');
```

Zarel:
ty!

PR #1824, opened 2021-06-06
Update password reset links

DaWoblefet:
Also cleans up a bit of grammar and updates a couple old links in htaccess. See also https://github.com/smogon/pokemon-showdown/pull/8358


AnnikaCodes:
I'm fairly sure XenForo just cares about the numbers, not the thread URL, but I guess it doesn't hurt.

PR #1825, opened 2021-06-12
Gen 5: Reveal Mail in team preview

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-58#post-8882546

Server Side: smogon/pokemon-showdown#8363

Mail graphic provided by Instruct


Zarel:
Is this just a drawn mail icon?

Zarel:
It looks pretty close but not exactly the same; in particular, it looks too tall...

src/battle.ts
```diff
@@ -3375,7 +3375,9 @@ class Battle {
 		}
 		case 'poke': {
 			let pokemon = this.rememberTeamPreviewPokemon(args[1], args[2])!;
-			if (args[3] === 'item') {
+			if (args[3] === 'mail') {
+				pokemon.item = 'Mail';
```
Zarel, line 3379: Is there only one Mail item in Gen 5? If not, this should probably be `(mail)`.

Karthik99999:
> Is this just a drawn mail icon?

Yep. Couldn't find the one that's in the game anywhere online, so this is the next best thing.

src/battle.ts
Karthik99999: Currently there is only one Mail item, but I guess it would be better to use `(mail)` for future proofing in case there's ever more mail types added.

Zarel: There are not going to be more mail types added to Gen 5. If Gen 5 has only one Mail item, it's fine to call it Mail.

Zarel:
The drawing is good enough; I won't block the PR on the drawing. We can replace it later.

I will need a separate copyright dedication for the mail drawing, though. (Art is just more of a legal issue than code when it comes to copyright.)

src/battle.ts
Zarel: Oh, I mean in Gen 5, not in PS.

Zarel: How many Mail items are in Black 2 and White 2 the actual games?

Karthik99999: https://bulbapedia.bulbagarden.net/wiki/Mail

Each Generation has its own set of mail that aren't present in the next generation. Gen 5 has 12 different types of mail. The only real reason to add more mail would be for item clause, which the standard Gen 5 tiers don't have.

Zarel:
Thanks!

PR #1828, opened 2021-06-24
Tooltips: Check if user is grounded for Terrain Pulse

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8890803

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -1316,13 +1316,16 @@ class BattleTooltips {
 	 * Gets the proper current type for moves with a variable type.
 	 */
 	getMoveType(move: Move, value: ModifiableValue, forMaxMove?: boolean | Move): [TypeName, 'Physical' | 'Special' | 'Status'] {
+		const pokemon = value.pokemon!;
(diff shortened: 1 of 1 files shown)


src/battle-tooltips.ts
```diff
@@ -1316,13 +1316,16 @@ class BattleTooltips {
 	 * Gets the proper current type for moves with a variable type.
 	 */
 	getMoveType(move: Move, value: ModifiableValue, forMaxMove?: boolean | Move): [TypeName, 'Physical' | 'Special' | 'Status'] {
+		const pokemon = value.pokemon!;
```
Zarel, line 1319:
```suggestion
		const pokemon = value.pokemon;
```

Pretty sure this works without the `!`

Zarel (approved):
Everything else looks good.

src/battle-tooltips.ts
Karthik99999: yeah you're right. actually there's a check to see if the variable is null later so it's better to not have it.

Zarel:
Thanks!

PR #1829, opened 2021-06-26
Add support for Let's Go move data overrides

Karthik99999:


Zarel:
I think before we deploy to client, it might be worth renaming all mods to the new standard of having a gen prefix (so: `gen7letsgo`). Is that ok with you?

Zarel:
If you don't want to actually rename it server-side, I'd be fine with doing it only in the client. I just don't want any _more_ stuff on the old client.

Karthik99999:
@Zarel anything else I need to do?

PR #1830, opened 2021-07-02
Add an administration panel for side servers

mia-pi-git:
Features requested / approved by policy upperstaff.
This PR adds a few things.
- Makes leaders/admins on the loginserver into sysops on side servers (barring Main and Smogtours)
- Makes it so side servers can only have one owner.
- Adds a required email field for side servers.
- Adds an administration panel. This can do a few things. Namely, ban/unban hosts from being used on a side server, ban/unban users from being owner on side servers, filter servers by different criteria, and reading modlogs for servers (editing, deletion, banning, addition, etc.)

### DIFF
--- a/lib/ntbb-servermodlog.sql
+++ b/lib/ntbb-servermodlog.sql
@@ -16,7 +16,6 @@ CREATE TABLE `ntbb_servermodlog` (
   KEY `ip` (`ip`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
 
-CREATE INDEX type_idx ON `ntbb_servermodlog` (`serverid`);
 CREATE INDEX server_idx ON `ntbb_servermodlog` (`serverid`, `type`);
 CREATE INDEX user_idx ON `ntbb_servermodlog` (`serverid`, `actorid`);
 CREATE INDEX mixed_search ON `ntbb_servermodlog` (`serverid`, `actorid`, `type`);
--- a/lib/ntbb-servermodlog.sql
+++ b/lib/ntbb-servermodlog.sql
@@ -1,0 +1,22 @@
+--
+-- Table structure for table `ntbb_servermodlog`
+-- Effectively taken from ntbb_usermodlog, as they serve a similar function.
+-- We just happen to use a `type` column to make it easier to narrow searches.
+
+CREATE TABLE `ntbb_servermodlog` (
+  `entryid` int(11) NOT NULL AUTO_INCREMENT,
+  `serverid` varchar(63) CHARACTER SET latin1 NOT NULL,
+  `actorid` varchar(63) CHARACTER SET latin1 NOT NULL,
+  `date` int(11) NOT NULL,
+  `ip` varchar(63) CHARACTER SET latin1 NOT NULL,
+  `type` varchar(63) CHARACTER SET latin1 NOT NULL,
+  `note` varchar(1023) CHARACTER SET latin1 NOT NULL,
+  PRIMARY KEY (`entryid`),
+  KEY `actorid` (`actorid`),
+  KEY `ip` (`ip`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE INDEX type_idx ON `ntbb_servermodlog` (`serverid`);
+CREATE INDEX server_idx ON `ntbb_servermodlog` (`serverid`, `type`);
+CREATE INDEX user_idx ON `ntbb_servermodlog` (`serverid`, `actorid`);
+CREATE INDEX mixed_search ON `ntbb_servermodlog` (`serverid`, `actorid`, `type`);
--- a/lib/ntbb-servermodlog.sql
+++ b/lib/ntbb-servermodlog.sql
@@ -17,5 +17,4 @@ CREATE TABLE `ntbb_servermodlog` (
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
 
 CREATE INDEX server_idx ON `ntbb_servermodlog` (`serverid`, `type`);
-CREATE INDEX user_idx ON `ntbb_servermodlog` (`serverid`, `actorid`);
 CREATE INDEX mixed_search ON `ntbb_servermodlog` (`serverid`, `actorid`, `type`);
--- a/lib/ntbb-session.lib.php
+++ b/lib/ntbb-session.lib.php
@@ -471,7 +471,12 @@ class NTBBSession {
 		if (($user['userid'] === $userid) && !empty($user['loggedin'])) {
 			// already logged in
 			$usertype = '2';
-			if (in_array($userid, $psconfig['sysops'], true)) {
+			$ismain = $serverhostname === 'sim3.psim.us' || $serverhostname === 'smogtours.psim.us';
+			if (
+				in_array($userid, $psconfig['sysops'], true) ||
+				// give admins and leaders sysop on side servers
+				(!$ismain && $users->isLeader())
+			) {
 				$usertype = '3';
 			} else {
 				// check autoconfirmed
--- a/website/servers/index.php
+++ b/website/servers/index.php
@@ -46,7 +46,7 @@ if (@$_POST['act'] === 'addserver') {
 		die("invalid server location");
 	}
 	$email = trim($_POST['email'] ?? '');
-	if (!$email || !strpos($email, '@')) {
+	if (!$email || strpos($email, '@') === false) {
 		die('Invalid email.');
 	}
 	$PokemonServers[$id] = [
--- a/website/servers/index.php
+++ b/website/servers/index.php
@@ -1,7 +1,6 @@
 <?php
 
 include_once 'servers.lib.php';
-include_once __DIR__ . '/../../lib/ntbb-database.lib.php';
 include '../style/wrapper.inc.php';
 
 $timenow = time();
--- a/website/servers/index.php
+++ b/website/servers/index.php
@@ -134,7 +134,7 @@ if ($users->isLeader()) {
 				</div>
 				<div class="buttonrow">
 					<button type="submit"><strong>Add server</strong></button>
-				</div></div>
+				</div>
 			</form>
 <?php
 }
--- a/website/servers/index.php
+++ b/website/servers/index.php
@@ -45,7 +45,7 @@ if (@$_POST['act'] === 'addserver') {
 	if (strpos($server, '.') === false) {
 		die("invalid server location");
 	}
-	$email = isset($_POST['email']) ? trim($_POST['email']) : false;
+	$email = trim($_POST['email'] ?? '');
 	if (!$email || !strpos($email, '@')) {
 		die('Invalid email.');
 	}
--- a/website/servers/index.php
+++ b/website/servers/index.php
@@ -1,6 +1,7 @@
 <?php
 
 include_once 'servers.lib.php';
+include_once __DIR__ . '/../../lib/ntbb-database.lib.php';
 include '../style/wrapper.inc.php';
 
 $timenow = time();
@@ -45,13 +46,25 @@ if (@$_POST['act'] === 'addserver') {
 	if (strpos($server, '.') === false) {
 		die("invalid server location");
 	}
+	$email = isset($_POST['email']) ? trim($_POST['email']) : false;
+	if (!$email || !strpos($email, '@')) {
+		die('Invalid email.');
+	}
 	$PokemonServers[$id] = [
 		'name' => $name,
 		'id' => $id,
 		'server' => $server,
-		'port' => $port
+		'port' => $port,
+		'email' => $email
 	];
 	if ($owner) $PokemonServers[$id]['owner'] = $owner;
+	$logMessage = "{$name} @ host '{$server}' (contact email: {$email}";
+	if ($owner) $logMessage .= ", owner: {$owner}";
+	$logMessage .= ")";
+	$psdb->query(
+		"INSERT INTO {$psdb->prefix}servermodlog (`serverid`, `actorid`, `date`, `ip`, `type`, `note`) VALUES (?, ?, ?, ?, ?, ?)",
+		[$id, $curuser['id'], time(), $users->getIp(), 'CREATESERVER', $logMessage]
+	);
 	saveservers();
 }
 
@@ -117,9 +130,11 @@ if ($users->isLeader()) {
 				<div class="formrow">
 					<label class="label">Owner's username: <input class="textbox" type="text" name="owner" placeholder="(Optional)" /><em>(separate multiple owners by commas)</em></label>
 				</div>
+				<div class="formrow">
+					<label class="label">Owner's email: <input class="textbox" type="text" name="email" /></label>
 				<div class="buttonrow">
 					<button type="submit"><strong>Add server</strong></button>
-				</div>
+				</div></div>
 			</form>
 <?php
 }
--- a/website/servers/index.php
+++ b/website/servers/index.php
@@ -131,6 +131,7 @@ if ($users->isLeader()) {
 				</div>
 				<div class="formrow">
 					<label class="label">Owner's email: <input class="textbox" type="text" name="email" /></label>
+				</div>
 				<div class="buttonrow">
 					<button type="submit"><strong>Add server</strong></button>
 				</div></div>
--- a/website/servers/manage.php
+++ b/website/servers/manage.php
@@ -159,11 +159,11 @@ if (isset($_REQUEST['unbanhosts']) && mb_strlen($_REQUEST['unbanhosts'])) {
 	writeserverinfo();
 	$scount = count($successes);
 	$fcount = count($failures);
-	echo('Removed ' . $scount . " host (s) from the blacklist. (" . implode(', ', $successes) . ")");
+	echo 'Removed ' . $scount . ' host(s) from the blacklist. (' . implode(', ', $successes) . ')<br />';
 	if ($fcount > 0) {
-		echo("Failed to remove " . $fcount . " host (s) from the blacklist. (" . implode(', ', $failures) . ")");
+		echo 'Failed to remove ' . $fcount . ' host(s) from the blacklist. (' . implode(', ', $failures) . ')<br />';
 	}
-	echo('</div>');
+	echo '</div>';
 }
 
 ?>
--- a/website/servers/manage.php
+++ b/website/servers/manage.php
@@ -57,7 +57,7 @@ if (isset($_REQUEST['banowners'])) {
 			$serverinfo['blacklist'] = [];
 		}
 		$failures = [];
-		if (!in_array($userid, $serverinfo['blacklist'])) {
+		if (!in_array($userid, $serverinfo['blacklist'], true)) {
 			$serverinfo['blacklist'][] = $userid;
 			$successes += 1;
 			$psdb->query(
@@ -81,11 +81,11 @@ if (isset($_REQUEST['banowners'])) {
 	}
 	echo "<strong>{$successes} user (s) added to blacklist.</strong><br />";
 	if (count($failures) > 0) {
-		echo ("(" . count($failures) . "user (s) could not be added: " . implode(", ", $failures) . ")<br />");
+		echo "(" . count($failures) . "user (s) could not be added: " . implode(", ", $failures) . ")<br />";
 	}
 }
 
-echo('<br />');
+echo '<br />';
 if (isset($serverinfo['blacklist']) && count($serverinfo['blacklist'])) {
 ?>
 	<hr />
@@ -173,13 +173,10 @@ if (isset($_REQUEST['unbanhosts']) && mb_strlen($_REQUEST['unbanhosts'])) {
 		<input type="text" name="unbanhosts" class="textbox" /><br /><br />
 		<button type="submit">Remove hosts from list</button>
 	</form>
+</td></tr>
+</table></div></div>
+<center>
 <?php
-
-echo('</td></tr>');
-
-echo('</table></div></div>');
-
-echo '<center>';
 
 if (isset($_REQUEST['ownedby']) || isset($_REQUEST['hostedby']) || isset($_REQUEST['lastactive'])) {
 	$owner = isset($_REQUEST['ownedby']) && mb_strlen($_REQUEST['ownedby']) ? $users->userid($_REQUEST['ownedby']) : NULL;
--- a/website/servers/manage.php
+++ b/website/servers/manage.php
@@ -52,7 +52,7 @@ if (isset($_REQUEST['banowners'])) {
 	$successes = 0;
 	foreach ($banned as $user) {
 		$userid = $users->userid($user);
-		if (!mb_strlen($userid)) continue;
+		if (!strlen($userid)) continue;
 		if (!isset($serverinfo['blacklist'])) {
 			$serverinfo['blacklist'] = [];
 		}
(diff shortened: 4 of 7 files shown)


mia-pi-git:
Whoops.

lib/ntbb-session.lib.php
```diff
@@ -475,7 +475,13 @@ function getAssertion($userid, $serverhostname, $user = null, $challengekeyid =
 		if (($user['userid'] === $userid) && !empty($user['loggedin'])) {
 			// already logged in
 			$usertype = '2';
-			if (in_array($userid, $psconfig['sysops'], true)) {
+			$isupperstaff = $curuser['group'] === 6 || $curuser['group'] === 2;
```
Zarel, line 478: probably these should all be stuck in `$curuser` instead of having a bunch more globals.

Zarel: Or maybe a separate `$curauth`.

Honestly I would design this so differently if I was writing this today...

website/serveradmin.php
```diff
@@ -0,0 +1,378 @@
+<?php
```
Zarel, line 1: I think there's no reason why this can't be a part of `servers/[servername]`

Plausibly we should have a `website/servers/index.php` and `website/servers/view.php` if we wanted to split things up that way.

Zarel:
OK, I refactored the file locations. You should be able to put your administration panel in `servers/server.php` now.

lib/ntbb-session.lib.php
```diff
@@ -475,7 +475,13 @@ function getAssertion($userid, $serverhostname, $user = null, $challengekeyid =
...
+			$ismain = $serverhostname === 'sim3.psim.us' || $serverhostname === 'smogtours.psim.us';
+			if (
+				in_array($userid, $psconfig['sysops'], true) ||
+				// give admins and leaders sysop on side servers
```
Zarel, line 482: no

Zarel: I really really don't like it when people "sneak" major policy changes in like this. Please actually say if you do something like this in the pull request description.

I think I would be amenable to this if it was discussed first.

mia-pi-git: Oh, sorry. I was told this was good to go. 
CC: Aeonic

website/serveradmin.php
mia-pi-git: I feel like this might be better suited to a separate page, still? Say, /servers/manage.php?
Having it integrated into the other pages seems like it would add clutter.

Zarel: You always seem to have a very different definition of clutter than I do. Putting the interface for editing a side server on the page for editing a side server makes more sense than making you dig for it.

mia-pi-git: I mean, this isn't for editing one side server so much as for general administration tasks (banning users from owning, reading server modlogs, banning hosts from being used.) To me, that seems better suited to another page.

lib/ntbb-session.lib.php
Zarel: OK, after upper staff discussion, I'm fine with this. The main thing I dislike is that this is nowhere in the pull request description, like you were trying to hide it from me.

It might be useful to compare to what it looks like when I make a pull request:

https://github.com/smogon/pokemon-showdown/pull/8267

I don't need _exactly_ this much detail, but you should at least describe any change with privacy or permissions implications, as well as the specific features being added (this would also prevent my confusion over why this isn't in `servers/server.php`).

website/serveradmin.php
Zarel: Oh, I see! Yeah, `servers/manage.php` seems like a good place.

lib/ntbb-session.lib.php
Zarel: Oh, I misunderstood this. Yeah, this is fine.

website/serveradmin.php
mia-pi-git: Alrighty then.

lib/ntbb-session.lib.php
mia-pi-git: Sorry; this change skipped my mind when I was making the PR description. I'll keep it more detailed in the future.

mia-pi-git:
Whoops.

mia-pi-git:
OK, updated to new /website/servers handling.

Zarel:
You still don't have a PR description.

mia-pi-git:
> You still don't have a PR description.

OK, done as requested. 

> This PR adds a few things.
> 
> * Makes leaders/admins on the loginserver into sysops on side servers (barring Main and Smogtours)
> * Makes it so side servers can only have one owner.
> * Adds a required email field for side servers.
> * Adds an administration panel. This can do a few things. Namely, ban/unban hosts from being used on a side server, ban/unban users from being owner on side servers, filter servers by different criteria, and reading modlogs for servers (editing, deletion, banning, addition, etc.)

mia-pi-git:
Bump?

lib/ntbb-servermodlog.sql
```diff
@@ -0,0 +1,22 @@
...
+  `ip` varchar(63) CHARACTER SET latin1 NOT NULL,
+  `type` varchar(63) CHARACTER SET latin1 NOT NULL,
+  `note` varchar(1023) CHARACTER SET latin1 NOT NULL,
+  PRIMARY KEY (`entryid`),
+  KEY `actorid` (`actorid`),
+  KEY `ip` (`ip`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE INDEX type_idx ON `ntbb_servermodlog` (`serverid`);
+CREATE INDEX server_idx ON `ntbb_servermodlog` (`serverid`, `type`);
```
Zarel, line 20:
```suggestion
CREATE INDEX server_idx ON `ntbb_servermodlog` (`serverid`, `type`);
```

These two are redundant.

lib/ntbb-session.lib.php
```diff
@@ -471,7 +471,12 @@ function getAssertion(string $userid, string $serverhostname, array $user = null
 		if (($user['userid'] === $userid) && !empty($user['loggedin'])) {
 			// already logged in
 			$usertype = '2';
-			if (in_array($userid, $psconfig['sysops'], true)) {
+			$ismain = $serverhostname === 'sim3.psim.us' || $serverhostname === 'smogtours.psim.us';
```
Zarel, line 474: This doesn't seem secure. Wouldn't this let you get sysop on Main by connecting through the IP address?

lib/ntbb-servermodlog.sql
```diff
@@ -0,0 +1,21 @@
...
+  `type` varchar(63) CHARACTER SET latin1 NOT NULL,
+  `note` varchar(1023) CHARACTER SET latin1 NOT NULL,
+  PRIMARY KEY (`entryid`),
+  KEY `actorid` (`actorid`),
+  KEY `ip` (`ip`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE INDEX server_idx ON `ntbb_servermodlog` (`serverid`, `type`);
+CREATE INDEX user_idx ON `ntbb_servermodlog` (`serverid`, `actorid`);
+CREATE INDEX mixed_search ON `ntbb_servermodlog` (`serverid`, `actorid`, `type`);
```
Zarel, line 21:
```suggestion
CREATE INDEX mixed_search ON `ntbb_servermodlog` (`serverid`, `actorid`, `type`);
```

These two are also redundant.

Remember, every `A, B, C` index also includes an `A, B` index.

website/servers/index.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 include_once 'servers.lib.php';
+include_once __DIR__ . '/../../lib/ntbb-database.lib.php';
```
Zarel, line 4: What is this for? `servers` already includes `session` which includes `database`.

Zarel:
```suggestion
```

lib/ntbb-session.lib.php
mia-pi-git: Oh. Yes, if one is already a leader. Is there a good way around this?

website/servers/index.php
```diff
@@ -45,13 +46,25 @@
 	if (strpos($server, '.') === false) {
 		die("invalid server location");
 	}
+	$email = isset($_POST['email']) ? trim($_POST['email']) : false;
```
Zarel, line 48:
```suggestion
	$email = trim($_POST['email'] ?? '');
```

website/servers/index.php
```diff
@@ -45,13 +46,25 @@
...
+	if (!$email || !strpos($email, '@')) {
```
Zarel, line 49:
```suggestion
	if (!$email || strpos($email, '@') === false) {
```

I understand that yours also handles the `0` case, but writing it this way is much more readable.

website/servers/server.php
```diff
@@ -82,13 +82,77 @@
...
-	foreach ($owners as $i=>$owner) $owners[$i] = $users->userid($owner);
-	$PokemonServers[$id]['owner'] = implode(',',$owners);
+		$cannotUpdate = false;
+		if (isset($serverinfo['bannedhosts']) && trim($server) !== trim($oldServer)) {
+			foreach ($serverinfo['bannedhosts'] as $bannedhost) {
+				$bannedhost = trim($bannedhost);
+				if ($bannedhost && strpos($server, $bannedhost) > -1) {
+					$cannotUpdate = true;
+					echo(
+						"<br /><strong style=\"color:red\">You cannot use the host " . $server .
```
Zarel, line 95:
```suggestion
						"<br /><strong style=\"color:red\">You cannot use the host " . htmlspecialchars ($server) .
```

website/servers/server.php
```diff
@@ -82,13 +82,77 @@
...
+		$oldServer = $PokemonServers[$id]['server'];
+		$oldPort = $PokemonServers[$id]['port'];
 
-	$owners = explode(',', $_POST['owner']);
-	foreach ($owners as $i=>$owner) $owners[$i] = $users->userid($owner);
-	$PokemonServers[$id]['owner'] = implode(',',$owners);
+		$cannotUpdate = false;
+		if (isset($serverinfo['bannedhosts']) && trim($server) !== trim($oldServer)) {
+			foreach ($serverinfo['bannedhosts'] as $bannedhost) {
+				$bannedhost = trim($bannedhost);
```
Zarel, line 91: Why isn't `$serverinfo['bannedhosts']` pre-trimmed?

website/servers/server.php
```diff
@@ -185,8 +260,10 @@
 				<h1>Edit server</h1>
 				<div class="formrow">
 					<p><label class="label">Location: <input class="textbox" type="text" name="loc" placeholder="http://<?php echo $entry['server']; if ($entry['port'] != 8000) echo '-',$entry['port']; ?>.psim.us/" size="60" /><em>(you can leave off http:// etc if you want)</em></label></p>
-					<p><label class="label">Owner: <input class="textbox" type="text" name="owner" value="<?php echo $entry['owner'] ?? '(none)'; ?>" size="60" /><em>(separate multiple owners by commas)</em></label></p>
-					<label class="label">Configuration:</label><?php
+					<p><label class="label">Owner: <strong><?php echo $entry['owner'] ?? '(none)'; ?></strong></label></p>
```
Zarel, line 263: Why isn't the owner editable anymore? PS admins like me should be able to edit it without having to SSH in...

mia-pi-git: Uh, it definitely still is. I think the button just got removed somehow.

website/servers/servers.lib.php
```diff
@@ -15,3 +15,19 @@ function saveservers() {
 $PokemonServers = '.var_export($GLOBALS['PokemonServers'], true).';
 ');
 }
+
+
+$serverinfo = json_decode(
+	file_exists(__DIR__ . '/../../config/serverinfo.json') ?
+		file_get_contents(__DIR__ . '/../../config/serverinfo.json') :
+		"{}",
```
Zarel, line 23:
```suggestion
	file_get_contents(__DIR__ . '/../../config/serverinfo.json') ?: "{}",
```

In general, any `exists` function is a race condition waiting to happen, and should not be used.

lib/ntbb-session.lib.php
Zarel: Pass a server ID here?

website/servers/server.php
mia-pi-git: It is, this was just a force-of-habit sanity check.

website/servers/server.php
mia-pi-git: Fixed.

website/servers/manage.php
```diff
@@ -0,0 +1,349 @@
+<?php
+
+include_once __DIR__ . '/../../config/config.inc.php';
+include_once __DIR__ . '/../../config/servers.inc.php';
+include_once '../../lib/ntbb-session.lib.php';
+include '../style/wrapper.inc.php';
+include_once __DIR__ . '/../../lib/ntbb-database.lib.php';
+include_once __DIR__ . '/servers.lib.php';
```
Zarel, line 8:
```suggestion
include_once './servers.lib.php';
```

All of these are already dependencies of `servers.lib.php`.

website/servers/index.php
```diff
@@ -117,9 +129,11 @@ function cmpByUsercount($b, $a) {
 				<div class="formrow">
 					<label class="label">Owner's username: <input class="textbox" type="text" name="owner" placeholder="(Optional)" /><em>(separate multiple owners by commas)</em></label>
 				</div>
+				<div class="formrow">
+					<label class="label">Owner's email: <input class="textbox" type="text" name="email" /></label>
 				<div class="buttonrow">
 					<button type="submit"><strong>Add server</strong></button>
-				</div>
+				</div></div>
```
Zarel, line 136:
```suggestion
				</div>
```

Why are you adding the `</div>` here? The indentation is wrong for this, anyway.

website/servers/index.php
```diff
@@ -117,9 +129,11 @@ function cmpByUsercount($b, $a) {
 				<div class="formrow">
 					<label class="label">Owner's username: <input class="textbox" type="text" name="owner" placeholder="(Optional)" /><em>(separate multiple owners by commas)</em></label>
 				</div>
+				<div class="formrow">
+					<label class="label">Owner's email: <input class="textbox" type="text" name="email" /></label>
```
Zarel, line 133:
```suggestion
				<div class="formrow">
					<label class="label">Owner's email: <input class="textbox" type="text" name="email" /></label>
				</div>
```

website/servers/server.php
```diff
@@ -185,8 +260,16 @@
...
 				<div class="formrow">
 					<p><label class="label">Location: <input class="textbox" type="text" name="loc" placeholder="http://<?php echo $entry['server']; if ($entry['port'] != 8000) echo '-',$entry['port']; ?>.psim.us/" size="60" /><em>(you can leave off http:// etc if you want)</em></label></p>
-					<p><label class="label">Owner: <input class="textbox" type="text" name="owner" value="<?php echo $entry['owner'] ?? '(none)'; ?>" size="60" /><em>(separate multiple owners by commas)</em></label></p>
-					<label class="label">Configuration:</label><?php
+						<p><label class="label">Owner: <?php
+							if ($users->isLeader()) {
+								echo ('<input class="textbox" type="text" name="owner" value="'.$entry['owner'] ?? '(none)' .'" size="60">');
+							} else {
+								echo '<em>' . $entry['owner'] ?? '(none)' . "</em>";
+							}
```
Zarel, line 268:
```suggestion
					<p><label class="label">Owner: <?php
						if ($users->isLeader()) {
							echo ('<input class="textbox" type="text" name="owner" value="'.$entry['owner'] ?? '(none)' .'" size="60">');
						} else {
							echo '<em>' . $entry['owner'] ?? '(none)' . "</em>";
						}
```

website/servers/server.php
```diff
@@ -82,13 +82,77 @@
...
+		$oldPort = $PokemonServers[$id]['port'];
 
-	$owners = explode(',', $_POST['owner']);
-	foreach ($owners as $i=>$owner) $owners[$i] = $users->userid($owner);
-	$PokemonServers[$id]['owner'] = implode(',',$owners);
+		$cannotUpdate = false;
+		if (isset($serverinfo['bannedhosts']) && trim($server) !== trim($oldServer)) {
+			foreach ($serverinfo['bannedhosts'] as $bannedhost) {
+				$bannedhost = trim($bannedhost);
+				if ($bannedhost && strpos($server, $bannedhost) > -1) {
```
Zarel, line 92:
```suggestion
				if ($bannedhost && strpos($server, $bannedhost) !== false) {
```

lib/ntbb-session.lib.php
mia-pi-git: We discussed in DMs and determined this actually wouldn't be an issue due to other guards against using IPs.

website/servers/server.php
```diff
@@ -82,13 +82,77 @@
...
+					$cannotUpdate = true;
+					echo(
+						"<br /><strong style=\"color:red\">You cannot use the host " . htmlspecialchars ($server) .
+						" as it is from a blacklisted host.</strong><br />"
+					);
```
Zarel, line 97:
```suggestion
					echo "<br /><strong style=\"color:red\">You can't use the host " . htmlspecialchars($server) . " because it is from a blacklisted host.</strong><br />"
```

website/servers/server.php
```diff
@@ -82,13 +82,77 @@
...
+			}
+		}
+
+		foreach ($PokemonServers as $curserver) {
+			if ($curserver['server'] === $server) {
+				$cannotUpdate = true;
+				echo(
+					"<br /><strong style=\"color:red\">You cannot use the host " . $server .
+					" as it is already in use by another server.</strong><br />"
+				);
```
Zarel, line 109:
```suggestion
				echo "<br /><strong style=\"color:red\">You can't use the host " . htmlspecialchars($server) . " because it is already in use by another server.</strong><br />";
```

website/servers/server.php
```diff
@@ -185,8 +260,16 @@
 				<h1>Edit server</h1>
 				<div class="formrow">
 					<p><label class="label">Location: <input class="textbox" type="text" name="loc" placeholder="http://<?php echo $entry['server']; if ($entry['port'] != 8000) echo '-',$entry['port']; ?>.psim.us/" size="60" /><em>(you can leave off http:// etc if you want)</em></label></p>
-					<p><label class="label">Owner: <input class="textbox" type="text" name="owner" value="<?php echo $entry['owner'] ?? '(none)'; ?>" size="60" /><em>(separate multiple owners by commas)</em></label></p>
-					<label class="label">Configuration:</label><?php
+					<p><label class="label">Owner: <?php
+						if ($users->isLeader()) {
+							echo ('<input class="textbox" type="text" name="owner" value="'.$entry['owner'] ?? '(none)' .'" size="60">');
```
Zarel, line 265:
```suggestion
							echo ('<input class="textbox" type="text" name="owner" value="'.$entry['owner'] ?? '' .'" placeholder="(none)" size="60">');
```

website/servers/server.php
```diff
@@ -82,13 +82,77 @@
...
+		}
+
+		if (isset($_POST['owner']) && $is_manager) {
+			$userid = $users->userid($_POST['owner']);
+			if (strlen($userid) && $PokemonServers[$id]['owner'] !== $userid) {
+				if (isset($serverinfo['blacklist']) && in_array($userid, $serverinfo['blacklist'])) {
+					echo(
+						'<strong style="color:red">' . $userid .
+						" is blacklisted from owning a server.</strong><br />"
+					);
```
Zarel, line 141:
```suggestion
					echo '<strong style="color:red">' . $userid . ' is blacklisted from owning a server.</strong><br />';
```

website/servers/manage.php
```diff
@@ -0,0 +1,344 @@
...
+	</form>
+<?php
+
+if (isset($_REQUEST['banowners'])) {
+	echo('<br />');
+	$banned = explode(',', $_REQUEST['banowners']);
+	$successes = 0;
+	foreach ($banned as $user) {
+		$userid = $users->userid($user);
+		if (!mb_strlen($userid)) continue;
```
Zarel, line 55: Why do you use `mb_strlen` so much? Just use `strlen`.

website/servers/manage.php
```diff
@@ -0,0 +1,344 @@
...
+		if (!isset($serverinfo['blacklist'])) {
+			$serverinfo['blacklist'] = [];
+		}
+		$failures = [];
+		if (!in_array($userid, $serverinfo['blacklist'])) {
```
Zarel, line 60:
```suggestion
		if (!in_array($userid, $serverinfo['blacklist'], true)) {
```

`in_array` in loose mode, like `==`, is responsible for lots of bugs. Always use strict mode.

Zarel: This is even more of a problem in PHP, where e.g. `'1' == ' 1'` is `true`. In JavaScript, loose and strict comparisons work the same if both sides are the same type, but string comparisons in PHP are even looser.

website/servers/manage.php
```diff
@@ -0,0 +1,344 @@
...
+				);
+			}
+		} else {
+			$failures[] = $userid;
+		}
+		writeserverinfo();
+	}
+	echo "<strong>{$successes} user (s) added to blacklist.</strong><br />";
+	if (count($failures) > 0) {
+		echo ("(" . count($failures) . "user (s) could not be added: " . implode(", ", $failures) . ")<br />");
```
Zarel, line 84:
```suggestion
		echo "(" . count($failures) . "user (s) could not be added: " . implode(", ", $failures) . ")<br />";
```

`echo` isn't a function.

website/servers/manage.php
```diff
@@ -0,0 +1,344 @@
...
+	}
+}
+
+echo('<br />');
```
Zarel, line 88:
```suggestion
echo '<br />';
```

website/servers/manage.php
```diff
@@ -0,0 +1,344 @@
...
+		}
+	}
+	writeserverinfo();
+	$scount = count($successes);
+	$fcount = count($failures);
+	echo('Removed ' . $scount . " host (s) from the blacklist. (" . implode(', ', $successes) . ")");
+	if ($fcount > 0) {
+		echo("Failed to remove " . $fcount . " host (s) from the blacklist. (" . implode(', ', $failures) . ")");
+	}
+	echo('</div>');
```
Zarel, line 166: I also don't understand why you're randomly switching between `'` and `"`. Generally I tend to stick with `'` for HTML, because `"` is very common in HTML.

```suggestion
	echo 'Removed ' . $scount . ' host(s) from the blacklist. (' . implode(', ', $successes) . ')<br />';
	if ($fcount > 0) {
		echo 'Failed to remove ' . $fcount . ' host(s) from the blacklist. (' . implode(', ', $failures) . ')<br />';
	}
	echo '</div>';
```

website/servers/manage.php
```diff
@@ -0,0 +1,344 @@
...
+		<input type="text" name="unbanhosts" class="textbox" /><br /><br />
+		<button type="submit">Remove hosts from list</button>
+	</form>
+<?php
+
+echo('</td></tr>');
+
+echo('</table></div></div>');
+
+echo '<center>';
```
Zarel, line 182: This is so confusing...

```suggestion
</td></tr>
</table></div></div>
<center>
<?php
```

mia-pi-git:
Bump?

mia-pi-git:
Bump x2?

lib/ntbb-session.lib.php
```diff
@@ -471,7 +471,12 @@ function getAssertion(string $userid, string $serverhostname, array $user = null
...
+			if (
+				in_array($userid, $psconfig['sysops'], true) ||
+				// give admins and leaders sysop on side servers
```
AnnikaCodes, line 477: The [last Upper Staff discussion](https://discord.com/channels/534298043176976395/565617827298541572/942897233207128124) on sysop for side servers concluded that it wasn't worth it, since side servers could still disable it and Main leadership shouldn't be in charge of moderating side servers anyway

monsanto:
This seems useful but
- I don't like PHP
- I don't like reviewing changes to untyped code that I am unfamiliar with
Does this have any overlap with the loginserver? I think the loginserver reads some PHP file this generates? If we could move this functionality to TypeScript that would be useful.

PR #1831, opened 2021-07-03
Add Wonder Launcher as an alternative to the 'Duel Disk' avatars

TheSupervillain:
It's the real name of the machine they have, so why not?


Zarel:
I don't see a point in this specific alias.

The way I see it, we have two options:

- client-side, rename dueldisk to wonderlauncher, and then alias the old names to the new names (and then also do the rename server-side)
- server-side, add an alias from wonderlauncher to dueldisk (simpler approach)

This isn't either of those options, and I see no reason to prefer it to that approach.

Zarel:
OK, I already did the rename in the client avatar files, so this is good now. Thanks! We'll still need the rename on the server-side, though.

PR #1833, opened 2021-07-06
Fix Smogon Analysis links in teambuilder

Karthik99999:
- Make the link redirect to the current tier for Gen 8 formats (was previously only doing it for Gen 7 and below)
- Fixes bh/ag getting appended to the species name without a forward slash (check this [bug report](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8903859))
- Uses the proper link for National Dex AG (previously was just using `ag` when it should be `national-dex-ag`)


Zarel:
ty!

PR #1839, opened 2021-07-15
Set team name to pokepaste title when importing

Karthik99999:
https://www.smogon.com/forums/threads/import-pokepaste-titles-as-the-title-of-a-given-team.3671574/


Zarel:
ty!

PR #1840, opened 2021-07-15
Support custom width/height attributes in twitch tags

mia-pi-git:
More convenient and generally useful feature.


AnnikaCodes (approved):
Looks good to me.

Zarel:
Sure

PR #1842, opened 2021-07-18
Start modularizing some battle files

Zarel:
Progress is very incomplete, but this is mostly an experiment to establish this as possible at all.

What's going on here is that we're using `remove-import-export` so the `import` and `export` statements just disappear after compiling (everything's still a global).

This allows us to piecewise convert files to modules.

So we're just using TypeScript to keep track of dependencies, and also to make it easier to use these files in other projects later on.

I've tried to avoid circular dependencies, but there's one between `battle-animations` and `battle-animations-moves`.

### DIFF
--- a/src/battle-animations-moves.ts
+++ b/src/battle-animations-moves.ts
@@ -9,7 +9,9 @@
  * @license CC0-1.0
  */
 
-const BattleMoveAnims: AnimTable = {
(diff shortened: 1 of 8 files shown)


Zarel:
I'm PRing this so everyone can see that this is possible and maybe work on this for me.

PR #1845, opened 2021-07-23
Check for Neutralizing Gas in tooltips

Karthik99999:


DaWoblefet:
What does this do?

Karthik99999:
> What does this do? Is this the Skill Swap not displaying the correct Ability thing?

Mainly accounts for Neutralizing Gas. I noticed that No Guard would still show up on the move tooltip when Neutralizing Gas was up.

src/battle-tooltips.ts
```diff
@@ -62,6 +62,7 @@ class ModifiableValue {
 			this.comment.push(` (${abilityName} suppressed by Gastro Acid)`);
 			return false;
 		}
+		if (!this.pokemon?.effectiveAbility(this.serverPokemon)) return false;
```
Zarel, line 65:
```suggestion
		// Check for Neutralizing Gas
		if (!this.pokemon?.effectiveAbility(this.serverPokemon)) return false;
```

Zarel:
https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md

> CM1. What, not how
>
> Commits should describe what the code _does_, not how it does it.

;)

Karthik99999:
> https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md
> 
> 
> 
> > CM1. What, not how
> 
> >
> 
> > Commits should describe what the code _does_, not how it does it.
> 
> 
> 
> ;)

You would think after all my PRs I would understand this lol

Zarel:
Thanks!

PR #1846, opened 2021-07-29
Teambuilder: Fix bugs and implement new policy for STABmons

Karthik99999:
https://www.smogon.com/forums/threads/stabmons.3656429/page-8#post-8931262
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8930934

- Implements new STABmons policy of accounting for past gen move and species typings
- Fixes Mega formes not showing all the moves they should have access to
- Fixes 'Past' moves showing up as illegal in Gen 7 (the way I did this is by adding `isNonstandard` to the list of properties to override for moves, which is why I removed the `dex.gen === 8` checks when checking for "Past" moves since they are unnecessary now)


KrisXV:
@Zarel bump

PR #1847, opened 2021-08-02
Teambuilder: List Shore Up as a good status move

Karthik99999:
Other recover clones are considered good/not useless so no reason why this shouldn't be as well


Zarel:
Good catch!

PR #1850, opened 2021-08-07
Tooltips: Fix Light Ball and Transform-item interaction

KrisXV:


src/battle-tooltips.ts
```diff
@@ -1934,6 +1934,13 @@ class BattleTooltips {
 			return value;
 		}
 
+		// Light ball is a base power modifier in gen 4 only
+		if (item.name === 'Light Ball' && this.battle.gen !== 4) return value;
+		if (item.name === 'Light Ball' && Dex.species.get(value.serverPokemon.speciesForme).baseSpecies === 'Pikachu') {
```
Marty-D, line 1939: Transform copied species prior to Gen 5, so a Light Ball Ditto Transformed into Pikachu would get the boost, etc. Though I suppose this mechanic isn't currently considered anywhere in tooltips if the one below is anything to go by.

KrisXV:
@Marty-D

PR #1852, opened 2021-08-15
Support find-a-user button in Preact

mia-pi-git:
Draft as I'm not sure the `PromptPanel` design is what I want to go with yet. Code feedback is appreciated though.

### DIFF
--- a/src/panels.tsx
+++ b/src/panels.tsx
@@ -321,6 +321,16 @@ class PSMain extends preact.Component {
 	}
 	handleButtonClick(elem: HTMLButtonElement) {
 		switch (elem.name) {
+		case 'finduser':
+			PS.addRoom({
+				id: 'finduser' as RoomID,
+				type: 'prompt',
+				message: 'Find a user',
+				parentElem: elem,
+				buttonName: "Open",
+			});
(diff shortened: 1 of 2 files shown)


src/panels.tsx
```diff
@@ -321,6 +321,16 @@ class PSMain extends preact.Component {
...
+		case 'finduser':
+			PS.addRoom({
+				id: 'finduser' as RoomID,
+				type: 'prompt',
+				message: 'Find a user',
+				parentElem: elem,
+				buttonName: "Open",
+			});
+			PS.update();
+			return true;
```
Zarel, line 333: Seems like this should be a link to the `finduser` room, rather than needing to manually dispatch it?

Zarel: Like just use `joinRoom` rather than `finduser`.

mia-pi-git: OK.

mia-pi-git: On second consideration, would that work? It would remove access to being able to set the message / buttonName params, no?

monsanto:
Seems unlikely to happen.

PR #1844, opened 2021-07-22
Add a friend/unfriend option to the UserOptions popup

mia-pi-git:
Useful feature that's already been requested a lot.

PR #1853, opened 2021-08-23
Add a Friends button to the main menu

Laevos:
Currently, the only way to access the Friends list is by using a `/command`. This requires that the user be in a room. It would be more convenient for users who do not normally join any chat rooms to be able to pull up their friends list from the main menu. To that end, this change adds a convenient "Friends" button, allowing the user to instantly check their friends list with a single click.

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -51,7 +51,8 @@
 			buf += '<p><button class="button mainmenu3" name="joinRoom" value="ladder">Ladder</button></p></div>';
 
 			buf += '<div class="menugroup"><p><button class="button mainmenu4 onlineonly disabled" name="joinRoom" value="battles">Watch a battle</button></p>';
-			buf += '<p><button class="button mainmenu5 onlineonly disabled" name="finduser">Find a user</button></p></div>';
+			buf += '<p><button class="button mainmenu5 onlineonly disabled" name="finduser">Find a user</button></p>';
+			buf += '<p><button class="button mainmenu6 onlineonly disabled" name="joinRoom" value="view-friends-all">Friends</button></p></div>';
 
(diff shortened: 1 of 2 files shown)


mia-pi-git:
Hi. I've stated as much, but this was purposefully not done until the UI / commands stabilize. See #1844.

mia-pi-git:
(It also doesn't require a user to be in a room, they can PM it to any user or to the & fakeuser.)

Laevos:
Hi @mia-pi-git! I appreciate the feedback; so for UI changes like this, may I ask what would indicate that the commands have stabilized?



> (It also doesn't require a user to be in a room, they can PM it to any user or to the & fakeuser.)

This is true, and I should have mentioned this in my description. In the future when the commands are stable, imo it would still be useful to have a button linking to it on the main menu, since new users might not be able to intuit the `/friends` command as easily as a clearly labeled button. 

All this being said, may I leave this PR open until such a time as the commands are considered stable? Having tested it locally, it works without issue, and I suspect it would continue to work unless the command for opening the "Friends" room changed, but I understand wanting UI changes to be fully integrated with everything else. 

Thanks for your time!

mia-pi-git:
> may I ask what would indicate that the commands have stabilized?

Effectively, as it's my code, when I and Zarel judge it to be good enough to be stable. 

You can leave the PR if you want, sure, but it may be a while before we're done with the rest.

Laevos:
Understandable, thanks for explaining! I appreciate the response. 😸  I'll go ahead and leave this open then, cheers!

AnnikaCodes:
I think the commands are fairly stable; the syntax hasn't changed in 6+ months.

AnnikaCodes (approved):

PR #1872, opened 2021-10-09
Teambuilder: Show Ferroseed as legal in Gen 8 LC

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9000038


Zarel:
ty!

PR #1873, opened 2021-10-09
Teambuilder: Don't show Gmax radio buttons for Gmax formes

Karthik99999:


Zarel:
I still feel like choosing Gmax pokemon was way less confusing before the current system, but I guess there's no going back now.

PR #1874, opened 2021-10-11
Teambuilder: Disable validate button when disconnected

KrisXV:
If you disconnect from the server (via restart or being DCed), currently the Validate button is still clickable despite not doing anything (because it can't send a command).


Zarel:
ty!

PR #1885, opened 2021-11-04
Improve Team Preview decision text

DaWoblefet:
In formats where you decide to bring more than just leads, currently it says you're sending out your 4-6 Pokemon "first". You can see this most clearly when choosing team slots for Zoroark or in VGC. This PR modifies the text to indicate more clearly which Pokemon are being led, and which Pokemon are in the back. It's a minor QOL improvement.
Before:
image
image
After:
image
image

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -906,11 +906,11 @@
 				for (var i = 0; i < leadCount; i++) {
 					leads.push(myPokemon[this.choice.teamPreview[i] - 1].speciesForme);
 				}
-				buf += this.createOxfordCommaList(leads) + ' will be sent out first';
+				buf += this.createCommaSeparatedList(leads) + ' will be sent out first';
 				for (var i = leadCount; i < this.choice.count; i++) {
 					back.push(myPokemon[this.choice.teamPreview[i] - 1].speciesForme);
 				}
-				if (back) buf += ', with ' + this.createOxfordCommaList(back) + ' in the back';
+				if (back.length) buf += ', with ' + this.createCommaSeparatedList(back) + ' in the back';
 				buf += '. <br />';
 			} else if (this.choice.choices && this.request && this.battle.myPokemon) {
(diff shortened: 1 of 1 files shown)


DaWoblefet:
Not sure if the list-creating bit needs to be its own separate function, or if it should, if it should live somewhere else other than client-battle.

Zarel:
Hm, I think my main concern here is extraneous text. Specifically, your text is:

> **A** and **B** will be sent out first, with **C** and **D** in the back.

My original text was:

> **A, B, C, D** will be sent out first.

I think one big difference is that with mine, the bolded information (the part that you need to reread every time you make a team preview choice) is all combined, while in yours, it's kind of scattered throughout the sentence, making the reread unnecessarily harder.

I understand yours is a grammatical sentence, but you can understand how having complete sentences in UI can be annoying, right? Can you imagine instead of "[Earthquake] [U-turn]" buttons, they instead were like "You could choose [Earthquake] but perhaps you might consider [U-turn]"?

Anyway, I think we might prefer something like:

> A, B, C, D will be your team order.

Maybe bold the leads or something?

Zarel:
If you really want to keep them separate, maybe put them on two lines?

> A, B will be sent out first.
> C, D in the back.

I think I'd still prefer all commas, rather than using "and" in the list. (I also prefer the Oxford comma, but fortunately my preferred format sidesteps that entirely.)

DaWoblefet:
> If you really want to keep them separate, maybe put them on two lines?
> 
> > A, B will be sent out first.
> > C, D in the back.
> 
> I think I'd still prefer all commas, rather than using "and" in the list. (I also prefer the Oxford comma, but fortunately my preferred format sidesteps that entirely.)

I like this compromise! I think it distinguishes even more clearly this way, which was my goal.

image
image

Zarel:
Looks good; thanks!

PR #1886, opened 2021-11-07
Teambuilder: Fix bug caused due to #1882

PartMan7:
I could've sworn I'd added this line; my bad
pasteNotes is a null, leading to the upload failing
Can someone merge this in


js/client-teambuilder.js
```diff
@@ -1147,6 +1147,7 @@
 				buf += '<input type="hidden" name="title" id="pasteTitle">';
 				buf += '<input type="hidden" name="paste" id="pasteData">';
 				buf += '<input type="hidden" name="author" id="pasteAuthor">';
+				buf += '<input type="hidden" name="author" id="pasteNotes">';
```
KrisXV, line 1150: the name should probably be different

PartMan7: Oh oops

PR #1887, opened 2021-11-08
Teambuilder: Set format when importing Pokepaste

Karthik99999:
I thought this would be a cool thing to add to go along with #1882. Code will have to be updated once actual notes gets implemented, but I imagine it'll be a while before those come out.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1290,17 +1290,11 @@
 						if (/^https?:\/\/pokepast\.es\/.*\/json\s*$/.test(url)) {
 							var teamData = JSON.parse(data);
 							// Split the notes on the first newline
-							var notes = teamData.notes.split(/\n(.+)/s);
-							// Splitting the array like we did above creates an empty array element at the end if there was a newline
-							if (notes.length > 1) notes.pop();
-							var teamNotes = ''; // Not implemented yet
+							var notes = teamData.notes.split(/\n([^]+)/);
 							if (notes[0].startsWith('Format: ')) {
 								var formatid = toID(notes[0].slice(8));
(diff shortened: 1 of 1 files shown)


js/client-teambuilder.js
```diff
@@ -1289,12 +1289,14 @@
 					success: function (data) {
 						if (/^https?:\/\/pokepast\.es\/.*\/json\s*$/.test(url)) {
 							var teamData = JSON.parse(data);
-							Storage.activeSetList = self.curSetList = Storage.importTeam(teamData.paste);
+							var format = window.BattleFormats && window.BattleFormats[toID(teamData.notes.slice(8))];
```
Zarel, line 1292: It's probably worth checking if the first 8 characters of `teamData.notes` is in fact `"Format: "`.

Also, can we slice it only to the next linebreak? I think both of these are probably useful future-proofing.

js/client-teambuilder.js
```diff
@@ -1289,8 +1289,12 @@
 					success: function (data) {
 						if (/^https?:\/\/pokepast\.es\/.*\/json\s*$/.test(url)) {
 							var teamData = JSON.parse(data);
-							var format = window.BattleFormats && window.BattleFormats[toID(teamData.notes.slice(8))];
-							if (format) self.changeFormat(format.id);
+							var notes = teamData.notes;
+							if (notes.startsWith('Format: ')) {
+								var formatid = toID(notes.slice(8, notes.indexOf('\n')));
```
Zarel: I think this breaks if `notes` doesn't contain `\n`.

Karthik99999: i fixed that in my latest commit

Zarel:
Thanks!

PR #1888, opened 2021-11-12
Allow chaos/Zarel to edit admin usergroups from the ACP

mia-pi-git:


Zarel (changes requested):
We discussed this and I'm against it.

Zarel:
Oh I think I didn't think through what this is changing. I think this is all right. Although it looks like it's missing UI.

PR #1892, opened 2021-11-21
Fix translation errors

Dirain1700:
There will be no more files change.

I'm room staff of Japanese, but I'll get approval from the other staff just in case.


Zarel:
ありがとうございます！

PR #1894, opened 2021-11-24
Teambuilder: Set Arceus forme item to Plate in Gen 4-6 & BDSP

Karthik99999:
I also slightly refactored how the Let's Go and BDSP mods are handled by the dex and build-indexes, making it easier to add mods to the dex in the future.


Karthik99999:
Weird, the build is failing on a file I didn't touch

Zarel:
Thanks!

PR #1897, opened 2021-11-29
Stop using /~~serverid/action.php

mia-pi-git:
Discussed in #loginserver-dev. This should make it easier to move to the new loginserver (and much less confusing when reading the code).


mia-pi-git:
(Preferably this shouldn't be merged until we're ready to deploy the new loginserver)

PR #1898, opened 2021-11-29
Fix uploading replays

mia-pi-git:
the `?act=uploadreplay` breaks when removing the /~~serverid/action.php magic.
It would be ideal if this could be merged and patched asap - rn replays aren't working for people on new clients.


monsanto:
Discussed this briefly in the staff room. I don't know if this works or not but if stuff is already broken then it can't hurt.

PR #1900, opened 2021-12-01
Add BDSP Monotype teambuilder support

KrisXV:
This is fairly high priority, as a BDSP Monotype format will exist in the next 24 hours, and with the current client, trying to make a team in "gen8bdspmonotype" crashes the client (shows a blank teambuilder); this also fixes future instances.


Zarel:
ty!

PR #1905, opened 2021-12-05
Teambuilder: Fix CAP LC & BDSP Hackmons

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9048414


Marty-D (approved):
Thanks!

PR #1907, opened 2021-12-08
Make hyperlinks more visible when displayed in /wall 

missyaria:
here's a small workaround regarding this : https://www.smogon.com/forums/threads/make-hyperlinks-more-visible-when-having-them-displayed-in-wall-or-announce.3693900/
(excuse the large amount of commits, I made a mess with my branches)


style/client.css
```diff
@@ -1355,6 +1355,9 @@ p.or:after {
 	padding: 0 4px 0 3px;
 	font-style: normal;
 }
+.chat span a small {
```
Zarel, line 1358: `span` is probably not the correct match. Doesn't `/wall` have a class on that span?

Zarel: This should probably be next to the thing that makes the regular link text gray, too.

Zarel: Ah, I see, it's right below this. This should be moved below `.chat small`, rather than above. Also, the selector should be `.chat .message-announce small`.

Zarel: Hm, the contrast here is kind of poor. Probably the background on announcements should be darker.

missyaria: Should I make the background color of `.message-announce` from battle-log.css darker ?

Zarel: I can handle that, no worries.

Zarel:
Thanks!

PR #1921, opened 2022-01-12
Add option to match system theme for light/dark mode

Karthik99999:
https://www.smogon.com/forums/threads/add-option-to-match-system-theme-for-light-dark-mode-in-client.3695930/

One question I have is do I have to check if `window.matchMedia` exists before using it? The support for it seems to go pretty far back but I check if it exists just to be safe. Please let me know if I don't need to do that.

### DIFF
--- a/js/storage.js
+++ b/js/storage.js
@@ -319,6 +319,17 @@ var updatePrefs = function () {
 			app.addPopupMessage('Your version of Chrome has a bug that makes animated GIFs freeze games sometimes, so certain animations have been disabled. Only some people have the problem, so you can experiment and enable them in the Options menu setting "Disable GIFs for Chrome 64 bug".');
 		});
 	}
+
+	var colorSchemeQuerySupported = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').media !== 'not all';
+	if (Storage.prefs('theme') === 'system' && !colorSchemeQuerySupported) {
+		Storage.prefs('theme', null);
+	}
+	if (Storage.prefs('dark') !== undefined) {
+		if (Storage.prefs('dark')) {
+			Storage.prefs('theme', 'dark');
+		}
+		Storage.prefs('dark', null);
+	}
 };
 Storage.whenPrefsLoaded(updatePrefs);
 
--- a/js/client-topbar.js
+++ b/js/client-topbar.js
@@ -447,13 +447,13 @@
 			'change input[name=blockchallenges]': 'setBlockchallenges',
 			'change input[name=blockpms]': 'setBlockpms',
 			'change input[name=inchatpm]': 'setInchatpm',
-			'change input[name=dark]': 'setDark',
 			'change input[name=temporarynotifications]': 'setTemporaryNotifications',
 			'change input[name=refreshprompt]': 'setRefreshprompt',
 			'change select[name=bg]': 'setBg',
 			'change select[name=timestamps-lobby]': 'setTimestampsLobby',
 			'change select[name=timestamps-pms]': 'setTimestampsPMs',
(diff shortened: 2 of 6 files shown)


Slayer95:
This should probably be made the default, right?

DaWoblefet:
I agree, I think it should default to system with the option to override.

js/storage.js
```diff
@@ -319,6 +319,13 @@ var updatePrefs = function () {
 			app.addPopupMessage('Your version of Chrome has a bug that makes animated GIFs freeze games sometimes, so certain animations have been disabled. Only some people have the problem, so you can experiment and enable them in the Options menu setting "Disable GIFs for Chrome 64 bug".');
 		});
 	}
+
+	var colorSchemeQuerySupported = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').media !== 'not all';
+	var theme = Storage.prefs('theme');
+	if ((!theme || theme === 'system') && !colorSchemeQuerySupported) {
+		Storage.prefs('theme', Storage.prefs('dark') ? 'dark' : 'light');
```
Slayer95, line 326: Might as well reset the setting for all users?

Zarel:
idk, I think PS's light theme is popular enough that it's not a bad default.

It's like how Discord defaults to dark even though it _could_ match the system theme.

Slayer95:
Afaict, the primary point of this feature is the capability to automatically set dark theme and play nice with the OS detection of ambient light.

Since that capability has biological/health implications, I'd argue it should be the default.

Incidentally, there used to be an API that more specifically caters to it (DeviceLightEvent), but it was deprecated... seemly in favor of matchMedia?

Zarel:
> I don't think it makes sense to add an option to match system theme if it's not the default.

Discord has the same option. Dark, Light, or Sync with Computer. I would personally use Sync, but I think PS's light mode is just really popular (and I intentionally designed it to be black-on-colors with very little pure white, compared to nearly any other app with a light/dark-mode sync).

Zarel:
We really do have a pretty similar situation to Discord. Discord's light mode sucks, so Discord defaulting to dark mode rather than automatic is nice, because most people who would normally prefer light mode would prefer Discord's dark mode.

PS's dark mode is good, but PS's light mode is really good; I'd argue much better than basically any other light mode on the market. A lot of people use dark mode because they don't like blinding pure-white interfaces like GitHub's or Google's or Windows Explorer's etc etc. But PS's isn't blinding pure-white! So I do want to default to PS's light mode.

DaWoblefet:
I use PS light mode myself too and personally prefer it to dark mode. I guess I just think it's weird to have a sync with system theme if in practice, it's going to be equivalent to just setting light/dark mode. It just seems to be an unnecessary additional option. Maybe people change their system light/dark theme more often than I'm anticipating; I would figure you just set system theme once and you'd keep it that way. But I may be wrong in that assumption, in which case keeping light as default sounds fine to me.

Zarel:
> I guess I just think it's weird to have a sync with system theme if in practice, it's going to be equivalent to just setting light/dark mode.

Oh, I have my system set up to be light in the daytime and dark at night. This is a common use-case, and I'd assume is why the thread is asking for a sync option.

DaWoblefet:
> Oh, I have my system set up to be light in the daytime and dark at night. This is a common use-case, and I'd assume is why the thread is asking for a sync option.

Yeah that makes a lot of sense; sorry for derailing the thread with bad assumptions, Karthik and Zarel. This sounds good to me!

Karthik99999:
went back to light theme being the default, should be good to go now

Zarel:
Thanks!

PR #1924, opened 2022-01-20
Teambuilder: Show gen 5 sprites when old graphics are preferred

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9090255


Zarel:
Thanks!

PR #1926, opened 2022-01-22
Fix PP display for some moves in Gen 1-2

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9093024

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -560,12 +560,11 @@
 		buf += '</span> ';
 
 		// power, accuracy, pp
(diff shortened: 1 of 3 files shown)


DaWoblefet:
Is this just in battles, or does it also get the teambuilder too?

Karthik99999:
> Is this just in battles, or does it also get the teambuilder too?

It gets fixed for both

PR #1931, opened 2022-01-25
Support syncing YouTube videos

mia-pi-git:
This makes it so the server can control <youtube> tags created in game room fields - it's meant for the group watch feature. 
I abstracted the code into a BattleGame class (in client-game.js) because GameRooms use BattleRoom, and putting game code in battle code seemed messy. This way, other users of GameRoom (or this user) can add more support as needed.

The BattleGame for youtube sets a listener for `|queryresponse|youtube` (which I used since it's easily supported and listened to within existing infrastructure), loads the YouTube IFrame API (necessary to control the video), creates a player object, and executes the commands sent by the server in the JSON blob. Currently supported: `at` (tells the client where the server is and syncs it if they're >2s apart), `pause` (pauses the video), and lastly `play` (starts the video).

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -50,6 +50,7 @@
 			this.$chat = this.$chatFrame.find('.inner');
 
 			this.$options = this.battle.scene.$options.html('<div style="padding-top: 3px; padding-right: 3px; text-align: right"><button class="icon button" name="openBattleOptions" title="Options">Battle Options</button></div>');
+			this.game = new BattleGame(this);
 		},
 		events: {
 			'click .replayDownloadButton': 'clickReplayDownloadButton',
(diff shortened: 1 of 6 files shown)


js/client-battle.js
```diff
@@ -50,6 +50,7 @@
 			this.$chat = this.$chatFrame.find('.inner');
 
 			this.$options = this.battle.scene.$options.html('<div style="padding-top: 3px; padding-right: 3px; text-align: right"><button class="icon button" name="openBattleOptions" title="Options">Battle Options</button></div>');
+			this.game = new BattleGame(this);
```
Zarel, line 53: It's concerning to me that you're just adding this to every single battle, rather than only YouTube rooms.

Anyway I don't think `|queryresponse|` and creating a new object is necessarily the right approach. I think probably a `|controlyoutube|` protocol message type, which is handled by the battle event handler is the right way to do it.

mia-pi-git: > It's concerning to me that you're just adding this to every single battle, rather than only YouTube rooms.

Yes, my intent was to make this a spot where plugins that use GameRoom could share code. (I also only attach YouTube code for YouTube rooms by way of the queryresponse).

> Anyway I don't think |queryresponse| and creating a new object is necessarily the right approach. I think probably a |controlyoutube| protocol message type, which is handled by the battle event handler is the right way to do it.

If you prefer that, I can do it. To me, it seems uglier to let plugin code bleed into battle code, though.

PR #1933, opened 2022-01-26
Tooltips: Fix small display issue in Gen 2

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9099699

Before
image
After
image


Zarel:
Thanks!

PR #1944, opened 2022-03-02
Rotation Battles

MathyFurret:
Server pull request: https://github.com/smogon/pokemon-showdown/pull/8677

Definitely not up to standards right now, just wanted to have something to show ~~im not great at frontend dev~~

Preview: https://www.youtube.com/watch?v=MuMJA8-qzec This shows what's currently working, and some of what still needs to be done

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -965,6 +965,7 @@
 						}
 						var target = '';
 						buf += myPokemon[i].speciesForme + ' will ';
+						if (this.choice.willRotate) buf += 'rotate in and ';
 						if (parts.length > 2) {
 							var targetPos = parts[2];
 							if (targetPos === 'mega') {
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -367,7 +367,8 @@
 					this.choice = {
 						choices: [],
 						switchFlags: {},
-						switchOutFlags: {}
+						switchOutFlags: {},
+						willRotate: ''
 					};
 				}
 				this.updateMoveControls(type);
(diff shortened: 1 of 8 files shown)


urkerab:
Just an observation from watching the video, but surely you should by default (assuming e.g. no Illusion users) pick your first three Pokémon when entering battle?

Slayer95:
I think showing the Pokémon in the Rotate Right/Left actions doesn't seem very intuitive, and can be confused with switching. BTW, remember to also show how the battle controls will be displayed in small screens.

MathyFurret:
Some issues/ideas I'm aware of
- The rotation options dont hide/show on the small screen version yet
- Switch options should be hidden when rotating
- On cartridge, you can't navigate from one sub-active to the other, you have to return to the front Pokemon (pressing B also returns you to the front Pokemon). I might consider matching this design, perhaps even displaying "**Back**" instead of the front Pokemon's name.
- The waiting screen should probably say something like "X will rotate in and use Y"

I'll fix the more nitpicky stuff once I get more feedback on the overall design.

MathyFurret:
image

Playing around with the design some more, this time on the small display. It pushes stuff up a lot, but there's really nowhere else I can put the buttons. (I could try squeezing them off to the side, but meh)

I still think putting the Pokemon in them is ideal -- it makes it clearer at a glance what you're rotating to (and it's clear which action you're doing with each Pokemon, since the corresponding switch buttons would be disabled), and a decently large button size is necessary for accessibility.

PR #1945, opened 2022-03-07
Don't remove type changes for Gmax forme changes

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9143591

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -2384,8 +2384,8 @@ export class Battle {
 			if (!poke.getSpeciesForme().endsWith('-Gmax') && !species.name.endsWith('-Gmax')) {
 				poke.removeVolatile('typeadd' as ID);
 				poke.removeVolatile('typechange' as ID);
+				if (this.gen >= 6) poke.removeVolatile('autotomize' as ID);
 			}
-			if (this.gen >= 7) poke.removeVolatile('autotomize' as ID);
 
 			if (!kwArgs.silent) {
 				this.activateAbility(poke, fromeffect);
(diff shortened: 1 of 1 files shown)


DaWoblefet:
This looks good to me. Ideally we'd not be sending Gigantamax as a formechange at all (since Dynamax isn't really a formechange), but that's a whole lot of work compared to the solution here, which works well.

src/battle.ts
```diff
@@ -2381,8 +2381,10 @@ export class Battle {
 			let species = Dex.species.get(args[2]);
 			let fromeffect = Dex.getEffect(kwArgs.from);
 			let isCustomAnim = species.name.startsWith('Wishiwashi');
-			poke.removeVolatile('typeadd' as ID);
-			poke.removeVolatile('typechange' as ID);
+			if (!poke.getSpeciesForme().endsWith('-Gmax') && !species.name.endsWith('-Gmax')) {
+				poke.removeVolatile('typeadd' as ID);
+				poke.removeVolatile('typechange' as ID);
+			}
 			if (this.gen >= 7) poke.removeVolatile('autotomize' as ID);
```
Karthik99999, line 2388: should I move the autotomize thing into the if statement as well (and also change it to check for gen 6 cause that's when forme changes started to get rid of autotomize according to bulbapedia)

DaWoblefet: Yep, they're not forme changes so weight's not reset either. That generation check isn't even being done server-side as far as I can tell. It is Gen 6 onward.

urkerab: > That generation check isn't even being done server-side as far as I can tell.

Server side, Gen 6+ autotomize directly modifies the Pokémon's weight, which means that a forme change will get rid of the effect, while Gen 5 adds a volatile (that isn't copied by Baton Pass) which reduces the weight.

Zarel:
(Yeah, what Wob said, Gmax should ideally just be a volatile status.)

PR #1946, opened 2022-03-07
Support Gmax sprites without formechange

Karthik99999:
Server Side: smogon/pokemon-showdown#8685

I thought of this as soon as my other pr got merged.

Should I get rid of the changes from my other pr? Or should I leave them in for old replays?

### DIFF
--- a/src/battle-dex.ts
+++ b/src/battle-dex.ts
@@ -495,8 +495,6 @@ const Dex = new class implements ModdedDex {
 			let isGigantamax = false;
 			if (pokemon.volatiles.dynamax) {
 				if (pokemon.volatiles.dynamax[1]) {
-					// Gmax sprites are already extremely large, so we don't need to double.
-					isDynamax = false;
 					isGigantamax = true;
 				} else if (options.dynamax !== false) {
 					isDynamax = true;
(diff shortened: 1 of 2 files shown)


DaWoblefet:
> Should I get rid of the changes from my other pr? Or should I leave them in for old replays?

They should still be there for backwards compatability.

src/battle-dex.ts
```diff
@@ -492,12 +492,19 @@ const Dex = new class implements ModdedDex {
...
+					isGigantamax = true;
+				} else if (options.dynamax !== false) {
+					isDynamax = true;
+				}
+			}
+			pokemon = pokemon.getSpeciesForme() + (isGigantamax ? '-Gmax' : '');
 		}
 		const species = Dex.species.get(pokemon);
-		// Gmax sprites are already extremely large, so we don't need to double.
-		if (species.name.endsWith('-Gmax')) isDynamax = false;
```
Zarel, line 500: This code should be left in to retain support for the old system, right? Otherwise, everything else looks good.

Karthik99999: oh right I didn't think about that

Marty-D (approved):

PR #1948, opened 2022-03-21
Gen 1: Fix Conversion type copying

Karthik99999:
server side: smogon/pokemon-showdown#8701


Karthik99999:
I realized that I don't have to make conversion send the target since it doesn't need the extra added type logic that reflect type has, so closing this

PR #1953, opened 2022-03-29
Gen I-II: Fix speed stat displayed when paralysed

DavidSyrup:
Fixing an issue with speed not being seen as lowered when paralyzed in gen 1 and 2.
Issue was up because of " if (pokemon.status === 'par' && ability !== 'quickfeet') {" and abilities were inexistant in the first two gens. My guess is that ability is at an undefined value at this moment but I'm not sure.

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -1204,6 +1204,10 @@ class BattleTooltips {
 			}
 		}
 
+		if (pokemon.status === 'par' && this.battle.gen < 3) {
+			stats.spe = Math.floor(stats.spe * 0.25)
+		}
+
 		return stats;
 	}
 
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -997,6 +997,8 @@ class BattleTooltips {
 
 			if (this.battle.gen > 2 && ability === 'quickfeet') {
 				stats.spe = Math.floor(stats.spe * 1.5);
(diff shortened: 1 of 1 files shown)


DaWoblefet:
I think you're on the right track! We'd probably want to know why the ``ability !== 'quickfeet'`` check is failing though and fix it in that if statement (or earlier, if necessary), rather than tacking on another one at the end.

DaWoblefet:
Aha! I found it. Gen 1 stats return a lot earlier in this function (around line 1004). This is before the paralysis code can be reached! As you correctly suspected, this is because there's no items / Abilities to mess with. For your purposes, what you need to do is add another else if after if statement at line 998. You can use the brn check above it as a model for your comparison.

DaWoblefet (approved):

src/battle-tooltips.ts
```diff
@@ -997,6 +997,8 @@ class BattleTooltips {
 
 			if (this.battle.gen > 2 && ability === 'quickfeet') {
 				stats.spe = Math.floor(stats.spe * 1.5);
+			} else if (this.battle.gen < 2 && pokemon.status === 'par') {
```
Marty-D, line 1000: Isn't this excluding Gen 2? I realize the burn check above does the same thing but that seems wrong too.

AnnikaCodes: The check above excludes Gen 1 and 2, but that should be fine since they don't have Abilities anyway.

PR #1954, opened 2022-03-30
TOOLTIPS/TEXT PARSER: Fix Nickname in tooltips and in Battle Log

DavidSyrup:
Fixed a "bug" (I think it wasn't just programmed just a feature not completely implemented).
For both of these fixes I just added conditions to take into account the "Ignore Nicknames" Option.


AnnikaCodes:
Thanks!

PR #1956, opened 2022-04-03
Tooltips: Trying to fix the name of the ability when changed and suppressed

DavidSyrup:
This is not a final version but a first draft of what I think should be done, basically what I understand from the architecture of the code is when you have to trigger an ability it goes from the switch in battle.ts and calls the server for changing the abilty and the name. But as of now the normal behavior is to normally not change the name of the ability as it is still just (suppressed), so for the moment I put a condition where if the ability name is '(suppressed)' it shouldn't switch to Mummy(base: "baseAbility"). I don't think it should be the final solution but I want to report what's the problem at its core.

Should be hard to tackle, might need some ideas for it

### DIFF
--- a/src/battle-text-parser.ts
+++ b/src/battle-text-parser.ts
@@ -505,10 +505,16 @@ class BattleTextParser {
 		}
 
 		case 'switchout': {
-			const [, pokemon] = args;
-			const side = pokemon.slice(0, 2);
+			const [, pokemon, details] = args;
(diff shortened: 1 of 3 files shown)


monsanto:
I'm sorry nobody replied with feedback at the time. I don't fully understand what is going on here, and you are saying this PR isn't actually ready to merge, and since it has been nearly a year since I am going to close it. If you are still interested in tackling this problem we can reopen.

PR #1961, opened 2022-04-13
Fix Ultra Burst text after selecting a move

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9170998


Marty-D (approved):
Thanks! 👍

Issue #1943, opened 2022-02-16
Current responsive page is hard to navigate


zurgeg:
Title says it all, but for additional clarification, this:
![Screen Shot 2022-02-16 at 14 07 37](https://user-images.githubusercontent.com/46549042/154337865-f15e77c8-7207-47ec-83cb-8d15cb984c0d.png)

is not an enjoyable experience (taken on Firefox responsive testing mode with "iPhone SE Gen 2" profile).

My suggestion is increasing overall element and tap target size when on mobile. Currently some of the targets (notably the Import button) are harder to hit then Gen V C-Gear seeds.

Another issue with the current experience is if we go onto a smaller device (OG iPhone SE), we run into this:
![Screen Shot 2022-02-16 at 14 12 35](https://user-images.githubusercontent.com/46549042/154338505-3bb406d4-2172-4e78-b1ee-4dddc5a3f436.png)

I would suggest removing the username bit while on mobile, so that we get something like this:
![image](https://user-images.githubusercontent.com/46549042/154338831-13a663b9-a240-4156-8248-93dd539941ae.png)

This way, the username doesn't overlap with the teambuilder button. 

I understand that Showdown isn't really made to be played on mobile, but a little extra responsiveness could help the experience on smaller screens.

thisisdylandev:
I'd love to help with this!

joebangara43903:
![image](https://user-images.githubusercontent.com/131228993/232965414-fd3e5e66-5ccc-4e87-8ff8-e5dec634be61.png)

Zarel:
The new Preact client has a teambuilder design that flexes down to mobile.

I've been waffling on whether to close reports fixed in Preact client, and I've decided on yes. Preact client is in the master branch of this repository and it's useful for the issue tracker to track whether things have been fixed in the codebase, as one of the biggest use cases is for devs to track what still needs doing.

As an end user, please do try out the alpha test: https://psim.us/preactalpha or wait for a stable release in a few months, tops.

Here are some screenshots of what it currently looks like on mobile (not final, but the design is unlikely to have major changes from here):

<img width="410" alt="Image" src="https://github.com/user-attachments/assets/e37c42e1-41fb-483b-b08c-9fb5ce5fbb48" />

<img width="410" alt="Image" src="https://github.com/user-attachments/assets/eceea26f-c64a-47aa-89e1-61507540d7b5" />

[...]

PR #1973, opened 2022-05-16
(fix): hide username on screens smaller than 500px

thisisdylandev:
fix for #1943 

First PR here, so open to constructive feedback on how to better do this, if this is applicable.

### DIFF
--- a/js/client-topbar.js
+++ b/js/client-topbar.js
@@ -36,7 +36,7 @@
 			if (!app.user.loaded) {
 				buf = '<button disabled>Loading...</button>';
 			} else if (app.user.get('named')) {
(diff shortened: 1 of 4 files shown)


DaWoblefet:
Can you provide screenshots of the new behavior? I have a pretty good idea of what it's supposed to be doing but screenshots are always nice :).

thisisdylandev:
@DaWoblefet yeah, sure thing. Thank you for the quick response!
Screen Shot 2022-05-18 at 8 31 36 AM
Screen Shot 2022-05-18 at 8 32 00 AM
Screen Shot 2022-05-18 at 8 32 14 AM

DaWoblefet:
I personally think this is OK. The only other idea I could think of would be to reduce font-size, or force it to say inside its container and add a text-overflow: ellipsis, or perhaps some combo of the two. 

Can we call the class something more descriptive? Maybe ``username-topbar``?

AnnikaCodes:
I like Wob's idea of using `text-overflow: ellipsis`; it would solve the navigation problems while still making it somewhat possible for users with multiple accounts to tell which alt they are on at a glance.

Zarel:
I think this change is better than nothing, so I'm merging for now.

Zarel (approved):

AnnikaCodes:
Hi, sorry about the wait to merge this. It should go live today!

PR #1974, opened 2022-05-17
Fix text when switching out after failing a switch move

Karthik99999:


src/battle.ts
```diff
@@ -867,15 +866,16 @@ export class Side {
...
+		const effect = Dex.getEffect(kwArgs.from);
 		if (pokemon.lastMove !== 'batonpass' && pokemon.lastMove !== 'zbatonpass') {
 			pokemon.clearVolatile();
 		} else {
 			pokemon.removeVolatile('transform' as ID);
 			pokemon.removeVolatile('formechange' as ID);
 		}
-		if (pokemon.lastMove === 'uturn' || pokemon.lastMove === 'voltswitch') {
-			this.battle.log(['switchout', pokemon.ident], {from: pokemon.lastMove});
+		if (['uturn', 'voltswitch', 'flipturn'].includes(effect.id)) {
```
DaWoblefet, line 877: Is there a way we can generalize this better? This is missing Parting Shot too, for example.

Karthik99999: how about this?
```js
if (pokemon.lastMove !== 'batonpass' && pokemon.lastMove !== 'zbatonpass') {
	this.battle.log(['switchout', pokemon.ident], {from: kwArgs.from});
}
```
`BattleTextParser` should automatically deal with the `from` effect not having its own switch out text (or 'from` not existing

DaWoblefet: Your solution here looks good.

Marty-D (approved):
Thanks! 👍

PR #1975, opened 2022-05-30
Change when accept challenge button gets disabled

Karthik99999:
https://www.smogon.com/forums/threads/dont-disable-accept-button-from-a-challenge-if-you-clicked-accept-without-a-team-for-the-format.3702601/


Marty-D (approved):
👍

PR #1982, opened 2022-07-18
Battle: Support Persistent Extensions

HisuianZoroark:
~~While Trick Room and the other effects don't naturally get extended, in CAP persistent gives them an extra 2 turns.~~

~~example after: https://user-images.githubusercontent.com/96159984/179448974-494c6f00-b17f-4f74-a16b-ac0553d2dfd9.png
(Note how the OU battle doesn't have this effect while the CAP match does.)~~

~~This was discussed with Marty and some members (notably spoo) of the CAP community. A tie-in pull request to remove the activation message for Persistent might be needed if this goes through.~~

See comment down below for what the code does now.


HisuianZoroark:
Instead of what I originally did, I discussed with spoo again, and now this just simply gives a more accurate for certain moves if activated with Persistent's extension.

PR #1987, opened 2022-08-09
Desktop: Use Electron instead of NW.js

AnnikaCodes:
Electron makes the desktop client smaller and reduces the number of workarounds needed to account for differences between the desktop and browser clients. Packaging is also simplified, with the `electron-forge` tool replacing the multi-step NW.js process.

All the dependencies and code to be bundled for the desktop client are in the `desktop/` folder, with packaging instructions in `desktop/DESKTOP.md`. I have also created a GitHub Actions workflow to automatically rebuild the desktop client whenever it's changed, but we can remove that if it's not wanted.

macOS codesigning seems to be simplified by Electron's tooling, but I didn't test it myself because I didn't want to spend $100 on a developer account.


NotTsunami:
Looking forward to this! Is there any additional testing or work that needs to be done to move forward?

AnnikaCodes:
pretty sure it's waiting for @Zarel to review it

desktop/package.json
```diff
@@ -1,16 +1,26 @@
...
+  "dependencies": {
+    "electron-squirrel-startup": "^1.0.0"
+  },
+  "devDependencies": {
+    "@electron-forge/cli": "^6.0.0-beta.65",
+    "@electron-forge/maker-deb": "^6.0.0-beta.65",
+    "@electron-forge/maker-dmg": "^6.0.0-beta.65",
+    "@electron-forge/maker-rpm": "^6.0.0-beta.65",
+    "@electron-forge/maker-squirrel": "^6.0.0-beta.65",
+    "@electron-forge/maker-zip": "^6.0.0-beta.65",
```
NotTsunami, line 23: Well, while we wait for review these can be bumped to the stable versions, and optionally Electron v22 has been released (v22 will be the last version supporting Windows 7/8/8.1, so it probably shouldn't be taken past that)

monsanto:
This seems like a good thing to do and I'm sorry Z never reviewed your PR but, I don't use the desktop client or find it useful, and I am looking to pare down this project to the essentials. So, if this is something you're motivated to do, I'll give you free reign on it. I think a GH action to update it is a good idea as otherwise I would forget to update it. If you or another trusted programmer aren't motivated to handle it on your own, I would just close this issue.

NotTsunami:
> This seems like a good thing to do and I'm sorry Z never reviewed your PR but, I don't use the desktop client or find it useful, and I am looking to pare down this project to the essentials. So, if this is something you're motivated to do, I'll give you free reign on it. I think a GH action to update it is a good idea as otherwise I would forget to update it. If you or another trusted programmer aren't motivated to handle it on your own, I would just close this issue + remove the desktop client from whatever.

I personally use the desktop client and believe it is great. This PR should make maintaining and packaging the desktop way easier, making it much less of a headache.

NotTsunami:
This might fix #2109

PR #1996, opened 2022-09-24
Let's Go: Mega Evolve Y

EvezaGit:
In Gen 7 Let's Go formats, these changes allow Pokemon with multiple Mega Evolutions to choose between their Mega Evolutions during battle. This is accurate to cartridge. Have tested fairly extensively on my side server, so hopefully it all works! 

This pull request has been made simultaneously with one for smogon/pokemon-showdown.


js/client-battle.js.bak
```diff
@@ -0,0 +1,1580 @@
+(function ($) {
```
DaWoblefet, line 1: Looks like you accidentally committed a backup file you made.

PR #2005, opened 2022-10-29
User and active battle counts update on an interval

StephenArg:
I added a feature to allow the user count and active battle counts to automatically update themselves every 20 seconds.

Currently these values only update on refocusing the room window after 60 seconds have passed since last updated.

### DIFF
--- a/js/client-rooms.js
+++ b/js/client-rooms.js
@@ -24,6 +24,12 @@
 			app.send('/cmd rooms');
 			app.user.on('change:named', this.updateUser, this);
 			this.update();
+			this.chatroomInterval = setInterval(() => {
+				if (app.curSideRoom?.id === 'rooms') {
+					app.send('/cmd rooms');
+					this.update()
+				}
+			}, 20000);
 		},
 		initSectionSelection: function () {
 			var buf = ['<option value="">(All rooms)</option>'];
@@ -166,6 +172,8 @@
 		},
 		closeHide: function () {
 			app.sideRoom = app.curSideRoom = null;
+			clearInterval(this.chatroomInterval);
+			this.chatroomInterval = null;
 			this.close();
 		},
 		finduser: function () {
--- a/js/client-rooms.js
+++ b/js/client-rooms.js
@@ -24,10 +24,10 @@
 			app.send('/cmd rooms');
 			app.user.on('change:named', this.updateUser, this);
 			this.update();
-			this.chatroomInterval = setInterval(() => {
-				if (app.curSideRoom?.id === 'rooms') {
+			this.chatroomInterval = setInterval(function () {
+				if (app.curSideRoom && app.curSideRoom.id === 'rooms') {
 					app.send('/cmd rooms');
-					this.update()
+					this.update();
 				}
 			}, 20000);
 		},


HoeenCoder (commented):
This seems fine in concept to me. I don't know if this will cause an significant increase in server load though since were asking the server to compile the room list for each user on the rooms page every 20 seconds.

I only have server write access anyways so I can't merge this, but I like the idea.

StephenArg:
> This seems fine in concept to me. I don't know if this will cause an significant increase in server load though since were asking the server to compile the room list for each user on the rooms page every 20 seconds.
> 
> I only have server write access anyways so I can't merge this, but I like the idea.

Yea I didn't consider the additional server usage. Maybe it would be something worth looking at after merging, and if it significantly added to the server usage then it can be reversed. Though you may be on to something; that could be the reason it's being implemented this way in the first place (only allowing it to refresh after 60 seconds and refocusing on that tab)

AnnikaCodes:
As far as I can tell, this only refreshes when the Rooms tab is in focus, right? If so, I'm not concerned about server. It would be great if you could fix the CI failures, though.

js/client-rooms.js
```diff
@@ -24,6 +24,12 @@
 			app.send('/cmd rooms');
 			app.user.on('change:named', this.updateUser, this);
 			this.update();
+			this.chatroomInterval = setInterval(() => {
+				if (app.curSideRoom?.id === 'rooms') {
```
MathyFurret, line 28:
```suggestion
			this.chatroomInterval = setInterval(function () {
				if (app.curSideRoom && app.curSideRoom.id === 'rooms') {
```

HoeenCoder:
> As far as I can tell, this only refreshes when the Rooms tab is in focus, right? If so, I'm not concerned about server. It would be great if you could fix the CI failures, though.

This could include a lot of users - anyone who doesn't use chatrooms and only battles/uses teambuilder. Then again, we probably get tons of these requests already, it could just be negligible.

> Yea I didn't consider the additional server usage. Maybe it would be something worth looking at after merging, and if it significantly added to the server usage then it can be reversed. Though you may be on to something; that could be the reason it's being implemented this way in the first place (only allowing it to refresh after 60 seconds and refocusing on that tab)

Its probably the reason yeah. Worst case we could tweak/revert.

MathyFurret (commented):
Our .js files aren't transpiled, so they still have to use ES3 syntax.

js/client-rooms.js
AnnikaCodes: i hate internet explorer users

StephenArg: Yea I caught it as you commented. Can't we use babel? It would automatically convert our code to something that would work with es3

HoeenCoder: That would be something to discuss with @Zarel. I don't know what the plan is for the preact rewrite.

StephenArg: Ok, I'll see if I can reach out to him again

StephenArg: Someone may need to rerun the workflow in the mean time. I can't trigger it because I'm a first time contributor

AnnikaCodes: Zarel is pretty busy right now, I'll ask him to consider using a transpiler for the client when he's finished with some more important tasks.

StephenArg: Thanks! I remember talking to him a year or two about helping out with the Preact rewrite. I be happy to contribute if he's still working on it

StephenArg:
Is there a way to check out the server usage? It might be good to write down the current average usage, then maybe in a few days we can see if it marginally increased

HoeenCoder:
Thats a good point, though it may not be accurate as overall memory/cpu usage is affected by many factors and can vary over time.

StephenArg:
> Thats a good point, though it may not be accurate as overall memory/cpu usage is affected by many factors and can vary over time.

I don't think it needs to be accurate by any means. Just looking for any potential spike in usage after merging.

js/client-rooms.js
```diff
@@ -24,6 +24,12 @@
 			app.send('/cmd rooms');
 			app.user.on('change:named', this.updateUser, this);
 			this.update();
+			this.chatroomInterval = setInterval(function () {
+				if (app.curSideRoom && app.curSideRoom.id === 'rooms') {
+					app.send('/cmd rooms');
+					this.update();
```
Zarel, line 30: Was this tested? This is a crash and also seems unnecessary because `/cmd rooms`'s response already triggers an update, and until it gets a response there's nothing to update.

PR #2010, opened 2022-11-15
Tooltips: Add RBY Recover failure indicator

MathyFurret:
https://www.smogon.com/forums/threads/visual-indicator-for-rby-recover-failure.3677391/

### DIFF
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -763,7 +763,7 @@ class BattleTooltips {
 			}
 			// RBY healing move glitch
 			if (this.battle.gen === 1 && !toID(this.battle.tier).includes('stadium') &&
-					['recover', 'softboiled', 'rest'].includes(move.id)) {
+				['recover', 'softboiled', 'rest'].includes(move.id)) {
 				const hpValues = [];
 				let hp = serverPokemon.maxhp - 255;
 				if (hp > 0 && serverPokemon.hp % 256 !== 0) {
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -772,7 +772,7 @@ class BattleTooltips {
 						hpValues.push(hp - 256);
 					}
 				}
-				if (hpValues.length) text += `<p>Will fail if HP is ${hpValues.join(', ')}</p>`;
+				if (hpValues.length) text += `<p>Will fail if HP is ${hpValues.join(' or ')}</p>`;
 			}
 		}
 		return text;
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -766,7 +766,7 @@ class BattleTooltips {
 					['recover', 'softboiled', 'rest'].includes(move.id)) {
 				const hpValues = [];
 				let hp = serverPokemon.maxhp - 255;
-				if (hp > 0 && hp % 256 !== 0) {
+				if (hp > 0 && serverPokemon.hp % 256 !== 0) {
 					hpValues.push(hp);
 					if (hp - 256 > 0) {
 						hpValues.push(hp - 256);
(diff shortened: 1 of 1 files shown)


MathyFurret:
Realized the server only checks two values so parity was kept here. This is now up to date w/ the server mechs

Slayer95:
My reading of https://www.smogon.com/forums/threads/visual-indicator-for-rby-recover-failure.3677391/#post-8738188 is that this PR hasn't been accepted because the current implementation depends on the current HP of the Pokémon, which it shouldn't.

MathyFurret:
^ That helped me realize another error, it's fixed now

Slayer95:
The current iteration fails to warn if the Pokemon's max HP is 511. Also, the code is made terribly confusing, by relying on the sign of the mod operator, or even performing mod with a maybe-negative number.

MathyFurret:
If the max HP is 511, [the glitch doesn't happen](https://github.com/smogon/pokemon-showdown/pull/9603).

If someone else complains about % with a possibly negative number, I'll make it clearer. A long time ago it would have confused me, too, but having the same sign rules as multiplication/division is pretty easy to learn.

Slayer95:
The related mechanics are already complex enough. Simpler code is much more likely to get reviews (something you can see in docs PRs everywhere), and 3 years after the PR was opened, I'd think that's something you'd consider important.

But don't take it as me pressuring you.

Zarel:
Personally I think I entirely missed this the first time around, I think this was PRed when I was burned out. I think this is exactly what I wanted and I'd be happy to take it.

play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -761,6 +761,19 @@ class BattleTooltips {
 			if (move.flags.wind) {
 				text += `<p class="movetag">&#x2713; Wind <small>(activates Wind Power and Wind Rider)</small></p>`;
 			}
+			// RBY healing move glitch
+			if (this.battle.gen === 1 && !toID(this.battle.tier).includes('stadium') &&
+				['recover', 'softboiled', 'rest'].includes(move.id)) {
+				const hpValues = [];
+				const hp = serverPokemon.maxhp - 255;
+				if (hp % 256 > 0) {
```
Zarel, line 769:
```suggestion
				// glitches at HP values equal to `maxHP + 1 % 256` unless they are `0 % 256`
				if (hp % 256 !== 0) {
```

Zarel, line 769:
```suggestion
				// glitches at HP values equal to `maxHP + 1 % 256` unless they are `0 % 256`
				const hp = serverPokemon.maxhp - 255;
				if (hp % 256 !== 0) {
```

MathyFurret: It's not guaranteed to be >0 here, so do you prefer `hp > 0 && hp % 256 !== 0`?

Zarel: On reflection, yes, I would. I think `%`'s behavior at negatives is really bad and unintuitive, and I'm unhappy it's what all modern languages use.

Zarel:
Update: I talked to @DaWoblefet and he said he's fine bolding the text if the move would currently fail, which I think would make a variety of people really happy.

play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -761,6 +761,19 @@ class BattleTooltips {
...
+					hpValues.push(hp);
+					if (hp - 256 > 0) {
+						hpValues.push(hp - 256);
+					}
+				}
+				if (hpValues.length) text += `<p>Will fail if current HP is ${hpValues.join(' or ')}</p>`;
```
Zarel, line 775:
```suggestion
				let failMessage = hpValues.length ? `Will fail if current HP is ${hpValues.join(' or ')}.` : '';
				if (hpValues.includes(serverPokemon.hp) failMessage = `<strong class="message-error">${failMessage}</strong>`;
				if (failMessage) text += `<p>${failMessage}</p>;
```

Zarel:
lol I accidentally put Thanks into the commit message instead of a comment. Oh well you know what I mean P;

PR #2021, opened 2022-11-21
Check [from] effect for switches instead of lastMove

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9404420

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -821,14 +821,15 @@ export class Side {
 		return poke;
 	}
 
-	switchIn(pokemon: Pokemon, slot = pokemon.slot) {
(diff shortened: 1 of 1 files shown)


urkerab:
What about replays from before 2018 which never have a `[from]`?

Karthik99999:
> What about replays from before 2018 which never have a `[from]`?

I guess I can add a check for that and use `lastMove` if there is no `[from]` effect.

Karthik99999:
Actually that wouldn't work since the regular switches don't have an effect and then the thing I'm trying to fix wouldn't get fixed.

PR #2023, opened 2022-11-22
Add support for Revival Blessing

HisuianZoroark:
Client side portion of smogon/pokemon-showdown#9035

Needs a lot more testing, but it is worthwhile to review now.


HoeenCoder (commented):
Excellent work! Just waiting on server side now.

PR #2030, opened 2022-11-26
Update and add new background music

kittenchilly:
Since the audio folder isn't included on the repo, I included it here: https://cdn.discordapp.com/attachments/906654781471359076/1045895175655936010/audio.zip
Adds LGPE and SWSH music. Also updates the older BGM to use better sources, primarily smashcustommusic.net, which should reduce file size and increase quality


Marty-D:
Like I mentioned before, this isn't happening.

PR #2034, opened 2022-11-28
Support changing namecolors through the loginserver

mia-pi-git:
Approved by US.
I specifically routed this through the loginserver in order to have it write to the usermodlog (requested by US).
This reads/writes to a `config/colors.json` file that clients request on startup, and then assign to the existing colors (I specifically chose a new JSON file because editing the config directly programatically would be a nightmare).
I'll be setting up the server-side command code shortly.

### DIFF
--- a/lib/dispatcher.lib.php
+++ b/lib/dispatcher.lib.php
@@ -112,7 +112,7 @@ class ActionDispatcher {
 			}
 			return null;
 		} else {
-			$server =& $PokemonServers[$serverid];
+			$server = &$PokemonServers[$serverid];
 			if (empty($server['skipipcheck']) && empty($server['token']) && $serverid !== 'showdown') {
 				if (!isset($server['ipcache'])) {
 					$server['ipcache'] = gethostbyname($server['server']);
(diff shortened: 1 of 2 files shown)


lib/dispatcher.lib.php
```diff
@@ -112,7 +112,7 @@ public function findServer() {
 			}
 			return null;
 		} else {
-			$server =& $PokemonServers[$serverid];
+			$server = &$PokemonServers[$serverid];
```
AnnikaCodes, line 115: is this change necessary?

lib/dispatcher.lib.php
```diff
@@ -356,6 +356,65 @@ public function updateuserstats($dispatcher, &$reqData, &$out) {
 		$dispatcher->setPrefix(''); // No need for prefix since only usable by server.
 	}
 
+	public function updatenamecolor($dispatcher, &$reqData, &$out) {
```
AnnikaCodes, line 359: where is the permissions check for this?

mia-pi-git: The permissions check on this end validates that it comes from sim3. Sim3 itself can handle permissions checking, since only that server (including an IP check) can send said request.

lib/dispatcher.lib.php
mia-pi-git: Nope. Entirely just a nitpick I saw while I was digging around.

lib/dispatcher.lib.php
mia-pi-git: Yeah, it validates both the server token and the IP address the request is coming from.

lib/dispatcher.lib.php
mia-pi-git, line 115:
```suggestion
			$server = & $PokemonServers[$serverid];
```

PR #2036, opened 2022-11-30
Rage fist tooltip

aviettran:
Implementing: https://www.smogon.com/forums/threads/updating-the-power-of-moves-like-rage-fist-dynamically.3711423/

Hopefully I haven't missed any edge cases. I had to implement it differently than the server.

image

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -103,6 +103,7 @@ export class Pokemon implements PokemonDetails, PokemonHealth {
 	/** [[moveName, ppUsed]] */
 	moveTrack: [string, number][] = [];
 	statusData = {sleepTurns: 0, toxicTurns: 0};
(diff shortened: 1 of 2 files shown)


src/battle.ts
```diff
@@ -1662,6 +1663,7 @@ export class Battle {
 					this.scene.runStatusAnim('cursed' as ID, [poke]);
 					break;
 				case 'confusion':
+					poke.timesAttacked += 1;
```
Karthik99999, line 1666: confusion hits don't count for rage fist, so you can get rid of this

aviettran: Smogon says it does: https://www.smogon.com/dex/sv/moves/rage-fist/. Is Smogon incorrect?

PR #2043, opened 2022-12-09
Add CSS class based on field slot (left-center-right)

harasuke:
While I was working on a custom theme, I noticed that it was not possible to tell which pokemon was positioned to the right or left during a double battle. This made it impossible to properly edit the "statbar" via css.

### DIFF
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -2651,9 +2651,9 @@ export class PokemonSprite extends Sprite {
 		// DOUBLES: Slot0 -> left / Slot1 -> Right
 		// TRIPLES: slot0 -> left / Slot1 -> Center / Slot2 -> Right
 		const position = [
-			' left',
-			this.scene.activeCount === 3 ? ' center' : ' right',
-			' right',
+			' leftstatbar',
+			this.scene.activeCount === 3 ? ' centerstatbar' : ' rightstatbar',
+			' rightstatbar',
 		];
 		return position[slot];
 	}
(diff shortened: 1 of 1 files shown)


DaWoblefet:
How does this work for Triple Battles or Free-for-Alls?

harasuke:
I haven't tested it for Triple Battles since I don't know how to... I couldn't find any live triple battle match to test on.

FFA works correctly. The Player and the opponents are treated by the game like in double battles. Opponent1 / Opponent2 as opponent right / left respectively. Opponent 3 is treated as the player right pokemon.

Opponent1     Opponent2

Player1           Opponent3

harasuke:
Now it does support triple battles

src/battle-animations.ts
```diff
@@ -2647,8 +2647,19 @@ export class PokemonSprite extends Sprite {
 	// Statbar
 	/////////////////////////////////////////////////////////////////////
 
+	getClassForPosition(slot: number) {
+		// DOUBLES: Slot0 -> left / Slot1 -> Right
+		// TRIPLES: slot0 -> left / Slot1 -> Center / Slot2 -> Right
+		const position = [
+			' left',
+			this.scene.activeCount === 3 ? ' center' : ' right',
+			' right',
```
HoeenCoder, line 2656: Maybe we should be using a different class name for these, `left`, `right`, and `center` are pretty generic terms that might get used elsewhere by mistake in the future.

Possibly `leftstatbar` or something similar? I'm open to discussion on this.

harasuke: You are right, so in the end it should be like this:

```suggestion
			' leftstatbar',
			this.scene.activeCount === 3 ? ' centerstatbar' : ' rightstatbar',
			' rightstatbar',
```

PR #2050, opened 2022-12-12
Fix rage fist counter triggering from substitute

aviettran:
Not super familiar with client code, but I think this should cover things like sub, belly drum, etc..

https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-97#post-9431398

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -1686,8 +1686,8 @@ export class Battle {
 					break;
 				}
 			} else {
-				if (Dex.moves.get(this.lastMove).category !== "Status") {
-					poke.timesAttacked += 1;
+				if (this.dex.moves.get(this.lastMove).category !== 'Status') {
+					poke.timesAttacked++;
 				}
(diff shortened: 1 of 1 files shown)


src/battle.ts
```diff
@@ -1686,7 +1686,9 @@ export class Battle {
 					break;
 				}
 			} else {
-				poke.timesAttacked += 1;
+				if (Dex.moves.get(this.lastMove).category !== "Status") {
+					poke.timesAttacked += 1;
```
Karthik99999, line 1690:
```suggestion
					poke.timesAttacked++;
```

src/battle.ts
```diff
@@ -1686,7 +1686,9 @@ export class Battle {
 					break;
 				}
 			} else {
-				poke.timesAttacked += 1;
+				if (Dex.moves.get(this.lastMove).category !== "Status") {
```
Karthik99999, line 1689:
```suggestion
				if (this.dex.moves.get(this.lastMove).category !== 'Status') {
```

Karthik99999:
I think this is a pretty good solution

aviettran:
thanks

PR #2051, opened 2022-12-12
Add accessible names to team builder controls

sivakusayan:
First pull request here. I think this is straightforward but open to how things should be done differently 🙂 

### Change
Previously, screen readers would not announce the name of certain controls in the team builder. Now, those controls will have a name announced by the screen reader.

### Testing
I tested with JAWs/Chrome, NVDA/Chrome, JAWs/Firefox, NVDA/Firefox, and Voiceover/Safari.
1. Launch your screen reader of choice.
2. Launch your browser of choice. _(Order is important here! Some browsers only check if they should turn accessibility mode on during start up, so launching a screen reader after the browser launched might not work)_
4. Add a new pokemon.
5. Tab through the various input fields (Item, Ability, Moves, EVs, IVs, etc.) 
6. Make sure the screen reader announces the name of each input field. _(For example, focusing the Ability field has the screen reader announce that it is an "Ability Input Field", rather than a nameless "Input Field". Verbiage can differ slightly between screen readers, and this is okay as long as the general idea is communicated.)_

### Out of Scope
This patch doesn't take internationalization into account.
This seems to [be a work in progress](https://www.smogon.com/forums/threads/translations.3650637/), so this patch will only work for English players.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -1181,7 +1181,7 @@
 			}
 			buf += '<div class="setmenu"><button name="copySet"><i class="fa fa-files-o"></i>Copy</button> <button name="importSet"><i class="fa fa-upload"></i>Import/Export</button> <button name="moveSet"><i class="fa fa-arrows"></i>Move</button> <button name="deleteSet"><i class="fa fa-trash"></i>Delete</button></div>';
 			buf += '<div class="setchart-nickname">';
-			buf += '<label>Nickname</label><input type="text" name="nickname" class="textbox" value="' + BattleLog.escapeHTML(set.name || '') + '" placeholder="' + BattleLog.escapeHTML(species.baseSpecies) + '" />';
+			buf += '<label for="nickname">Nickname</label><input type="text" id="nickname" name="nickname" class="textbox" value="' + BattleLog.escapeHTML(set.name || '') + '" placeholder="' + BattleLog.escapeHTML(species.baseSpecies) + '" />';
 			buf += '</div>';
 			buf += '<div class="setchart" style="' + Dex.getTeambuilderSprite(set, this.curTeam.gen) + ';">';
 
@@ -1192,7 +1192,7 @@
 			} else {
 				buf += '<div class="setcell-sprite"></div>';
 			}
-			buf += '<div class="setcell setcell-pokemon"><label>Pok&eacute;mon</label><input type="text" name="pokemon" class="textbox chartinput" value="' + BattleLog.escapeHTML(set.species) + '" autocomplete="off" /></div></div>';
+			buf += '<div class="setcell setcell-pokemon"><label for="pokemon">Pok&eacute;mon</label><input type="text" id="pokemon" name="pokemon" class="textbox chartinput" value="' + BattleLog.escapeHTML(set.species) + '" autocomplete="off" /></div></div>';
 
 			// details
 			buf += '<div class="setcol setcol-details"><div class="setrow">';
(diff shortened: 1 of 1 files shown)


js/client-teambuilder.js
```diff
@@ -1181,7 +1181,7 @@
 			}
 			buf += '<div class="setmenu"><button name="copySet"><i class="fa fa-files-o"></i>Copy</button> <button name="importSet"><i class="fa fa-upload"></i>Import/Export</button> <button name="moveSet"><i class="fa fa-arrows"></i>Move</button> <button name="deleteSet"><i class="fa fa-trash"></i>Delete</button></div>';
 			buf += '<div class="setchart-nickname">';
-			buf += '<label>Nickname</label><input type="text" name="nickname" class="textbox" value="' + BattleLog.escapeHTML(set.name || '') + '" placeholder="' + BattleLog.escapeHTML(species.baseSpecies) + '" />';
+			buf += '<label for="nickname">Nickname</label><input type="text" id="nickname" name="nickname" class="textbox" value="' + BattleLog.escapeHTML(set.name || '') + '" placeholder="' + BattleLog.escapeHTML(species.baseSpecies) + '" />';
```
urkerab, line 1184: You're using `for`/`id` because implicit labels don't work in Voiceover/Safari or for some other reason? (Just wondering.)

sivakusayan: Yeah, implicit labels don't work in any browser actually. These relationships need to be made explicit so browsers know what name to expose to the screen reader (or whatever assistive technology is being used).

mia-pi-git: Maybe we should modify our Caja hooks so that all labels have their `for` values set to the inner content of the label? Would make a lot more stuff accessible without having to change every label.

sivakusayan: That would work, although each `<input />` would still need an `id` attribute since `for` only accepts IDs. So maybe the Caja hooks are also modified so that all `<input />` are given an `id` equal to their `name`, if it exists. Two questions if that's okay:

- Is it a fair assumption in this codebase that the `name` attribute for an input is unique, since the `id` needs to be unique? If so this approach would work, otherwise inputs will still need a manual `id` given.
- Can you point me to where the Caja hooks are? A naive search for "caja" in the repo didn't seem to point me to anything useful (besides `battle-log.ts` which I assume is specific to the battle-log)

mia-pi-git: > Is it a fair assumption in this codebase that the name attribute for an input is unique, since the id needs to be unique? If so this approach would work, otherwise inputs will still need a manual id given.

No, PS's element names are generally not unique. We frequently use names for event handlers, so there can (for example) be several buttons named "send" or "receive". Inputs are more often unique, but it's not a given (especially in server pages)

> Can you point me to where the Caja hooks are? A naive search for "caja" in the repo didn't seem to point me to anything useful (besides battle-log.ts which I assume is specific to the battle-log)

https://github.com/smogon/pokemon-showdown-client/blob/master/src/battle-log.ts#L778

sivakusayan: Unless I'm missing something, I don't think the teambuilder HTML goes through Caja. Breakpoints placed in `BattleLog.tagPolicy` don't seem to get hit when opening the team builder. Did we want to start calling `BattleLog.sanitizeHTML` on the teambuilder HTML? I think that feels heavy-handed for strings that aren't user input.

If we wanted to reduce duplication maybe it makes sense to have some function like: 
```
generateInputLabelHTML({ 
    name: string, 
    value?: string, 
    placeholder?: string, 
    autocomplete?: bool 
})
```
and replacing those strings with calls to `generateInputLabelHTML`, but I wasn't sure if that change would be too much for a small fix like this.

PR #2053, opened 2022-12-13
Support viewing Tera Types for inactive Pokemon

Karthik99999:
Server Side: smogon/pokemon-showdown#9168

### DIFF
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -2748,7 +2748,7 @@ export class PokemonSprite extends Sprite {
 			status += '<span class="frz">FRZ</span> ';
 		}
 		if (pokemon.terastallized) {
-			status += '<img src="' + Dex.resourcePrefix + 'sprites/types/' + encodeURIComponent(pokemon.terastallized) + '.png" alt="' + pokemon.terastallized + '" class="pixelated" /> ';
+			status += `<img src="${Dex.resourcePrefix}sprites/types/${encodeURIComponent(pokemon.terastallized)}.png" alt="${pokemon.terastallized}" class="pixelated" /> `;
 		} else if (pokemon.volatiles.typechange && pokemon.volatiles.typechange[1]) {
 			const types = pokemon.volatiles.typechange[1].split('/');
 			for (const type of types) {
--- a/src/battle-animations.ts
+++ b/src/battle-animations.ts
@@ -2670,8 +2670,8 @@ export class PokemonSprite extends Sprite {
 		if (symbol) {
 			buf += ` <img src="${Dex.resourcePrefix}sprites/misc/${symbol}.png" alt="${symbol}" style="vertical-align:text-bottom;" />`;
 		}
-		if (pokemon.teraType) {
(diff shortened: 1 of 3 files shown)


src/battle-tooltips.ts
```diff
@@ -795,12 +795,19 @@ class BattleTooltips {
...
-			if (clientPokemon && (clientPokemon.volatiles.typechange || clientPokemon.volatiles.typeadd)) {
+			if (pokemon.terastallized) {
+				text += `<small>(Terastallized)</small><br />`;
+			} else if (clientPokemon?.volatiles.typechange || clientPokemon?.volatiles.typeadd) {
 				text += `<small>(Type changed)</small><br />`;
 			}
 			text += types.map(type => Dex.getTypeIcon(type)).join(' ');
+			if (pokemon.terastallized) {
+				text += ` <small>(base: ${this.getPokemonTypes(pokemon, true).map(type => Dex.getTypeIcon(type)).join(' ')})</small>`;
+			} else if (serverPokemon) {
```
urkerab, line 808: `serverPokemon?.teraType`? (I'm assuming here you're only sending the Tera type for Gen 9 battles.)

src/battle.ts
```diff
@@ -994,6 +996,8 @@ export interface ServerPokemon extends PokemonDetails, PokemonHealth {
 	pokeball: string;
 	/** false if the pokemon cannot gigantamax, otherwise a string containing the full name of its G-max move */
 	gigantamax: string | false;
+	teraType: string;
+	terastallized: boolean;
```
Marty-D, line 1000: I find it pretty confusing to have `terastallized` defined as a boolean here but in other contexts it's a string. Is it possible to rename one of them?

Karthik99999: I was thinking that as well, just thought it would be pointless to send the type twice for both properties when a mon is terastallized. I could just have the server send over the type instead of a boolean if that's preferred.

Marty-D: The boolean could be sent from the server with a name other than "teraType" or "terastallized"? Or, have the `terastallized` that's an actual type name which is being repurposed from `teraType` renamed to "terastallizedType" or something.

Karthik99999: I went with just having `terastallized` be a string that contains the tera type if the mon is terastallized (and an empty string if it isn't)

src/battle-animations.ts
```diff
@@ -2747,7 +2747,9 @@ export class PokemonSprite extends Sprite {
 		} else if (pokemon.status === 'frz') {
 			status += '<span class="frz">FRZ</span> ';
 		}
-		if (pokemon.volatiles.typechange && pokemon.volatiles.typechange[1]) {
+		if (pokemon.terastallized) {
+			status += '<img src="' + Dex.resourcePrefix + 'sprites/types/' + encodeURIComponent(pokemon.terastallized) + '.png" alt="' + pokemon.terastallized + '" class="pixelated" /> ';
```
AnnikaCodes, line 2751:
```suggestion
			status += `<img src="${Dex.resourcePrefix}sprites/types/${encodeURIComponent(pokemon.terastallized)}.png" alt="${pokemon.terastallized}" class="pixelated" /> `;
```

AnnikaCodes:
Have you tested this?

Karthik99999:
> Have you tested this?

Yep, here are some screenshots

image
image

AnnikaCodes:
OK, let's do this.

PR #2059, opened 2022-12-16
Update formats with public inputlog

adrivrie:
The check of whether a format should allow the inputlog to be public was outdated, I tried to make it up-to-date and generalize it a little by removing gens and combining BF and BSSF.

Almost entirely done by pattern matching, sorry if I did something wrong.

### DIFF
--- a/replays/battle.log.php
+++ b/replays/battle.log.php
@@ -48,7 +48,7 @@ if ($replay['password'] ?? null) {
 $replay['log'] = str_replace("\r","",$replay['log']);
 
 if ($replay['inputlog']) {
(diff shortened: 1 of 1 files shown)


adrivrie:
A better fix would be to check if `"random"`, `"challengecup"`, `"hackmonscup"`, `"factory"` or `"computergenerated"` is contained in the `formatid`, but I'm too scared to break something...

Best fix would be to check whether the format requires a team, but I have no idea how to do this.

In the meantime, updating a little would make debugging a little easier.

NotTsunami:
> A better fix would be to check if `"random"`, `"challengecup"`, `"hackmonscup"`, `"factory"` or `"computergenerated"` is contained in the `formatid`, but I'm too scared to break something...

PHP 8 offers [str_contains](https://www.php.net/manual/en/function.str-contains.php) if this project can use PHP 8 features (unsure of any restrictions or much of the internals, very infrequent contributor).

PR #2064, opened 2022-12-22
Update to handle esbuild migration

mia-pi-git:
Will get this merged when it's time for esbuild to go live.


AnnikaCodes:
Thanks!

PR #2073, opened 2022-12-30
Add a link to the staff list

KennedyLFC:
[very serious comment]

The text in the first paragraph says "A full staff list can be found below." but there was no staff list below lmao.

### DIFF
--- a/website/pages/staff.md
+++ b/website/pages/staff.md
@@ -6,7 +6,7 @@ Staff are represented by the symbols next to their names and appear at the top o
 
 ### What's the difference between room staff and global staff?
 
-Room staff have permissions tied specifically to the room they were promoted in, whereas global staff have increased permissions across the server. Room staff can only perform moderation duties within their rooms, whereas global staff have the ability to globally lock or ban users and thus should be sought instead of room staff for PM or battle abuse. Global staff may also hold higher positions as room staff. Global staff can be identified by the "Global" indicator when clicking on their username.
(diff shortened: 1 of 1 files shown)


website/pages/staff.md
```diff
@@ -6,7 +6,7 @@ Staff are represented by the symbols next to their names and appear at the top o
 
 ### What's the difference between room staff and global staff?
 
-Room staff have permissions tied specifically to the room they were promoted in, whereas global staff have increased permissions across the server. Room staff can only perform moderation duties within their rooms, whereas global staff have the ability to globally lock or ban users and thus should be sought instead of room staff for PM or battle abuse. Global staff may also hold higher positions as room staff. Global staff can be identified by the "Global" indicator when clicking on their username.
+Room staff have permissions tied specifically to the room they were promoted in, whereas global staff have increased permissions across the server. Room staff can only perform moderation duties within their rooms, whereas global staff have the ability to globally lock or ban users and thus should be sought instead of room staff for PM or battle abuse. Global staff may also hold higher positions as room staff. Global staff can be identified by the "Global" indicator when clicking on their username, you can also refer to the [staff list](https://www.smogon.com/sim/staff_list) to see all Global Staff members.
```
NotTsunami, line 9: It would make more sense to add this to the comment question above.

KennedyLFC: I think it makes sense in the second paragraph after the distinction is made between global staff and room staff. If anything I'd just remove the sentence in the first paragraph that refers to the staff list tbh

NotTsunami: I don't think it does. I'm not trying to be overly pedantic, but in the question that asks who the staff are, it makes significantly more sense to link there.

KrisXV:
```suggestion
Room staff have permissions tied specifically to the room they were promoted in, whereas global staff have increased permissions across the server. Room staff can only perform moderation duties within their rooms, whereas global staff have the ability to globally lock or ban users and thus should be sought instead of room staff for PM or battle abuse. Global staff may also hold higher positions as room staff. Global staff can be identified by the "Global" indicator when clicking on their username. You can also refer to the [staff list](https://www.smogon.com/sim/staff_list) to see all Global Staff members. 
```

AnnikaCodes:
This is fine, I suppose. Sorry that you had wait half a year for it to be merged.

PR #2074, opened 2022-12-30
Build: Fix compiled TS file count

Karthik99999:


monsanto:
What is the problem, and how does this fix it?

Zarel:
This seems to make it so multiple `combineResults` calls can run concurrently, which I believe could improve performance, but I don't believe it fixes the compiled file count at all. I also don't think we have performance issues?

Karthik99999:
Currently, the file count looks something like this

```
(0[object Promise][object Promise] files in 16.69s) DONE
```

This happens because `compileToFile` is an async function, but the return value from `compileToFile` is being treated as if it is already a number and not a promise to a number. I could have either resolved the promises where `compileToFile` is called or changed it to a sync function and only make one `.then` call, which is what I believed look cleaner (and does actually fix the display issue above).

Zarel:
Oh, good to know; thanks!

PR #2078, opened 2023-01-06
Bypass https on localhost

mia-pi-git:
This fixes accessing localhost. Previously, browsers forced http to https, meaning the WebSocket connection was made to port 443 on https instead of the correct http connection to whatever port. This ensures localhost is always connecting to http and on the correct port, which is added to the querystring by the redirect in `ps/server/static/index.html`.

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -758,7 +758,7 @@ function toId() {
 						// We need to bypass the port as well because on most modern browsers, http gets forced
 						// to https, which means a ws connection is made to port 443 instead of wherever it's actually running,
 						// thus ensuring a failed connection.
-						var port = possiblePort ? possiblePort[1] : Config.server.port;
+						var port = possiblePort || Config.server.port;
 						console.log("Bypassing SockJS for localhost");
 						var url = 'ws://' + Config.server.host + ':' + port + Config.sockjsprefix + '/websocket';
 						console.log(url);
--- a/js/client.js
+++ b/js/client.js
@@ -393,6 +393,7 @@ function toId() {
 		},
 		focused: true,
 		initialize: function () {
+			this.query = window.location.search;
 			window.app = this;
 			this.initializeRooms();
 			this.initializePopups();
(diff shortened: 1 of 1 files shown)


js/client.js
```diff
@@ -753,11 +754,15 @@ function toId() {
...
+						// to https, which means a ws connection is made to port 443 instead of wherever it's actually running,
+						// thus ensuring a failed connection.
+						var port = possiblePort ? possiblePort[1] : Config.server.port;
 						console.log("Bypassing SockJS for localhost");
-						console.log('ws' + protocol.slice('4') + '://' + Config.server.host + ':' + Config.server.port + Config.sockjsprefix + '/websocket');
-						return new WebSocket(
-							'ws' + protocol.slice('4') + '://' + Config.server.host + ':' + Config.server.port + Config.sockjsprefix + '/websocket'
-						);
+						var url = 'ws://' + Config.server.host + ':' + port + Config.sockjsprefix + '/websocket';
+						console.log(url);
```
DaWoblefet, line 764: Extra debug print

mia-pi-git: This was already there - I just reformatted it so it was more consistent.

js/client.js
```diff
@@ -753,11 +754,15 @@ function toId() {
 						// anyway, this affects SockJS because it makes HTTP requests to localhost
 						// but it turns out that making direct WebSocket connections to localhost is
 						// still supported, so we'll just bypass SockJS and use WebSocket directly.
+						var possiblePort = /port=([0-9]+)/.exec(self.query || "");
```
monsanto, line 757: https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams#examples

mia-pi-git: OK. Did.

js/client.js
```diff
@@ -753,11 +754,15 @@ function toId() {
 						// anyway, this affects SockJS because it makes HTTP requests to localhost
 						// but it turns out that making direct WebSocket connections to localhost is
 						// still supported, so we'll just bypass SockJS and use WebSocket directly.
+						var possiblePort = new URL(document.location + self.query).searchParams.get('port');
+						// We need to bypass the port as well because on most modern browsers, http gets forced
+						// to https, which means a ws connection is made to port 443 instead of wherever it's actually running,
+						// thus ensuring a failed connection.
+						var port = possiblePort ? possiblePort[1] : Config.server.port;
```
monsanto, line 761: I think you forgot to update this after the searchParams change

mia-pi-git: Yeah. 
```suggestion
						var port = possiblePort || Config.server.port;
```

js/client.js
```diff
@@ -753,11 +754,15 @@ function toId() {
 						// anyway, this affects SockJS because it makes HTTP requests to localhost
 						// but it turns out that making direct WebSocket connections to localhost is
 						// still supported, so we'll just bypass SockJS and use WebSocket directly.
+						var possiblePort = new URL(document.location + self.query).searchParams.get('port');
```
monsanto, line 757: The example suggests that it isn't necessary to append `window.location.search` to `document.location`. Also this would make `this.query` above a dead variable.

mia-pi-git: `this.query` is necessary since the querystring gets removed from the URL early on (I think backbone does it?) - as such, it has to be cached as soon as possible.

monsanto: Oh I see, maybe a comment would help, unless this is documented elsewhere.

mia-pi-git: Sure. Done.

monsanto:
It looks OK to me, so if you've tested it then feel free to merge.

PR #2079, opened 2023-01-06
Tournaments: Support team locking

MathyFurret:
Client side for https://github.com/smogon/pokemon-showdown/pull/9283

This adds a warning for team-locked tournaments, and hides the team select after a player's team is locked.

### DIFF
--- a/js/client-chat-tournament.js
+++ b/js/client-chat-tournament.js
@@ -97,6 +97,8 @@
 					'<div class="tournament-tools">' +
 						'<div class="tournament-team"></div>' +
 						'<button class="button tournament-join">Join</button><button class="button tournament-validate"><i class="fa fa-check"></i> Validate</button> <button class="button tournament-leave">Leave</button>' +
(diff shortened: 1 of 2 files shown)


js/client-chat-tournament.js
```diff
@@ -357,6 +375,17 @@
 					}
 					break;
 
+				case 'teamlock':
+					if (data[0] === 'on') {
+						this.room.$chat.append("<div class=\"notice tournament-message-teamlock\">This tournament is now team-locked (Tournament players can't change teams between games)</div>");
+						this.teamLocked = true;
+					} else if (data[0] === 'off') {
+						this.room.$chat.append("<div class=\"notice tournament-message-teamlock\">This tournament is no longer team-locked.</div>");
```
AnnikaCodes, line 383: Why does one of these have a period, but not the other?

MathyFurret: Because other lines that end in parentheticals don't have a period while the rest do

PR #2085, opened 2023-01-11
Tooltips: Align tera reminders with types, Space

qwertystrargs:
The current- and modified tooltips, from left to right, respectively:
image

### DIFF
--- a/style/battle.css
+++ b/style/battle.css
@@ -789,8 +789,8 @@ License: GPLv2
 #tooltipwrapper .tooltip p.section {
 	border-top: 1px solid #888888;
 }
-#tooltipwrapper .tooltip small.type-aligned {
-	vertical-align: 3px;
+#tooltipwrapper .tooltip span.textaligned-typeicons {
+	vertical-align: -3px;
 }
 #tooltipwrapper.tooltip-locked {
 	pointer-events: auto;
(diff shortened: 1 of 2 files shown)


qwertystrargs:
Fixed. I assumed there was an issue with introducing space under the type-state reminder.

qwertystrargs:
Results (previous version on the left):
image

style/battle.css
```diff
@@ -789,6 +789,9 @@ License: GPLv2
 #tooltipwrapper .tooltip p.section {
 	border-top: 1px solid #888888;
 }
+#tooltipwrapper .tooltip span.textaligned-typeicons {
+	vertical-align: -3px;
+}
```
urkerab, line 794: Rather than hard-coding an adjustment on the wrapper, wouldn't it be better to add a class to the type icons themselves so that you could style them with `vertical-align: text-bottom;` or `middle` (not sure which one would be better)?

urkerab: Although the type icon in the status indicator gets away without using a class on the icon itself, so maybe just leave the span in and style the icon using
```suggestion
#tooltipwrapper .tooltip .textaligned-typeicons img {
	vertical-align: text-bottom;
}
```

qwertystrargs:
With text-bottom:
image

Marty-D:
Looks good to me, thanks!

PR #2086, opened 2023-01-12
Tooltips: Hide Tera Type for Terastal Clause

MathyFurret:
so annoyed seeing it in cc1v1


Zrp200:
Wouldn't it have been better if it just hid tera type for those that can't tera?

PR #2092, opened 2023-01-19
Teambuilder: Don't modify IVs for Hidden Power on import

EvGym:
addresses my issue here: https://github.com/smogon/pokemon-showdown-client/issues/2091

I wanted to see if I could make it depend on the format of the team it was being imported to but it doesn't appear that the areas of code responsible for this issue have access to that information. This seems like the best fix I could find. This shouldn't pose a huge issue to those using earlier formats for the following reasons:
1) Teams they are importing were probably already created in the proper gen of showdown and won't need the extra correction
2) validation will catch the issue and re-adding the hidden power type will fix the stats.
3) you have always been able to import illegal sets so this change won't catch anyone off-guard. If people even realize it is gone.

### DIFF
--- a/js/storage.js
+++ b/js/storage.js
@@ -1246,13 +1246,6 @@ Storage.importTeam = function (buffer, teams) {
 			if (line.substr(0, 14) === 'Hidden Power [') {
 				var hptype = line.substr(14, line.length - 15);
 				line = 'Hidden Power ' + hptype;
-				var type = Dex.types.get(hptype);
-				if (!curSet.ivs && type) {
-					curSet.ivs = {};
-					for (var stat in type.HPivs) {
-						curSet.ivs[stat] = type.HPivs[stat];
-					}
-				}
 			}
 			if (line === 'Frustration' && curSet.happiness === undefined) {
 				curSet.happiness = 0;
--- a/src/panel-teamdropdown.tsx
+++ b/src/panel-teamdropdown.tsx
@@ -398,13 +398,6 @@ class PSTeambuilder {
 			if (line.startsWith('Hidden Power [')) {
 				const hpType = line.slice(14, -1) as TypeName;
 				line = 'Hidden Power ' + hpType;
-				if (!set.ivs && Dex.types.isName(hpType)) {
-					set.ivs = {hp: 31, atk: 31, def: 31, spa: 31, spd: 31, spe: 31};
-					const hpIVs = Dex.types.get(hpType).HPivs || {};
-					for (let stat in hpIVs) {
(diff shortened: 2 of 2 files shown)


AnnikaCodes:
Shouldn't this still adjust the IVs for formats where Hyper Training isn't available? Also, did you make a thread on the [PS Suggestions forum](https://www.smogon.com/forums/forums/suggestions.517/) to allow for community input on the change?

EvGym:
> 

I don't think so. This only applies for importing teams and how often will you be importing teams from higher gens into earlier gens? If you add the hidden power in the format you will get the required IV modification and the IV modifications are listed when you export a team. In the ends, it seems a rather pointless feature that will cause issues 99 out of 100 times it comes into effect

EvGym:
https://www.smogon.com/forums/threads/default-to-31-ivs-on-pokemon-with-hidden-power-gen-7-and-national-dex.3692969/

It has been brought up before, but I did not want to necro post it. I went ahead and made the fix as it has been effecting some of my friends

AnnikaCodes:
Since PS doesn't specify generation in the team text and previously exported teams may rely on the automatic IV adjustment, we unfortunately cannot accept this change.

EvGym:
> Since PS doesn't specify generation in the team text and previously exported teams may rely on the automatic IV adjustment, we unfortunately cannot accept this change.

As every future generation has hyper training, this feels really silly. As I explained above, the ONLY time this matters would be if you are importing a team from a future generation or if you built the team outside the teambuilder someone. Even then, this is the work of the team validator. At no point should importing a team modify what you are trying to import. That is an inherently confusing and misleading process that is only going to cause issues.

If we are going to default to anything, why are we defaulting to previous generations and not the modern ones?

HoeenCoder:
It may be better to re-attempt this by wrapping the hidden power logic in a generational check eg:
```js
if (!hyperTrainingSupported) {
  // adjust IVs
}
```
(the above is just psudeocode to be clear). Assume the current generation if its not specified for the team.

EvGym:
If we can get that, it would be great. The teambuilder appears to be agnostic to the format you are currently important for so I do not think this is possible.

You have been able to import illegal sets into the validator for a very long time. Heck, you can import mon that do not exist. Instead of trying to give the importer half enforcement powers, I think we should make it consistent with the rest of its functionality and let the validator yell at the player for having wrong IVs.

We will need to change it so that the proper IVs are exported for mon hidden powers as it appears it cuts that out at the moment.

EvGym:
I did some testing and my fix also makes it so it properly exports the IVs necessary for the hidden powers.

PR #2095, opened 2023-01-21
Add revealed Tera Types to battle tooltips

vivapuffs:
This code depends on Marty's Tera Type preview code for the server which can be found here: https://discord.com/channels/630837856075513856/906654781471359076/1066202399225688185Z
edit: now depends on https://github.com/smogon/pokemon-showdown/pull/9350

### DIFF
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -777,9 +777,9 @@ class BattleTooltips {
 		const gender = pokemon.gender;
 
 		if (this.battle.rules['Tera Type Preview']) {
-			const substring = pokemon.details.match(/.*teratype:(?:*)/);
-			if (substring != null) {
-				revealedTeraType = substring[0];
+			const splitDetails = pokemon.details.split(':');
+			if (splitDetails != null) {
+				revealedTeraType = splitDetails[splitDetails.length - 1];
 			}
 		}
 		if (gender === 'M' || gender === 'F') {
--- a/src/battle-tooltips.ts
+++ b/src/battle-tooltips.ts
@@ -821,7 +821,8 @@ class BattleTooltips {
 
 			if (pokemon.terastallized) {
 				text += `&nbsp; &nbsp; <small>(base: <span class="textaligned-typeicons">${this.getPokemonTypes(pokemon, true).map(type => Dex.getTypeIcon(type)).join(' ')}</span>)</small>`;
-			} else if (serverPokemon?.teraType && !this.battle.rules['Terastal Clause']) {
+			} else if (serverPokemon?.teraType && !this.battle.rules['Terastal Clause']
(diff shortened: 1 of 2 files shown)


src/battle-tooltips.ts
```diff
@@ -773,7 +773,15 @@ class BattleTooltips {
 		const pokemon = clientPokemon || serverPokemon!;
 		let text = '';
 		let genderBuf = '';
+		let revealedTeraType = '';
 		const gender = pokemon.gender;
+
+		if (this.battle.rules['Tera Type Preview']) {
+			const substring = pokemon.details.match(/(?<=teratype\:).*/g);
```
urkerab, line 780: I don't believe all clients support lookbehinds yet. (Using a capture group might even be more readable anyway.) Also you don't need the `g` flag, do you?

src/battle-tooltips.ts
```diff
@@ -809,6 +817,16 @@ class BattleTooltips {
 				text += `&nbsp; &nbsp; <small>(Tera Type: <span class="textaligned-typeicons">${Dex.getTypeIcon(serverPokemon.teraType)}</span>)</small>`;
 			}
 			text += `</h2>`;
+
+			if (!pokemon.terastallized && this.battle.rules['Tera Type Preview']) {
+				if (revealedTeraType !== '' || serverPokemon?.teraType) {
+					let displayedTeraType = revealedTeraType;
+					if (serverPokemon?.teraType) {
+						displayedTeraType = serverPokemon.teraType;
+					}
```
urkerab, line 826:
```suggestion
				let displayedTeraType = serverPokemon?.teraType || revealedTeraType;
				if (displayedTeraType) {
```

src/battle-tooltips.ts
vivapuffs: hmm i dont think so since details is just one string right? if i am honest i am super unfamiliar with regex so i just went with the first thing that worked

src/battle-tooltips.ts
```diff
@@ -777,7 +777,7 @@ class BattleTooltips {
 		const gender = pokemon.gender;
 
 		if (this.battle.rules['Tera Type Preview']) {
-			const substring = pokemon.details.match(/(?<=teratype\:).*/g);
+			const substring = pokemon.details.match(/.*teratype:(?:*)/);
```
vivapuffs: this does not work correctly

MathyFurret:
The `teratype:` part is only sent once by the server, so you should just save it the first time and remove it from the details string instead of handling it every time the Pokemon is sent out.

qwertystrargs:
If adding a `teraType` property to the Pokemon class to address Mathy's suggestion, would it be appropriate to keep the revealed type separate from the details string when sending, as with `'mail'` and `'item'` in gen 4/5/7 team preview mods?

PR #2097, opened 2023-01-27
Tooltips: Fix Steely Spirit to boost ally moves

MathyFurret:


src/battle-tooltips.ts
```diff
@@ -1912,6 +1909,10 @@ class BattleTooltips {
 					if (ally !== pokemon) {
 						value.modify(1.3, 'Power Spot');
 					}
+				} else if (allyAbility === 'Steely Spirit') {
+					if (move.type === 'Steel') {
+						value.modify(1.5, 'Steely Spirit');
+					}
```
Karthik99999, line 1915:
```suggestion
				} else if (moveType === 'Steel' && allyAbility === 'Steely Spirit') {
					value.modify(1.5, 'Steely Spirit');
```

MathyFurret: several of the surrounding ifs are nested , i will just go ahead and fix those as well (in the morning im so tired)

Karthik99999:
btw you are still using `move.type` instead of `moveType` (which was the thing I was actually trying to fix with my suggestion lol)

PR #2100, opened 2023-02-23
Palafin Hero Cry on switch-in

JagSamurott:
Upon switching or fainting, the cry of hero palafin will now play, but still nothing plays during terastallization


JagSamurott:
trying to look into the tera cry issue

PR #2113, opened 2023-04-15
Dark Mode: Use dark grey and blue instead of white and orange

PartMan7:
See #chat-dev on Discord for screenshots

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -3433,10 +3433,61 @@ a.ilink.yours {
 
 /* main menu */
 
+.dark .pm-window h3,
+.dark .pm-window button {
+	color: #BBB;
+}
+
+.dark .pm-window .pm-buttonbar button {
+	color: #DDD;
+}
+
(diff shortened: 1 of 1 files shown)


AnnikaCodes:
I looked at [the discussion on Discord](https://discord.com/channels/630837856075513856/906654781471359076/1098447528577466401), and it seems to me that Zarel would prefer to keep the orange for highlighted PM boxes, and change the dark mode room highlights to orange instead. Can you change this PR to do that? I agree that it'd be good to be consistent.

PartMan7:
My takeaway from the discussion was that there were varying opinions about it and that it'd be best to have a poll in <<lobby>> for a consensus; I'll see if I can make mockups when I get time off work if that's still the case.

Zarel:
So I did a dark mode overhaul but it didn't touch PMs here very much. If you update this PR, I'd still have use for it.

PR #2118, opened 2023-05-10
Highlight timer messages below 10 seconds

dingdongg:
[Link to suggestion thread in forum](https://www.smogon.com/forums/threads/embolden-timer-warnings-when-they-dip-below-10-seconds.3700824/)

This PR adds yellow highlighting to timer messages to make them more visually obvious, which should help indicate the urgency of timer messages at lower values.

My implementation makes it so that, if a player has not made a move for 30 seconds in the current turn AND the timer value is <= 10, all subsequent timer messages for that user are highlighted with a yellow backdrop.


DaWoblefet:
We should already have a highlight class (should just be "highlighted"). It'd be good to use the same yellow for consistency. I also think it should only appear highlighted for the player low on time.

Do we want this to also create a browser notification, like how we do it for tournaments / start of battles? I'm not sure how much of that was part of the suggestion, but as it stands this is only a visual indicator if you're in the current tab / battle.

dingdongg:
Ahh I see, I'll use the existing class instead then.

> I also think it should only appear highlighted for the player low on time.

I bumped up the timer values to speed up development so ignore that, but did you mean something like this?

Screenshot 2023-05-09 at 8 55 45 PM

Screenshot 2023-05-09 at 8 55 53 PM

> Do we want this to also create a browser notification, like how we do it for tournaments / start of battles?

That's a good idea, I didn't think of it but it should be super helpful for people that are navigated away from the tab. I added a call to the `app.rooms[roomid].notifyOnce()` function, but I couldn't actually get the notifications to pop up on the `testclient.html` file. I know it's not an issue with my browser settings because they pop up for things like DMs and challenges on `play.pokemonshowdown.com`. Are notifications not supposed to work locally or am I missing something else? Thanks!

dingdongg:
Hey @DaWoblefet, when you have the time could you take another look at this?

src/battle-log.ts
```diff
@@ -210,6 +210,17 @@ export class BattleLog {
 
 		case 'error': case 'inactive': case 'inactiveoff':
 			divClass = 'chat message-error';
+			battle = this.scene?.battle;
+
+			if (battle?.kickingInactive && battle.timerMessageHighlighted) {
+				divClass += ' highlighted';
+				if (this.scene?.battle?.roomid) {
+					const title = 'Make your move!';
+					const body = `${battle?.kickingInactive || 'Only a few'} seconds remaining!`;
```
AnnikaCodes, line 219: You check `battle?.kickingInactive` in the if statement, so it should always be truthy. Is the "Only a few" fallback just to satisfy TypeScript, or is there a situation where it would be used that I'm not seeing?

AnnikaCodes (commented):
This looks pretty good to me; have you tested it locally?

src/battle.ts
```diff
@@ -1085,6 +1085,9 @@ export class Battle {
 	endLastTurnPending = false;
 	totalTimeLeft = 0;
 	graceTimeLeft = 0;
+	currTurnTimeElapsed = 0;
+	timerMessageHighlighted = false;
```
Zarel, line 1089: I don't think this needs two new properties on Battle? Like we're already displaying "User has X seconds left", we can just check how many seconds are left and use that to decide how pushy to make the message, right?

PR #2127, opened 2023-06-16
Accept pre-game info about individual Pokemon

qwertystrargs:
Updates Pokemon with data from rules like Open Team Sheets or Tera Type Preview.
This is dependent on #2131 and #9605

Related suggestions:
https://www.smogon.com/forums/threads/auto-adding-moveset-tera-hovering-on-pokemon-when-accepting-team-sheets.3718545/

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -91,6 +91,7 @@ export class Pokemon implements PokemonDetails, PokemonHealth {
 	prevItem = '';
 	prevItemEffect = '';
 	terastallized: string | '' = '';
+	teraType = '';
 
 	boosts: {[stat: string]: number} = {};
 	status: StatusName | 'tox' | '' | '???' = '';
@@ -358,6 +359,18 @@ export class Pokemon implements PokemonDetails, PokemonHealth {
 		this.ability = ability;
 		if (!this.baseAbility && !isNotBase) {
 			this.baseAbility = ability;
+		}
+	}
+	rememberSetData(kwArgs: KWArgs) {
+		if (!kwArgs) return;
+		if (kwArgs.name) this.name = kwArgs.name;
+		if (kwArgs.item) this.item = Dex.items.get(kwArgs.item).name;
+		if (kwArgs.ability) this.rememberAbility(kwArgs.ability);
+		if (kwArgs.teraType) this.teraType = Dex.types.get(kwArgs.teraType).name;
+		if (kwArgs.moveList) {
+			for (const moveName of kwArgs.moveList.split(`;`)) {
+				this.rememberMove(moveName, 0);
+			}
 		}
 	}
 	getBoost(boostStat: BoostStatName) {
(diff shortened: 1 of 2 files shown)


src/battle.ts
```diff
@@ -3539,10 +3546,15 @@ export class Battle {
 		}
 		case 'poke': {
 			let pokemon = this.rememberTeamPreviewPokemon(args[1], args[2])!;
-			if (args[3] === 'mail') {
-				pokemon.item = '(mail)';
-			} else if (args[3] === 'item') {
-				pokemon.item = '(exists)';
+			if (args[3]) {
+				const notes = JSON.parse(args[3]);
+				Object.assign(pokemon, notes);
```
Slayer95, line 3551: Merging a malformed object received from the server may break the ``pokemon`` instance.

Slayer95: Specifically, before I forget, ``pokemon.item`` must be escaped.

qwertystrargs: I will restore the existing code for the mods and add checks for Tera Type Preview and Force Open Team Sheets.

Assuming that a `details`-esque string is sent instead of JSON, should the name be sent as an additional `poke` parameter in case the `item` contains something that invalidates splitting `name` and `item` at `lastIndexOf(';item:')` ?

Are there other formatting concerns in this context, such as other properties that might include `;`?

Slayer95: ``;`` doesn't have a reserved meaning for any of PS protocols, IIRC, so it would be best not to use it. Instead, ``,`` or ``|`` can be used. Or even hyphen ``-`` or spaces, if the context is non-ambiguous.

Normally, we'd just add a new ``args`` parameter with ``|``, but you may prefer using ``kwArgs`` (i.e. ``|tera: Poison``), specially since it seems that ``|poke|`` may get an increasingly amount of extra specifiers in further generations.

Slayer95: (You may also use JSON, too, but the amount of safeguards that need to be added make it not a very pretty approach.)

Slayer95: Sorry, @qwertystrargs , it seems like I forgot how kwargs work. Syntax is ``|[tera] Poison``. Then you may access ``kwArgs.tera == 'Poison'``, and ``args[3]`` can still be used for items exclusively.

src/battle.ts
```diff
@@ -3546,23 +3546,24 @@ export class Battle {
...
+			if (args[3] === 'mail') {
+				pokemon.item = '(mail)';
+			} else if (args[3] === 'item') {
+				pokemon.item = '(exists)';
+			}
+			if (this.rules['Force Open Team Sheets']) {
+				pokemon.name = kwArgs.name;
+				pokemon.ident = `${pokemon.side.sideid}: ${pokemon.name}`;
+				pokemon.searchid = `${pokemon.ident}|${pokemon.details}`;
+				pokemon.item = kwArgs.item;
```
Slayer95, line 3558: In addition to ``item``, I will trouble you to investigate which of these need ``escapeHTML``. Likely, the answer is "all of them".

Tip: Most (if not all) of effect types in dex-data.ts have their .name property escaped.

qwertystrargs: From what I could tell, `name` is stored as is and is escaped later for display purposes, such as in `showPokemonTooltip` of `battle-tooltips.ts`. The rest definitely need to be cleaned.

src/battle.ts
```diff
@@ -3544,6 +3556,17 @@ export class Battle {
 			} else if (args[3] === 'item') {
 				pokemon.item = '(exists)';
 			}
+			if (this.rules['Force Open Team Sheets'] && kwArgs.name) {
+				pokemon.name = kwArgs.name;
+				if (kwArgs.item) pokemon.item = Dex.items.get(Dex.sanitizeName(kwArgs.item)).name;
```
Slayer95, line 3561: You may use either ``.get().name`` or ``sanitizeName()``, but not both. That's double escaping.

Slayer95:
This approach looks better. But could you add some comments explaining the new code branches in ``|updatepoke|`` ?

PR #2130, opened 2023-07-10
Ladder: Set current gen to 9 for decay purposes

adrivrie:
Truthfully I'm not sure what this code does: why does the client require the Elo decay rules? Regardless, rules have changed in the loginserver so they should probably change here too.

loginserver commit: https://github.com/smogon/pokemon-showdown-loginserver/commit/1186bcfb4ab5839fc0d3dcf6a92b3377665f0011

### DIFF
--- a/lib/ntbb-ladder.lib.php
+++ b/lib/ntbb-ladder.lib.php
@@ -316,8 +316,8 @@ class NTBBLadder {
 						$decay = 1 + intval(($elo-1400)/50);
 					}
 					switch ($this->formatid) {
-					case 'gen8randombattle':
-					case 'gen8ou':
+					case 'gen9randombattle':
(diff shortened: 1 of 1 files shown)


mia-pi-git:
This is the old loginserver code, it's not in use anymore. We're just keeping it around as a backup.

adrivrie:
Ah that makes sense. If it's not useful enough to edit feel free to close, and have a nice day :)

AnnikaCodes:
Yeah, this code isn't in use, but if we're keeping it in the repo in case we need to switch to PHP we might as well keep it up to date. I don't know how long we should keep the old code around; I believe the new loginserver was deployed in spring 2023, so it's still fairly recent.

PR #2131, opened 2023-07-22
Denote once-terastallized Pokemon via teraType

qwertystrargs:
This displays the "Tera Type: " reminder for pokemon which were previously terastallized.
image


AnnikaCodes:
Let's hope this works!

Issue #1991, opened 2022-08-28
[Teambuilder] Filtering moves by type/category overwrites existing moves


Rebmiami:
Filtering moves by type or category in the team builder will seemingly delete the move from your Pokémon's moveset, however it still appears highlighted in the moves list. Attempting to delete the "phantom" move will cause strange effects when you go back to the team view, as shown in this recording:

https://user-images.githubusercontent.com/59275598/187088226-dcee32e4-3de1-4dc8-99d4-0a42027b7cb7.mp4

Occurs on nwjs app and Firefox on Windows 10. Have not tested on other platforms.

joebangara43903:
![image](https://user-images.githubusercontent.com/131228993/232965391-e2ead865-858e-480a-a5fd-fd3faae24a65.png)

PR #2134, opened 2023-08-01
Fix phantom move bugs in teambuilder

Solirs:
There are currently two annoying bugs in the teambuilder for moves.

1. Replacing an already used move with itself doesn't actually remove it and instead overwrites another move after saving.
2. Moves that are overwritten by changing the filter create phantom moves with similar behavior after saving if we try to re add them.

Bug examples:

https://github.com/smogon/pokemon-showdown-client/assets/88553777/63aa4bd6-704c-4643-8eca-2644def98034
https://user-images.githubusercontent.com/59275598/187088226-dcee32e4-3de1-4dc8-99d4-0a42027b7cb7.mp4

This PR fixes both issues with an elegant two lines solution.
The array containing the moves should be updated after overwriting and moves should be selected and not overwritten when changing filters (this also makes more sense imo).

This PR would fix/close issue  #1991 .

After the patch:

https://github.com/smogon/pokemon-showdown-client/assets/88553777/b7db47ed-64eb-4052-8e2d-88943215f7aa

Thanks to anyone who tests this PR so i can be sure i didn't mess anything up.

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -2851,8 +2851,8 @@
 						this.unChooseMove(curVal);
 						$inputEl.val('');
 						delete this.search.cur[toID(val)];
-						this.curSet.moves.splice(i - 1, 1);
-
+						// Remove all instances of the duplicate move from the moves array.
+						this.curSet.moves = this.curSet.moves.filter(e => e != curVal);
 					} else if (curVal) {
 						moves.push(curVal);
 					}
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -2835,7 +2835,7 @@
 		},
 		chartClick: function (e) {
 			if (this.search.addFilter(e.currentTarget)) {
-				this.$('input[name=' + this.curChartName + ']').val('').select();
+				this.$('input[name=' + this.curChartName + ']').select();
 				this.search.find('');
 				return;
 			}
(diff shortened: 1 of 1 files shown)


Solirs:
Anyone up to check this?

js/client-teambuilder.js
```diff
@@ -2835,7 +2835,7 @@
 		},
 		chartClick: function (e) {
 			if (this.search.addFilter(e.currentTarget)) {
-				this.$('input[name=' + this.curChartName + ']').val('').select();
+				this.$('input[name=' + this.curChartName + ']').select();
```
Zarel, line 2838: This is necessary when searching for filters. i.e. if you want to search for Fire moves, you would type "Fire", click on "Fire", and then you would want "Fire" to no longer be in the move box afterwards.

js/client-teambuilder.js
```diff
@@ -2851,6 +2851,8 @@
 						this.unChooseMove(curVal);
 						$inputEl.val('');
 						delete this.search.cur[toID(val)];
+						// Remove all instances of the duplicate move from the moves array.
+						this.curSet.moves = this.curSet.moves.filter(move => move != curVal);
```
Zarel, line 2855: Hm, this works but uses modern JavaScript that would break older browsers. The CI should have caught it; I'm not sure why the CI isn't running on your code.

Zarel:
I've fixed these properly now. Sorry I took so long to see this!

Please don't be discouraged that your PR wasn't merged. It was a huge help pointing me in the correct direction for what was wrong, and how to fix it, and I'd love more PRs like this.

PR #2141, opened 2023-08-28
Fix leaveroom confirm when the room no longer exists 

singiamtel:
When using the setting `Confirmation before leaving a room`, there's a bug that prevents you from leaving if you have this setting turned on. I haven't found any regression, let me know if you want me to test a specific use case

Example on how to replicate the bug:
- Toggle the setting `Confirmation before leaving a room` to true 
- Create groupchat with ``/mgc``
- Destroy it with ``/dgc``
- Try to leave

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1511,8 +1511,8 @@
 			this.close();
 		},
 		leaveRoom: function (data) {
-			this.room.send('/noreply /leave');
 			this.close();
+			return app.removeRoom(this.room.id);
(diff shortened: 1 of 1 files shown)


singiamtel:
Also when I try to refresh the site with this setting on I get a browser alert with

```
This page is asking you to confirm that you want to leave — information you’ve entered may not be saved.
```

Is this intended?

mia-pi-git:
Thanks!

PR #2143, opened 2023-08-29
Fixing variable scoping error

EvGym:
Apparently I cannot test locally. Noted for the future. Sorry about that chaps


HoeenCoder:
You will need to re-add your previous changes too as they were reverted.

I'll take the time to help fix your problems with testing when I am able too. Earliest would probably be tomorrow night assuming I still have power (hurricane).

EvGym:
Easy enough, had them stashed. I didn't re-add them because they were still there when I pulled changes

DaWoblefet:
As it is, this PR makes it so IVs are no longer auto-set in generations without hyper training on import, when they should be. You should using ``node build`` after making changes and testing the behavior via ``testclient.html``.

I would strongly recommend testing the following:

In Generations without hyper training (Gens 2-6)
- Importing a team without specific IVs and Hidden Power should automatically attempt to set appropriate Hidden Power IVs
- IVs already set in the export should be respected on import and not modified

In generations with hyper training (Gen 7+)
- Importing a team without specific IVs and Hidden Power should not automatically attempt to set Hidden Power IVs

Make sure to test for both Hidden Power [type] and just Hidden Power. Check for various ways of importing, e.g. importing a single team, importing teams in bulk, and importing from pokepaste.

singiamtel:
Make sure to clear cache between tests too

EvGym:
> 

I did do that, Something is wrong with my local test copy and I am doing a deep purge and re-installing node to see if it will fix it. What I had thought was just standard reduced functionality with a local test turned out to be more severe than anticipated. 

GOOD NEWS! Purging everything with fire and re-installing everything worked! Your guide was very helpful to ensure I didn't forget to test anything

Problems:
- The teambuilder is aggressively agnostic about formats. So it will only take it into account when importing from backup. I tried to route information about the format to it but I was not able to make any progress. People more experienced than I would need to do this.

Every other corrective measure for IVs when using hidden power still works. For example, editing your move set still changes IVs as it should.

So what does this actually change?
 - Sets directly imported into teams, or teams directly imported can no longer have their ivs tampered with. In any generation. (Importing your backups is left unchanged). The validator still catches this issue. You have always been able to import illegal sets through this method anyway so this won't catch people off guard.

EvGym:
I removed how the teambuilder would reset the IVs of mon that had them match the IV set for their hidden power type due to the blind spot with importing teams individually. This should make it so you can spam the import button on a team member in your roster and not  have their IVs reset

EvGym:
One slight fix, during retesting I realized I forgot to include my fix for pokepaste imports

EvGym:
I am still watching this by the way. Hopefully, this will be able to be merged soon as this bug is extremely frustrating.

EvGym:
bumping this. I would still love to see this bug finally get fixed

Zarel:
Sorry about never getting around to this. Reviewing this sort of code is difficult. The Preact client takes a new approach to the teambuilder which should in theory fix this sort of thing.

EvGym:
Hopefully. This bug is extremely annoying and the "feature" that caused it was simply never necessary

Zarel:
The Preact client is here, if you would like to test it and see if it fixes this:

https://psim.us/preactbeta

EvGym:
I realized I never actually stated.
1) It does fix it. 🎉 
2) Damn the new client looks good

PR #2145, opened 2023-09-01
Fix level setting for LC variants

KrisXV:


KrisXV:
i swear i tested this i just didnt get crashes from the typo bc its in a failsafe

PR #2150, opened 2023-09-14
Fix Hospitality activation message

Karthik99999:
I didn't want to hardcode this, but it seems like absorb abilities like volt absorb and dry skin depend on the client not using the pokemon from `[of]` to properly show their activation message. That feels like something that should be fixed server side (the source of healing for those abilities should be the same as the ability holder), but if I did fix that old replays would have incorrect activations unless I hardcoded removing `[of]` from said absorption abilities (which is what used to be done).

### DIFF
--- a/src/battle-text-parser.ts
+++ b/src/battle-text-parser.ts
@@ -930,7 +930,7 @@ class BattleTextParser {
 		case '-heal': {
 			let [, pokemon] = args;
 			let template = this.template('heal', kwArgs.from, 'NODEFAULT');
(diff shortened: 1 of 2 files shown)


urkerab:
Shouldn't this be done in `upgradeArgs` where it could affects both places?

Karthik99999:
> Shouldn't this be done in `upgradeArgs` where it could affect both places?

Yeah i guess upgradeArgs is better. I think I will go back to hardcoding the absorb abilities though because those are the ones that are incorrect.

PR #2163, opened 2023-10-16
Prevent remote teams from overwriting local teams

Karthik99999:
The extra `loaded` check seems to be allowing local teams to be overwritten with remote team data on every refresh, which doesn't seem to be the intended behavior given this comment in `storage.js`.
```
// prioritize locally saved teams over remote
// as so to not overwrite changes
```

### DIFF
--- a/js/client-teambuilder.js
+++ b/js/client-teambuilder.js
@@ -142,7 +142,7 @@
 		update: function () {
 			teams = Storage.teams;
 			if (this.curTeam) {
-				if (this.curTeam.loaded === false || (this.curTeam.teamid && !this.curTeam.loaded)) {
+				if (this.curTeam.loaded === false) {
 					this.loadTeam();
 					return this.updateTeamView();
 				}


mia-pi-git:
This check is necessary to ensure remote teams actually get loaded. `teamid` only exists on the team if it's been remotely loaded at one point or another.

mia-pi-git (changes requested):
This won't fix the issue described.

Karthik99999:
> This check is necessary to ensure remote teams actually get loaded. `teamid` only exists on the team if it's been remotely loaded at one point or another. 

I tested the change on the testclient and the team loaded in just fine initially. Is there an edge case that I missed?

mia-pi-git:
it works fine with actual remote teams, the problem is non-remote (local-only) teams, which it tries to load forever without this check.

Karthik99999:
I'm a bit confused on this. My local only teams seemed to work just fine as well, and wouldn't local only teams not have a `teamid` anyway?

PR #2164, opened 2023-10-18
Fix `showteam` in replays

Karthik99999:
- Moves the HTML generation to `BattleLog`
- Creates functions for `unpackTeam` and `exportTeam` under `Battle` as replays do not have access to the `Storage` functions.

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -3325,6 +3325,256 @@ export class Battle {
 		} as any;
 	}
 
+	// Taken from Storage, they need to be implemented here so `showteam` can work
+	unpackTeam(buf: string) {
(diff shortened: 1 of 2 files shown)


src/battle.ts
```diff
@@ -3325,6 +3325,256 @@ export class Battle {
 		} as any;
 	}
 
+	// Taken from Storage, they need to be implemented here so `showteam` can work
```
Zarel, line 3328: Hm, I think it's best not to duplicate this code.

Zarel: Centralizing in `battle-dex.ts` makes sense. It'd be loaded before Storage.

PR #2165, opened 2023-10-20
Fix CSS on downloaded replays

Karthik99999:
I did not notice the Times New Roman font in the tooltips on my other PR.

### DIFF
--- a/js/replay-embed.template.js
+++ b/js/replay-embed.template.js
@@ -29,6 +29,8 @@ function requireScript(url) {
 }
 
 linkStyle('https://play.pokemonshowdown.com/style/font-awesome.css?');
+linkStyle('https://pokemonshowdown.com/theme/panels.css?');
+linkStyle('https://pokemonshowdown.com/theme/main.css?');
 linkStyle('https://play.pokemonshowdown.com/style/battle.css?a7');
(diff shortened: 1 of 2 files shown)


Zarel:
https://github.com/smogon/pokemon-showdown-client/pull/2162/files#diff-aba53ffa06492e2f12ffc7597a0bbeff1a3b5e5d74b7e932c73a149c01622c00L1145

This would be fixed if we just reverted that, right?

Zarel:
I tested and it does seem to be sufficient.

Zarel:
Okay, I pushed e3b82f2181d2723b8b3 which fixes the specific bug described in your PR OP.

I'm still confused why you removed all the CSS from downloaded replays, so I'm leaving this PR open until you tell me why you made the change and whether any changes still need to be made.

Karthik99999:
> Okay, I pushed e3b82f2181d2723b8b3 which fixes the specific bug described in your PR OP.
> 
> 
> 
> I'm still confused why you removed all the CSS from downloaded replays, so I'm leaving this PR open until you tell me why you made the change and whether any changes still need to be made.

I just noticed some discrepancies (mainly font size) from the downloaded and online replays, and after I removed all the css I was like "hey it still works" so I just left it like that. I can see why you had it there though, so yeah this pr isn't needed anymore

PR #2166, opened 2023-10-20
Make popups scrollable internally instead of adding scrollbar to body

lorumic:
- Changed some CSS so that popups with an overflowing content can be scrollable internally, instead of adding a scrollbar to the body, so that the layout of the page is kept and does not seem broken.
- Browser compatibility is not altered by the changes in this PR.

Before:

Peek 2023-10-20 09-26

After:

Peek 2023-10-20 09-27

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -706,7 +706,9 @@ span.header-username:hover {
 	z-index: 20;
 }
 .ps-popup {
-	position: absolute;
+	position: fixed;
+	max-height: 99vh;
+	overflow-y: auto;
 	top: auto; right: auto; left: auto; bottom: auto;
 	text-align: left;
 	background: #E1E8E8;


Zarel:
I don't particularly like internal scrollbars for popups. I guess the current system is a little ugly, but I'm happy with how convenient/usable it is.

(It was also necessary on mobile back when internal scrolling was unnecessarily difficult on mobile.)

I guess one compromise would be to put it in an overlay and have the overlay be scrollable. You'd have to darken the overlay but that might not be a bad idea for Options.

lorumic:
> I guess one compromise would be to put it in an overlay and have the overlay be scrollable. You'd have to darken the overlay but that might not be a bad idea for Options.

Thanks for the feedback @Zarel. Just to check that I understood correctly, you mean using, for this popup, the same approach that is used for the avatar selection, right?

Peek 2023-11-21 21-45

PR #2170, opened 2023-10-25
Fix formats not found message not appearing

Karthik99999:
well actually "No formats found" whatever. Maybe it shouldn't rely on `bufs` being empty?

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -1263,6 +1263,7 @@
 			}
 
 			for (var i in this.starred) {
+				if (!bufs[1]) bufs[1] = '';
 				var format = BattleFormats[i];
 				if (!format) {
 					delete this.starred[i];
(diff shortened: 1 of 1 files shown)


mia-pi-git:
Probably should check when looping over `bufs` to see if there's any text in the string to begin with.

js/client-mainmenu.js
```diff
@@ -1318,7 +1318,7 @@
 				);
 			}
 			var html = '';
-			if (!bufs.length) {
+			if (!bufs.every(function (buf) { return !!buf; })) {
```
urkerab: This logic is wrong, which is why you added the length check to try to fix it.

js/client-mainmenu.js
```diff
@@ -1318,7 +1318,7 @@
 				);
 			}
 			var html = '';
-			if (!bufs.length) {
+			if (!bufs.length || !bufs.every(function (buf) { return !!buf; })) {
```
urkerab, line 1321:
```suggestion
			if (bufs.every(function (buf) { return !buf; })) {
```

js/client-mainmenu.js
Karthik99999: I still need the length check in case `bufs` is empty

Karthik99999: actually nvm I see what you mean now

DaWoblefet:
Thanks all!

PR #2171, opened 2023-10-25
Item ratings/isNonstandard

EnPassant4264:
As per [this Issue](https://github.com/smogon/pokemon-showdown-client/issues/2156) (that I wish anyone had responded to), this code adds an overridable rating variable for items to determine which section of the default teambuilder list it appears in, rather than a lengthy, hardcoded switch statement. The Unreleased items section has been removed, since it was only coded to appear in Gen 6 and it would always be empty since its items were Illegal and therefore don't display in the list. Like the prior algorithm, it defaults to the Items section, and all items not in that section have been given ratings in the [server PR](https://github.com/smogon/pokemon-showdown/pull/9856). Pokemon-specific items continue to be determined by itemUser or megaStone, and Poke Balls and TRs are still automatically listed as Useless; both ignore the rating variable. Booster Energy has been made to be specific to Paradox Pokemon.

This code also collects and displays mods overriding an item's isNonstandard value on the client side, which is presently useful in Mix and Mega.

### DIFF
--- a/play.pokemonshowdown.com/src/battle-log.ts
+++ b/play.pokemonshowdown.com/src/battle-log.ts
@@ -916,11 +916,13 @@ export class BattleLog {
 
 				const src = getAttrib('src') || '';
 				// Google's ToS requires a minimum of 200x200
-				let width = '320';
-				let height = '200';
-				if (window.innerWidth >= 400) {
-					width = '400';
-					height = '225';
+				let width = getAttrib('width') || '0';
+				let height = getAttrib('height') || '0';
+				if (Number(width) < 200) {
+					width = window.innerWidth >= 400 ? '400' : '320';
+				}
+				if (Number(height) < 200) {
+					height = window.innerWidth >= 400 ? '225' : '200';
 				}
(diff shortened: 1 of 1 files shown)


Zarel:
You seem to have closed the server PR?

EnPassant4264:
I physically couldn't update my Showdown repos until my PR was either accepted or my commits discarded due to merge conflicts. I guess it closed automatically when I did that. This is why I created that Issue I mention in the summary before coding and creating the PR. It has been sitting here untouched since before DLC1 dropped, and all the files I changed clientside have been moved to another folder anyway, so the odds of it being accepted were very low.

Zarel:
Oh, yeah. Sorry about that. I spent some time away and missed this PR.

In the future, to deal with the issue where you couldn't update, you should PR a branch rather than master.

EnPassant4264:
How would I PR a branch of the main repo? Wouldn't making a new branch on mine still need to be PR'd into an existing branch, or otherwise have the step of being merged afterward?

urkerab:
You make a branch in your repo before committing your changes. (If you're working locally, you push that branch to GitHub.) You can then PR that branch directly to here. (You need to do that anyway because this repo uses squash merges so you wouldn't have been able to merge your master back in even if your PR had been accepted.)

PR #2172, opened 2023-10-30
Give JSON files the Access-Control-Allow-Origin: * header 

LumarisX:
Updating the .htaccess file to allow moves, abilities, and learnsets json files to be accessed across domain by adding the Access-Control-Allow-Origin: * header. 

From the research I did on the .htaccess file, this should solve the issue, but I couldn't see these files in my test environment to confirm that this is a fix.


singiamtel:
Doesn't the browser ignore CORS when requesting JSONs?

LumarisX:
Unless I misunderstood something in my research, no. When I pull moves.json this is the error I see.
image

When I pull news.json however, it works. Only difference being the header.

singiamtel:
Yeah you're right, my bad

Zarel:
Thanks!

PR #2173, opened 2023-10-31
Support more variable best-of challenges

DaWoblefet:
Right now, the checkbox used for team-locked best-of-3 play is limited to VGC formats. This change allows the server's format list to determine which formats have the checkbox by default. This was initially requested by side servers, but also fixes a very minor bug of team-locked bo3 challenges showing up for Gen 4 VGC, where you weren't team-locked (you could move around items in between a best-of-three and change team order). I'm sure we'll eventually have a non-VGC format that wants best-of-n support on main as well.

It also allows the challenger to select any of the supported best-of-n options, rather than exclusively being best-of-3.
image

I'm not super sure about the policy of inline CSS; it's needed for the number input. I can move it to a proper CSS file if desired.

### DIFF
--- a/js/client-mainmenu.js
+++ b/js/client-mainmenu.js
@@ -943,12 +943,12 @@
 			var teamIndex = $pmWindow.find('button[name=team]').val();
 			var privacy = this.adjustPrivacy($pmWindow.find('input[name=private]').is(':checked'));
 
-			var bestof = $pmWindow.find('input[name=bestof]').is(':checked');
-			var bestofvalue = $pmWindow.find('input[name=bestofvalue]').val();
-			if (bestof && bestofvalue) {
+			var bestOf = $pmWindow.find('input[name=bestof]').is(':checked');
+			var bestOfValue = $pmWindow.find('input[name=bestofvalue]').val();
+			if (bestOf && bestOfValue) {
 				var hasCustomRules = format.includes('@@@');
 				format += hasCustomRules ? ', ' : '@@@';
-				format += 'Best of = ' + bestofvalue;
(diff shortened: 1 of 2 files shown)


js/client-mainmenu.js
```diff
@@ -943,10 +944,11 @@
 			var privacy = this.adjustPrivacy($pmWindow.find('input[name=private]').is(':checked'));
 
 			var bestof = $pmWindow.find('input[name=bestof]').is(':checked');
```
KrisXV, line 946:
```suggestion
			var bestOf = $pmWindow.find('input[name=bestof]').is(':checked');
```

js/client-mainmenu.js
```diff
@@ -943,10 +944,11 @@
...
-			var hasCustomRules = format.includes('@@@');
-			if (bestof) {
+			var bestofvalue = $pmWindow.find('input[name=bestofvalue]').val();
+			if (bestof && bestofvalue) {
```
KrisXV, line 948:
```suggestion
			var bestOfValue = $pmWindow.find('input[name=bestofvalue]').val();
			if (bestOf && bestOfValue) {
```

js/client-mainmenu.js
```diff
@@ -943,10 +944,11 @@
...
+				var hasCustomRules = format.includes('@@@');
 				format += hasCustomRules ? ', ' : '@@@';
-				format += 'Best of = 3';
+				format += 'Best of = ' + bestofvalue;
```
KrisXV, line 951:
```suggestion
				format += 'Best of = ' + BestOfValue;
```

PR #2174, opened 2023-11-02
Change footer to fixed positioning

FeluciaPS:
When you have a lot of PM windows open, the footer now probably stays put when you scroll through them.

### DIFF
--- a/style/client.css
+++ b/style/client.css
@@ -944,7 +944,7 @@ p.or:after {
 	min-height: 100%;
 }
 .mainmenufooter {
-	position: absolute;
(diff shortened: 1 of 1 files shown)


Zarel:
I'm a little confused. Why?

Zarel:
Oh, I see what's going on.

Unfortunately, this doesn't work well if the window is less tall than the footer.

image

Zarel:
You could solve this with a bunch of media queries, I suppose. Your screen would need to be tall enough to fit the main menu and wide enough to make it two-column.

PR #2175, opened 2023-11-02
Design refresh

Zarel:
Felucia mentioned she had some custom CSS in dark mode to make it darker, so I figured I'd try my hand at a design refresh to make it not quite so bright. In the process, I've made the buttons look more 3D, matching light mode. Sorry, any dark mode users who prefer flat design.

This also updates all remaining OS-style buttons into PS-style buttons, so they'll be properly dark-mode.

This refresh focuses on dark mode, but light mode gets some improvements as well.

image
(top: old, bottom: new)

image
(left: old, right: new)

image
(left: old, right: new)

image
(left: old, right: new)

image
(left: old, right: new)

image
(left: old, right: new)

image
(left: old, right: new)

image
(left: old, right: new)

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -276,10 +276,18 @@
 					// spectator
 					if (this.battle.paused) {
 						// paused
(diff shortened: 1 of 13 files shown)


Zarel:
I've deleted 350 lines from client.css but there are still 3300 left... Oh well, the remaining CSS should be a lot less jank.

Zarel:
After some consideration, I'm going with this as the in-sim replay layout:

image

Zarel:
For posterity, here's all the layouts I considered:

image

image

image

image

image

image

image

The fifth isn't very pretty, but it's very usable so that's the one I went with in the end.

WallyTheBully:
Very good dark mode improvements

PR #2177, opened 2023-11-04
CSS : Minor fix

Distrib-ps:
imageoldimagenew

imageoldimagenew

imageoldimagenew

imageoldimagenew

### DIFF
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1512,7 +1512,7 @@
 			if (this.gameType === 'battle' && this.room.battle && !this.room.battle.rated) {
 				buf += '<button type="button" name="replacePlayer" class="button">Replace player</button> ';
 			}
-			buf += '<button type="button" name="close" class="button" class="button">Cancel</button></p></form>';
+			buf += '<button type="button" name="close" class="button">Cancel</button></p></form>';
 			this.$el.html(buf);
 		},
 		replacePlayer: function (data) {
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1512,7 +1512,7 @@
 			if (this.gameType === 'battle' && this.room.battle && !this.room.battle.rated) {
 				buf += '<button type="button" name="replacePlayer" class="button">Replace player</button> ';
 			}
-			buf += '<button type="button" name="close" class="button">Cancel</button></p></form>';
+			buf += '<button type="button" name="close" class="autofocus button">Cancel</button></p></form>';
 			this.$el.html(buf);
 		},
 		replacePlayer: function (data) {
--- a/js/client-battle.js
+++ b/js/client-battle.js
@@ -1512,7 +1512,7 @@
 			if (this.gameType === 'battle' && this.room.battle && !this.room.battle.rated) {
 				buf += '<button type="button" name="replacePlayer" class="button">Replace player</button> ';
 			}
-			buf += '<button type="button" name="close" class="autofocus" class="button">Cancel</button></p></form>';
+			buf += '<button type="button" name="close" class="button" class="button">Cancel</button></p></form>';
 			this.$el.html(buf);
 		},
 		replacePlayer: function (data) {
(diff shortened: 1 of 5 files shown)


js/client-battle.js
```diff
@@ -1512,7 +1512,7 @@
 			if (this.gameType === 'battle' && this.room.battle && !this.room.battle.rated) {
 				buf += '<button type="button" name="replacePlayer" class="button">Replace player</button> ';
 			}
-			buf += '<button type="button" name="close" class="autofocus" class="button">Cancel</button></p></form>';
+			buf += '<button type="button" name="close" class="button" class="button">Cancel</button></p></form>';
```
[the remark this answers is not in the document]
Distrib-ps: oops, thx

js/client-battle.js
```diff
@@ -1512,7 +1512,7 @@
 			if (this.gameType === 'battle' && this.room.battle && !this.room.battle.rated) {
 				buf += '<button type="button" name="replacePlayer" class="button">Replace player</button> ';
 			}
-			buf += '<button type="button" name="close" class="autofocus" class="button">Cancel</button></p></form>';
+			buf += '<button type="button" name="close" class="button">Cancel</button></p></form>';
```
urkerab, line 1515: I'm pretty sure the intent was to add the `button` class without removing the `autofocus` class. Compare https://github.com/smogon/pokemon-showdown-client/blob/f9bc940eb5b2f2d55f7ea9b5c8c82e9ffe62c3c0/js/client.js#L2981
```suggestion
			buf += '<button type="button" name="close" class="autofocus button">Cancel</button></p></form>';
```
https://github.com/smogon/pokemon-showdown-client/blob/f9bc940eb5b2f2d55f7ea9b5c8c82e9ffe62c3c0/js/client.js#L2549 looks like it still needs fixing though.

style/client.css
```diff
@@ -247,7 +247,6 @@ i.subtle:hover {
 	position: relative;
 	padding: 3px 2px 3px 2px;
 	margin: 0 -1px 0 0;
-	top: 1px;
```
Zarel, line 250: ???

style/battle-log.css
```diff
@@ -163,6 +163,9 @@ button:disabled {
 	text-shadow: none;
 	color: #F9F9F9;
 }
+.dark select.button {
+	background-color: #2b2c31 !important;
```
Zarel, line 167: I have a guess as to what's going on here but what browser is this bad???

js/client-teambuilder.js
```diff
@@ -1202,12 +1202,12 @@
 				}
 				for (i = 0; i < this.curSetList.length; i++) {
 					if (this.curSetList.length < this.curTeam.capacity && this.deletedSet && i === this.deletedSetLoc) {
-						buf += '<li><button name="undeleteSet"><i class="fa fa-undo"></i> Undo Delete</button></li>';
+						buf += '<li><button name="undeleteSet" class="button"><i class="fa fa-undo"></i> Undo Delete</button></li>';
```
Zarel, line 1205: This looks better as not a button...

style/client.css
Distrib-ps: I thought the spacing was intentional

style/battle-log.css
Distrib-ps: I use chrome

js/client-teambuilder.js
Distrib-ps: Would you prefer to keep it white? 
image
image

style/client.css
Zarel: Sorry, I'm still confused. What is this changing and why? Do you have before/after screenshots?

Distrib-ps: Yeah, 
image
It's the little gap at the top

js/client-teambuilder.js
Zarel: It shouldn't be white. I must be mixing this up with something else.

Zarel: Yeah, I got this mixed up with undeleting teams, which already has styling.

style/battle-log.css
Zarel: Hah, it turns out that Chrome for Windows has this problem, but Chrome for Mac doesn't.

Anyway, interesting solution, but it turns out styling `option` works, and doesn't require your hack that makes select buttons uglier.

PR #2178, opened 2023-11-05
Fix unstyled buttons when "Group by Folders" is unchecked

shrianshChari:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9849881


Zarel:
Thanks!

PR #2180, opened 2023-11-06
Fix CSS on games under "Games:" section

Hypersonicly:
[https://i.snipboard.io/Q5vVdz.jpg](https://i.snipboard.io/Q5vVdz.jpg) Fixes bad looking links to battles you're in.


Distrib-ps:
Wouldn't something like this be better? image

Hypersonicly:
Yeah that would look better I think

PR #2181, opened 2023-11-06
CSS: Fix CSS on "Games"  section

Distrib-ps:
image

image


Zarel:
Fixed in 9219fb31b2d9b1e1447dcc8f42.

Thanks for drawing attention to this!

PR #2183, opened 2023-11-07
 Give \data files Access-Control-Allow-Origin Header

LumarisX:
[Gives all \data js and json files Access-Control-Allow-Origin Header to be requested across domains. Update to pull request 60.](https://github.com/smogon/pokemon-showdown-client/pull/2182)


.htaccess
```diff
@@ -91,7 +91,7 @@ RewriteCond %{HTTP_HOST} ^www\.play\.pokemonshowdown\.com$ [NC]
 RewriteRule ^(.*) https://play.pokemonshowdown.com/$1 [R=301,L]
 
 RewriteRule ^style/fonts?/.*?\.(eot|svg|ttf|woff|woff2)$ - [E=SAFE_RESOURCE:1]
-RewriteRule ^data/(moves|abilities|learnsets).json$ - [E=SAFE_RESOURCE:1]
+RewriteRule ^data\/\w*?\.js(on)?$ - [E=SAFE_RESOURCE:1]
```
Zarel, line 94: Seems like an excessive amount of processing. Why not just

```suggestion
RewriteRule ^data\/.*\.js(?:on)?$ - [E=SAFE_RESOURCE:1]
```

LumarisX: I agree, with your edits. That would reduce the amount of processing. My only note would be to change it to ^data\/**\w***\.js(?:on)?$ if you don't want subfolders to also have their files given the header. If you are fine with that then your suggestion is great.

Zarel: We don't have any subfolders I'd want to protect in `data/`.

Zarel: The only reason you'd want to leave off ACAO is for a dynamic file, anything `.php` or in `/api/`. That's not anything in there.

Zarel:
tyty

PR #2184, opened 2023-11-13
Fix Tourney Bracket panning for touch devices

SirzBenjie:
Mobile devices that don't have a mouse-attached can't pan the tournament bracket view.  This was because the listeners were registered for specifically _mouse_ instead of _pointer_

This presents a fix.
I tested this for mobile versions of chrome, firefox, and opera (dev tools).  The panel can now be panned with touch.

-Quazizata


Zarel:
Thanks!

PR #2187, opened 2023-11-16
Reorganize directories

Zarel:
Files meant to be served have been moved into `play.pokemonshowdown.com/` and `pokemonshowdown.com/`.

We now have three directories for the three subdomains handled by this repo:

- `pokemonshowdown.com/`
- `play.pokemonshowdown.com/`
- `replay.pokemonshowdown.com/`

Naming them after the subdomains will make it much easier to tell where the files for each go.

The diff is probably useless, it'll be easier if you just look at the new tree: https://github.com/smogon/pokemon-showdown-client/tree/reorganize


Zarel:
(This PR is really only to notify everyone; this is already deployed and seems to be working fine.)

PR #2195, opened 2023-12-12
Fix copy replay button functionality

malaow3:
Copy replay link button functionality was previously not working. This might not be the best way to "fix" that, as there is an existing copyReplayLink function present in the same file. 

This is my quick-and-inelegant solution with the purpose of possibly providing a short term solution until a better one can be found.

Thanks!

### DIFF
--- a/play.pokemonshowdown.com/js/client.js
+++ b/play.pokemonshowdown.com/js/client.js
@@ -2555,6 +2555,21 @@ function toId() {
 			this.$el.html('<form><p style="white-space:pre-wrap;word-wrap:break-word">' + (data.htmlMessage || BattleLog.parseMessage(data.message)) + '</p><p class="buttonbar">' + (data.buttons || '<button type="button" name="close" class="button autofocus"><strong>OK</strong></button>') + '</p></form>').css('max-width', data.maxWidth || 480);
 		},
 
+		copyText: function (value, target) {
+			var dummyInput = document.createElement("input");
+			// This is a hack. You can only "select" an input field.
+			//  The trick is to create a short lived input element and destroy it after a copy.
+			// (stolen from the replay code, obviously --mia)
+			dummyInput.id = "dummyInput";
+			dummyInput.value = value || target.value || target.href || "";
+			dummyInput.style.position = 'absolute';
+			target.appendChild(dummyInput);
+			dummyInput.select();
+			document.execCommand("copy");
+			target.removeChild(dummyInput);
+			$(target).text('Copied!');
+		},
+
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/js/client.js
```diff
@@ -2555,6 +2555,21 @@ function toId() {
...
+			//  The trick is to create a short lived input element and destroy it after a copy.
+			// (stolen from the replay code, obviously --mia)
+			dummyInput.id = "dummyInput";
+			dummyInput.value = value || target.value || target.href || "";
+			dummyInput.style.position = 'absolute';
+			target.appendChild(dummyInput);
+			dummyInput.select();
+			document.execCommand("copy");
+			target.removeChild(dummyInput);
+			$(target).text('Copied!');
```
HoeenCoder, line 2570: I will ask you to make a change here though. The best way to handle this is going to be calling the existing method rather than copy/pasting the code. This is easier to maintain because if any changes need to be made to the code for copying text, it only needs to be changed in one place. This can be done with this:

```suggestion
			app.curRoom.copyText(value, target);
```

HoeenCoder (changes requested):
Hi @malaow3, sorry for the delay on this. I've noticed that you have opened a PR to fix the same issue as I did, but since you opened this first your welcome to finish this up and take the credit.

Closes #2207

play.pokemonshowdown.com/js/client.js
malaow3: Appreciate the review, can get this change done tomorrow AM!

HoeenCoder (approved):
Thanks!

play.pokemonshowdown.com/js/client.js
```diff
@@ -2555,6 +2555,10 @@ function toId() {
 			this.$el.html('<form><p style="white-space:pre-wrap;word-wrap:break-word">' + (data.htmlMessage || BattleLog.parseMessage(data.message)) + '</p><p class="buttonbar">' + (data.buttons || '<button type="button" name="close" class="button autofocus"><strong>OK</strong></button>') + '</p></form>').css('max-width', data.maxWidth || 480);
 		},
 
+		copyText: function (value, target) {
+			app.curRoom.copyText(value, target)
```
HoeenCoder, line 2559:
```suggestion
			app.curRoom.copyText(value, target);
```
This is what I get for forgetting to approve github actions before reviewing the code.

malaow3:
Thanks again for the review (and for fixing my missing semicolon!)

PR #2205, opened 2024-01-01
Tooltips: Fix Hadron Engine/Orichalcum Pulse modifier

MathyFurret:
Fixes https://www.smogon.com/forums/threads/hadron-engine-oricalcum-pulse-show-wrong-modifier-in-tooltips.3733481/


DaWoblefet:
We use 1.3333x in the Ability descriptions (which is very close to 5461/4096), so we may as well be consistent here.

DaWoblefet:
Thanks Mathy!

PR #2206, opened 2024-01-03
Fix client sending the wrong team to the server in some cases

HoeenCoder:
Bug fix for this report: https://www.smogon.com/forums/threads/bug-report-team-selector-after-creating-teams.3733626/

Teams are stored as an array in `Storage.teams`. The team selection dropdown determines which team to send to the server when a player hits "Battle!" by using its `value` attribute as an index for the teams array. When a new team is created at the top of the team list, its placed at the front of this array becoming index 0 and shifting all other indexes up by 1. However, the team selection drop down's value was not updated resulting in the client sending the wrong team to the server. This fixes this by incrementing the value when a new team is added to the top of the list.

### DIFF
--- a/play.pokemonshowdown.com/js/client-teambuilder.js
+++ b/play.pokemonshowdown.com/js/client-teambuilder.js
@@ -794,12 +794,11 @@
 			var newTeam = this.createTeam(null, type === "box");
 			teams.unshift(newTeam);
 			for (var room in app.rooms) {
-				var teamSelectButton = app.rooms[room].$('button.teamselect');
-				var selection = teamSelectButton.val();
+				var selection = app.rooms[room].$('button.teamselect').val();
 				if (!selection || selection === 'random') continue;
 				var obj = app.rooms[room].id === "" ? app.rooms[room] : app.rooms[room].tournamentBox;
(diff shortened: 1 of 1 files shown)


urkerab:
`obj.curTeamIndex` is never used for some reason. Maybe it was originally supposed to have the team index, but the code to use it was never written, and it needs to be (!)

If not, there are several other places where `obj.curTeamIndex` is changed, and those should probably be updated too.

HoeenCoder:
> `obj.curTeamIndex` is never used for some reason. Maybe it was originally supposed to have the team index, but the code to use it was never written, and it needs to be (!)
> 
> If not, there are several other places where `obj.curTeamIndex` is changed, and those should probably be updated too.

I did some digging, its used (outside of the teambuilder code) to determine the value of the team selection buttons (including tournament ones). I should of checked that originally, my bad. Regardless, a simple call to updateTeams on the room object in question fixes the issue, so I've updated my code to do just that.

urkerab:
That certainly looks much nicer but I'm still worried about the other cases of `obj.curTeamIndex` in this file.

DaWoblefet:
This should have been merged ages ago!

PR #2207, opened 2024-01-04
Fix copy buttons within popups

HoeenCoder:
Fixes bug report: https://www.smogon.com/forums/threads/bug-report-copy-button-for-replays-does-not-work.3733583/

Popups lack support for copy buttons currently, this adds support by simply delegating the work to the current room's `copyText` method.


HoeenCoder:
Fixed in #2195

PR #2211, opened 2024-01-12
Wrap long format names

FeluciaPS:
image

I noticed certain format names in the formats dropdown are too long and end up cutting off, this change removes no-wrap from .option to make sure they do.

I went through every usage of the `.option` class and most of them have static-length content that is not affected by this change, the only other place this matters is for long team names in the team dropdown, which seems like a neutral (maybe slightly positive?) outcome for usability


DaWoblefet:
Thanks Felucia!

PR #2212, opened 2024-01-12
Add horizontal scroll bars to wide infoboxes on small screens

FeluciaPS:
implements the [suggestion to add a horizontal scrollbar for boxes on smaller resolution screens](https://www.smogon.com/forums/threads/add-a-horizontal-scrollbar-for-boxes-on-smaller-resolution-screens.3671644/)

This only affects boxes that have a set width that is already overflowing, and as such shouldn't have any implications for boxes that didn't already look horrid to begin with. This would put burden on writers of infoboxes to use width or min-width properties if they don't function below a certain width, but they should've been doing that anyway so I don't think it's within the scope of this change to do it for them.

### DIFF
--- a/play.pokemonshowdown.com/style/battle-log.css
+++ b/play.pokemonshowdown.com/style/battle-log.css
@@ -749,6 +749,7 @@ details.readmore[open] summary:hover:before {
 	border: 1px solid #7799BB;
 	padding: 2px 4px;
 	border-radius: 4px;
+	overflow-x: auto;
 }
 .dark .infobox {
 	border-color: #506070;


FeluciaPS:
Accidentally deleted the branch and closed the PR along with it. I'm sorry!

Zarel:
Thanks!

Zarel:
Wait! I thought roomintros specifically are already scrollable, right?

Zarel:
Yeah. Roomintros have always been scrollable. If there's an issue, it needs to be dealt with elsewhere.

Zarel:
Okay. I think I see. Roomintros are scrollable on desktop but not on mobile.

PR #2213, opened 2024-01-12
Teambuilder: Don't allow `|` in folder names

HisuianZoroark:
https://www.smogon.com/forums/threads/bug-report-teambuilder.3734292/

https://www.smogon.com/forums/threads/teambuilder-should-stop-you-from-creating-folders-that-contain.3657717


DaWoblefet:
Thanks HiZo!

PR #2214, opened 2024-01-13
Teambuilder: Fix IV values for certain Indigo Disk Pokemon

HisuianZoroark:
streamlines validation progress for the new Min-20/15 IV pokemon given to us in Indigo Disk DLC

### DIFF
--- a/play.pokemonshowdown.com/js/client-teambuilder.js
+++ b/play.pokemonshowdown.com/js/client-teambuilder.js
@@ -3286,6 +3286,9 @@
 				minSpe = true;
 			}
 
+			// only available through an event with 31 Spe IVs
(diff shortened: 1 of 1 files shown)


HisuianZoroark:
> I’m surprised there aren’t any other Pokémon with similar restrictions handled here, I could have sworn there were.

yeah I'm surprised as well. Afaik from this code Koraidon/Miraidon and Ash-Gren have 31 Atk IV restrictions and are handled in a similar manner.

DaWoblefet:
Since this doesn't actually affect validation, just the initial convenience, I'm OK with merging as-is since the Battle Bond and bike legend check earlier doesn't respect Hackmons either.

PR #2218, opened 2024-01-27
Fix to Best-Of checkbox

Akiak7:
The best-of-X checkbox when challenging someone does not work if the format is already set to Bo3 or Bo5. You get this error if you try to use it:

`- Rule "bestof=5" conflicts with "bestof=3" (Use "!! bestof=5" to override "bestof=3".)`

This fix solves this issue by making sure the best-of command always overrides any pre-existing best-of commands.


play.pokemonshowdown.com/js/client-mainmenu.js
```diff
@@ -949,7 +949,7 @@
 			if (bestOf && bestOfValue) {
 				var hasCustomRules = format.includes('@@@');
 				format += hasCustomRules ? ', ' : '@@@';
-				format += 'Best of = ' + bestOfValue;
+				format += '!!Best of = ' + bestOfValue;
```
KrisXV, line 952: Should there be logic here to only add `!!` if the best of rule already exists?

Akiak7: > Should there be logic here to only add `!!` if the best of rule already exists?

I don't see why it would make a difference personally.

KrisXV: image
because of this

Akiak7: I didn't know that was a thing, my bad.

Akiak7: I did try to make it conditional but I couldn't really figure out how to make sure it would cover all possible syntaxes for the best-of command, then I thought it wouldn't be needed so I let if be.

Zarel:
The hacky way to do it is to see if the format name ends in `Bo3`

iforgetwhyimhere:
Is this relevant anymore? I'm not aware of any formats that use Best Of in `ruleset` and also have `bestOfDefault`.

PR #2224, opened 2024-02-11
Support no animation during forme changes

HisuianZoroark:
Was requested for smogon/pokemon-showdown#10166

Probably not the best way to do it but it certainly works.


alphazeba:
I'd worked on a similar idea to handle dynamax animations.
https://github.com/smogon/pokemon-showdown-client/pull/2217

However, if there is a need for providing animation override flags, you could consider using an "override enum" as an optional input.  You could use "none" or something, i could update to use "dynamax" and it could also be extended in the future for other universal overrides without adding more flags.

PR #2225, opened 2024-02-14
Allow for importing via PokeBin

malaow3:
See [https://www.smogon.com/forums/threads/pokebin.3736569/](https://www.smogon.com/forums/threads/pokebin.3736569/) for information regarding PokeBin

### DIFF
--- a/play.pokemonshowdown.com/js/client-teambuilder.js
+++ b/play.pokemonshowdown.com/js/client-teambuilder.js
@@ -1387,7 +1387,7 @@
 					type: 'GET',
 					url: url,
 					success: function (data) {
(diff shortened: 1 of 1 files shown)


DaWoblefet:
I'm not even able to access this domain.
image

The reason we supported pokepaste was because it was becoming widely adopted. We already support pokepaste and server-side teams which perform this functionality. If pokebin becomes more popular, we could support it, but this is too soon into the project's lifespan as it is.

malaow3:
Not sure why you cannot access, will look into that! And totally fair reasoning, I can close the PR for now if that is preferred!

DaWoblefet:
I'd be happy to share any info with you that I can. I'll PM you on Smogon.

As it is, yeah, I think I'll close it out.

PR #2226, opened 2024-03-02
Teambuilder: Generate user samples from boxes

im-tofa:
This is an implementation of [a suggestion to display user sample sets from boxes](https://www.smogon.com/forums/threads/teambuilder-improvement-user-sets.3730629/).

The first time a user clicks "Import/Export" in a set view of the teambuilder, this addition will scan through the teambuilder for any boxes with the appropriate format, and cache all sets as user samples. These will then be displayed as buttons. User samples with duplicate names receive a number suffix. 

The cache for a given format is cleared if the user exits a box in said format. Caching reduces repetitive scans each time the user wants to import a set.

I did a quick "performance" test by creating 5000 teams, does not cause hiccups for me at least. Tested on Firefox. The bottleneck, if any, has been the smogon samples fetched using HTTP.

Let me know if I can improve the code quality or functionality, thanks!

This is what it looks like in my test client, see the bottom of the screenshot:

This was the corresponding box:

### DIFF
--- a/play.pokemonshowdown.com/js/client-teambuilder.js
+++ b/play.pokemonshowdown.com/js/client-teambuilder.js
@@ -1786,7 +1786,7 @@
 			// this into an object based on the content-type header.
 			$.get('https://' + Config.routes.client + '/data/sets/' + format + '.json', {}, function (data) {
 				try {
-					self.smogonSets[format] = $.extend(self.smogonSets[format], JSON.parse(data));
+					self.smogonSets[format] = JSON.parse(data);
 				} catch (e) {
 					// An error occured. Mark this as false, so that we don't try to reimport sets for this format
 					// in the future.
@@ -1798,12 +1798,16 @@
 		},
 		getUserSets: function (format) {
 			this.smogonSets[format] = $.extend(this.smogonSets[format], {"user": {}});
+
+			var duplicateNameIndices = {};
 			for(const team of teams) {
 				if(team.format === format && team.capacity === 24) {
 					const setList = Storage.unpackTeam(team.team);
 					for(const pokemon of setList) {
-						console.log(pokemon);
-						this.smogonSets[format]['user'][pokemon.species] = $.extend(this.smogonSets[format]['user'][pokemon.species], {[pokemon.name] : pokemon});
+						var name = pokemon.name + " " + (duplicateNameIndices[pokemon.name] || "");
+						var sets = this.smogonSets[format]['user'][pokemon.species];
+						this.smogonSets[format]['user'][pokemon.species] = $.extend({}, sets, {[name] : pokemon});
+						duplicateNameIndices[pokemon.name] = 1 + (duplicateNameIndices[pokemon.name] || 0);
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/js/client-teambuilder.js
```diff
@@ -1789,34 +1792,71 @@
 					// in the future.
 					self.smogonSets[format] = false;
 				}
+				self.getUserSets(format);
 				self.importSetButtons();
 			}, 'text');
 		},
+		getUserSets: function (format) {
```
im-tofa, line 1799: Should perhaps place an upper limit on the number of sets per Pokémon? Just so someone doesn't have like 45 samples of a Pokémon in a given format, idk what would be a reasonable limit though

play.pokemonshowdown.com/js/client-teambuilder.js
```diff
@@ -1792,31 +1797,76 @@
...
+			this.userSets[format] = this.userSets[format] || {};
+
+			var duplicateNameIndices = {};
+			for(const team of teams) {
+				if(team.format === format && team.capacity === 24) {
+					const setList = Storage.unpackTeam(team.team);
+					for(const pokemon of setList) {
+						var name = pokemon.name + " " + (duplicateNameIndices[pokemon.name] || "");
+						var sets = this.userSets[format][pokemon.species];
+						this.userSets[format][pokemon.species] = $.extend(sets || {}, {[name] : pokemon});
```
Zarel, line 1813: `{[name] : pokemon}` is not ES3, you'll need to do this manually.

Zarel, line 1813:
```suggestion
						var sets = this.userSets[format][pokemon.species] || {};
						sets[name] = pokemon;
						this.userSets[format][pokemon.species] = sets;
```

I think this is what you're trying to do?

play.pokemonshowdown.com/js/client-teambuilder.js
```diff
@@ -1792,31 +1797,76 @@
...
+						duplicateNameIndices[pokemon.name] = 1 + (duplicateNameIndices[pokemon.name] || 0);
+					}
+				}
+			}
+		},
+		clearCachedUserSetsIfNecessary: function (format) {
+			// clear cached user sets if we have just been in a box for given format
+			if(this.curTeam?.capacity === 24 && this.userSets?.[format]) {
```
Zarel, line 1821: `?.` is also not ES3.

Is something wrong with our CI? It should be catching this...

Zarel: Oh, CI needs to be manually approved.

play.pokemonshowdown.com/js/client-teambuilder.js
im-tofa: Exactly, thanks! :)

im-tofa:
Thanks for the feedback @Zarel! I've fixed the linting issues now (no more use of `?.`, `for-of`, `let`, `const` or `{ [name] : pokemon }`), made sure to try it out locally first now which I should've done the first time, my bad :)

After some thinking, I realised this feature is not very easy to discover or "self-explanatory". In an attempt to improve this, I changed the displayed text from "User sets" to "Box sets" and added a hint if no box sets are available for the format, see bottom of below screenshot. Let me know if that's not desired or if there are better alternatives:

bild

PR #2234, opened 2024-03-25
Support rematching with custom rules

Karthik99999:
Server Side: smogon/pokemon-showdown#10221

https://www.smogon.com/forums/threads/rematching-with-custom-rules.3672547/
https://www.smogon.com/forums/threads/when-clicking-rematch-have-it-re-challenge-with-mods-if-there-are-any.3715507/


DaWoblefet:
Thanks Karthik!

PR #2235, opened 2024-03-25
Add support for adding badges in battle

mia-pi-git:
Client end for https://github.com/smogon/pokemon-showdown/pull/10222.


play.pokemonshowdown.com/src/battle.ts
```diff
@@ -3535,6 +3536,15 @@ export class Battle {
 			this.scene.updateSidebar(side);
 			break;
 		}
+		case 'badge': {
+			const slot = args[1] as SideID;
+			var side = this[slot];
+			if (!side) return; // weird
```
Karthik99999, line 3542:
```suggestion
			let side = this.getSide(args[1]);
```

PR #2240, opened 2024-04-03
Teambuilder: Add EV optimization checker

Karthik99999:
https://www.smogon.com/forums/threads/smart-ev-management-in-the-builder.3736254/

image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -23,7 +23,7 @@ module.exports = {
 		"BattleFormats": false, "BattleFormatsData": false, "BattleLearnsets": false, "BattleItems": false, "BattleMoveAnims": false, "BattleMovedex": false, "BattleNatures": false,
 		"BattleOtherAnims": false,  "BattlePokedex": false,"BattlePokemonSprites": false, "BattlePokemonSpritesBW": false, "BattleSearchCountIndex": false, "BattleSearchIndex": false, "BattleArticleTitles": false,
 		"BattleSearchIndexOffset": false, "BattleSearchIndexType": false, "BattleStatIDs": false, "BattleStatNames": false, "BattleStatusAnims": false, "BattleStatuses": false, "BattleTeambuilderTable": false,
-		"ModifiableValue": false, "BattleStatGuesser": false, "BattleText": true, "BattleTextAFD": false, "BattleTextNotAFD": false,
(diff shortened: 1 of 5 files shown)


Karthik99999:
> It’s not necessary to use a brute force approach, or even any kind of iterative approach. The plus, if it moves at all, will always optimally move to the highest raw stat that’s been invested in. The point of moving the plus is to get the highest absolute stat increase from the 10% bonus, regardless of whether that comes from the base stat or EV. The minus, similarly, should only move to the lowest raw stat that’s been invested in to at most what it would be at 252-. Furthermore, the plus can never move from a stat that is boosted further than it could be at 252 neutral (you at least thought of this) while the minus should never move from an uninvested stat as there are no EVs in that stat that can be redistributed for optimization purposes.

Yeah I know brute force is inefficient here. If anyone can implement this without brute force that would be cool, I just did this cause it was easier.

> The focus of this optimizer should be on saving EVs and not on increasing stats. Sometimes improving the stats happens as a bonus when the stat we’re already moving the plus to either happens to be 1 less than a multiple of 11 or is a high base stat that was originally invested to less than what it would be at 0+, but if we can either provide the exact same stat spread for fewer EVs or a slightly stronger spread for the same EVs, we should be giving the spread that uses fewer EVs every time and leaving it up to the user to determine where the remainder goes. After all, once they reinvest their saved EVs after optimizing their spread, the optimizer will automatically run again anyway.

Not sure what you mean by this, since I am trying to return the spread that saved the most EVs. Did you find a case where it returned a spread with higher stats instead of more saved EVs?

Karthik99999:
Ah, I see. Honestly could just rid of the stat check, it finds the right set without it
image

The only reason I had it in the first place was for this Amoonguss set
>Amoonguss  
Ability: Effect Spore  
Level: 50  
Tera Type: Grass  
EVs: 252 HP / 100 Def / 156 SpD  
Bold Nature

Using a +Sp. Def nature on this set leads to no EVs being saved but you have a Sp. Def stat of 121 instead of 120.

>These problems are best avoided by simply not brute forcing in the first place.

I'll try to investigate later this if I have the time. Again, you're more than welcome to make a code suggestion here, as I would also love for this to not be brute forced.

DaWoblefet:
Good work Karthik and pyuk!

PR #2243, opened 2024-04-12
client: escape dots in URL regex

laggron42:
This line looks for URLs without a port specified and sets the default port of 8000. However, in the case of IP subdomains (such as `127.0.0.1-12345.psim.us`), the port `12345` is overwritten because the `-` is ignored by the unescaped dot.

This PR escapes the dots forming the IP address and ensures the port is not being overwritten when specified.

Tested in the console's browser: 
image


Zarel:
Thanks!

PR #2244, opened 2024-04-25
Clarify 2D/3D Model Options

Distrib-ps:
PR for: https://www.smogon.com/forums/threads/have-option-use-bw-sprites-instead-of-xy-models-say-use-2d-sprites-instead-of-3d-models.3741313/


DaWoblefet:
Thanks Distrib!

PR #2248, opened 2024-05-15
fix redundant string slicing and explicit checks for protosynthesis and quarkdrive

Jerick120:
PR: https://github.com/smogon/pokemon-showdown/pull/10300 needs to be merged along with this

### DIFF
--- a/play.pokemonshowdown.com/src/battle-text-parser.ts
+++ b/play.pokemonshowdown.com/src/battle-text-parser.ts
@@ -620,7 +620,7 @@ class BattleTextParser {
 				templateId += 'FromItem';
 			}
 			const template = this.template(templateId, kwArgs.from, effect);
-			return line1 + template.replace('[POKEMON]', this.pokemon(pokemon)).replace('[EFFECT]', this.effect(effect)).replace('[MOVE]', arg3).replace('[STAT]', this.stat(arg3)).replace('[SOURCE]', this.pokemon(kwArgs.of)).replace('[ITEM]', this.effect(kwArgs.from));
+			return line1 + template.replace('[POKEMON]', this.pokemon(pokemon)).replace('[EFFECT]', this.effect(effect)).replace('[MOVE]', arg3).replace('[STAT]', BattleTextParser.stat(arg3)).replace('[SOURCE]', this.pokemon(kwArgs.of)).replace('[ITEM]', this.effect(kwArgs.from));
 		}
 
 		case '-end': {


Zarel:
You say "fix" but this just changes the protocol.

I have two objections here:

- When we change the protocol, the client needs to keep support for the old protocol. If we didn't, old replays would all be broken.
- I don't see any reason to change the protocol. The old one's string slicing may be ugly but it works.

Zarel:
Hm. Yeah, I think that's a good point. Capitalization would also need to be fixed, and we'd need to update the volatile status display on the HP bar, but I think I could see that as being worth it.

play.pokemonshowdown.com/src/battle-text-parser.ts
```diff
@@ -602,11 +602,6 @@ class BattleTextParser {
 				const template = this.template('activate', 'perishsong');
 				return line1 + template.replace('[POKEMON]', this.pokemon(pokemon)).replace('[NUMBER]', num);
 			}
-			if (id.startsWith('protosynthesis') || id.startsWith('quarkdrive')) {
-				const stat = id.slice(-3);
-				const template = this.template('start', id.slice(0, id.length - 3));
-				return line1 + template.replace('[POKEMON]', this.pokemon(pokemon)).replace('[STAT]', BattleTextParser.stat(stat));
-			}
 			let templateId = 'start';
```
Zarel, line 605:
```suggestion
			if (id.startsWith('protosynthesis') || id.startsWith('quarkdrive')) {
				const stat = id.slice(-3);
				id = id.slice(0, 3);
				if (stat) arg3 = stat;
			}
			let templateId = 'start';
```

play.pokemonshowdown.com/src/battle-text-parser.ts
```diff
@@ -620,7 +615,7 @@ class BattleTextParser {
 				templateId += 'FromItem';
 			}
 			const template = this.template(templateId, kwArgs.from, effect);
-			return line1 + template.replace('[POKEMON]', this.pokemon(pokemon)).replace('[EFFECT]', this.effect(effect)).replace('[MOVE]', arg3).replace('[SOURCE]', this.pokemon(kwArgs.of)).replace('[ITEM]', this.effect(kwArgs.from));
+			return line1 + template.replace('[POKEMON]', this.pokemon(pokemon)).replace('[EFFECT]', this.effect(effect)).replace('[MOVE]', arg3).replace('[STAT]', this.stat(arg3)).replace('[SOURCE]', this.pokemon(kwArgs.of)).replace('[ITEM]', this.effect(kwArgs.from));
```
Zarel, line 618:
```suggestion
			return line1 + template.replace('[POKEMON]', this.pokemon(pokemon)).replace('[EFFECT]', this.effect(effect)).replace('[MOVE]', arg3).replace('[STAT]', BattleTextParser.stat(arg3)).replace('[SOURCE]', this.pokemon(kwArgs.of)).replace('[ITEM]', this.effect(kwArgs.from));
```

PR #2249, opened 2024-05-16
Add Elo filter 1000

Distrib-ps:
PR for https://www.smogon.com/forums/threads/give-an-option-for-1000-minimum-elo-in-the-battles-tab-to-filter-out-challenge-battles.3743305/


Zarel:
Thanks!

PR #2251, opened 2024-05-21
Fix tooltips getting cleared when using switch moves

Karthik99999:
https://www.smogon.com/forums/threads/reset-tooltips-on-pivot-move-faint-replacement.3733788/


DaWoblefet:
Thanks Karthik!

PR #2252, opened 2024-06-01
Override Fling and Natural Gift data for items in past gens

Karthik99999:
https://www.smogon.com/forums/threads/ddp-natural-gift-shows-wrong-base-power-in-tooltip.3744325/


DaWoblefet:
Thanks Karthik!

PR #2255, opened 2024-06-24
Replays: Fix More Replays button on mobile

Karthik99999:
Currently, the `‹ More Replays` button that is shown on mobile always takes you back to the main replays page. This change should preserve the original search query.

### DIFF
--- a/replay.pokemonshowdown.com/src/replays-battle.tsx
+++ b/replay.pokemonshowdown.com/src/replays-battle.tsx
@@ -494,7 +494,7 @@ export class BattlePanel extends preact.Component<{id: string}> {
         {/* {} <code>{this.keyCode}</code> */}
       </p> : <p>&nbsp;</p>}
       {!PSRouter.showingLeft() && <p>
-        <a href={`./${PSRouter.leftLoc || ''}`} class="button"><i class="fa fa-caret-left"></i> More replays</a>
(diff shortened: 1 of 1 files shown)


replay.pokemonshowdown.com/src/replays-battle.tsx
```diff
@@ -494,7 +494,7 @@ export class BattlePanel extends preact.Component<{id: string}> {
         {/* {} <code>{this.keyCode}</code> */}
       </p> : <p>&nbsp;</p>}
       {!PSRouter.showingLeft() && <p>
-        <a href="." class="button"><i class="fa fa-caret-left"></i> More replays</a>
+        <a href={`./${PSRouter.leftLoc || ''}`} class="button"><i class="fa fa-caret-left"></i> More replays</a>
```
Zarel, line 497: Good catch! But probably more complex than it needs to be.

```suggestion
        <a href={PSRouter.leftLoc || '.'} class="button"><i class="fa fa-caret-left"></i> More replays</a>
```

Zarel:
Good catch! Thanks!

replay.pokemonshowdown.com/src/replays-battle.tsx
Karthik99999: Are you sure this change is correct? This is what I tried at first but it kept appending the query to the end of the replay link rather than going back to the base link of replay.pokemonshowdown.com.

PR #2256, opened 2024-07-03
Properly display legal egg moves in teambuilder

dot-Comfey:
https://www.smogon.com/forums/threads/bloodmoon-ursaluna-egg-moves.3745741/


DaWoblefet:
Thanks dot Comfey!

PR #2259, opened 2024-07-08
Fix issue where 1.2x damage calculation is not showing properly

earljoves:
Based on the description laid out in this thread: https://www.smogon.com/forums/threads/add-the-1-2x-damage-calculation-of-ogerpon-wellspring-hearthflame-cornerstone-mask-to-move-ui.3746567/

Local testing LGTM!

Please send any feedback my way, more than willing to learn here!

### DIFF
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -2381,9 +2381,9 @@ class BattleTooltips {
 		}
 		if (speciesName === 'Ogerpon') {
 			const speciesForme = value.pokemon.getSpeciesForme();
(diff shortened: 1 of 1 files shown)


DaWoblefet:
I just tried pulling your changes down and don't notice any difference. The base power boost doesn't show up before or after Terastallization.
image
image

DaWoblefet:
This does fix it for base Ogerpon forms, but not Terastallized ones.
image

earljoves:
Thanks for the quick feedback, @DaWoblefet! Just pushed a new commit with the changes, and Terastallized forms look good on local machine after some quick happy path testing.

DaWoblefet:
Good work! Thanks, earljoves!

PR #2260, opened 2024-07-09
Fix crash (client)

dot-Comfey:
No crash here, but a change made to sync the teambuilder after a server-side change to fix a validator crash.


DaWoblefet:
Thanks dot Comfey!

PR #2262, opened 2024-07-11
Add Gen 5 ZU Support

AlexMMcConnell:
Hello! This is my best attempt at adding TeamBuilder support for Gen 5 ZU. I wasn't seeing any changes that I made within the build index occurring on my local client (even when I tried to mess with already functional formats to break the), so this fix admittedly could be incomplete - if there's a way to make the changes reflected in my local client I'd love to see it. If you'd like me to do some rebasing of commits or anything of that sort as well, please let me know. Thank you!

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -565,8 +565,10 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 			BattleTeambuilderTable.monotypeBans = monotypeBans;
 			BattleTeambuilderTable.formatSlices = formatSlices;
 		} else {
-			BattleTeambuilderTable.gen5zuBans = gen5zuBans;
 			BattleTeambuilderTable[gen] = {};
+			if (genNum === 5) {
+				BattleTeambuilderTable[gen].gen5zuBans = gen5zuBans;
+			}
 			BattleTeambuilderTable[gen].overrideTier = overrideTier;
 			BattleTeambuilderTable[gen].tiers = tiers;
 			BattleTeambuilderTable[gen].items = items;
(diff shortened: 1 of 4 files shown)


play.pokemonshowdown.com/src/battle-dex-search.ts
```diff
@@ -1061,6 +1060,13 @@ class BattlePokemonSearch extends BattleTypedSearch<'pokemon'> {
 				});
 			}
 		}
+		if (format === 'zu' && dex.gen === 5) {
+			console.log(table.gen5zuBans)
+			// tierSet = tierSet.filter(([type, id]) => {
+			// 	if (id in table.gen5zuBans) return false;
+			// 	return true;
+			// });
+		}
```
KrisXV, line 1069:
```suggestion
		if (format === 'zu' && dex.gen === 5 && table.gen5zuBans) {
			tierSet = tierSet.filter(([type, id]) => {
				if (id in table.gen5zuBans) return false;
				return true;
			});
		}
```

build-tools/build-indexes
```diff
@@ -558,6 +565,7 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 			BattleTeambuilderTable.monotypeBans = monotypeBans;
 			BattleTeambuilderTable.formatSlices = formatSlices;
 		} else {
+			BattleTeambuilderTable.gen5zuBans = gen5zuBans;
```
KrisXV, line 568:
```suggestion
			if (genNum === 5) BattleTeambuilderTable[gen].gen5zuBans = gen5zuBans;
```

KrisXV:
try testing it now with the changes i just pushed. this should be a step in the right direction at the very least?

@AlexMMcConnell 
(also sorry for no reply on discord, been busy)

PR #2263, opened 2024-07-14
Chat: Support nameless /html messages

Karthik99999:
Currently, if you send an `/html` message with `&` as the identity, it treats `&` as a username and shows it before the message with a colon.
image
This change should allow the message to look like this instead
image
I would like this feature for the auctions plugin so that I can send messages with username colors without sacrificing the timestamp like I would be doing by sending `/raw`.


DaWoblefet:
Thanks Karthik!

PR #2264, opened 2024-07-18
Fix getAllyAbility in Gen < 7

Karthik99999:


DaWoblefet:
This should also fix ``getAllyAbility`` with Neutralizing Gas etc. Thanks Karthik!

PR #2267, opened 2024-07-24
Don't guess spreads when there are empty moves

DaWoblefet:
Currently, when the teambuilder supplies a guessed spread, it will consider moves with the empty string for its check. This prevents the teambuilder from guessing spreads after the user only puts in 3 moves.

I'm not sure if the correct fix is here or in some way deleting the empty string move from ``set.moves`` altogether at some other point.

### DIFF
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -2612,9 +2612,10 @@ class BattleStatGuesser {
 
 		if (set.moves.length < 1) return '?';
 		let needsFourMoves = !['unown', 'ditto'].includes(species.id);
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -2612,9 +2612,10 @@ class BattleStatGuesser {
 
 		if (set.moves.length < 1) return '?';
 		let needsFourMoves = !['unown', 'ditto'].includes(species.id);
+		let hasFourValidMoves = (set.moves.length >= 4) && !set.moves.includes('');
```
KrisXV, line 2615:
```suggestion
		let hasFourValidMoves = set.moves.length >= 4 && !set.moves.includes('');
```

PR #2270, opened 2024-07-27
Teambuilder: Debounce searching teams

PartMan7:
Wasn't fun having my browser frozen for half a minute every time I typed more than two letters in the teambuilder search; this PR introduces debounced calls to updateTeamList() instead of direct calls.

Tested locally on Chrome 127.0.6533.72

### DIFF
--- a/play.pokemonshowdown.com/js/client-teambuilder.js
+++ b/play.pokemonshowdown.com/js/client-teambuilder.js
@@ -3241,10 +3241,7 @@
 			this.chartSet(val, selectNext);
 		},
 		searchChange: function (e) {
-			if (this.searchTimeout) clearTimeout(this.searchTimeout);
-
-			var searchVal = e.currentTarget.value;
-
+			var DEBOUNCE_THRESHOLD_TEAMS = 500;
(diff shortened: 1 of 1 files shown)


Zarel:
Ten years ago this was instantaneous enough, and computers have only gotten better since then. Probably this should at least be a setting?

PartMan7:
Takes about 5-10s per character input on my device (which has fairly decent specs); this isn't an issue for people who don't store teams, but for people with a lot (I'm around 9K or so), it is... problematic, to say the least. I'm not sure if there's any benefit to _not_ having it debounced (except the removal of the 300ms typing delay), but it offers pretty noticeable benefits for users that do.

data/pokemon-showdown
PartMan7, line 1: Wait just saw this; will remove

Zarel:
Why does this only get laggy at a high team count? Is the team list still hidden somewhere in the DOM?

Zarel:
Probably _that's_ the bug to fix.

Zarel:
Thanks! This is a good stopgap until we get a real fix in.

PR #2272, opened 2024-08-03
Support tabcompleting commands

mia-pi-git:
Self-explanatory. Will merge in a day or so if no one has any complaints, so please eyeball sooner rather than later.

### DIFF
--- a/play.pokemonshowdown.com/js/client-chat.js
+++ b/play.pokemonshowdown.com/js/client-chat.js
@@ -366,7 +366,6 @@
 					if (this.tabComplete.searchPending) return true; // wait
 					this.tabComplete.isCommand = true;
 					this.tabComplete.searchPending = true;
-					app.send('/crq cmdsearch ' + text);
 					var self = this;
 					app.once('response:cmdsearch', function (data) {
 						delete self.tabComplete.searchPending;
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/js/client-chat.js
```diff
@@ -357,43 +357,73 @@
...
+					return x.startsWith(prefix);
+				}).length;
+				var isCommandSearch = text.startsWith('/') || text.startsWith('!');
+				if (isCommandSearch && shouldSearchCommands) {
+					if (this.tabComplete.searchPending) return true; // wait
+					this.tabComplete.isCommand = true;
+					this.tabComplete.searchPending = true;
+					app.send('/crq cmdsearch ' + text);
+					var self = this;
+					app.once('response:cmdsearch', function (data) {
```
Karthik99999, line 371: Isn't the handler supposed to be defined before the `/crq` is sent?

mia-pi-git: It works either way. But I suppose there's no real harm in being cautious.

DaWoblefet:
Typing /z (or anything else that doesn't return any commands) sends the request twice.

mia-pi-git:
> Typing /z (or anything else that doesn't return any commands) sends the request twice.

This has been fixed.

PR #2276, opened 2024-08-16
Fix formatting of Glicko-1 on ladder page

Karthik99999:
Before:
image
After:
image

Is there a reason that the widths are even specified though? Getting rid of all the width values also makes the table keep everything on one line.


Zarel:
Thanks!

Zarel:
I think the widths are nice here; keeps things consistent.

PR #2277, opened 2024-08-30
Teambuilder: Pass `this` context correctly to updateTeamList

PartMan7:
No idea how this worked locally before; tested again and works about right now


mia-pi-git:
Thanks!

PR #2279, opened 2024-09-02
Use only current line in tab-completing commands

dot-Comfey:
https://www.smogon.com/forums/threads/bug-report.3749269/

At this point, it's a bit more than a bug fix. It seems intuitive that attempting to perform a tab-complete on one line should search for commands on that line, but it doesn't because it's searching for commands using the entire text box, which was also causing the issue of sending lines into chat.

Besides that, I also reverted the behavior of both commands and usernames being able to be candidates of the same autocomplete, while retaining the ability to autocomplete usernames if placed after the command; I'm not sure how it was helpful to go between commands and usernames.


mia-pi-git:
LGTM, thanks!

PR #2280, opened 2024-09-10
Display exact HP when using Exact HP Mod

HisuianZoroark:
Exact HP mod sends current HP and Max HP to both the user and the opponent via server protocol messages, but displays percent regardless unless the user turns on hardcore mode, making it information almost only accessible via the console. This pull request displays the current hp in the hp bar and the current and Max HP in the pokemon's tooltips for easier accessibility.

Before:
https://github.com/user-attachments/assets/606f60bc-a4ee-4666-9144-c5432f517855

After:
https://github.com/user-attachments/assets/d103e177-005f-4132-a425-0d14e49a2008

### DIFF
--- a/play.pokemonshowdown.com/src/battle.ts
+++ b/play.pokemonshowdown.com/src/battle.ts
@@ -582,9 +582,10 @@ export class Pokemon implements PokemonDetails, PokemonHealth {
 		return percentage * maxWidth / 100;
 	}
 	getHPText(precision = 1) {
-		return Pokemon.getHPText(this, precision);
+		return Pokemon.getHPText(this, this.side.battle.reportExactHP, precision);
 	}
-	static getHPText(pokemon: PokemonHealth, precision = 1) {
+	static getHPText(pokemon: PokemonHealth, exactHP: boolean, precision = 1) {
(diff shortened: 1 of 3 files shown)


play.pokemonshowdown.com/src/battle.ts
```diff
@@ -3442,6 +3444,7 @@ export class Battle {
 				this.messageFadeTime = 40;
 				this.isBlitz = true;
 			}
+			if (ruleName === 'Exact HP Mod') this.reportExactHP = true;
```
Karthik99999, line 3447: why are you making a new property instead of checking `this.rules['Exact HP Mod']`

HisuianZoroark: was primarily was looking at something else was implemented and decided to mirror, if it should be checked that way then sure

Karthik99999: I left blitz and species clause when I added the `rules` object just to minimize refactoring at the time, but I think any new code that needs rules should just use the `rules` object if the only thing that needs to be checked is if the rule is in play.

PR #2282, opened 2024-09-15
Update ladder names

Glazelf:
VGC 2023 --> VGC 2024
gen9vgc2023regulationc --> gen9vgc2024regh
Removed NFE and BSS since they don't seem to be playable ladders anymore so the ladder returns a dead list of players with a rank of 1500. If this causes issues we should replace them with newer metagames but I'm not too wellversed in Smogon othermetas.


pokemonshowdown.com/users.php
```diff
@@ -63,17 +63,23 @@
 	'gen9pu' => 'PU',
 	'gen9lc' => 'Little Cup',
 	'gen9monotype' => 'Monotype',
```
KrisXV, line 65:
```suggestion
	'gen9monotype' => 'Monotype',
	'gen9bssregh' => 'Battle Stadium Singles Regulation H',
```

PR #2283, opened 2024-09-21
Fixed missing HTML sanitization

swordfishtr:
It was possible to run arbitrary javascript from restoring a crafted teambuilder backup. Let me know if a Smogon bug report with clarification is needed.

A sample: paste the following into the backup menu and restore, the effect is immediate.

=== [<script>alert("hi i'm from an arbitrary script!");</script>] Untitled 1 ===

### DIFF
--- a/play.pokemonshowdown.com/js/client-teambuilder.js
+++ b/play.pokemonshowdown.com/js/client-teambuilder.js
@@ -494,7 +494,7 @@
 					// support dragging and dropping buttons.
 					buf += '<li><div name="edit" data-value="' + i + '" class="team';
 					if (team.capacity === 24) buf += ' pc-box';
-					buf += '" draggable="true">' + formatText + '<strong>' + BattleLog.escapeHTML(team.name) + '</strong><br /><small>';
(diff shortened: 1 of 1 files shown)


Slayer95:
Confirmed.

A tangential concern is that importing https://pokepast.es/17af0dd9b7bcc5a8 as a single team hangs the browser.

swordfishtr:
> A tangential concern is that importing https://pokepast.es/17af0dd9b7bcc5a8 as a single team (using the URL) hangs the browser.

oh shit, oh fuck, that's caused by an unterminated `while(true)` look here:

https://github.com/smogon/pokemon-showdown-client/blob/5512e76866463722c60ec09f1ab6f5c1e31db67c/play.pokemonshowdown.com/js/storage.js#L1095

the stuff to do with `misc` is irrelevant, check the break line after those. importing so much as `]` bypasses it. i'll try to fix this in another pr (good catch!)

swordfishtr:
> A tangential concern
I have a fix for this issue now but i've messed up by creating this pr from my main branch (web interface is confusing and i didn't know any better). Closing to reopen properly, sorry!

PR #2285, opened 2024-09-25
Fixed a case of endless loop

swordfishtr:
It was possible to freeze the tab by importing a team/set containing `]` with no `|` after.

(only if the whole import is 1 line or 2 lines with the 2nd one being blank - this leads to the importer using `Storage.unpackTeam()` which is intended functionality.)

I took care to not conflict with #2222 . hopefully we can discuss rewriting the packed team format after these prs are resolved ...

credit for finding this bug: https://github.com/smogon/pokemon-showdown-client/pull/2283#issuecomment-2370371178

### DIFF
--- a/play.pokemonshowdown.com/js/storage.js
+++ b/play.pokemonshowdown.com/js/storage.js
@@ -1107,7 +1107,7 @@ Storage.unpackTeam = function (buf) {
 			set.dynamaxLevel = (misc[4] ? Number(misc[4]) : 10);
 			set.teraType = misc[5];
 		}
-		if (j < 0) break;
+		if (j < 0 || buf.indexOf('|', j) < 0) break;
 		i = j + 1;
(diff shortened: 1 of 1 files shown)


Slayer95:
> rewriting the packed team format after these prs are resolved ...

Team packing originally seemed like a great idea, since teams should be able to be efficiently stored in ``localStorage`` to prevent browser-initiated cleanups (some people have thousands of teams.

However, newer gens have released all sorts of new mechanics that mess with teams formats. Additionally, there has been some developments in data compression, such as:
- Pokémon Showdown server adopting the per-message deflate websocket extension.
- Browser vendors implementing ``CompressionStream`` and ``DecompressionStream``, exposing native ``gzip`` and ``deflate`` support.

You can probably bring this discussion to the Smogon Development Discord server.

PR #2291, opened 2024-10-21
Responsive behaviour of the main tab bar 

christopher-dev-91:
Responsive behaviour of the main tab bar (https://github.com/smogon/pokemon-showdown-client/issues/2290)


krst221:
did you only approve one of the commits? it's pointless without the other one

PR #2300, opened 2024-11-24
News: Add support for minisprites

KrisXV:
I used the syntax of `[psicon thing="object"]` to somewhat mirror the html syntax and for ease of replace for regex


mia-pi-git:
LGTM, ty!

PR #2302, opened 2024-11-27
Support blank target in -item message

MathyFurret:
For Gen 4-5 Frisk. If no target is sent with an -item message, the client first checks if there is exactly 1 possible target, and updates its item.

### DIFF
--- a/play.pokemonshowdown.com/src/battle.ts
+++ b/play.pokemonshowdown.com/src/battle.ts
@@ -2225,10 +2225,24 @@ export class Battle {
 			break;
 		}
 		case '-item': {
-			let poke = this.getPokemon(args[1])!;
+			let poke = this.getPokemon(args[1]);
 			let item = Dex.items.get(args[2]);
 			let effect = Dex.getEffect(kwArgs.from);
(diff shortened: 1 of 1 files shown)


MathyFurret:
Guess we're on an older TS version, hold on

MathyFurret:
https://youtu.be/kSglNQBFGQg

play.pokemonshowdown.com/src/battle.ts
```diff
@@ -2229,6 +2229,20 @@ export class Battle {
 			let item = Dex.items.get(args[2]);
 			let effect = Dex.getEffect(kwArgs.from);
 			let ofpoke = this.getPokemon(kwArgs.of);
+			if (!poke) {
+				if (effect.id === 'frisk') {
+					const possibleTargets = ofpoke!.side.foe.active.filter(p => p !== null);
```
urkerab, line 2234: Wouldn't it be easier to use `find` instead of `filter`?

MathyFurret: find doesn't account for having >1 possible target.

PR #2310, opened 2024-12-19
Staff FAQ: Fix missed Global Admin symbol updates 

bingingem:
This was reported in help room, and I think I caught all of them.


DaWoblefet:
Thank you much!

PR #2312, opened 2024-12-27
Complementary patch for 35 Pokes

swordfishtr:
All context here: https://github.com/smogon/pokemon-showdown/pull/10772

This PR should only be merged after the linked one; the teambuilder tables won't build without it.


swordfishtr:
Fixed by Zarel: https://github.com/smogon/pokemon-showdown/pull/10932

PR #2317, opened 2025-01-17
Teambuilder: Fix HPIvs ignoring hypertraining

thejetou:
There were two cases where hidden powers IVs would be written to a set that could be hypertrained:

1. Set/team imports
2. When maxing the IVs of a stat (eg. when a physical move is chosen)

The first case has been fixed by making `Storage#importTeam` take a `format` paramater to determine whether hypertraining is possible. The `format` is only needed when the import is a set since the format is not written to a set so it needs to be specified.

In the second case, its a simple change in `TeamBuilderRoom#changeMove` to adjust maxing a stat to set it to `31` if it can be hypertrained.

### DIFF
--- a/play.pokemonshowdown.com/js/client-teambuilder.js
+++ b/play.pokemonshowdown.com/js/client-teambuilder.js
@@ -1149,19 +1149,11 @@
 				var reader = new FileReader();
 				var self = this;
 				reader.onload = function (e) {
-					var team;
-					try {
-						team = Storage.packTeam(Storage.importTeam(e.target.result));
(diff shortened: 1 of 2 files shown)


Zarel:
If I'm understanding correctly, this is a duplicate of #2143, and is fixed in the Preact client. Please reopen if I'm wrong.

Zarel:
I fixed it in bfb2813c720d61524a1f2, around 5 days ago, around when you said it was still broken. Good to know there's nothing left to do there.

PR #2318, opened 2025-01-18
Add National Dex Ubers UU teambuilder support

Zrp200:
It seems the only thing actually necessary to make this work was to make sure `ubersUUBans` was generated for Natdex as well. No other file changes seem to be necessary.

Tested locally, note the absence of Mega Mewtwo Y.
image

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -465,7 +465,7 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 			tierTable[tier].push(id);
 
 			if (genNum === 9) {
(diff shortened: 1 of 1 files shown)


Zrp200:
@KrisXV bumping. National Dex Ubers UU is set to have a ladder this month, and this would greatly aid teambuilding for this tier at very little overall cost.

As it is now, many of the viable picks are in the "Ubers" section, but all of the bans are also in that section, making teambuilding actively confusing without the builder aid.

I don't mind if this support is yanked at the end of the month when we lose the ladder, but I would like to try anything I can to ensure the ladder gets an adequate amount of activity and the tier gets the exposure intended from the spotlight.

PR #2319, opened 2025-01-19
Use most recent format as default format

bingingem:
Followed DaWoblefet's suggested approach to just use the most recently selected format as the default format. 
https://www.smogon.com/forums/threads/select-default-format.3670473/

### DIFF
--- a/play.pokemonshowdown.com/js/client-mainmenu.js
+++ b/play.pokemonshowdown.com/js/client-mainmenu.js
@@ -701,7 +701,7 @@
 			if (!this.searching || $.isArray(this.searching) && !this.searching.length) {
 				var format = $formatButton.val();
 				var teamIndex = $teamButton.val();
-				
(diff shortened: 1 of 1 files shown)


bingingem:
There's currently an issue where the first team in the teambuilder is selected on non-random formats. It's because a non-undefined ``teamIndex`` is being passed into renderTeams (""), and ``this.curTeamIndex`` is set to 0. I'm gonna look into that

Slayer95:
This is missing logic in case ``Storage`` loads after the mainmenu.
- See the ``Storage.whenPrefsLoaded()`` handler and augment it.
- Format selection and LFAB should probably be disabled until prefs are loaded (or a timeout is exceeded.)

bingingem:
Thanks for the review and suggestions, Slayer. I'll take a look and update once I get those cases handled.

PR #2324, opened 2025-02-11
Don't force Tera types because of Hackmons

### LINKED ISSUE #2314: Server compatibility: [Hackmons] allow Ogerpon and Terapagos to choose other Tera types
In the cartrige, any Ogerpon form is able to Terastallize into Fire, Grass, Rock or Water types, while Terapagos is able to Terastallize into any type.

If [this PR](https://github.com/smogon/pokemon-showdown/pull/10814) is accepted, it will correctly handle those transformations.

Currently, the server does not support changing the Tera types of these Pokémon, so adding that functionality would be beneficial (at least for some metagames like Hackmons).

andrebastosdias:
In the cartrige, any Ogerpon form is able to Terastallize into Fire, Grass, Rock or Water types, while Terapagos is able to Terastallize into any type. Currently, the server does not support changing the Tera types of these Pokémon, so adding that functionality is crucial for metagames like Hackmons.

This PR should be merged together with [this server PR](https://github.com/smogon/pokemon-showdown/pull/10814), that it will correctly handle those transformations.

Closes #2314


Zarel:
I suppose it's a better name. I'm fine with this...

Zarel:
Preact teambuilder was written by me and thus already works as it should, and doesn't need changes besides the rename. :p

andrebastosdias:
Just because I changed it in the server PR to align with `requiredAbility`, `requiredItem`/`requiredItems`, and `requiredMoves`. In the end, it works the same way as those.

andrebastosdias:
Except in some weird cases, the game will softlock if Ogerpon tries to Terastallize into a Type different than Grass, Fire, Rock or Water. In the server, I implemented it like the Desync Mod, and just ignore the Terastallization and send a  I don't know if you want to make

Zarel:
Thanks!

PR #2326, opened 2025-02-22
Update to ESLint 9

Zarel:
(See also https://github.com/smogon/pokemon-showdown/pull/10926 )

This finally removes the tslint dependency and switches to eslint.

There are a lot of other changes here, too, to bring the codebase up to server standards. TSLint never had much in the way of indentation enforcement.

Not very happy about eslint splitting itself up over 6 dependencies, or its documentation over three websites, nor how poorly documented the new flat config is, but I mean, eslint's gonna eslint. Customizing would be even harder if we tried to use Biome or something. They mostly seem to go full Prettier.


Zarel:
TODO:
- decide whether to move to Prettier style for curly braces (spaces around curlies but not around square brackets): `{ a: 1, b: 2 }` but `[1, 2]`
- decide whether we're going to support no-type linting

urkerab:
Removing the trailing object commas from `.js` files is to restore support for IE8?

Zarel:
> Removing the trailing object commas from `.js` files is to restore support for IE8?

At the point we originally decided to go ES3, it was for Safari 5 (the last Safari version supported by the iPad 1).

Now? I don't even know. A while ago someone showed off PS running on their refrigerator. We've always tried to support ES3 and the broken eslint config was the only reason we hadn't already, so we might as well. New code should mostly be in the TS files, where ES3 support isn't actually difficult.

Zarel:
Being able to lint with `npx eslint` will be really nice in the future, though. No more need for complicated command-line invocations. I don't think we'll need the multiple config files on server, either.

Zarel:
`eslint-ps-standard.mjs` is intended to be shared between client and server. It'd be _nice_ to have some automated way to keep them in sync, but for now they'll just be duplicated.

PR #2327, opened 2025-02-24
Simulate Gen 1 Fight Button for Wrap

HisuianZoroark:
Adds a warning if you are partial trapping/partially trapped, also collaborated with @inashellnut

To be merged with smogon/pokemon-showdown#10936


SerineMolecule:
These feel like they should be in `finalDecisionMove`, rather than their own variables? Maybe `finalDecisionMove` should be a string enum of possible reasons?

Zarel:
These feel like they should be in finalDecisionMove, rather than their own variables? Maybe finalDecisionMove should be a string enum of possible reasons?

Ditto for `maybeDisabled` and `maybeTrapped` server-side.

HisuianZoroark:
> Ditto for `maybeDisabled` and `maybeTrapped` server-side.

I mean, sure? My biggest concern is that maybePartiallyTrapping takes a 'string' argument for a move so im not sure how to go about it

Zarel:
We got to an understanding on Discord. The client only needs to know whether it's possible to cancel after using a move, and whether it's possiblce to cancel after switching, and a reason for each.

play.pokemonshowdown.com/js/client-battle.js
```diff
@@ -1296,6 +1303,14 @@
 			this.choice.choices[this.choice.choices.length - 1] += ' ' + posString;
 			this.chooseMove();
 		},
+		chooseFight: function () {
+			if (!this.choice) return;
+			this.tooltips.hideTooltip();
+
+			// TODO?: change this action
+			this.choice.choices.push('fight');
```
Zarel, line 1311:
```suggestion
			this.choice.choices.push('testfight');
```

Zarel:
Current design:

image

Please ignore the Kabutops, I just played Gen 1 randbat and used surgery to graft Wrap onto the first pokemon I saw.

Zarel:
To be clear, the text is not finalized, but it makes sense to make the current version playable while waiting for the RBY council to slowly deliberate on it.

PR #2328, opened 2025-02-27
Fix crash with tooltips, potientially causing unclickable moves

HisuianZoroark:


HisuianZoroark:
Relevant console error: ``Uncaught TypeError: Dex.Move is not a constructor``

PR #2331, opened 2025-03-02
Move room owners above admins once again

DaWoblefet:
Regression from the transition back to ~ for admin symbol; see https://github.com/smogon/pokemon-showdown-client/pull/1618


Zarel:
Looks good!

PR #2332, opened 2025-03-03
Teambuilder: Fix error on set import for formats without dex sets

FeluciaPS:
Currently throws a console error when you click "Import/Export" on a pokemon set when building a team for a format that doesn't have sets on the smogdex (Example: Gen 4 1v1). Notably this prevents box sets from showing up for those formats


Zarel:
Thanks!

PR #2335, opened 2025-03-10
Remove broadcast Restriction on avatar.

WifiLatency:
Depends on https://github.com/smogon/pokemon-showdown/pull/10961

Client end of implementing the suggestion: https://www.smogon.com/forums/threads/avatar-avatar-to-be-broadcastable-showing-the-avatar-sprite-and-the-credit.3759095/

Removes the restriction on sending a broadcast prefix '!' with the avatar command to allow for users to broadcast it in chat. Prevents caching the selected avatar when using the '!' prefix to send the command.

### DIFF
--- a/play.pokemonshowdown.com/js/client-chat.js
+++ b/play.pokemonshowdown.com/js/client-chat.js
@@ -1101,15 +1101,14 @@
 				return text;
 
 			case 'avatar':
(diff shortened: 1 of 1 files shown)


Zarel:
Thanks!

PR #2343, opened 2025-04-02
Implement Flame/Toxic orb texts

Karthik99999:
https://www.smogon.com/forums/threads/wrong-text-for-flame-orb.3762624/

Apparently the templates already exist server side and are just not implemented? Better late than never I guess.


Zarel:
Thanks!

PR #2344, opened 2025-04-03
Add a volume slider to Replays

Confettimaker:
Added a volume slider to the replay client, updated the `.optgroup` class to adjust the look.

image

### DIFF
--- a/replay.pokemonshowdown.com/src/replays-battle.tsx
+++ b/replay.pokemonshowdown.com/src/replays-battle.tsx
@@ -341,6 +341,14 @@ export class BattlePanel extends preact.Component<{ id: string }> {
 		BattleSound.setBgmVolume(muted === 'musicoff' ? 0 : 65.88125800126558);
 		this.forceUpdate();
 	};
+	changeVolume = (e: Event) => {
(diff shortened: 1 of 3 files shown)


replay.pokemonshowdown.com/src/replays-battle.tsx
```diff
@@ -341,6 +341,14 @@ export class BattlePanel extends preact.Component<{ id: string }> {
 		BattleSound.setBgmVolume(muted === 'musicoff' ? 0 : 65.88125800126558);
 		this.forceUpdate();
 	};
+	changeVolume = (e: Event) => {
+		const volume = Number((e.target as HTMLSelectElement).value);
+		BattleSound.setBgmVolume(volume);
+		BattleSound.setEffectVolume(volume);
+		this.battle?.setMute(true);
+		this.battle?.setMute(false);
```
Confettimaker, line 349: I have not found a way to get around doing this. Calling the Update function has no effect.

Zarel: Yeah, I'll probably have to fix it in `BattleSound` at some point.

Zarel:
Thanks!

PR #2345, opened 2025-04-09
Preact: Fix UserPanel

thejetou:
- User search works on the first search
- Stop sending the server empty `/cmd userdetails`
- User search popup uses the server username instead of the searched name


Zarel:
Heh it looks like I already fixed the big part, although these other fixes are still good.

Zarel:
Ugh this is also not very compatible with the thing where I just moved the file for popups... Probably it'd be easiest if I just manually merged this PR.

PR #2348, opened 2025-04-11
Teams: Use logical `&&` operators for inline if statements

KrisXV:


Zarel:
Btw @mia-pi-git for future reference, `null` does the same thing as `<></>` except it doesn't have object creation overhead.

PR #2352, opened 2025-04-12
Preact: Add /showjoins and other client side commands

ISenseAura:
for the /rank command, i mostly copy pasted the code from old client, it uses jquery ($) to request for the userrank details.
let me know if we dont want that

### DIFF
--- a/play.pokemonshowdown.com/src/client-main.ts
+++ b/play.pokemonshowdown.com/src/client-main.ts
@@ -801,10 +801,10 @@ export class PSRoom extends PSStreamModel<Args | null> implements RoomOptions {
 				} else {
 					serverShowjoins[room] = 0;
 				}
-				this.receiveLine([`c`, "", `Join/leave messages on room ${room}: AUTOMATIC`]);
+				this.receiveLine([`c`, "", `Join/leave messages on room ${room}: OFF`]);
 			} else {
 				serverShowjoins = { global: 0 };
-				this.receiveLine([`c`, "", `Join/leave messages: AUTOMATIC`]);
+				this.receiveLine([`c`, "", `Join/leave messages: OFF`]);
 
 			}
 			showjoins[PS.server.id] = serverShowjoins;
@@ -1186,25 +1186,6 @@ export const PS = new class extends PSModel {
 				continue;
 			}
 
-			case 'j':
-			case 'J': {
-				if (!room) continue;
-				let showjoins = this.prefs.showjoins || {};
-				let serverShowjoins = showjoins[this.server.id] || {};
-				if (!serverShowjoins[room.id] && !serverShowjoins.global) continue;
-				room.receiveLine([`j`, args[1]]);
-				break;
-			}
-
-			case 'l':
-			case 'L': {
-				if (!room) continue;
-				let showjoins = this.prefs.showjoins || {};
-				let serverShowjoins = showjoins[this.server.id] || {};
-				if (!serverShowjoins[room.id] && !serverShowjoins.global) continue;
-				room.receiveLine([`l`, args[1]]);
-				break;
-			}
 			}
 			room?.receiveLine(args);
 		}
--- a/play.pokemonshowdown.com/src/client-main.ts
+++ b/play.pokemonshowdown.com/src/client-main.ts
@@ -574,15 +574,6 @@ class PSServer {
 	getGroup(symbol: string | undefined) {
 		return this.groups[(symbol || ' ').charAt(0)] || this.defaultGroup;
 	}
-	getActionPHP() {
-		let ret = '/~~' + this.id + '/action.php';
-		if (Config.testclient) {
-			ret = 'https://' + Config.routes.client + ret;
-		}
-		return (this.getActionPHP = function () {
-			return ret;
-		})();
-	}
 }
 
 /**********************************************************************
--- a/play.pokemonshowdown.com/src/client-main.ts
+++ b/play.pokemonshowdown.com/src/client-main.ts
@@ -770,8 +770,59 @@ export class PSRoom extends PSStreamModel<Args | null> implements RoomOptions {
 				PS.join('login' as RoomID);
 			}
 			return true;
+		case 'open':
+		case 'user': {
+			let roomid = `user-${toID(target)}` as RoomID;
+			PS.join(roomid, {
+				rightPopup: true,
+				args: { username: target },
+			});
+			return true;
 		}
-		return false;
+
+		case 'showjoins': {
+			let showjoins = PS.prefs.showjoins || {};
+			let serverShowjoins = showjoins[PS.server.id] || {};
+			if (target) {
+				let room = toID(target);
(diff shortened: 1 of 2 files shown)


play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -136,6 +136,130 @@ export class ChatRoom extends PSRoom {
...
+					formats[toID(targets[i])] = 1;
+				}
+				formatTargeting = true;
+			}
+
+			// let self = this;
+			$.get(PS.server.getActionPHP(), {
+				act: 'ladderget',
+				user: targets[0],
+			}, (data: string) => {
```
Zarel, line 174: The easiest way to do this query is with `LoginServer`. Look at how `LoginPopup` works in Preact (`panel-popups.tsx`) vs old client.

The best way would be to rewrite it to pull from `https://pokemonshowdown.com/users/[userid].json`. Look at how `panel-ladders.tsx` works in Preact client.

Zarel:
By the way, if you use Visual Studio Code and install the ESLint plugin, it will be able to highlight and automatically fix most of these linter errors.

play.pokemonshowdown.com/src/panel-chat.tsx
ISenseAura: `Net('https://play.pokemonshowdown.com/~~showdown/action.php?act=ladderget&user=aurastic')
.get()
.then(data => {...do stuff})`

this should be it, right?
I tried this inside the browser console itself and it fetched the data we needed

ISenseAura: `https://pokemonshowdown.com/users/[userid].json` does not provide enough data, it misses values for wins, losses and suspect.

Zarel: Not `Net`. `LoginServer`.

Zarel: image

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -136,6 +136,130 @@ export class ChatRoom extends PSRoom {
...
+				if (data.startsWith(']')) data = data.substr(1);
+				data = JSON.parse(data);
+				console.log(data);
```
Zarel, line 177:
```suggestion
			PSLoginServer.query(
				'ladderget', { user: targets[0] }
			).then((data: any) => {
```

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -136,6 +136,130 @@ export class ChatRoom extends PSRoom {
...
+					buffer += `<tr><td colspan="8"><button name="showOtherFormats"> ${hiddenFormats.slice(0, 3).join(', ') + (hiddenFormats.length > 3 ? ' and ' + String(hiddenFormats.length - 3) + ' other formats' : '')} not shown</button></td></tr>'`;
+				}
+				let userid = toID(targets[0]);
+				let registered = PS.user.registered;
+				if (registered && PS.user.userid === userid) {
+					buffer += '<tr><td colspan="8" style="text-align:right"><a href="//' + PS.routes.users + '/' + userid + '">Reset W/L</a></tr></td>';
+				}
+				buffer += '</table></div>';
+				this.receiveLine(['raw', buffer]);
+			}, 'text');
```
Zarel, line 259:
```suggestion
			});
```

to go with the `PSLoginServer` change above.

play.pokemonshowdown.com/src/panel-chat.tsx
Zarel: Note that you'll need to import `PSLoginServer` as well. And get rid of the `getActionPHP` stuff.

play.pokemonshowdown.com/src/panel-chat.tsx
Zarel: > `https://pokemonshowdown.com/users/[userid].json` does not provide enough data, it misses values for wins, losses and suspect.

I fixed this. You should probably use this now, because it works in testclient.

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -1133,6 +1185,26 @@ export const PS = new class extends PSModel {
...
+
+			case 'l':
+			case 'L': {
+				if (!room) continue;
+				let showjoins = this.prefs.showjoins || {};
+				let serverShowjoins = showjoins[this.server.id] || {};
+				if (!serverShowjoins[room.id] && !serverShowjoins.global) continue;
+				room.receiveLine([`l`, args[1]]);
+				break;
+			}
```
Zarel, line 1207: These guys would be best implemented in `panel-chat`'s `receiveLine`. You can make them `return` instead of `break` to prevent `log` from getting them.

Zarel: The problem is that this interception is too early. If you do it here, `ChatUserList` won't be able to update the userlist. Doing it there also makes sure you can do it after you update the userlist.

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -761,6 +761,58 @@ export class PSRoom extends PSStreamModel<Args | null> implements RoomOptions {
...
+			let showjoins = PS.prefs.showjoins || {};
+			let serverShowjoins = showjoins[PS.server.id] || {};
+			if (target) {
+				let room = toID(target);
+				if (!serverShowjoins['global']) {
+					delete serverShowjoins[room];
+				} else {
+					serverShowjoins[room] = 0;
+				}
+				this.receiveLine([`c`, "", `Join/leave messages on room ${room}: AUTOMATIC`]);
```
Zarel, line 804: _Are_ they automatic? I know they are in regular client, but are they here?

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -761,6 +761,58 @@ export class PSRoom extends PSStreamModel<Args | null> implements RoomOptions {
 				PS.join('login' as RoomID);
 			}
 			return true;
+		case 'open':
+		case 'user': {
+			let roomid = `user-${toID(target)}` as RoomID;
+			PS.join(roomid, {
+				rightPopup: true,
```
Zarel, line 768:
```suggestion
```

`rightPopup` opens popups to the right of the thing you clicked, it's used for user lists. It doesn't make sense for a text command.

Zarel:
Looks good! I do still want a future change to use the `https://pokemonshowdown.com/users/[userid].json` API rather than `PSLoginServer`, though. That'll get it working in testclient.

ISenseAura:
> Looks good! I do still want a future change to use the `https://pokemonshowdown.com/users/[userid].json` API rather than `PSLoginServer`, though. That'll get it working in testclient.

ah yes, i will do that soon, currently focusing on adding more to the options panel

Zarel:
Thanks!

PR #2353, opened 2025-04-12
Preact: Fix instant replay not showing controls in battles

ISenseAura:
After clicking "Instant Replay" it shows the moves menu instead of the replay control menu. 
this fixes it.

Steps to reproduce:
1. start a battle
2. forfeit
3. press rematch
4. go back to that battle
5. click Instant Replay

### DIFF
--- a/play.pokemonshowdown.com/src/panel-battle.tsx
+++ b/play.pokemonshowdown.com/src/panel-battle.tsx
@@ -141,6 +141,7 @@ class BattleRoom extends ChatRoom {
 			if (this.battle.atQueueEnd) {
 				this.battle.reset();
 			}
+			this.request = null;
 			this.battle.play();
 			this.update(null);
 			return true;


ISenseAura:
it happens when you close and reopen an ended battle so theres probably a missing `request = null` or something while creating a battleroom of an ended battle

Zarel:
Seems to be working fine for me.

Zarel:
> +Aurastic ✧: Ah okay, I got it! It's reproducible like this: start a battle, forfeit it, press rematch, go back to that battle, and then click Instant Replay

Oh! Yeah I can reproduce that. That looks like a server-side bug.

Zarel:
https://github.com/smogon/pokemon-showdown/commit/14ec0787460591955fb39d30d30b4a8b103b4cf0

And fixed. Will need to wait for a restart, though. Let me see if I can monkeypatch something.

Zarel:
Okay. I think I successfully monkeypatched a fix. Thanks for finding this and also for figuring out how to repro.

PR #2354, opened 2025-04-13
Preact: Add /ignore and /unignore

dot-Comfey:
Implements basic functionality of the ignore commands. I also added the ^ (prize winner) rank to the ignorable ranks since the inability for this rank to be ignored has been [reported as a bug](https://www.smogon.com/forums/threads/ignoring-prize-winners.3759839/).

### DIFF
--- a/play.pokemonshowdown.com/src/battle-log.ts
+++ b/play.pokemonshowdown.com/src/battle-log.ts
@@ -142,9 +142,7 @@ export class BattleLog {
 			if (battle?.ignoreOpponent) {
 				if ('\u2605\u2606'.includes(rank) && toUserid(name) !== app.user.get('userid')) return;
 			}
-			let ignore = PS.prefs.ignore || {};
-			let serverIgnore = ignore[PS.server.id] || {};
-			if (serverIgnore[toUserid(name)] && ' +^\u2605\u2606'.includes(rank)) return;
+			if ((window.app?.ignore?.[toUserid(name)] || PS?.prefs?.ignore?.[toUserid(name)]) && ' +^\u2605\u2606'.includes(rank)) return;
 			let isHighlighted = window.app?.rooms?.[battle!.roomid].getHighlight(message);
 			[divClass, divHTML, noNotify] = this.parseChatMessage(message, name, '', isHighlighted);
 			if (!noNotify && isHighlighted) {
--- a/play.pokemonshowdown.com/src/battle-log.ts
+++ b/play.pokemonshowdown.com/src/battle-log.ts
@@ -17,6 +17,7 @@ import type { Battle } from './battle';
 import type { BattleScene } from './battle-animations';
 import { Dex, Teams, toID, toRoomid, toUserid, type ID } from './battle-dex';
 import { BattleTextParser, type Args, type KWArgs } from './battle-text-parser';
+import { PS } from "./client-main";
 
 // Caja
 declare const html4: any;
@@ -141,7 +142,9 @@ export class BattleLog {
 			if (battle?.ignoreOpponent) {
 				if ('\u2605\u2606'.includes(rank) && toUserid(name) !== app.user.get('userid')) return;
 			}
-			if (window.app?.ignore?.[toUserid(name)] && ' +\u2605\u2606'.includes(rank)) return;
+			let ignore = PS.prefs.ignore || {};
+			let serverIgnore = ignore[PS.server.id] || {};
+			if (serverIgnore[toUserid(name)] && ' +^\u2605\u2606'.includes(rank)) return;
 			let isHighlighted = window.app?.rooms?.[battle!.roomid].getHighlight(message);
 			[divClass, divHTML, noNotify] = this.parseChatMessage(message, name, '', isHighlighted);
 			if (!noNotify && isHighlighted) {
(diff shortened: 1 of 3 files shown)


play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -60,6 +60,12 @@ class PSPrefs extends PSStreamModel<string | null> {
 	 * reasons.
 	 */
 	showjoins: { [serverid: string]: { [roomid: string]: 1 | 0 } } | null = null;
+	/**
+	 * List of users whose messages should be ignored. serverid:userid
+	 * table. Uses 1 and 0 instead of true/false for JSON packing
+	 * reasons.
+	 */
+	ignore: { [serverid: string]: { [userid: string]: 1 | 0 } } | null = null;
```
Zarel, line 68: Why _was_ this per-server? Feels like a weird way to do things...

play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -141,7 +142,9 @@ export class BattleLog {
 			if (battle?.ignoreOpponent) {
 				if ('\u2605\u2606'.includes(rank) && toUserid(name) !== app.user.get('userid')) return;
 			}
-			if (window.app?.ignore?.[toUserid(name)] && ' +\u2605\u2606'.includes(rank)) return;
```
Zarel, line 144: We shouldn't _drop_ support for the old client, we're still using that!

play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -141,7 +142,9 @@ export class BattleLog {
...
+			let ignore = PS.prefs.ignore || {};
```
Zarel, line 145: `PS` isn't guaranteed to exist here, unfortunately. For instance, Replays uses this code.

Zarel: You're going to need to redo the `window.PS?.prefs` thing, sadly.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -135,6 +135,35 @@ export class ChatRoom extends PSRoom {
...
 			return false;
+		} case 'ignore': {
+			let ignore = PS.prefs.ignore || {};
+			let serverIgnore = ignore[PS.server.id] || {};
+			if (!target) return true;
+			if (toID(target) === PS.user.userid) {
+				this.receiveLine([`You are not able to ignore yourself.`]);
+			} else if (serverIgnore[toID(target)]) {
+				this.receiveLine([`c`, "", `User '` + target + `' is already on your ignore list.` +
+				`(Moderator messages will not be ignored.)`]);
```
Zarel, line 146:
```suggestion
				this.receiveLine(["", `User '` + target + `' is already on your ignore list.` +
				`(Moderator messages will not be ignored.)`]);
```
This is the standard way to do unformatted text. No `c`.

play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -141,7 +142,7 @@ export class BattleLog {
 			if (battle?.ignoreOpponent) {
 				if ('\u2605\u2606'.includes(rank) && toUserid(name) !== app.user.get('userid')) return;
 			}
-			if (window.app?.ignore?.[toUserid(name)] && ' +\u2605\u2606'.includes(rank)) return;
+			if ((window.app?.ignore?.[toUserid(name)] || PS?.prefs?.ignore?.[toUserid(name)]) && ' +^\u2605\u2606'.includes(rank)) return;
```
Zarel, line 145:
```suggestion
			const ignoreList = window.app?.ignore || window.PS?.prefs?.ignore;
			if (ignoreList?.[toUserid(name)] && ' +^\u2605\u2606'.includes(rank)) return;
```

It needs to be `window.PS?`, not just `PS?`. The latter will still crash if it's not defined.

Zarel: This is the last thing left! After this fix, this'll be ready to merge!

play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -17,6 +17,7 @@ import type { Battle } from './battle';
 import type { BattleScene } from './battle-animations';
 import { Dex, Teams, toID, toRoomid, toUserid, type ID } from './battle-dex';
 import { BattleTextParser, type Args, type KWArgs } from './battle-text-parser';
+import { PS } from "./client-main";
```
Zarel, line 20:
```suggestion
```

Also we don't need the import; this isn't a real dependency.

Zarel:
> @dot-Comfey
> [No need for server-based ignores](https://github.com/smogon/pokemon-showdown-client/pull/2354/commits/cc4601e52b9a4155668e964fb7d3182e25aefb89) 
> [cc4601e](https://github.com/smogon/pokemon-showdown-client/pull/2354/commits/cc4601e52b9a4155668e964fb7d3182e25aefb89)
> I didn't like it anyway, I just copied from showjoins.

I want to be clear, we want 100% compatibility with 2013 client for now, and 2013 client's ignore list doesn't have server IDs.

Zarel:
Anyway, thanks! More Preact progress!!!

PR #2359, opened 2025-04-14
Preact: Add autojoin functionality

ISenseAura:


play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -60,6 +60,12 @@ class PSPrefs extends PSStreamModel<string | null> {
 	 * reasons.
 	 */
 	showjoins: { [serverid: string]: { [roomid: string]: 1 | 0 } } | null = null;
+
+	/**
+	 * List of rooms to autojoin
+	 * different for each server
+	 */
+	autojoin: { [serverid: string]: string } | null = null;
```
Zarel, line 68: This is a `,` delimited string. It needs to match 2013 client.

Zarel: Oh! It uses either.

```suggestion
	autojoin: { [serverid: string]: string } | string | null = null;
```

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -1077,6 +1087,17 @@ export const PS = new class extends PSModel {
 			}, true);
 		}
 
+		/*
+		* Create rooms before /autojoin is sent to the server
+		*/
+		let autojoin = this.prefs.autojoin;
+		if (autojoin) {
+			let rooms = autojoin[this.server.id] || '';
+			rooms.split(",").forEach(title => {
```
Zarel, line 1096: We always use `for...of` instead of `forEach`

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -1077,6 +1087,17 @@ export const PS = new class extends PSModel {
...
+				this.addRoom({ id: String(toID(title)) as RoomID, title, connected: true }, true);
```
Zarel, line 1097:
```suggestion
				this.addRoom({ id: toID(title) as RoomID, title, connected: true }, true);
```

The `String` cast isn't necessary.

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -1793,6 +1815,46 @@ export const PS = new class extends PSModel {
...
 			this.removeRoom(room);
 			this.update();
+			this.updateAutojoin();
+		}
+	}
+
+	updateAutojoin() {
+		if (!PS.server.registered) return;
+		if (!this.prefs.autojoin) this.prefs.autojoin = {};
+		if (!this.prefs.autojoin[this.server.id]) this.prefs.autojoin[this.server.id] = '';
```
Zarel, line 1825:
```suggestion
		this.prefs.autojoin ||= {};
		this.prefs.autojoin[this.server.id] ||= '';
```

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -1793,6 +1815,46 @@ export const PS = new class extends PSModel {
...
+				curAutojoin = { showdown: curAutojoin };
+				if (!autojoins.length) return;
+				curAutojoin[this.server.id] = autojoins.join(',');
+			} else {
+				if (curAutojoin === autojoins.join(',')) return;
+				curAutojoin[this.server.id] = autojoins.join(',') as never;
+			}
 		}
+		this.prefs.set('autojoin', curAutojoin);
+
```
Zarel, line 1858:
```suggestion

		const thisAutojoin = autojoins.join(',') || undefined;
		let autojoin = this.prefs.autojoin || undefined;
		if (this.server.id === 'showdown' && typeof this.prefs.autojoin !== 'object') {
			// Main server only mode
			if (autojoin === thisAutojoin) return;

			this.prefs.set('autojoin', thisAutojoin || null);
		} else {
			// Multi server mode
			autojoin = typeof autojoin === 'string' ? { showdown: autojoin } : autojoin || {};
			if (autojoin[this.server.id] === thisAutojoin) return;

			autojoin[this.server.id] = thisAutojoin || undefined;
			this.prefs.set('autojoin', autojoin);
		}
```

play.pokemonshowdown.com/src/client-main.ts
Zarel:
```suggestion
```

Actually, with my new code, this isn't even necessary.

play.pokemonshowdown.com/src/panel-topbar.tsx
```diff
@@ -29,6 +29,7 @@ window.addEventListener('drop', e => {
 	// in the tab, which is generally undesirable anyway.
 	e.preventDefault();
 	PS.dragging = null;
+	PS.updateAutojoin();
```
Zarel, line 32: Only after dragging? Maybe also in the router after URL updates?

play.pokemonshowdown.com/src/panels.tsx
```diff
@@ -299,6 +299,14 @@ export class PSMain extends preact.Component {
 		super();
 		PS.subscribe(() => this.forceUpdate());
 
+		let autojoin = PS.prefs.autojoin;
+		if (autojoin) {
+			let rooms = autojoin[PS.server.id] || '';
+			if (rooms.length) {
+				PS.rooms['']?.sendDirect(`/autojoin ${rooms}`);
+			}
+		}
```
Zarel, line 308: This should probably happen before `new PSRouter()`, so it can happen before the link URL takes effect.

Zarel: The earliest it can happen is after `new PSConnection()`.

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -1793,6 +1815,46 @@ export const PS = new class extends PSModel {
...
+		if (!PS.server.registered) return;
+		if (!this.prefs.autojoin) this.prefs.autojoin = {};
+		if (!this.prefs.autojoin[this.server.id]) this.prefs.autojoin[this.server.id] = '';
+		let autojoins: string[] = [];
+		let autojoinCount = 0;
+		let rooms = this.rightRoomList;
+		for (let roomid of rooms) {
+			let room = PS.rooms[roomid];
+			if (!room) return;
+			if (room.type !== 'chat') continue;
```
Zarel, line 1832:
```suggestion
			if (room.type !== 'chat' || room.pmTarget) continue;
```

PM rooms are chatrooms now and need to be manually skipped.

play.pokemonshowdown.com/src/client-main.ts
ISenseAura: i get this if i dont cast it to string
`Conversion of type 'ID' to type 'RoomID' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.`

ISenseAura: guess i should just convert it to unknown

play.pokemonshowdown.com/src/panels.tsx
ISenseAura: > The earliest it can happen is after `new PSConnection()`. For perf reasons, you might even want to move it right there.

so in client-connection.ts right after `new PSConnection()` ?

ISenseAura:
smh i dont know why im not getting this specific error in vscode

Zarel:
The error message looks straightforward to me. Wonder why it's not appearing for you. Try `npm ci` to sync dependencies, then `npm test` to see if it appears from there?

PR #2362, opened 2025-04-15
Teams: Add icon for Tera type

shrianshChari:


shrianshChari:
After using enough electricity to cause PG&E's rates to double and enough water to turn the San Luis Reservoir into a puddle, I have screenshots!

image

PR #2363, opened 2025-04-15
Teams: Change move's type icon if it does in-battle

shrianshChari:


KrisXV:
Multi-Attack too maybe?

shrianshChari:
Sure! Should I also do Techno Blast and Natural Gift?

Zarel:
Should this share code with client? Ideally this sort of thing can all go into BattleTooltips or something.

shrianshChari:
Yeah I copied a bunch of this code from `BattleTooltip.showMoveTooltip`, though it had to be adapted a bit because here we're given a `Dex.PokemonSet` object instead of a `Pokemon` object. If the code could somehow be adapted or unified, that would be ideal.

shrianshChari:
Implemented this using `BattleTooltips` (with a messy implementation, IMO) and can show screenshots.

image

PR #2382, opened 2025-04-19
Preact: Support tabcomplete

ISenseAura:
TODO
tabcomplete for commands, this is a tricky since it relies on the response of /crq server command.
i have something 'experimental' in my mind for this but thats really not the priority so ill keep that for later

### DIFF
--- a/play.pokemonshowdown.com/src/panel-chat.tsx
+++ b/play.pokemonshowdown.com/src/panel-chat.tsx
@@ -710,7 +710,10 @@ export class ChatTextEntry extends preact.Component<{
 		// Don't tab complete at the start of the text box.
 		let idx = this.getSelection().start;
 		if (idx === 0) return false;
-		let users = (this.props.room as ChatRoom)?.users || (PS.rooms['lobby'] ? (PS.rooms['lobby'] as ChatRoom).users : {});
+		let users = (this.props.room as ChatRoom)?.users ||
+			(PS.rooms['lobby'] ? (
+				PS.rooms['lobby'] as ChatRoom).users :
+			{});
 		let text = this.miniedit?.getValue() || '';
 		let prefix = text.slice(0, idx);
 		if (this.tabComplete.cursor !== null && prefix === this.tabComplete.cursor) {
--- a/play.pokemonshowdown.com/src/panel-chat.tsx
+++ b/play.pokemonshowdown.com/src/panel-chat.tsx
@@ -776,8 +776,6 @@ export class ChatTextEntry extends preact.Component<{
 			return this.historyDown(true);
 		} else if (ev.keyCode === 27) { // esc
 			if (this.undoTabComplete()) {
-				ev.preventDefault();
-				ev.stopPropagation();
 				return true;
 			}
 			if (PS.room !== PS.panel) { // only close if in mini-room mode
--- a/play.pokemonshowdown.com/src/panel-chat.tsx
+++ b/play.pokemonshowdown.com/src/panel-chat.tsx
@@ -899,7 +899,7 @@ class ChatTextBox extends preact.Component<{ placeholder: string, disabled?: boo
 	override render() {
 		return <pre
 			class={`textbox textbox-empty ${this.props.disabled ? ' disabled' : ' autofocus'}`} placeholder={this.props.placeholder}
-			onFocus={this.handleFocus} onBlur={this.handleBlur} name="textbox"
+			onFocus={this.handleFocus} onBlur={this.handleBlur}
 		>{'\n'}</pre>;
 	}
 }
--- a/play.pokemonshowdown.com/src/panel-chat.tsx
+++ b/play.pokemonshowdown.com/src/panel-chat.tsx
@@ -787,9 +787,7 @@ export class ChatTextEntry extends preact.Component<{
 		name = Dex.getShortName(name).replace(/^[^a-zA-Z0-9]+|[^a-zA-Z0-9]+$/g, "");
 		let prefixIndex = candidate[1].toString().startsWith('/') ? prefix.lastIndexOf('\n') + 1 : candidate[1];
 		let fullPrefix = this.tabComplete.prefix?.substr(0, prefixIndex as number) + name;
-		this.miniedit?._setContent(fullPrefix);
-		let pos = fullPrefix.length;
-		this.setSelection(pos, pos);
+		this.setValue(fullPrefix, { start: fullPrefix.length, end: fullPrefix.length });
 		this.miniedit?.element.focus();
 		this.tabComplete.cursor = text.length;
 		return true;
--- a/play.pokemonshowdown.com/src/panel-chat.tsx
+++ b/play.pokemonshowdown.com/src/panel-chat.tsx
@@ -702,12 +702,10 @@ export class ChatTextEntry extends preact.Component<{
 		}
 		return false;
 	}
-	/**
-	 * Taken from the old client — just tweaked a bit to:
-     * - Work with <pre> instead of <textarea>
-     * - Fit the way `room` now handles user info
-	 * TODO - add support for commands tabcomplete
-     */
+	// Taken from the old client — just tweaked a bit to:
+	// - Work with <pre> instead of <textarea>
+	// - Fit the way `room` now handles user info
+	// TODO - add support for commands tabcomplete
 	handleTabComplete(reverse: boolean) {
 		// Don't tab complete at the start of the text box.
 		let idx = this.getSelection().start;
--- a/play.pokemonshowdown.com/src/panel-chat.tsx
+++ b/play.pokemonshowdown.com/src/panel-chat.tsx
@@ -578,9 +578,8 @@ export class ChatTextEntry extends preact.Component<{
 		this.subscription = PS.user.subscribe(() => {
 			this.forceUpdate();
 		});
-		const textbox = (this.base!.children[0].children[1] ||
-			this.base!.querySelector('textarea[name=textbox]')) as HTMLElement;
-		if (textbox.tagName === 'TEXTAREA' || textbox.tagName === 'PRE') this.textbox = textbox as HTMLTextAreaElement;
+		const textbox = this.base!.children[0].children[1] as HTMLElement;
+		if (textbox.tagName === 'TEXTAREA') this.textbox = textbox as HTMLTextAreaElement;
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -546,12 +559,28 @@ export class ChatTextEntry extends preact.Component<{
...
+		};
 	override componentDidMount() {
 		this.subscription = PS.user.subscribe(() => {
 			this.forceUpdate();
 		});
-		const textbox = this.base!.children[0].children[1] as HTMLElement;
-		if (textbox.tagName === 'TEXTAREA') this.textbox = textbox as HTMLTextAreaElement;
+		const textbox = (this.base!.children[0].children[1] ||
+			this.base!.querySelector('textarea[name=textbox]')) as HTMLElement;
+		if (textbox.tagName === 'TEXTAREA' || textbox.tagName === 'PRE') this.textbox = textbox as HTMLTextAreaElement;
```
Zarel, line 583: Oh nooooo. If it's a `pre` you shouldn't be manipulating its contents directly. It's tracked in `room.miniedit`. See how `toggleFormatChar` does it.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -649,14 +677,18 @@ export class ChatTextEntry extends preact.Component<{
...
+			const reverse = !!ev.shiftKey; // Shift+Tab reverses direction
+			return this.handleTabComplete(reverse);
 		} else if (ev.keyCode === 38 && !ev.shiftKey && !ev.altKey) { // Up key
 			return this.historyUp();
 		} else if (ev.keyCode === 40 && !ev.shiftKey && !ev.altKey) { // Down key
 			return this.historyDown();
 		} else if (ev.keyCode === 27) { // esc
+			if (this.undoTabComplete()) {
+				ev.preventDefault();
+				ev.stopPropagation();
```
Zarel, line 690:
```suggestion
				ev.stopPropagation();
				return true;
```

Seems like a rather obscure keyboard shortcut, but oh well.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -669,6 +701,107 @@ export class ChatTextEntry extends preact.Component<{
 		}
 		return false;
 	}
+	/**
+	 * Taken from the old client — just tweaked a bit to:
+     * - Work with <pre> instead of <textarea>
+     * - Fit the way `room` now handles user info
+	 * TODO - add support for commands tabcomplete
+     */
```
Zarel, line 710:
```suggestion
	// Taken from the old client — just tweaked a bit to:
	// - Work with <pre> instead of <textarea>
	// - Fit the way `room` now handles user info
	// TODO - add support for commands tabcomplete
```
Comments starting with `/**` are doc comments and are for describing what the function does and how to use it. Other notes should be regular comments.

ISenseAura: got it

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -669,6 +701,107 @@ export class ChatTextEntry extends preact.Component<{
...
+	handleTabComplete(reverse: boolean) {
+		// Don't tab complete at the start of the text box.
+		let idx = this.getSelection().start;
+		if (idx === 0) return false;
+		let users = (this.props.room as ChatRoom)?.users || (PS.rooms['lobby'] ? (PS.rooms['lobby'] as ChatRoom).users : {});
+		let text = this.miniedit?.getValue() || '';
```
Zarel, line 714: You're still not using the abstractions I taught you...

Zarel:
```suggestion
		let text = this.getValue();
```

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -669,6 +701,107 @@ export class ChatTextEntry extends preact.Component<{
...
+		let prefix = text.substr(0, idx);
```
Zarel, line 715:
```suggestion
		let prefix = text.slice(0, idx);
```

`substr` is deprecated; I've been mostly using `slice` in new code.

play.pokemonshowdown.com/src/panel-chat.tsx
ISenseAura: > You're still not using the abstractions I taught you...

line 714? mb lemme fix it

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -669,6 +701,107 @@ export class ChatTextEntry extends preact.Component<{
...
+		let substituteUser = users[substituteUserId] || candidate[1];
+		if (!substituteUser) return true;
+		let name = substituteUser;
+		// Remove rank and busy charaxcters
+		name = Dex.getShortName(name).replace(/^[^a-zA-Z0-9]+|[^a-zA-Z0-9]+$/g, "");
+		let prefixIndex = candidate[1].toString().startsWith('/') ? prefix.lastIndexOf('\n') + 1 : candidate[1];
+		let fullPrefix = this.tabComplete.prefix?.substr(0, prefixIndex as number) + name;
+		this.miniedit?._setContent(fullPrefix);
+		let pos = fullPrefix.length;
+		this.setSelection(pos, pos);
```
Zarel, line 792: How did you even find the `_setContent` function??? Anyway, if a function starts with `_`, that's a good sign you're using the wrong one.

```suggestion
		this.setValue(fullPrefix, { start: fullPrefix.length, end: fullPrefix.length });
```

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -669,6 +701,107 @@ export class ChatTextEntry extends preact.Component<{
...
+		this.miniedit?.element.focus();
```
Zarel, line 793:
```suggestion
```

play.pokemonshowdown.com/src/panel-chat.tsx
ISenseAura: lmaoo my editor autofilled it and it didnt warn me so i thought eh, it worked tho but yeah i shouldnt use it

play.pokemonshowdown.com/src/panel-chat.tsx
ISenseAura: this is required to keep the textbox focus otherwise it focuses the userlist

Zarel: You weren't using `setValue`. Now that you are, this should no longer be an issue.

Zarel: Oh, wait. Wait.

Zarel: I'm pretty sure returning `true` should be enough, actually. Let me test this.

Zarel: Yeah it works correctly in my tests.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -757,6 +791,106 @@ export class ChatTextEntry extends preact.Component<{
...
+		let substituteUser = users[substituteUserId] || candidate[1];
+		if (!substituteUser) return true;
+		let name = substituteUser;
+		// Remove rank and busy charaxcters
+		name = Dex.getShortName(name).replace(/^[^a-zA-Z0-9]+|[^a-zA-Z0-9]+$/g, "");
+		let prefixIndex = candidate[1].toString().startsWith('/') ? prefix.lastIndexOf('\n') + 1 : candidate[1];
+		let fullPrefix = this.tabComplete.prefix?.substr(0, prefixIndex as number) + name;
+		this.setValue(fullPrefix, fullPrefix.length);
+		this.miniedit?.element.focus();
+		this.tabComplete.cursor = text.length;
```
Zarel, line 880: This should actually be `fullPrefix` and not `text.length`. This is why your `cursor` was `string | number` and not `string`.

ISenseAura: > This should actually be `fullPrefix` and not `text.length`. This is why your `cursor` was `string | number` and not `string`.

i actually changed the original code from old client
it wasnt working properly but with this change it did

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -738,14 +767,19 @@ export class ChatTextEntry extends preact.Component<{
...
+			const reverse = !!ev.shiftKey; // Shift+Tab reverses direction
+			return this.handleTabComplete(reverse);
 		} else if (ev.keyCode === 38 && !ev.shiftKey && !ev.altKey) { // Up key
 			return this.historyUp(true);
 		} else if (ev.keyCode === 40 && !ev.shiftKey && !ev.altKey) { // Down key
 			return this.historyDown(true);
 		} else if (ev.keyCode === 27) { // esc
+			if (this.undoTabComplete()) {
+				ev.preventDefault();
+				ev.stopPropagation();
```
Zarel, line 780:
```suggestion
```

These are unnecessary, `return true` does the same thing.

Zarel: Also, for future reference, always use `stopImmediatePropagagion` instead of `stopPropagation`, unless for some reason you actually want to stop some but not all propagation.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -757,6 +791,105 @@ export class ChatTextEntry extends preact.Component<{
...
 	}
+	// Taken from the old client — just tweaked a bit to:
+	// - Work with <pre> instead of <textarea>
+	// - Fit the way `room` now handles user info
+	// TODO - add support for commands tabcomplete
+	handleTabComplete(reverse: boolean) {
+		// Don't tab complete at the start of the text box.
+		let idx = this.getSelection().start;
+		if (idx === 0) return false;
+		let users = (this.props.room as ChatRoom)?.users ||
```
Zarel, line 800: This is a sign that `this.props.room` is typed incorrectly. There's no need for an unsafe cast here.

Zarel:
Thanks!

PR #2383, opened 2025-04-19
Client: Fix volume slider resetting

bingingem:
https://www.smogon.com/forums/threads/bug-report-client.3762992/

Fixes the issue where volume sliders reset to 50 when volumes are set from local storage

### DIFF
--- a/play.pokemonshowdown.com/js/client-topbar.js
+++ b/play.pokemonshowdown.com/js/client-topbar.js
@@ -423,7 +423,7 @@
 			Storage.prefs('effectvolume', volume);
 		},
 		getEffectVolume: function () {
-			var volume = Dex.prefs('effectvolume');
+			var volume = Storage.prefs('effectvolume');
 			return typeof volume === 'number' ? volume : 50;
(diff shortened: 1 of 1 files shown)


Zarel:
Have you tested this? It doesn't seem to me like this would do anything.

bingingem:
Yes I tested it, and made sure it didn't affect Preact. I'm not entirely sure why it started misbehaving, but if you look at the `Set()` functions, they're saving the values to `Storage.prefs` which was my first hint.  The issue was `Dex.prefs` was returning undefined which defaulted to 50 in the next line.

Zarel:
image

I can't figure out how to make `Dex.prefs` return something different from `Storage.prefs`. Do you have repro steps?

Zarel:
Oh! I see! They're different at 0 specifically.

bingingem:
> Oh! I see! They're different at 0 specifically.

Yep! That seems to be the only difference. Let me know if this change works if you'd prefer anything different. Thanks for taking the time to look!

PR #2384, opened 2025-04-20
Preact: Add format resource loading to teambuilder

mia-pi-git:
For parity with current builder.

### DIFF
--- a/play.pokemonshowdown.com/src/panel-teambuilder-team.tsx
+++ b/play.pokemonshowdown.com/src/panel-teambuilder-team.tsx
@@ -256,7 +256,7 @@ class TeamPanel extends PSRoomPanel<TeamRoom> {
 					/>
 				</label>
 				<TeamTextbox team={team} />
-				{info && (info.resources.length || info.url) && (
+				{!!(info && (info.resources.length || info.url)) && (
 					<>
 						<br /><br />
 						<div style={{ paddingLeft: "5px" }}>
--- a/play.pokemonshowdown.com/src/panel-teambuilder-team.tsx
+++ b/play.pokemonshowdown.com/src/panel-teambuilder-team.tsx
@@ -205,7 +205,7 @@ class TeamPanel extends PSRoomPanel<TeamRoom> {
 	static readonly Model = TeamRoom;
 	static readonly title = 'Team';
 
-	resources!: FormatResource;
+	resources?: FormatResource;
 
 	constructor(props?: { room: TeamRoom }) {
 		super(props);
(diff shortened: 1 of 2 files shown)


play.pokemonshowdown.com/src/panel-teambuilder-team.tsx
```diff
@@ -205,6 +205,22 @@ class TeamPanel extends PSRoomPanel<TeamRoom> {
 	static readonly Model = TeamRoom;
 	static readonly title = 'Team';
 
+	resources!: FormatResource;
```
Zarel, line 208:
```suggestion
	resources?: FormatResource;
```

play.pokemonshowdown.com/src/panel-teamdropdown.tsx
```diff
@@ -532,6 +535,24 @@ export class PSTeambuilder {
...
+				result => {
+					this.formatResources[format] = JSON.parse(result);
+					resolve(this.formatResources[format]);
+				},
+				err => {
+					this.formatResources[format] = null;
+					resolve(this.formatResources[format]);
+				}
+			);
+		});
```
Zarel, line 554: I guess you're not used to pre-async/await Promises. Valid, they both landed roughly the same time.

```suggestion
		if (format in this.formatResources) return Promise.resolve(this.formatResources[format]);
		return Net('https://www.smogon.com/dex/api/formats/by-ps-name/' + format).get()
			.then(result => {
				this.formatResources[format] = JSON.parse(result);
				return this.formatResources[format];
			}).catch(err => {
				this.formatResources[format] = null;
				return this.formatResources[format];
			});
```

mia-pi-git: Nope, that was literally just me writing `new Promise`, remembering `Net` is already promise, and then forgetting to remove it.

play.pokemonshowdown.com/src/panel-teambuilder-team.tsx
```diff
@@ -239,6 +256,23 @@ class TeamPanel extends PSRoomPanel<TeamRoom> {
 					/>
 				</label>
 				<TeamTextbox team={team} />
+				{info && (info.resources.length || info.url) && (
```
Zarel, line 259:
```suggestion
				{!!(info && (info.resources.length || info.url)) && (
```

The `&&` pattern only works with bools. You could accidentally insert a `0` into your output if you're not careful.

mia-pi-git: Right, right. I always forget that sometimes that returns numbers.

Zarel:
Looks great, thanks!

PR #2385, opened 2025-04-20
Preact: Support custom colors

mia-pi-git:
Not wedded to where this is actually _done_, but battle-log made the most sense since it actually handles colors. Open to opinions otherwise.

### DIFF
--- a/play.pokemonshowdown.com/src/battle-log.ts
+++ b/play.pokemonshowdown.com/src/battle-log.ts
@@ -1790,9 +1790,8 @@ export class BattleLog {
 	}
 }
 
-const net = window.Net as typeof Net | undefined;
-if (net) {
-	net(`/config/colors.json`).get().then(response => {
+if (window.Net) {
+	Net(`/config/colors.json`).get().then(response => {
 		const data = JSON.parse(response);
 		Object.assign(Config.customcolors, data);
 	}).catch(() => {});
--- a/play.pokemonshowdown.com/src/battle-log.ts
+++ b/play.pokemonshowdown.com/src/battle-log.ts
@@ -17,6 +17,7 @@ import type { Battle } from './battle';
 import type { BattleScene } from './battle-animations';
 import { Dex, Teams, toID, toRoomid, toUserid, type ID } from './battle-dex';
 import { BattleTextParser, type Args, type KWArgs } from './battle-text-parser';
+import { Net } from './client-connection'; // optional
 
 // Caja
 declare const html4: any;
--- a/play.pokemonshowdown.com/src/battle-log.ts
+++ b/play.pokemonshowdown.com/src/battle-log.ts
@@ -17,7 +17,6 @@ import type { Battle } from './battle';
 import type { BattleScene } from './battle-animations';
 import { Dex, Teams, toID, toRoomid, toUserid, type ID } from './battle-dex';
 import { BattleTextParser, type Args, type KWArgs } from './battle-text-parser';
-import type { Net } from './client-connection';
 
 // Caja
 declare const html4: any;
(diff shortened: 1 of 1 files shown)


Zarel:
Hm. What's your plan for testclient?

mia-pi-git:
> Hm. What's your plan for testclient?

Probably just .catch it? I guess we could call out specifically to `Config.routes.client/config/colors.json`, but I don't really see the point there. Not like testclient hugely _needs_ custom color data.

Zarel:
neat. I think the bottom of panel-topbar is good. early in the loading process, but still after everything important is loaded.

Zarel:
It would also save us the maybe-loaded maybe-not thing going on here.

Zarel:
On the other hand, putting it here means it'd work in Replays...

play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -1788,3 +1789,11 @@ export class BattleLog {
 		return 'data:text/plain;base64,' + encodeURIComponent(btoa(unescape(encodeURIComponent(replayFile))));
 	}
 }
+
+const net = window.Net as typeof Net | undefined;
+if (net) {
+	net(`/config/colors.json`).get().then(response => {
```
Zarel, line 1795:
```suggestion
if (window.Net) {
	Net(`/config/colors.json`).get().then(response => {
```

I'd just do this tbh.

play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -17,6 +17,7 @@ import type { Battle } from './battle';
 import type { BattleScene } from './battle-animations';
 import { Dex, Teams, toID, toRoomid, toUserid, type ID } from './battle-dex';
 import { BattleTextParser, type Args, type KWArgs } from './battle-text-parser';
+import type { Net } from './client-connection';
```
urkerab, line 20: Out of interest, does this work differently to jQuery in that it doesn't set the `X-Requested-With` header?

play.pokemonshowdown.com/src/battle-log.ts
urkerab: Does optional chaining not compile well here?

play.pokemonshowdown.com/src/battle-log.ts
Zarel: I suppose it does. It's using raw XHR. You can just look at the source.

play.pokemonshowdown.com/src/battle-log.ts
Zarel: I think we have `window: any` declared, so it at least wouldn't be typesafe the way this one is.

Zarel: And "not typesafe" means "TS will force us to declare the type of `response`"

urkerab: Ah, so because the original code had declared the type of `net` it could have used ``net?.(`/config/colors.json`).get().then(response => {`` (but not having to declare the type in the first place is better still of course).

play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -17,6 +17,7 @@ import type { Battle } from './battle';
 import type { BattleScene } from './battle-animations';
 import { Dex, Teams, toID, toRoomid, toUserid, type ID } from './battle-dex';
 import { BattleTextParser, type Args, type KWArgs } from './battle-text-parser';
+import { Net } from './client-connection'; // optional
```
Zarel, line 20: Oh, I guess you didn't know. In client, import/export are entirely ignored. That's why `if (window.Net)` works in the first place.

Zarel:
(Thanks!)

PR #2386, opened 2025-04-20
Preact: Fix confirmleaveroom popup

ISenseAura:


play.pokemonshowdown.com/src/panels.tsx
```diff
@@ -416,6 +416,8 @@ export class PSView extends preact.Component {
 				if (elem.getAttribute('data-cmd')) {
 					const cmd = elem.getAttribute('data-cmd')!;
 					const room = PS.getRoom(elem) || PS.mainmenu;
+					// if room is a leaveroom popup, close it
+					if (room.id === "confirmleaveroom") PS.closePopup();
```
Zarel, line 420: Okay we're not doing this kind of hardcode. We're adding `data-closeand`.

ISenseAura: much needed

ISenseAura: want me to do it?

Zarel: I already did it.

PR #2387, opened 2025-04-20
Refactor BattlePanel controls and fix volume slider bug.

TurboRx:
Modernizes keyboard event handling, improves control rendering structure, and fixes the volume slider bug. Includes other minor usability and accessibility enhancements.

### DIFF
--- a/replay.pokemonshowdown.com/src/replays-battle.tsx
+++ b/replay.pokemonshowdown.com/src/replays-battle.tsx
@@ -77,7 +77,7 @@ export class BattlePanel extends preact.Component<{ id: string }> {
 	override componentDidMount() {
 		this.loadBattle(this.props.id);
 		showAd('LeaderboardBTF');
-		window.onkeydown = this.keyPressed;
+		window.addEventListener('keydown', this.keyPressed);
 	}
 	override componentWillReceiveProps(nextProps: this['props']) {
 		if (this.stripQuery(this.props.id) !== this.stripQuery(nextProps.id)) {
@@ -145,7 +145,7 @@ export class BattlePanel extends preact.Component<{ id: string }> {
 	override componentWillUnmount(): void {
 		this.battle?.destroy();
 		(window as any).battle = null;
-		window.onkeydown = null;
+		window.removeEventListener('keydown', this.keyPressed);
 	}
 	override componentDidUpdate(): void {
 		if (this.autofocusTurnView === 'select') {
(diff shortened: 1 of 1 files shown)


Zarel:
Well, this does more than nothing. Unfortunately, most of the decisions we made are made for reasons.

replay.pokemonshowdown.com/src/replays-battle.tsx
```diff
@@ -77,7 +77,7 @@ export class BattlePanel extends preact.Component<{ id: string }> {
 	override componentDidMount() {
 		this.loadBattle(this.props.id);
 		showAd('LeaderboardBTF');
-		window.onkeydown = this.keyPressed;
+		window.addEventListener('keydown', this.keyPressed);
```
Zarel, line 80: Would drop IE8 support for no benefit.

replay.pokemonshowdown.com/src/replays-battle.tsx
```diff
@@ -159,17 +159,22 @@ export class BattlePanel extends preact.Component<{ id: string }> {
 			this.autofocusTurnView = null;
 		}
 	}
+	// @ts-expect-error really wish they let me assert that the target is an HTMLElement
```
Zarel, line 162: This would clearly crash, there's no error to expect.

replay.pokemonshowdown.com/src/replays-battle.tsx
```diff
@@ -159,17 +159,22 @@ export class BattlePanel extends preact.Component<{ id: string }> {
...
+		if (e.key === 'Escape' && this.turnView) {
 			this.closeTurn();
 			return;
 		}
-		// @ts-expect-error really wish they let me assert that the target is an HTMLElement
-		if (e.target?.tagName === 'INPUT' || e.target?.tagName === 'SELECT') return;
-		switch (e.keyCode) {
-		case 75: // k
+
+		switch (e.key) {
```
Zarel, line 175: https://caniuse.com/?search=KeyboardEvent.key

Actually doesn't support _any_ Edge version before 79. Relatively bad compatibility in general.

replay.pokemonshowdown.com/src/replays-battle.tsx
```diff
@@ -178,22 +183,40 @@ export class BattlePanel extends preact.Component<{ id: string }> {
...
 			break;
-		case 190: // . (>)
+		case '.':
 			if (e.shiftKey) this.stepSpeed(1);
 			break;
-		case 191: // / (?)
+		case '/':
+		case '?':
 			if (e.shiftKey) {
+				console.log(
```
Zarel, line 207: Does not belong in production code.

replay.pokemonshowdown.com/src/replays-battle.tsx
```diff
@@ -486,61 +544,65 @@ export class BattlePanel extends preact.Component<{ id: string }> {
...
-				<BattleLogDiv />
-				{this.renderControls()}
-				<div id="LeaderboardBTF"></div>
-			</div>
-		</div>;
-	}
-}
+		return <div class={PSRouter.showingLeft() ? 'mainbar mainbar-hassidebar' : 'mainbar'} style={position}>
+			<div class="battle-wrapper" style={{ position: 'relative' }}>
+				<BattleDi
```
Zarel, line 608: ...Actually this is straight up a syntax error.

Zarel:
Look, please stop PRing while not knowing what you're doing. I didn't mind the first time if you were just learning and didn't understand, and I wouldn't say I'm mad at you or anything, but I do not think you're at a point where you should be PRing. If you're doing it by accident and can't figure out how to stop, I could also simply block you (no hard feelings).

PR #2390, opened 2025-04-21
Preact: Support remote teams

mia-pi-git:
Porting old features.


play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -261,6 +263,9 @@ export interface Team {
 	/** The icon cache must be cleared (to `null`) whenever `packedTeam` is modified */
 	iconCache: preact.ComponentChildren;
 	key: string;
+	teamid?: number;
```
Zarel, line 266: I guess this is the ID from the teams database? Plausibly these should be documented.

Zarel:
I am putting off reviewing this for a few days so I can actually work on teambuilder. A cursory glance didn't reveal anything obviously wrong. I mainly think some documentation would make it much easier to read.

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -373,6 +387,69 @@ class PSTeams extends PSStreamModel<'team' | 'format'> {
...
 	}
+	loadRemoteTeams() {
+		PSLoginServer.query('getteams', {}).then(data => {
+			if (!data) return;
+			if (data.actionerror) {
+				return PS.alert('Error loading uploaded teams: ' + data.actionerror);
+			}
+			for (const team of data.teams as ((Team & { team: string })[])) {
+				let matched = false;
+				for (const curTeam of this.list) {
```
Zarel, line 398: O(n^2)? oof. You could bring it down to O(n) after sorting both lists.

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -373,6 +387,69 @@ class PSTeams extends PSStreamModel<'team' | 'format'> {
...
+	}
+	compareTeams(serverTeam: Team & { team: string }, localTeam: Team) {
+		// if titles match exactly and mons are the same, assume they're the same team
+		// if they don't match, it might be edited, but we'll go ahead and add it to the user's
+		// teambuilder since they may want that old version around. just go ahead and edit the name
+		const mons = serverTeam.team.split(',');
+		let matches = 0;
+		const otherMons = PSTeambuilder.unpackTeam(localTeam.packedTeam);
+		for (const mon of mons) {
+			for (const otherMon of otherMons) {
```
Zarel, line 435: This is also O(n^2). Can't you just compare the exact strings? Or are you trying to do something like "same species good enough"?

Zarel: In which case both teams can be mapped to species and then sorted and then compared.

mia-pi-git: "Same species, good enough" is generally the approach, yes.

Zarel: Okay. In which case yeah, just map to species, sort, join, and compare.

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -373,6 +387,69 @@ class PSTeams extends PSStreamModel<'team' | 'format'> {
...
+			}
+		}
+		let sanitize = (name: string) => (name || "").replace(/\s+\(server version\)/g, '').trim();
+		const nameMatches = sanitize(serverTeam.name) === sanitize(localTeam.name);
+		if (!(nameMatches && serverTeam.format === localTeam.format)) {
+			return false;
+		}
+		// if it's been edited since, invalidate the team id on this one (count it as new)
+		// and load from server
+		if (mons.length !== otherMons.length || matches !== otherMons.length) return 'rename';
```
Zarel, line 449:
```suggestion
		let sanitize = (name: string) => (name || "").replace(/\s+\(server version\)/g, '').trim();
		const nameMatches = sanitize(serverTeam.name) === sanitize(localTeam.name);
		if (!(nameMatches && serverTeam.format === localTeam.format)) {
			return false;
		}
		// if it's been edited since, invalidate the team id on this one (count it as new)
		// and load from server
		const mons = serverTeam.team.split(',').map(toID).sort().join(',');
		const otherMons = PSTeambuilder.unpackTeam(localTeam.packedTeam).map(p => toID(p.species)).sort().join(',');
		if (mons !== otherMons) return 'rename';
```

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -373,6 +387,69 @@ class PSTeams extends PSStreamModel<'team' | 'format'> {
...
+		if (serverTeam.teamid === localTeam.teamid && localTeam.teamid) return true;
+		return true;
```
Zarel, line 451: ...What? This is just `return true`, right?

Zarel:
> I am putting off reviewing this for a few days so I can actually work on teambuilder. A cursory glance didn't reveal anything obviously wrong. I mainly think some documentation would make it much easier to read.

As promised, I'm ready to review this.

Zarel:
My main thing is that upload privacy doesn't make much sense as a global preference. It seems to me that "Save this in my account" and "Upload to Pokepaste 2" are two very different actions that should simply get different buttons.

play.pokemonshowdown.com/src/panel-teambuilder-team.tsx
```diff
@@ -137,6 +177,20 @@ class TeamPanel extends PSRoomPanel<TeamRoom> {
 					</p>
 				</div>
 			)}
+			<br />
+			<button class="button exportbutton" onClick={this.uploadTeam} disabled={this.exported}>
+				<i class="fa fa-upload"></i> Upload to Showdown database (saves across devices)
+			</button>
+			<label>
```
Zarel, line 184:
```suggestion
			<label class="label" style="display: inline-block">
```

KrisXV: are we no longer using `style={{ display: 'inline-block' }}`? or was that only ever a server thing

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -384,6 +398,60 @@ class PSTeams extends PSStreamModel<'team' | 'format'> {
...
+				team.loaded = false;
+				if (!matched) {
+					// hack so that it shows up in the format selector list
+					team.folder = '';
+					// team comes down from loginserver as comma-separated list of mons
+					// to save bandwidth
+					let mons = team.team.split(',').map((m: string) => ({ species: m, moves: [] }));
+					team.packedTeam = PSTeambuilder.packTeam(mons);
+					team.iconCache = null;
+					this.push(team);
```
Zarel, line 433: Oh, this doesn't work, you need to also set the key.

play.pokemonshowdown.com/src/client-main.ts
Zarel: I ended up not doing this. The code got too complex, and with my other optimizations, I think it's unnecessary.

Zarel:
Whew! Code is looking great. This is ready!

PR #2391, opened 2025-04-21
Make team box height flexible

andrebastosdias:
I use boxes as a way to store a lot of Pokémon with different sets, and they always end up overlapping.
Personally, I’d prefer not to have a minimum height and let the box adjust automatically, but I understand it's meant to indicate the maximum number of Pokémon in a team.
That said, it still wouldn't allow someone to manually add more than 24 Pokémon to a box.

### DIFF
--- a/play.pokemonshowdown.com/style/client.css
+++ b/play.pokemonshowdown.com/style/client.css
@@ -2251,7 +2251,8 @@ p.or:after {
 	color: transparent;
 }
 .teamlist .team.pc-box {
-	height: 139px;
+	min-height: 139px;
+	height: auto;
 }
 
 .teamedit {


andrebastosdias:
An example from one of my projects that automatically generates sets for the main tiers, using the most commonly used Pokémon and their stats for each cycle.
<details>
  <summary>Click to view screenshots</summary>

  image  
  image

</details>

Zarel:
I think the original reason we didn't do this was something about drag/drop bugs? If someone else wants to vouch that there's no problem with this, I can merge. Basically all our client effort is in the Preact rewrite, though.

andrebastosdias:
https://github.com/smogon/pokemon-showdown-client/pull/1222 The original implementation was done by @pyuk-bot 
I didn’t find anything related to a bug there, and the drag-and-drop functionality seems fine with these changes

Zarel:
Hm. Good enough for me, I suppose.

PR #2392, opened 2025-04-22
Replay: Add button to turn headers to go to that turn

davisow18:
The h2 element for turn headers in replays can now be clicked to go to that turn in the replay.

### DIFF
--- a/play.pokemonshowdown.com/src/battle-log.ts
+++ b/play.pokemonshowdown.com/src/battle-log.ts
@@ -354,33 +354,33 @@ export class BattleLog {
 
 		case 'turn':
 			const goToTurnButton = document.createElement('button');
-            const h2elem = document.createElement('h2');
-            goToTurnButton.className = 'subtle';
-            h2elem.className = 'battle-history';
-            let turnMessage;
-            let turnNumber: number;
-            if (this.battleParser) {
-                turnMessage = this.battleParser.parseArgs(args, {}).trim();
-                if (!turnMessage.startsWith('==') || !turnMessage.endsWith('==')) {
-                    throw new Error("Turn message must be a heading.");
-                }
-                turnMessage = turnMessage.slice(2, -2).trim();
-                turnNumber = parseInt(turnMessage.split(' ')[1].trim());
-                this.battleParser.curLineSection = 'break';
-            } else {
-                turnMessage = `Turn ${args[1]}`;
-                turnNumber = parseInt(args[1].split(' ')[1].trim());
-            }
-            goToTurnButton?.addEventListener?.('click', e => {
-                e.preventDefault();
-                this.scene?.battle.seekTurn(turnNumber);
-            });
-            goToTurnButton.innerHTML = BattleLog.escapeHTML(turnMessage);
-            h2elem.appendChild(goToTurnButton);
+			const h2elem = document.createElement('h2');
+			goToTurnButton.className = 'subtle';
+			h2elem.className = 'battle-history';
+			let turnMessage;
+			let turnNumber: number;
+			if (this.battleParser) {
+				turnMessage = this.battleParser.parseArgs(args, {}).trim();
+				if (!turnMessage.startsWith('==') || !turnMessage.endsWith('==')) {
+					throw new Error("Turn message must be a heading.");
+				}
+				turnMessage = turnMessage.slice(2, -2).trim();
+				turnNumber = parseInt(turnMessage.split(' ')[1].trim());
+				this.battleParser.curLineSection = 'break';
+			} else {
+				turnMessage = `Turn ${args[1]}`;
+				turnNumber = parseInt(args[1]);
+			}
+			goToTurnButton?.addEventListener?.('click', e => {
+				e.preventDefault();
+				this.scene?.battle.seekTurn(turnNumber);
+			});
+			goToTurnButton.innerHTML = BattleLog.escapeHTML(turnMessage);
+			h2elem.appendChild(goToTurnButton);
 
-            this.addSpacer();
-            this.addNode(h2elem);
-            break;
+			this.addSpacer();
+			this.addNode(h2elem);
+			break;
 
 		default:
 			if (this.addAFDMessage(args, kwArgs)) return;
--- a/play.pokemonshowdown.com/src/battle-log.ts
+++ b/play.pokemonshowdown.com/src/battle-log.ts
@@ -353,23 +353,34 @@ export class BattleLog {
 			break;
 
 		case 'turn':
-			const h2elem = document.createElement('h2');
-			h2elem.className = 'battle-history';
-			let turnMessage;
-			if (this.battleParser) {
-				turnMessage = this.battleParser.parseArgs(args, {}).trim();
-				if (!turnMessage.startsWith('==') || !turnMessage.endsWith('==')) {
-					throw new Error("Turn message must be a heading.");
-				}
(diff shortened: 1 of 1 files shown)


Zarel:
Hm. The messed-up indentation is making it hard to read the diff. Let me fix that for you.

play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -353,20 +353,31 @@ export class BattleLog {
...
 				turnMessage = this.battleParser.parseArgs(args, {}).trim();
 				if (!turnMessage.startsWith('==') || !turnMessage.endsWith('==')) {
 					throw new Error("Turn message must be a heading.");
 				}
 				turnMessage = turnMessage.slice(2, -2).trim();
+				turnNumber = parseInt(turnMessage.split(' ')[1].trim());
 				this.battleParser.curLineSection = 'break';
 			} else {
 				turnMessage = `Turn ${args[1]}`;
+				turnNumber = parseInt(args[1].split(' ')[1].trim());
```
Zarel, line 372:
```suggestion
				turnNumber = parseInt(args[1]);
```

play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -353,20 +353,31 @@ export class BattleLog {
...
 			}
-			h2elem.innerHTML = BattleLog.escapeHTML(turnMessage);
+			goToTurnButton?.addEventListener?.('click', e => {
+				e.preventDefault();
+				this.scene?.battle.seekTurn(turnNumber);
```
Zarel, line 376: Hm. This would probably be frustrating if someone clicked it in the middle of actively playing a battile. Maybe we should make an exception for that.

davisow18: Might I ask what about it could be frustrating? I think I'm not quite seeing what you are

Zarel: Say you're at turn 500 of a long stall game. You accidentally click on turn 16 while you're going back to recap yourself on what happened in the beginning. Now you have to hope your computer is fast enough that you don't stall out while you fast forward back to the latest turn.

Zarel: Yes, the "skip to end" button on long battle (say, 500 turns) takes a long time, and you could lose to timer before it finisheds.

davisow18: I see, I misunderstood how that button worked then. So my thought is that being able to rewind like this during a battle would be a nice feature, especially for the spectators, and I do think that this might be a niche case. I've had a couple of ideas for possible solutions and would love to know what you think is best:

1. set a rewind limit for active players, but not spectators (the button can't be clicked beyond current_turn - x)
2. the above option for players and spectators
3. remove the button altogether for the players, not spectators
4. remove the button altogether, it only appears for replays and not for live battles

Zarel: I think 3 is the least likely to be a problem. 2 and 4 are too drastic (spectators don't need to be limited in any way). I could be talked into 1 (if, for instance, you talked with competitive players and they said they wanted it) but we've never had replay controls for players at all and I don't think anyone wants them?

davisow18: Cool, I'll write 3 into the changes for now and send some feelers out and see if anyone might be interested in option 1. Thanks!

davisow18:
> Bump? I believe we're waiting on you to:
> 
> > remove the button altogether for the players, not spectators

Hi! Apologies for the delay, I had exams and was travelling. I will be back to finishing this week!

Zarel:
No worries about the delay; take as long as you like; I just wanted to make sure you hadn't forgotten.

davisow18:
Okay, I tested this and it now removes the functionality for active battlers. There might be a better way of implementing this but I looked through for multiple hours and then decided this was fine after not finding anything simpler.

Zarel:
Hmm. This can't possibly be the best way to do it. I bet checking `battle.myPokemon` would work.

play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -353,20 +353,44 @@ export class BattleLog {
...
+			let userId: string | undefined = undefined;
+			if (window.app?.user?.get) {
+				userId = window.app.user.get('userid');
+			}
+			if (!this.scene?.battle.isReplay && userId && [
+				this.scene?.battle.p1.id,
+				this.scene?.battle.p2.id,
+				this.scene?.battle.p3?.id,
+				this.scene?.battle.p4?.id,
+			].includes(userId)) {
```
Zarel, line 383:
```suggestion
			if (this.scene?.battle.myPokemon) {
```

I'm 80% sure this works. Can you test it and tell me if it does?

play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -353,20 +353,44 @@ export class BattleLog {
...
+				h2elem.innerHTML = BattleLog.escapeHTML(turnMessage);
+			} else {
+				goToTurnButton?.addEventListener?.('click', e => {
```
Zarel, line 386:
```suggestion
				goToTurnButton.addEventListener('click', e => {
```

I'm pretty sure `goToTurnButton` is guaranteed to exist here, so the optional chaining feels unnecessary.

play.pokemonshowdown.com/src/battle-log.ts
davisow18: So I tested this originally but I found that there is strange behavior with when myPokemon is null. For example, if you refresh the browser mid battle it will temporarily be null which will then add the button to the headers. So theoretically, if you played straight through the battle and never refreshed or clicked off it would be fine but if you did you would wind up with mixed headers (some being buttons, some not).
Also when you did refresh, any headers that were not buttons will become buttons up to the current turn.

I also wrote this solution off because it felt like an "indirect" solution, if that makes sense?

Zarel: Yeah that makes sense. Frustrating. On the other hand we do want battlers to be able to go to turn after the game is over. Probably the solution is that they should always be buttons, but they should be unclickable while the battle is ongoing.

davisow18: I think that is probably the best way forward. 

My knowledge of the code in its entirety is a little limited so I might be wrong here, but from my understanding: the battlelog has elements appended to it (like messages or turn headers) which are rendered on the DOM.
If this is the case then the turn headers would not be updated after the battle is finished unless the instant replay button was clicked?
Unless we wanted to, upon the battle ending, call something that goes and updates the turn headers I'm not sure how the button would become clickable without re-rendering the log by clicking the instant replay button.

Zarel: The simplest would be to make it always clickable and simply give an error to players in a battle.

Zarel: A more complex way would be go and update the turn headers with a script.

davisow18: I guess my thought process here is: why make it clickable to give an error for the users when it is already set to be non-clickable for users in battle? This adds extra clutter when the current change accomplishes the same thing without the extra noise. To be able to click the turn button, the players will have to click the replay button either way (excluding a script that updates the turn header)

PR #2393, opened 2025-04-22
Teambuilder: Update minAtk logic to account for attack evs

davisow18:
This is another PR for the same thing I previously created, I think I somehow deleted it on accident.

Per this approved suggestion: https://www.smogon.com/forums/threads/fix-the-0-ivs-for-attack-stat.3682315/

In some formats/niche situations, it makes sense to have a Pokémon with no physical moves but some attack EVs. In this situation, the current teambuilder logic will set the attack IV to a minimum.

This pull request would fix this and leave attack IVs unchanged for any Pokémon with attack EVs invested.

### DIFF
--- a/play.pokemonshowdown.com/js/client-teambuilder.js
+++ b/play.pokemonshowdown.com/js/client-teambuilder.js
@@ -3458,7 +3458,7 @@
 
 			if (this.curTeam.format === 'gen7hiddentype') return;
 
-			var minAtk = true;
+			var minAtk = set.evs && !set.evs['atk'];
 			// only available through an event with 31 Atk IVs
 			if (set.ability === 'Battle Bond' || ['Koraidon', 'Miraidon'].includes(set.species)) minAtk = false;
 			var hpModulo = (this.curTeam.gen >= 6 ? 2 : 4);


davisow18:
> Yeah, I think you deleted it by accident because you PRed your master branch. In the future, make sure you're in a feature branch before PRing. It's not too hard to branch after-the-fact, after you commit even, which I can walk you through if necessary.

Yeah that is exactly what happened lol, also why the indentation was weird on my other PR since I had to separate the commits from the two features. I got it all sorted now, thank you!

Zarel:
Unfortunately, like all the other teambuilder IV issues, this boils down to the teambuilder not really having a way to represent whether IVs are "at default" or "customized"; something that the new teambuilder fixes. I think this makes sense to implement into the Preact teambuilder but liable to cause weird interactions in the old teambuilder.

Zarel:
[Preact teambuilder does not currently have min atk IV logic at all, so this is not yet relevant, but I'll keep this mind.]

Zarel:
[Preact teambuilder now has min Atk IV logic, and doesn't have this problem]

PR #2394, opened 2025-04-23
Preact: Add support for hovering over roomtabs with notifications

KrisXV:
I have the `title` attribute set to `{hoverTitle || undefined}` so the title attribute is properly removed when there's no notifications instead of just being empty.


Zarel:
Thanks!

PR #2395, opened 2025-04-24
Preact: Add battle options and timer

ISenseAura:


play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -595,8 +658,6 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
 				const moveName = choices.getChosenMove(choices.current, choices.index()).name;
 				return <div class="controls">
 					<div class="whatdo">
-						<button name="openTimer" class="button disabled timerbutton"><i class="fa fa-hourglass-start"></i> Timer</button>
-						{this.renderOldChoices(request, choices)}
```
Zarel, line 599: This was probably a mistake?

play.pokemonshowdown.com/src/panel-popups.tsx
```diff
@@ -997,7 +997,7 @@ class BattleForfeitPanel extends PSRoomPanel {
 				</p>
 				<p>
 					<button type="submit" class="button"><strong>Forfeit</strong></button> {}
-					{!battleRoom.battle.rated && <button type="button" data-href="replaceplayer" class="button">
+					{!battleRoom?.battle?.rated && <button type="button" data-href="replaceplayer" class="button">
```
Zarel, line 1000: Why `?`?

I think crashing when something that shouldn't happen happens is fine. It'll just make it easier to track down the bug later.

play.pokemonshowdown.com/src/panel-popups.tsx
```diff
@@ -1425,6 +1425,202 @@ class LeaveRoomPanel extends PSRoomPanel {
...
+				</label>
+			</p>
+			<p>
+				<label class="checkbox">
+					<input
+						type="checkbox"
+						onChange={this.handleAllSettings}
+						name="ignorespects" checked={PS.prefs.ignorespects || false}
+					/> Ignore
+					spectators
```
Zarel, line 1587: I'm a little confused why you make things like this two lines. Like it's not consistent, either: some of these are two-lines and some of these are same-line.

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -813,8 +876,18 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
 			</ChatLog>
 			<ChatTextEntry room={room} onMessage={this.send} onKey={this.onKey} left={640} />
 			<ChatUserList room={room} left={640} minimized />
+			<button
+				class="icon button"
+				name="openBattleOptions"
```
Zarel, line 881:
```suggestion
```

You don't need `name` when you're using `data-href` or `data-cmd`.

We do still use `name`/`value` for the format/team dropdowns, but they're not used for stuff like this.

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -800,6 +849,20 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
 				</ChatLog>
 				<ChatTextEntry room={room} onMessage={this.send} onKey={this.onKey} left={0} />
 				<ChatUserList room={room} top={this.battleHeight} minimized />
+				<button
+					class="icon button"
+					name="openBattleOptions"
+					title="Options"
```
Zarel, line 855: The button itself already says "Battle Options". The title "Options" feels redundant with that.

play.pokemonshowdown.com/src/panel-popups.tsx
ISenseAura: my formatter does that whenever i manually split a `button`'s argument into newlines, 
gonna disable it.

play.pokemonshowdown.com/src/panel-popups.tsx
```diff
@@ -1512,6 +1708,30 @@ class RoomTabListPanel extends PSRoomPanel {
...
+		if (value === "stoptimer") room.send("/timer stop");
+		this.close();
+	};
+
+	override render() {
+		const room = this.props.room.getParent() as BattleRoom;
+		return <PSPanelWrapper room={this.props.room}><div class="pad">
+			{ room.timerInterval ?
+				<button class="button" value="stoptimer" onClick={this.triggerTimer}> Stop Timer</button> :
+				<button class="button" value="starttimer" onClick={this.triggerTimer}> Start Timer</button>}
```
Zarel, line 1730:
```suggestion
			{room.timerInterval ? (
				<button class="button" data-cmd="/closeand /timer stop"> Stop Timer</button>
			) : (
				<button class="button" data-cmd="/closeand /timer start"> Start Timer</button>
			)}
```

Oh I guess the problem is that the `/timer start` goes in the wrong room. Maybe we need a `/closeandinopener`.

play.pokemonshowdown.com/src/panel-battle.tsx
ISenseAura: yup that was a mistake

play.pokemonshowdown.com/src/panel-popups.tsx
ISenseAura: Yeah, but do we really need it? I don't think there are enough use cases to justify it, and to be honest now that i think about it, the logic for what happens after clicking the button should remain within the popup panel method imo

ISenseAura: oh nvm, i ignored the fact that this button only needs to send 1 command in a specifric room

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -283,13 +285,75 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
 			this.battleHeight = 360;
 		}
 	}
+	getTimerHTML(nextTick?: boolean) {
```
Zarel, line 288: Is there a reason this function returns HTML and not JSX?

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -283,13 +285,75 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
...
+		let time = 'Timer';
+		let room = this.props.room;
+		if (!room.battle)
+			return `<button name="openTimer" class="button timerbutton">
+			  <i class="fa fa-hourglass-start"></i> Timer
+			</button>`;
```
Zarel, line 294: This should simply never happen, right?

play.pokemonshowdown.com/src/panel-battle.tsx
ISenseAura: `button.setHTMLUnsafe?.(this.getTimerHTML(true))` does not accept jsx

Zarel: Oh. I'd just make the timer a Preact component and `forceUpdate()` every time it ticks.

play.pokemonshowdown.com/src/panel-battle.tsx
ISenseAura: yeah

play.pokemonshowdown.com/src/panel-battle.tsx
ISenseAura: ig i could do that

play.pokemonshowdown.com/src/panel-popups.tsx
```diff
@@ -1425,6 +1425,202 @@ class LeaveRoomPanel extends PSRoomPanel {
...
+				</label>
+			</p>
+			<p>
+				<label class="checkbox">
+					<input
+						type="checkbox"
+						onChange={this.handleAllSettings}
+						name="rightpanel" checked={PS.prefs.rightpanelbattles || false}
+					/> Open new battles on
+					the right side
```
Zarel, line 1616:
```suggestion
					/> Open new battles on the right side
```

play.pokemonshowdown.com/src/panel-popups.tsx
```diff
@@ -1425,6 +1425,202 @@ class LeaveRoomPanel extends PSRoomPanel {
...
+				</label>
+			</p>
+			<p>
+				<label class="checkbox">
+					<input
+						type="checkbox"
+						onChange={this.handleAllSettings}
+						name="autotimer" checked={PS.prefs.autotimer || false}
+					/> Automatically start
+					timer
```
Zarel, line 1606:
```suggestion
					/> Automatically start timer
```

play.pokemonshowdown.com/src/panel-popups.tsx
Zarel, line 1587:
```suggestion
					/> Ignore spectators
```

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -800,6 +845,19 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
 				</ChatLog>
 				<ChatTextEntry room={room} onMessage={this.send} onKey={this.onKey} left={0} />
 				<ChatUserList room={room} top={this.battleHeight} minimized />
+				<button
+					class="icon button"
+					name="openBattleOptions"
+					style={{ position: 'absolute', right: '75px', top: this.battleHeight }} data-href="battleoptions"
```
Zarel, line 851: What is the `icon` class supposed to do?

ISenseAura: that was from the old client

Zarel, line 851:
```suggestion
					data-href="battleoptions" class="button"
					style={{ position: 'absolute', right: '75px', top: this.battleHeight }} 
```

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -813,8 +871,17 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
 			</ChatLog>
 			<ChatTextEntry room={room} onMessage={this.send} onKey={this.onKey} left={640} />
 			<ChatUserList room={room} left={640} minimized />
+			<button
+				class="icon button"
+				title="Options"
+				style={{ position: 'absolute', right: '15px' }}
+				data-href="battleoptions"
```
Zarel, line 878:
```suggestion
				data-href="battleoptions" class="button"
				style={{ position: 'absolute', right: '15px' }}
```

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -813,8 +875,15 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
 			</ChatLog>
 			<ChatTextEntry room={room} onMessage={this.send} onKey={this.onKey} left={640} />
 			<ChatUserList room={room} left={640} minimized />
+			<button
+				data-href="battleoptions" class="button"
+				style={{ position: 'absolute', right: '15px' }}
+			>Battle Options
```
Zarel, line 881:
```suggestion
			>
				Battle Options
```

Zarel:
Thanks!

PR #2402, opened 2025-05-03
Nickname Dingbat support

oshaboy:
In the games you can give your Pokémon Emoticons that aren't in Unicode. I implemented them. It's a bit of a hack and I didn't test it a lot so maybe this should be a draft. 

The font is not made by me I downloaded it from Bulbapedia.  

image


Zarel:
Interesting idea. I'm guessing they're in Unicode's reserved space? Probably we'd want a way to input them...

oshaboy:
Some of them are over regular codepoints while others are in the PUA

oshaboy:
Probably downloading a font from Bulbapedia's internals which I don't know where it came from and putting it into the files verbatim isn't the best idea. And just adding it to the css as a font is too intrusive because it applies to non-PUA codepoints. I should try to find how to make a PUA only version. 

How would a keyboard for dingbats even integrate into the existing UI?

oshaboy:
The font might have been taken from the internal files of Pokémon Global Link

oshaboy:
The game let you type "halfwidth" dingbats in Western versions of the game and "fullwidth" dingbats in Eastern versions which are internally stored as separate codes to preserve the nicknames. Is this distinction important for Pokémon showdown?

Also Gen V used a different codepoint assignment from the Modern Games. PGLDings on Bulbapedia actually has both of them. 

I would like to clarify that by "intrusive". If you already used certain emoji they will be converted to the PGLDings form upon update. For example, if you have a snowman emoji in your Pokémon nickname will become the two blobs from PGLDings. So either we copy all the dingbats into the PUA or just accept that some nicknames will change.

Zarel:
"Keyboard for dingbats" doesn't seem _too_ hard to integrate into the existing UI, just have some buttons for them. The big question is whether this is worth it in the first place, which I have confused feelings about. I'd be curious how PKHeX handles this.

Zarel:
Okay, the Bulbapedia article has more details and leans me towards "sure, why not".

Zarel:
I guess "PGLDings" comes from some web version of pokemon home that displays them? That seems like a thing we might as well use.

oshaboy:
Well for now I will just move all the dingbats to the PUA in fontforge

kwsch:
> I guess "PGLDings" comes from some web version of pokemon home that displays them?

Was from [PGL](https://bulbapedia.bulbagarden.net/wiki/Pok%C3%A9mon_Global_Link); the font file is what was used by the site to render player-assigned strings from the game.

> I'd be curious how PKHeX handles this.

PKHeX serves the font file for its Nickname/OT text entry fields. It also has implemented the remappings when strings are transferred from one context to the next ([7->8](https://github.com/kwsch/PKHeX/blob/a85f91963052e05a6b1e045f633b0857c00a365d/PKHeX.Core/PKM/Strings/StringConverter8.cs#L191-L230)) as well as the [Font's available characters](https://github.com/kwsch/PKHeX/tree/a85f91963052e05a6b1e045f633b0857c00a365d/PKHeX.Core/PKM/Strings/Font).

For a popup list of the "extra" glyphs not accessible in QWERTY/etc keyboards, PKHeX provides a list of chars that are added to the Trash Byte Editor popup window:
https://github.com/kwsch/PKHeX/blob/a85f91963052e05a6b1e045f633b0857c00a365d/PKHeX.WinForms/Subforms/PKM%20Editors/Text.cs#L227-L334

oshaboy:
I changed my mind about the PUA assignments.

oshaboy:
image

Here's the "Keyboard"

play.pokemonshowdown.com/style/teambuilder.css
```diff
@@ -188,6 +188,7 @@
 	padding: 0;
 }
 
+
```
Zarel, line 191:
```suggestion
```

play.pokemonshowdown.com/js/client-teambuilder.js
```diff
@@ -3680,7 +3687,61 @@
...
+				"&#x2600;", "&#x2601;", "&#x2602;", "&#x2603;",
+				"&#xE081;", "&#xE082;", "&#xE083;", "&#xE084;", "&#xE087;", "&#xE085;", "&#xE086;",
+				"\n",
+				"&#x2642;","&#x2640;"
+			];
+			this.i = data.i;
+			this.team = data.team;
 
+			var buf = '<ul name="dingbats" class="popupmenu"><li>';
+			for (var i=0; i<dingbats_halfwidth.length; i++) {
```
Zarel, line 3716:
```suggestion
			for (var i = 0; i < dingbats_halfwidth.length; i++) {
```

play.pokemonshowdown.com/js/client-teambuilder.js
```diff
@@ -3680,7 +3687,61 @@
...
+				if (dingbats_halfwidth[i] == '\n'){
+					buf +='</li><li>'
```
Zarel, line 3718:
```suggestion
					buf += '</li><li>'
```

play.pokemonshowdown.com/js/client-teambuilder.js
```diff
@@ -3680,7 +3687,61 @@
...
+				"&#xE081;", "&#xE082;", "&#xE083;", "&#xE084;", "&#xE087;", "&#xE085;", "&#xE086;",
+				"\n",
+				"&#x2642;","&#x2640;"
+			];
+			this.i = data.i;
+			this.team = data.team;
 
+			var buf = '<ul name="dingbats" class="popupmenu"><li>';
+			for (var i=0; i<dingbats_halfwidth.length; i++) {
+				if (dingbats_halfwidth[i] == '\n'){
```
Zarel, line 3717:
```suggestion
				if (dingbats_halfwidth[i] == '\n') {
```

play.pokemonshowdown.com/js/client-teambuilder.js
```diff
@@ -3680,7 +3687,61 @@
...
+					buf += '<button name="insertDingbat" class="dingbat-button"';
+					buf += " halfwidth=" + dingbats_halfwidth[i] 
+					buf += " fullwidth=" + dingbats_fullwidth[i];
+					buf += " value=" + dingbats_halfwidth[i];
+					buf += ">" + dingbats_halfwidth[i] + "</button>";
+				}
+			}
+			buf += '</li><input type="checkbox" name="wideCheckbox"></input>Wide</ul>';
+			this.$el.html(buf);
+			var el=this.$el;
```
Zarel, line 3729: We generally would name this variable `$el`, like we do for other jQuery objects.

play.pokemonshowdown.com/js/client-teambuilder.js
```diff
@@ -3680,7 +3687,61 @@
...
+					buf += ">" + dingbats_halfwidth[i] + "</button>";
+				}
+			}
+			buf += '</li><input type="checkbox" name="wideCheckbox"></input>Wide</ul>';
+			this.$el.html(buf);
+			var $el = this.$el;
+			this.$el.find("input[name=wideCheckbox]").change(
+				function (e) {
+					var width = e.currentTarget.checked?"fullwidth":"halfwidth";
+					$el.find("button[name=insertDingbat]").each(function (_,elem) {
```
Zarel, line 3733:
```suggestion
					$el.find("button[name=insertDingbat]").each(function (_, elem) {
```

play.pokemonshowdown.com/style/client.css
```diff
@@ -489,6 +493,16 @@ p.or:after {
 .popupmenu .option {
 	width: 204px;
 }
+.popupmenu .dingbat-button {
+	font-family: PGLDings;
+	border: transparent;
+	background-color: transparent;
+}
+.dingbat-button:hover{
```
Zarel, line 501:
```suggestion
.dingbat-button:hover {
```

Zarel:
Okay. I've gotten new information that these don't exist in Gen 8 and later, which makes me less inclined to add them. Although maybe it'd still be neat in past gens.

oshaboy:
Gen 8 and 9 use the build in Switch keyboard for nicknames. But it's not like Pokémon Showdown really checks nicknames for in-game legality (For example, It lets you use characters in nicknames that aren't available on the Nintendo Switch keyboard). So I don't see why it shouldn't allow Gen 4 nicknames to be used in Gen 9. 

Though Gen 1 also lets you use characters that aren't in Unicode or Gen 4-7 (Specifically <sup>P</sup><sub>K</sub> and <sup>M</sup><sub>N</sub> which I am simulating using Markdown). The only reason I didn't add them is because PGLDings doesn't support them and I am unaware of a font that does.

Zarel:
Okay, so, how it works in newclient is great. The oldclient functionality, I'd like the button to only appear after the Nickname box is focused.

oshaboy:
I don't see why you think that little smiley face is cluttered. Also wouldn't clicking the button unfocus the nickname box and therefore make the button disappear. Seems like it's more intuitive to keep it always enabled.

Zarel:
Doesn't actually seem to work reliably. Editing the text in the box immediately make them work correctly, though.

image

Zarel:
`PGLDings, Verdana` instead of `Verdana, PGLDings` seems to fix it... but at what cost?

PR #2403, opened 2025-05-04
Preact: Support trapping/locking/disabling in battles

ISenseAura:


Zarel:
What's the status of maybeDisabled? Already implemented, or planning on implementing later?

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -152,23 +152,39 @@ class BattleDiv extends preact.Component<{ room: BattleRoom }> {
...
 }) {
 	const pokemon = props.pokemon;
 	if (!pokemon) {
 		return <button
-			data-cmd={props.cmd} class={`${props.disabled ? 'disabled ' : ''}has-tooltip`}
-			style={{ opacity: props.disabled === 'fade' ? 0.5 : 1 }} data-tooltip={props.tooltip}
+			data-href={props.trapped ? 'popup-cantswitch' : null}
+			data-message={props.trapped ? 'You are trapped and cannot switch' : null}
+			data-cmd={props.disabled ? null : props.cmd}
+			class={`${props.disabled ? 'disabled ' : ''}has-tooltip`}
```
Zarel, line 185: Hm

Zarel: I think plausibly we want to put this inside the command handler, rather than here.

Zarel: Probably command handlers need a button that they come from...

ISenseAura:
> What's the status of maybeDisabled? Already implemented, or planning on implementing later?

it is implemented, no?

ISenseAura:
> What's the status of maybeDisabled? Already implemented, or planning on implementing later?

```
{maybeDisabled &&
		<em class="movewarning">
			You <strong>might</strong> have some moves disabled, so you won't be able to cancel an attack!
		</em>}
```

play.pokemonshowdown.com/src/panel-battle.tsx
ISenseAura: the popup?

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -526,15 +590,29 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
...
 
-		return request.side.pokemon.map((serverPokemon, i) => {
-			const cantSwitch = trapped || i < numActive || choices.alreadySwitchingIn.includes(i + 1) || serverPokemon.fainted;
+		return (<div> {(trapped || maybeTrapped) &&
+			(maybeTrapped ?
+				<em class="movewarning">You <strong>might</strong> be trapped, so you won't be able to cancel a switch! <br /></em> :
+				<em class="movewarning">You <strong>are</strong> trapped and cannot switch! <br /></em>)}
+
+		{request.side.pokemon.map((serverPokemon, i) => {
+			const cantSwitch = maybeTrapped ||
```
Zarel, line 602: This seems wrong.  You should still be able to switch on a `maybeTrapped`, so it shouldn't be disabled in that case.

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -526,15 +590,29 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
...
-
+		const maybeTrapped = choices.currentMoveRequest()?.maybeTrapped;
 		const trapped = choices.currentMoveRequest()?.trapped;
 
-		return request.side.pokemon.map((serverPokemon, i) => {
-			const cantSwitch = trapped || i < numActive || choices.alreadySwitchingIn.includes(i + 1) || serverPokemon.fainted;
+		return (<div> {(trapped || maybeTrapped) &&
+			(maybeTrapped ?
+				<em class="movewarning">You <strong>might</strong> be trapped, so you won't be able to cancel a switch! <br /></em> :
+				<em class="movewarning">You <strong>are</strong> trapped and cannot switch! <br /></em>)}
```
Zarel, line 599:
```suggestion
				<em class="movewarning">You're <strong>trapped</strong> and cannot switch!<br /></em>)}
```

This message does not necessarily appear after a `maybeTrapped` (for instance, Fire Spin will just directly trap you), so bolding "are" could be confusing here.

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -475,13 +511,28 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
...
+				<em class="movewarning">
+					You <strong>might</strong> have some moves disabled, so you won't be able to cancel an attack!
+				</em>}
+			{maybeLocked &&
+				<em class="movewarning">
+					You <strong>might</strong> be locked into a move. <button
+						class="button"
+						onClick={this.chooseFight}
+					>
+						Try Fight button</button> (prevents switching if you\'re locked)
```
Zarel, line 525:
```suggestion
					You <strong>might</strong> be locked into a move. {}
					<button class="button" onClick={this.chooseFight}>Try Fight button</button> {}
					(prevents switching if you're locked)
```

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -152,10 +152,17 @@ class BattleDiv extends preact.Component<{ room: BattleRoom }> {
...
+	moveData: { pp: number, maxpp: number, disabled?: boolean },
+	type: Dex.TypeName,
+	tooltip: string,
 }) {
 	return <button
-		data-cmd={props.cmd} class={`movebutton type-${props.type} has-tooltip`} data-tooltip={props.tooltip}
+		data-cmd={props.cmd}
+		class={`movebutton type-${props.type} has-tooltip`}
+		data-tooltip={props.tooltip}
+		disabled={props.moveData.disabled}
```
Zarel, line 165: Unfortunately you can't disable buttons in the movemenu, because that suppresses tooltips and some other things that are important to us.

ISenseAura: huh, how do we achieve this then, add a class like to blur the button or something?

Zarel: We already have `class="disabled"` for this situation. Look at how I did it for `PokemonButton`. You don't need to do this, by the way, I'm about to push my fix.

ISenseAura: ok im dumb
thanks

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -359,7 +361,11 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
 			return;
 		case 'error':
 			if (args[1].startsWith('[Invalid choice]') && room.request) {
+				let oldChoices: string[] = [];
+				// theres probably a better way to do this?
+				if (room.choices?.choices.includes('testfight')) oldChoices.push('testfight');
 				room.choices = new BattleChoiceBuilder(room.request);
+				if (oldChoices) room.choices.choices = [...oldChoices];
```
Zarel, line 368:
```suggestion
				room.choices = new BattleChoiceBuilder(room.request);
```

Zarel (approved):
All right, this feels good to me now.

Zarel:
Thanks!

PR #2406, opened 2025-05-06
Preact: Add highlight, receivepopup and other client commands

ISenseAura:


play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -2243,4 +2355,44 @@ export const PS = new class extends PSModel {
 			this.prefs.set('autojoin', autojoin);
 		}
 	}
+	getHighlight(message: string, roomid: string) {
```
Zarel, line 2358: I think this and `highlightRegExp` stuff makes more sense as a static property of `ChatPanel` than here.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -176,16 +177,43 @@ export class ChatRoom extends PSRoom {
...
+		if (PS.getHighlight(message, this.id)) {
+			if (mayNotify) this.notify({
 				title: `Mentioned by ${name} in ${this.id}`,
 				body: `"${message}"`,
 				id: 'highlight',
 			});
 			return true;
 		}
 		/*
+		// i moved the following logic to ``PS.getHighlight``
```
Zarel, line 216: You can just remove this (and the comment after it, no need to keep it around after this.

play.pokemonshowdown.com/src/client-main.ts
ISenseAura: why static tho?

Zarel: So it's shared between all chatrooms, so it doesn't need to be tracked separately on each one.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -179,36 +181,81 @@ export class ChatRoom extends PSRoom {
...
-		return (((globalHighlightsRegExp &&
-		 globalHighlightsRegExp.test(message)) ||
-		  (roomHighlightsRegExp && roomHighlightsRegExp.test(message))));
-		*/
+		const id = PS.server.id + '#' + roomid;
+		const globalHighlightsRegExp = this.highlightRegExp?.['global'];
+		const roomHighlightsRegExp = this.highlightRegExp?.[id];
+		return (((globalHighlightsRegExp?.test(message)) || (roomHighlightsRegExp?.test(message))));
+	}
+	static updateHighlightRegExp(highlights: Record<string, string[]>) {
```
Zarel, line 210: Okay, actually this one shouldn't be static because it's called by commands. It's just the `highlightRegExp` itself that should be static. My bad.

Zarel: Sorry, no, I'm going back and forth on this. It's fine as-is.

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -926,6 +938,11 @@ export class PSRoom extends PSStreamModel<Args | null> implements RoomOptions {
 			this.send(target);
 			PS.leave(this.id);
 		},
+		'receivepopup'(target) {
+			PS.join(`popup-message` as RoomID, {
+				args: { message: target },
+			});
```
Zarel, line 944:
```suggestion
			PS.alert(target);
```

Zarel:
Thanks!

PR #2407, opened 2025-05-06
Preact: Fix users list not updating on autoreconnect

ISenseAura:
The autoreconnect logic was removing all message lines except `|c|` and possibly `|raw|`, causing the `|users|` line to be ignored. This resulted in the users list not updating correctly after a reconnect.
This fix ensures the `|users|` message is added early


Zarel:
Thanks!

PR #2409, opened 2025-05-07
Preact: Support auto-reconnecting through web worker

ISenseAura:
it worked as expected in my local  but still im not sure about the 
``this.reconnectWorker = new Worker('/js/reconnect-worker.js');``
is passing path like that fine?

----

https://www.smogon.com/forums/threads/the-disconnect-reconnect-process.3738903/


Zarel:
I'm confused, what's this for?

ISenseAura:
> I'm confused, what's this for?

Worker thread
It attempts to reconnect after every 5 seconds, i could have done it without the worker thread as well but its just to be on the safe side

Zarel:
Probably it makes more sense for the first connection to be on the worker thread? Also for the worker thread to be in a try block so it can fall back to a regular connection.

ISenseAura:
> Probably it makes more sense for the first connection to be on the worker thread? Also for the worker thread to be in a try block so it can fall back to a regular connection.

what does that mean? do you want to move the entire connection logic inside the worker thread or what?

Zarel:
Hm. If it's unclear, probably I can leave it for later.

play.pokemonshowdown.com/src/client-connection.ts
```diff
@@ -59,7 +78,7 @@ export class PSConnection {
 	}
 	disconnect() {
 		this.socket.close();
-		PS.connection = null;
+		// PS.connection = null;
```
Zarel, line 81: I still don't understand the value of keeping an old connection around.m

play.pokemonshowdown.com/src/reconnect-worker.ts
```diff
@@ -0,0 +1,17 @@
+let interval: number | null = null;
+
+self.onmessage = (e: MessageEvent) => {
+	if (e.data === 'start') {
+		if (interval === null) {
+			console.log("Attempting to reconnect in 5 seconds..");
+			interval = setInterval(() => {
+				postMessage('reconnect-check');
+			}, 5000); // or do we want it immediately
```
Zarel, line 9: I'd say immediate on the first try, with exponential backoff.

ISenseAura:
image
refer image - 1st connection is on the worker thread, further reconnections are on main thread.

need help -  how do i roomadd a message after reconnect

play.pokemonshowdown.com/src/client-connection.ts
```diff
@@ -13,70 +13,173 @@ declare const POKEMON_SHOWDOWN_TESTCLIENT_KEY: string | undefined;
 export class PSConnection {
 	socket: any = null;
 	connected = false;
-	queue = [] as string[];
+	queue: string[] = [];
+	private reconnectDelay = 1000;
+	private reconnectCap = 15000;
+	private shouldReconnect = true;
```
Zarel, line 19: `shouldReconnect` should probably be a function checking if uptime is under a day.

play.pokemonshowdown.com/src/client-connection.ts
```diff
@@ -13,70 +13,173 @@ declare const POKEMON_SHOWDOWN_TESTCLIENT_KEY: string | undefined;
...
+			return true;
+		} catch {
+			console.warn('Worker connection failed, falling back to regular connection.');
+			this.worker = null;
+			return false;
+		}
+	}
+
 	connect() {
+		this.worker = null; // ensure worker isn't used. we can keep
```
Zarel, line 88: This doesn't actually stop the worker from being used, though. I think `if (this.worker) return` is safer.

play.pokemonshowdown.com/src/reconnect-worker.ts
```diff
@@ -0,0 +1,67 @@
+declare const SockJS: any;
+
+interface ServerInfo {
+	protocol: string;
+	host: string;
+	port: string;
+	prefix: string;
+}
```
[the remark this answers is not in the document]
Zarel: Apparently this wasn't even typed before. I pushed some types for you to use.

Zarel:
Requesting a review while you have an unfixed build error. Bold. ;)

play.pokemonshowdown.com/src/panel-mainmenu.tsx
```diff
@@ -596,7 +596,9 @@ class MainMenuPanel extends PSRoomPanel<MainMenuRoom> {
 					</span>, " Disconnected"] : "Connecting..."}</em>
 				</button>
 				{PS.isOffline && <p class="buttonbar">
-					<button class="button" data-cmd="/reconnect"><i class="fa fa-plug" aria-hidden></i> <strong>Reconnect</strong></button>
+					<button class="button" data-cmd="/reconnect">
+						<i class="fa fa-plug" aria-hidden></i> <strong>Reconnect Now</strong>
+					</button>
```
Zarel, line 601: Why change this?

I prefer "Reconnect" because it's shorter, and shorter means easier to skim.

play.pokemonshowdown.com/src/client-connection.ts
```diff
@@ -189,7 +317,7 @@ export class PSStorage {
 
 				// in Safari, cross-origin local storage is apparently treated as session
 				// storage, so mirror the storage in the current origin just in case
-				if (document.location.hostname === Config.routes.client) {
+				if (document.location.hostname === window.Config.routes.client) {
```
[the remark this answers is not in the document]
Zarel: Oh, you didn't know I changed it so you're supposed to import `Config`

ISenseAura:
> Requesting a review while you have an unfixed build error. Bold. ;)

Nah that was a mistake, i meant to request review for the select default format PR but anyways thanks! :D

PR #2422, opened 2025-05-15
Preact Teambuilder: Add pokepaste link import functionality

Daniel1464:
Added the ability to directly import via pokepaste link

### DIFF
--- a/play.pokemonshowdown.com/src/battle-team-editor.tsx
+++ b/play.pokemonshowdown.com/src/battle-team-editor.tsx
@@ -15,7 +15,6 @@ import { PSSearchResults } from "./battle-searchresults";
 import { BattleNatures, BattleStatNames, type StatName } from "./battle-dex-data";
 import { BattleStatGuesser, BattleStatOptimizer } from "./battle-tooltips";
 import { PSModel } from "./client-core";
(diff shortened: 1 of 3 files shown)


Zarel:
This doesn't work like this on the old teambuilder...

Zarel:
Looks like the old teambuilder, you just paste the pokepaste link into the import/export screen and it happens automatically.

Daniel1464:
> Looks like the old teambuilder, you just paste the pokepaste link into the import/export screen and it happens automatically.

I was gonna do that but thought that having an explicit field for it was probably a better solution as it makes it slightly more obvious what you can do.

It wouldn't be hard to change it to the old way though if thats what you're thinking

Zarel:
Yeah, I was thinking about that concern. My main concern is UI clutter. I'll spend some time thinking about this one.

Daniel1464:
If i were to do it the old way, would it be fine to have an "import from link" button without the input label at the bottom? Or should i just do like a regex check when the user enters a link inside of the textarea

Zarel:
I did the regex check.

I think probably the thing to do here is to have some text "You can paste exported teams, JSON, packed teams, or pokepaste URLs in here" if the textbox is blank.

PR #2425, opened 2025-05-15
Preact: Teambuilder: Make link import change team title & format

Daniel1464:
Also adds a placeholder msg to the team textbox.
Btw is there any reason why we still use the net lib instead of the JS fetch api?


play.pokemonshowdown.com/src/battle-team-editor.tsx
```diff
@@ -854,6 +855,17 @@ class TeamTextbox extends preact.Component<{ editor: TeamEditorState, onChange?:
...
+        const notes = paste["notes"] as string;
+				if (notes.startsWith("Format: ")) {
+					const formatid = toID(notes.slice(8));
+					if (BattleFormats[formatid]) {
+						this.editor.setFormat(formatid);
+					}
+				}
+				const title = paste["title"] as string;
+				if (title && !title.startsWith('Untitled')) {
+					this.editor.team.name = title.replace(/[\|\\\/]/g, '');
```
Zarel, line 867: tbh there's an interesting question of, what if this team already has a name, which should win? Let's poll it.

play.pokemonshowdown.com/js/client-teambuilder.js
```diff
@@ -1422,7 +1422,7 @@
 					url: url,
 					success: function (data) {
 						if (/^https?:\/\/pokepast\.es\/.*\/json\s*$/.test(url)) {
-
+
```
Zarel, line 1425:
```suggestion

```

Zarel:
Drafting. You know how to run unit tests, right? `npm test`

Daniel1464:
> Drafting; please undraft once this passes tests. You know how to run unit tests, right? `npm test`

It doesn't pass lint because i think the lint isnt configured well for regex.
Escapes are necessary for I and / in regex but not in strings so it might be confused on that? maybe im confused

Zarel:
I don't think it's ever been wrong about escapes. There are cases where `|` and `/` don't need to be escaped inside `[]` if I remember correctly. Try it.

play.pokemonshowdown.com/src/battle-team-editor.tsx
```diff
@@ -1955,6 +1963,9 @@ class TeamWizard extends preact.Component<{
 	override render() {
 		const { editor } = this.props;
 		if (this.innerFocus) return this.renderInnerFocus();
+		if (editor.isLinkFetching) {
+			return <div class="teameditor">Fetching Paste...</div>;
+		}
```
Zarel, line 1968: I think this may have been a mistake. It's in the wrong place to do anything.

Zarel: Oh! Interesting workaround.

Zarel: You'd still need some way to update once the fetch was done, though...

Daniel1464: > You'd still need some way to update once the fetch was done, though...

Yeah i realized that - the newest commit should fix this.
Still bummed that the solution to re-renders in the client is seemingly spamming .forceUpdate() or rendering children - using react's builtin setState()/useState() or another state solution for auto rerendering seems like a good idea(but since the codebase is too big, idk if its worth it or not)

Zarel: You can't `setState` on a parent element without tons of code, though. This approach sidesteps that.

play.pokemonshowdown.com/src/panel-teambuilder-team.tsx
```diff
@@ -208,6 +210,7 @@ class TeamPanel extends PSRoomPanel<TeamRoom> {
 			</label>
 			<TeamEditor
 				team={team} onChange={this.save} readonly={!!team.teamid && !team.uploadedPackedTeam} resources={this.renderResources()}
+				fetching={this.fetching} setFetching={this.setFetching}
```
Zarel, line 213: ಠ_ಠ

Zarel: As far as I can tell, there's no reason to pull this all the way out here.

play.pokemonshowdown.com/src/battle-team-editor.tsx
Zarel: Like, to be clear, there's nothing preventing you from using `state`/`setState`. I don't, but there's nothing stopping you from doing it if you prefer it. I just see zero benefit to it.

Zarel:
> Btw is there any reason why we still use the net lib instead of the JS fetch api?

Historical reasons. Fetch didn't exist back when Showdown was first created. We still support browsers that don't have fetch.

Sure, we could polyfill it without too much difficulty, but now that we wrote Net, we might as well use it; it's got nicer syntax for request data, and the capitalization makes it more recognizable as "not a local variable".

play.pokemonshowdown.com/src/battle-team-editor.tsx
```diff
@@ -863,11 +873,30 @@ class TeamTextbox extends preact.Component<{ editor: TeamEditorState, onChange?:
...
-				// make sure it's still there:
-				const valueIndex = this.textbox.value.indexOf(value);
-				this.replace(paste.paste.replace(/\r\n/g, '\n'), valueIndex, valueIndex + value.length);
+				const pasteTxt = paste.paste.replace(/\r\n/g, '\n');
+				if (this.textbox) {
+					// make sure it's still there:
+					const valueIndex = this.textbox.value.indexOf(value);
+					this.replace(paste.paste.replace(/\r\n/g, '\n'), valueIndex, valueIndex + value.length);
+				} else {
+					this.editor.import(pasteTxt);
```
Zarel, line 885: ? probably it should do nothing if it's no longer there.

Zarel: I guess this is for "if we've switched to the wizard" again...

play.pokemonshowdown.com/src/battle-team-editor.tsx
```diff
@@ -863,11 +873,30 @@ class TeamTextbox extends preact.Component<{ editor: TeamEditorState, onChange?:
...
+				}
+				const notes = paste["notes"] as string;
+				if (notes.startsWith("Format: ")) {
+					const formatid = toID(notes.slice(8));
+					if (BattleFormats[formatid]) {
+						this.editor.setFormat(formatid);
+					}
```
Zarel, line 892:
```suggestion
					this.editor.setFormat(formatid);
```
Probably we should set this for past formats, too.

play.pokemonshowdown.com/style/battle-log.css
```diff
@@ -351,6 +351,10 @@ select.button {
 .value {
 	float: right;
 }
+.inputLabel {
+  font-family: Verdana, Helvetica, Arial, sans-serif;
+	font-size: 10pt;
+}
```
Zarel, line 357:
```suggestion
```
This appears unused (and is not capitalized correctly anyway).

play.pokemonshowdown.com/src/battle-team-editor.tsx
Daniel1464: yeah its for that

Zarel:
Thanks!

PR #2426, opened 2025-05-16
Preact: Teambuilder: team textbox fixes

Daniel1464:
Fixes for these "bugs":
1. When one of the team textbox selection panels are open and the user deletes the last set from the textbox, an error was thrown.
2. After selecting a panel, you can't switch to another panel by clicking on it(you have to click close on the first panel before doing so)


Zarel:
*gestures at build errors*

Daniel1464:
> _gestures at build errors_
> 
> It looks like you should configure your editor to infer indentation, or just use tabs in this project. Maybe make it respect editorconfig.

Sorry i usually add formatting in later commits not the initial one

Daniel1464:
My bad - i usually create a separate commit that applies lint onto code(out of habit)
I don't know how to resolve the issue in config.js though - i never changed it so idk how to fix it

Zarel:
Oh. That's just how Windows treats symlinks. I'll manually exclude them from lint, I suppose.

Zarel:
Here, this should make it possible for you to run unit tests. I'll merge this PR soon, so then your future PRs will run them automatically, which will be even nicer.

play.pokemonshowdown.com/src/battle-team-editor.tsx
```diff
@@ -925,6 +925,9 @@ class TeamTextbox extends preact.Component<{ editor: TeamEditorState, onChange?:
 					this.updateSearch();
 					this.setDirty = true;
 				}
+				if (selectionStart === 0 && selectionEnd === 0) {
+					this.clearInnerFocus();
+				}
```
Zarel, line 930: As far as I can tell, this would kick you out of inner focus if you cleared the input box while selecting your first pokemon (which, in addition to being a thing you might want to do normally, also happens if you select a search filter), which we don't want.

Zarel, line 930:
```suggestion
```
I think the other fixes are good, and I want one of your PRs merged so GitHub will stop needing approval for running tests in your PRs, so I'm just going to back out this specific change and merge this. If this is doing something important, talk to me about this.

Daniel1464: my bad - i don't quite remember what that was for, but I'll like notify u if its important

PR #2427, opened 2025-05-16
Silent detailschange

andrebastosdias:
Closes https://github.com/smogon/pokemon-showdown-client/issues/2323

Is it just this?


Zarel:
Thanks!

PR #2430, opened 2025-05-18
Add Home sprites

andrebastosdias:
image

I don't know if there is interest in changing this to a dropdownlist with
```
Default
Home sprites
3D sprites
2D sprites
```
@Zarel I think we could still create stills from the animated models, so we have 3D sprites for the Switch games and keep all options open.

### DIFF
--- a/play.pokemonshowdown.com/js/client-teambuilder.js
+++ b/play.pokemonshowdown.com/js/client-teambuilder.js
@@ -3714,7 +3714,7 @@
 				var spriteSize = data.spriteDir === 'sprites/dex' ? 120 : 96;
 				maxSpriteSize = Math.max(maxSpriteSize, spriteSize);
 				var spriteDim = 'width: ' + spriteSize + 'px; height: ' + spriteSize + 'px;';
-				var resize = (data.h ? 'background-size ' + data.h + 'px;' : '');
+				var resize = (data.h ? 'background-size:' + data.h + 'px;' : '');
 				buf += '<button name="setForm" value="' + form + '" style="';
 				buf += 'background-image: url(' + Dex.resourcePrefix + data.spriteDir + '/' + spriteid + '.png); ' + spriteDim + resize + '" class="option';
 				buf += (form === (species.forme || '') ? ' cur' : '') + '"></button>';
--- a/play.pokemonshowdown.com/js/client-teambuilder.js
+++ b/play.pokemonshowdown.com/js/client-teambuilder.js
@@ -3695,22 +3695,11 @@
 			this.room = data.room;
 			this.curSet = data.curSet;
 			this.chartIndex = data.index;
-			var species = this.room.curTeam.dex.species.get(this.curSet.species);
+			var dex = this.room.curTeam.dex;
+			var species = dex.species.get(this.curSet.species);
 			var baseid = toID(species.baseSpecies);
 			var forms = [baseid].concat(species.cosmeticFormes.map(toID));
-			var spriteDir = Dex.resourcePrefix + 'sprites/';
-			var spriteSize = 96;
-			var spriteDim = 'width: 96px; height: 96px;';
-
-			var gen = Math.max(this.room.curTeam.gen, species.gen);
-			var dir = gen > 5 ? 'dex' : 'gen' + gen;
-			if (Dex.prefs('nopastgens')) dir = 'dex';
-			if ((Dex.prefs('bwgfx') && dir === 'dex') || species.gen >= 8) dir = 'gen5';
-			spriteDir += dir;
-			if (dir === 'dex') {
-				spriteSize = 120;
-				spriteDim = 'width: 120px; height: 120px;';
-			}
+			var maxSpriteSize = 96;
 
 			var buf = '';
 			buf += '<p>Pick a variant or <button name="close" class="button">Cancel</button></p>';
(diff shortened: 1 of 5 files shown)


Zarel:
I've never thought we needed more fine-grained controls than "sprites or not". The last time we created stills from the models, we had a lot of people dislike them. I think the current approach is fine.

play.pokemonshowdown.com/src/battle-dex.ts
```diff
@@ -884,8 +884,17 @@ export const Dex = new class implements ModdedDex {
 			y: -3,
 		};
 		if (pokemon.shiny) spriteData.shiny = true;
-		if (Dex.prefs('nopastgens')) gen = 6;
+		if (Dex.prefs('nopastgens')) gen = 9;
 		if (Dex.prefs('bwgfx') && gen > 5) gen = 5;
+		let homeExists = (!species.isNonstandard || !['CAP', 'Custom'].includes(species.isNonstandard) ||
+			species.id === "xerneasneutral") && ![
+			// TODO check which forms are missing
+			"eeveestarter", "floetteeternal", "pichuspikyeared", "pikachubelle", "pikachucosplay", "pikachulibre", "pikachuphd", "pikachupopstar", "pikachurockstar", "poltchageistartisan", "polteageistantique", "sinistchamasterpiece", "sinisteaantique", "toxtricitylowkeygmax",
```
Marty-D, line 892:
```suggestion
			"floetteeternal", "pichuspikyeared", "pikachubelle", "pikachucosplay", "pikachulibre", "pikachuphd", "pikachupopstar", "pikachurockstar",
```

Marty-D:
The images are now in `/sprites/home` and `/sprites/home-shiny`.

andrebastosdias:
image
image
image
image
image
image
image

Mostly looks fine; some parts are a bit weird because of the text boxes.

andrebastosdias:
image

play.pokemonshowdown.com/js/client-teambuilder.js
```diff
@@ -1297,7 +1297,7 @@
 			buf += '<div class="setchart-nickname">';
 			buf += '<label>Nickname</label><input type="text" name="nickname" class="textbox" value="' + BattleLog.escapeHTML(set.name || '') + '" placeholder="' + BattleLog.escapeHTML(species.baseSpecies) + '" />';
 			buf += '</div>';
-			buf += '<div class="setchart" style="' + Dex.getTeambuilderSprite(set, this.curTeam.gen) + ';">';
+			buf += '<div class="setchart" style="' + Dex.getTeambuilderSprite(set, this.curTeam.dex.modid) + ';">';
```
Zarel, line 1300: At that point you could just send the whole dex in, so you don't need to extract the gen from the modid.

Zarel: I'll do this myself.

play.pokemonshowdown.com/src/battle-dex.ts
```diff
@@ -880,11 +898,12 @@ export const Dex = new class implements ModdedDex {
...
 
-	getTeambuilderSprite(pokemon: any, gen = 0) {
+	getTeambuilderSprite(pokemon: any, dex?: ModdedDex) {
 		if (!pokemon) return '';
-		const data = this.getTeambuilderSpriteData(pokemon, gen);
+		const data = this.getTeambuilderSpriteData(pokemon, dex);
 		const shiny = (data.shiny ? '-shiny' : '');
-		return `background-image:url(${Dex.resourcePrefix}${data.spriteDir}${shiny}/${data.spriteid}.png);background-position:${data.x}px ${data.y}px;background-repeat:no-repeat`;
+		const resize = (data.h ? `background-size:${data.h}px` : '');
+		return `background-image:url(${Dex.resourcePrefix}${data.spriteDir}${shiny}/${data.spriteid}.png);background-position:${data.x}px ${data.y}px;background-repeat:no-repeat;${resize}`;
```
Zarel, line 906:
```suggestion
		return `background-image:url(${Dex.resourcePrefix}${data.spriteDir}${shiny}/${data.spriteid}.png);background-position:${data.x + xOffset}px ${data.y + yOffset}px;background-repeat:no-repeat;${resize}`;
```

play.pokemonshowdown.com/src/battle-dex.ts
```diff
@@ -880,11 +898,12 @@ export const Dex = new class implements ModdedDex {
 		return spriteData;
 	}
 
-	getTeambuilderSprite(pokemon: any, gen = 0) {
+	getTeambuilderSprite(pokemon: any, dex?: ModdedDex) {
```
Zarel, line 901:
```suggestion
	getTeambuilderSprite(pokemon: any, dex?: ModdedDex, xOffset = 0, yOffset = 0) {
```

andrebastosdias:
Those suggested changes would cause other Pokémon like Serperior and Landorus-T to have their heads hidden on the Preact. Currently, only a few Pokémon are slightly obscured **under the bottom textbox in the Preact client** (such as Flutter Mane and Chi-Yu). All sprites seem fine with the old client. I’d suggest leaving it as is for now, and later hardcoding specific cases—like what's already done for Gen 6 sprites—once the Preact client becomes the main client. Or just rearrange the teambuilder a bit—I’m not sure if this is the final version set in stone.

Zarel:
These suggested changes would not change the xOffset or yOffset at all...

Anyway, they can still be used in `TeamTextbox` without hiding anything.

Zarel:
Thanks! I might tweak this later but this is good enough to deploy.

PR #2432, opened 2025-05-20
Preact: Notify users for joined tours despite global setting

ISenseAura:
forum approved
https://www.smogon.com/forums/threads/if-you-join-a-tournament-with-tour-notifications-turned-off-you-should-get-the-notifications-from-that-tournament.3675730/#post-10566479

### DIFF
--- a/play.pokemonshowdown.com/src/client-main.ts
+++ b/play.pokemonshowdown.com/src/client-main.ts
@@ -140,7 +140,7 @@ class PSPrefs extends PSStreamModel<string | null> {
 	 * Uses 1 and 0 instead of true/false for JSON packing reasons.
 	 */
 	ignore: { [userid: string]: 1 | 0 } | null = null;
-	tournaments: 'hide' | 'notify' | null = null;
+	tournaments: 'hide' | 'notify' | 'joined' | 'off' = 'notify'; // on by default
 	/**
 	 * true = one panel, false = two panels, left and right
 	 */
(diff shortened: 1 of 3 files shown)


play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -140,7 +140,7 @@ class PSPrefs extends PSStreamModel<string | null> {
 	 * Uses 1 and 0 instead of true/false for JSON packing reasons.
 	 */
 	ignore: { [userid: string]: 1 | 0 } | null = null;
-	tournaments: 'hide' | 'notify' | null = null;
+	tournaments: 'hide' | 'notify' | 'off' = 'notify'; // on by default
```
Zarel, line 143: We don't support settings that are anything but `null` by default.

Zarel: I think we should probably standardize on `null` by default. So that changing something to the default will remove the setting from storage.

Zarel: In general I'm concerned that these names are unclear.

ISenseAura: previously “null” was being used for No notifications which shouldnt be default

Zarel:
I think "Notify when joined" makes sense as a default. The other settings, `"notify"` and `"hide"` can stay as-is.

ISenseAura:
so
hide - hide tour 
notify - notify always
null - notify when joined

Zarel:
Yeah, that sounds good to me.

Zarel:
Thanks!

PR #2433, opened 2025-05-20
Preact: Fix confirmleaveroom popup bug

ISenseAura:
fixes the bug where closing room from roomtablist popup was throwing error

### DIFF
--- a/play.pokemonshowdown.com/src/panel-popups.tsx
+++ b/play.pokemonshowdown.com/src/panel-popups.tsx
@@ -1423,7 +1423,9 @@ class LeaveRoomPanel extends PSRoomPanel {
 
 	override render() {
 		const room = this.props.room;
-		const parentRoomId = (this.props.room.parentElem as HTMLInputElement).value;
+		// for some reason `parentElem` is null when trying to close room from `roomtablist` popup
+		// hence we use `parentRoomId` in that case
+		const parentRoomId = (this.props.room.parentElem as HTMLInputElement)?.value || room.parentRoomid;
 		return <PSPanelWrapper room={room} width={480}><div class="pad">
 			<p>Are you sure you want to exit this room?</p>
 			<p class="buttonbar">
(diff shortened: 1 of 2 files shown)


Zarel:
It seems better to fix the parent roomid not being correct...

Zarel:
Like this.

play.pokemonshowdown.com/src/panel-popups.tsx
```diff
@@ -1423,7 +1423,9 @@ class LeaveRoomPanel extends PSRoomPanel {
 
 	override render() {
 		const room = this.props.room;
-		const parentRoomId = (this.props.room.parentElem as HTMLInputElement).value;
+		// for some reason `parentElem` is null when trying to close room from `roomtablist` popup
+		// hence we use `parentRoomId` in that case
+		const parentRoomId = (this.props.room.parentElem as HTMLInputElement)?.value || room.parentRoomid as string;
 		return <PSPanelWrapper room={room} width={480}><div class="pad">
 			<p>Are you sure you want to exit this room?</p>
```
Zarel, line 1430: Specifying the roomid here is more readable _and_ would have made debugging easier.

Zarel:
(My commit also fixes the forfeit popup which I think you overlooked.)

play.pokemonshowdown.com/src/panel-popups.tsx
```diff
@@ -1423,7 +1423,9 @@ class LeaveRoomPanel extends PSRoomPanel {
 
 	override render() {
 		const room = this.props.room;
-		const parentRoomId = (this.props.room.parentElem as HTMLInputElement).value;
+		// for some reason `parentElem` is null when trying to close room from `roomtablist` popup
+		// hence we use `parentRoomId` in that case
+		const parentRoomId = (this.props.room.parentElem as HTMLInputElement)?.value || room.parentRoomid as string;
```
Zarel, line 1428: Wait. Why was this using `parentElem.value` in the first place? It should just always `parentRoomid`.

PR #2439, opened 2025-05-23
Add missing trim() to handleSend

bingingem:
Fixes issue where trailing spaces with /avatar (and possibly other commands)


Zarel:
Fair enough, thanks!

PR #2442, opened 2025-05-25
Preact: Improve room leave handling

Distrib-ps:
I added the room ID because, for example, if we take ticket and then leave; client send `/noreply /leave`, so we don't actually leave it properly. By specifying the room ID, it will be handled correctly and we'll properly leave the ticket.

Before
image

After
image


Zarel:
Thanks!

PR #2446, opened 2025-05-29
Preact: Do not highlight on 'raw' messages

ISenseAura:


play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -183,7 +183,16 @@ export class BattleLog {
...
+				if (window.PS) {
+					const room = window.PS.rooms[battle?.roomid || ''];
+					room.notify({
+						title: notifyTitle,
+						body: `"${message}"`,
+						id: 'highlight',
+					});
+				} else {
+					window.app?.rooms[battle?.roomid || '']?.notifyOnce(notifyTitle, "\"" + message + "\"", 'highlight');
+				}
```
Zarel, line 195: I guess there's still no better way than `window.PS`...

ISenseAura: hm?

Zarel:
PMs with your name should be highlighted, shouldn't they?

ISenseAura:
> PMs with your name should be highlighted, shouldn't they?

No, in the old client they are not

Zarel:
Hm. Well. This is an improvement, I'll take it. Thanks!

PR #2449, opened 2025-06-01
Preact: Fix DM highlight bugs

ISenseAura:
Fixes:

- clicking a PM header (to minimise) doesn't remove the highlight
- Home still highlighted after accepting a challenge from a user
- Receiving a PM while the box is focused creates a highlight


play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -2255,6 +2255,12 @@ export const PS = new class extends PSModel {
 			}
 			this.room = room;
 		}
+		if (this.room.type === "mainmenu") {
+			// dismiss all DM notifications when clicking Home
+			for (roomid of PS.miniRoomList) {
+				this.rooms[roomid]?.autoDismissNotifications();
+			}
+		}
```
Zarel: I don't know if this should be automatic...

Zarel: Convince me. Why not just do it on focus? That's what the old client does.

play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -31,6 +31,7 @@ declare function formatText(input: string, isTrusted?: boolean): string;
 export { MD5, formatText };
 
 export class BattleLog {
+	roomid: string | null;
```
Zarel: (Merge conflict because I already fixed this in batch 25)

play.pokemonshowdown.com/src/client-main.ts
ISenseAura: well, old client removes all notifications from mainmenu when you click the Home tab, current Preact client doesnt

play.pokemonshowdown.com/src/battle-log.ts
ISenseAura: oh, so we keep my change right?

play.pokemonshowdown.com/src/client-main.ts
ISenseAura: > well, old client removes all notifications from mainmenu when you click the Home tab, current Preact client doesnt

removes all notifications as in the Home tab doesnt highlight

Zarel: Individual DMs still highlight, though. I don't mind making it so individual DMs need to be focused.

play.pokemonshowdown.com/src/battle-log.ts
Zarel: I would say the `panels.tsx` change is the only one I'd want to keep.

ISenseAura: gotcha, reverting other changes

PR #2452, opened 2025-06-05
Move Animations: Take Heart

slainey:
https://github.com/user-attachments/assets/39b0d6d0-3095-4c20-9426-220297a1c565


Zarel:
Looks nice. Could we get an MIT license from you? Post here:

> I, [name or nickname], hereby license all my contributions to the Pokémon Showdown project, past and future, under the MIT license.

PR #2456, opened 2025-06-10
Move Animations: Lunar Blessing

slainey:
**Work in progress**: animation has a thumbs up from kalaloki and tb0lt, but still needs dhelmise's review

_https://youtu.be/bk8l9NEld1g (file size too large, this test shows doubles and singles)
**this video is old ^^^ see my newest comment for the current animation**_


slainey:
Suggestions from Tb0lt were to make lunar blessing animation apply to ally defenders in double battles, and due to this also make the animation smaller overall.
Here's what that looks like: https://youtu.be/S2wqKa-PELU?si=f9SIEnaiYjTQHZKx 

Got the thumbs up for tb0lt and kalalokki, so Lunar Blessing just needs a dhelmise review now

PR #2457, opened 2025-06-10
Preact: Add auto-rejoin for battle rooms on reconnect

DieterReinert:
This PR implements automatic rejoining of battle rooms after disconnection, as suggested by jetou.

## What it does
When spectating multiple battles and getting disconnected, users previously had to manually rejoin each battle room. This feature automatically tracks open battle rooms during disconnection and rejoins them upon reconnection.

## Why this is useful
As noted by jetou:
> when you're spectating a few battles it's very annoying to dc and have to rejoin all

This feature improves the spectator experience by removing the manual work of rejoining multiple battle rooms after network issues.

### DIFF
--- a/play.pokemonshowdown.com/src/client-connection.ts
+++ b/play.pokemonshowdown.com/src/client-connection.ts
@@ -19,6 +19,7 @@ export class PSConnection {
 	private shouldReconnect = true;
 	reconnectTimer: ReturnType<typeof setTimeout> | null = null;
 	private worker: Worker | null = null;
+	battleRoomsToRejoin: string[] = [];
 
 	constructor() {
 		const loading = PSStorage.init();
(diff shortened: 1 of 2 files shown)


play.pokemonshowdown.com/src/client-connection.ts
```diff
@@ -19,6 +19,7 @@ export class PSConnection {
 	private shouldReconnect = true;
 	reconnectTimer: ReturnType<typeof setTimeout> | null = null;
 	private worker: Worker | null = null;
+	battleRoomsToRejoin: string[] = [];
```
Zarel: I don't think this should be state. You can just iterate all rooms when you need this list.

DieterReinert:
@Zarel thanks for the feedback. You were right. I adjusted to code to compute it on demand.

Zarel:
Thanks!

PR #2462, opened 2025-06-18
Tooltip: Ignore nicknames on hover

hammy358:
Screenshot from 2025-06-18 15-55-11
Screenshot from 2025-06-18 15-54-36

### DIFF
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -315,12 +315,12 @@ export class BattleTooltips {
 				let index = 1;
 				for (const otherPokemon of side.pokemon) {
 					if (otherPokemon.getBaseSpecies().baseSpecies === species) {
-						buf += this.showPokemonTooltip(otherPokemon, null, false, index);
+						buf += this.showPokemonTooltip(otherPokemon, null, false, index,side);
 						index++;
 					}
 				}
 			} else {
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -820,21 +820,26 @@ export class BattleTooltips {
...
 		const gender = pokemon.gender;
 		if (gender === 'M' || gender === 'F') {
 			genderBuf = ` <img src="${Dex.fxPrefix}gender-${gender.toLowerCase()}.png" alt="${gender}" width="7" height="10" class="pixelated" /> `;
 		}
-
-		let name = BattleLog.escapeHTML(pokemon.name);
-		if (pokemon.speciesForme !== pokemon.name) {
+		//
+		
+		let name = pokemonSide !== this.battle.mySide && (this.battle.ignoreOpponent || this.battle.ignoreNicks) ? pokemon.speciesForme : BattleLog.escapeHTML(pokemon.name);
```
Karthik99999, line 836: I think it would be easier to check for the existence of `serverPokemon` rather than creating a whole new parameter.

hammy358: I did try this earlier, but my issue was that it would cause your own nicknames on the sidebar to not show up, since as far as I could tell there wasn't a differentiator between your opponents sidebar and your own.

PR #2463, opened 2025-06-20
Fix HP bar colours

andrebastosdias:
Server PR: https://github.com/smogon/pokemon-showdown/pull/11213

### DIFF
--- a/play.pokemonshowdown.com/src/battle.ts
+++ b/play.pokemonshowdown.com/src/battle.ts
@@ -148,25 +148,23 @@ export class Pokemon implements PokemonDetails, PokemonHealth {
 		}
 		return 'hpbar';
 	}
-	static getPixelRange(pixels: number, color: HPColor | '', gen?: number): [number, number] {
+	static getPixelRange(pixels: number, color: HPColor | ''): [number, number] {
 		let epsilon = 0.5 / 714;
 
 		if (pixels === 0) return [0, 0];
 		if (pixels === 1) return [0 + epsilon, 2 / 48 - epsilon];
-		if (!gen || gen >= 5) {
-			if (pixels === 9) {
-				if (color === 'y') { // ratio is > 0.2
-					return [0.2 + epsilon, 10 / 48 - epsilon];
-				} else { // ratio is <= 0.2
-					return [9 / 48, 0.2];
-				}
+		if (pixels === 9) {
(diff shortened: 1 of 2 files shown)


play.pokemonshowdown.com/src/battle.ts
```diff
@@ -148,23 +148,25 @@ export class Pokemon implements PokemonDetails, PokemonHealth {
 		}
 		return 'hpbar';
 	}
-	static getPixelRange(pixels: number, color: HPColor | ''): [number, number] {
+	static getPixelRange(pixels: number, color: HPColor | '', gen?: number): [number, number] {
```
Zarel: I'm a little concerned that this needs a gen number. In theory we should be able to calculate everything from numerator, denominator, and color.

andrebastosdias: You can't make those assumptions about the 9 and 24 pixels for gens 1 to 4. 9 pixels is always red and 24 pixels is always yellow, and they should include the full pixel range.

andrebastosdias: Oh wait, is that color the one forced by the server? Is it absent if the server doesn't send one?

andrebastosdias: We could avoid using the generation by forcing the client to send `r` for 9 pixels, or `y` for 24 pixels if we want to reduce the range. In generations 1 to 4, no color would be sent, so the client would use the full range.

andrebastosdias: Solved with the server.

Zarel: Gen 1 to 4 shouldn't send `r` `y` `g` then; that should fix it?

Zarel: > We could avoid using the generation by forcing the client to send `r` for 9 pixels, or `y` for 24 pixels if we want to reduce the range. In generations 1 to 4, no color would be sent, so the client would use the full range.

Yeah, that'd be my preferred solution.

Zarel (approved):

play.pokemonshowdown.com/src/battle.ts
andrebastosdias: Yup, already updated the server PR.

Zarel:
Thanks!

PR #2464, opened 2025-06-21
Tooltip: Show tera clearer on faint

hammy358:
Screenshot from 2025-06-24 23-11-22

### DIFF
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -849,8 +849,8 @@ export class BattleTooltips {
 
 			let types = serverPokemon?.terastallized ? [serverPokemon.teraType] : this.getPokemonTypes(pokemon);
 			let knownPokemon = serverPokemon || clientPokemon!;
-
-			if (pokemon.terastallized) {
+			
+			if (pokemon.terastallized || (pokemon.hasEverTerastallized && !isActive)) {
 				text += `<small>(Terastallized)</small><br />`;
 			} else if (clientPokemon?.volatiles.typechange || clientPokemon?.volatiles.typeadd) {
 				text += `<small>(Type changed)</small><br />`;
--- a/play.pokemonshowdown.com/src/battle.ts
+++ b/play.pokemonshowdown.com/src/battle.ts
@@ -96,6 +96,7 @@ export class Pokemon implements PokemonDetails, PokemonHealth {
 	prevItemEffect = '';
 	terastallized = '';
 	teraType = '';
+	hasEverTerastallized = false;
 
 	boosts: { [stat: string]: number } = {};
 	status: Dex.StatusName | 'tox' | '' | '???' = '';
@@ -878,8 +879,10 @@ export class Side {
 			pokemon.statusData = { ...oldpokemon.statusData };
 			if (oldpokemon.terastallized) {
 				pokemon.terastallized = oldpokemon.terastallized;
+				pokemon.hasEverTerastallized = oldpokemon.hasEverTerastallized;
 				pokemon.teraType = oldpokemon.terastallized;
 				oldpokemon.terastallized = '';
+				oldpokemon.hasEverTerastallized = false;
 				oldpokemon.teraType = '';
 			}
 			// we don't know anything about the illusioned pokemon except that it's not fainted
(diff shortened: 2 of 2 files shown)


play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -850,15 +850,20 @@ export class BattleTooltips {
 			let types = serverPokemon?.terastallized ? [serverPokemon.teraType] : this.getPokemonTypes(pokemon);
 			let knownPokemon = serverPokemon || clientPokemon!;
 
-			if (pokemon.terastallized) {
+			if (pokemon.terastallized || (knownPokemon.teraType && !this.battle.rules['Terastal Clause'] && !serverPokemon)) {
```
andrebastosdias: Won't this make it look like all Pokémon are Terastallized on open team sheets or best-of-3?

andrebastosdias:
I think what you want is to split `terastallized` into two separate variables—one that resets on fainting and one that doesn't. Or just change `terastallized` from `string` to `string | null` or something like that.

play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -849,16 +849,22 @@ export class BattleTooltips {
 
 			let types = serverPokemon?.terastallized ? [serverPokemon.teraType] : this.getPokemonTypes(pokemon);
 			let knownPokemon = serverPokemon || clientPokemon!;
-
-			if (pokemon.terastallized) {
+			const terastallizedAndFainted = knownPokemon.teraType && !this.battle.rules['Terastal Clause'] && !serverPokemon && pokemon.fainted;
```
andrebastosdias: You can't do this because in formats with Open Team Sheets (like VGC), all `Pokemon#teraType`'s are known/available. With your condition `terastallizedAndFainted = knownPokemon.teraType && ... && pokemon.fainted`, any Pokemon that faints will appear as if it had terastallized.

You have to add a boolean parameter `hasTerastallized`, or whatever you want to call it, in the `Pokemon` class next to the `terastallized` parameter.

hammy358:
> This is how it currently looks
> Images
> 
> I agree that you may keep the `(Terastallized)` tag if you terastallized before fainting, but you should revert the types to their original values and only keep `(Tera Type: <type>)`, instead of showing `(base: <types>)`. The latter makes it seem like you're still terastallized, which you’re **_not_** in case you're revived.

I did it like that since that's pretty much how the original suggestion put it, but I do agree going from teraed to not teraed and back to teraed with revival blessing is a little weird so it should be fixed.

play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -849,8 +849,8 @@ export class BattleTooltips {
 
 			let types = serverPokemon?.terastallized ? [serverPokemon.teraType] : this.getPokemonTypes(pokemon);
 			let knownPokemon = serverPokemon || clientPokemon!;
-
-			if (pokemon.terastallized) {
+			
+			if (pokemon.terastallized || (pokemon.hasEverTerastallized && !isActive)) {
 				text += `<small>(Terastallized)</small><br />`;
```
andrebastosdias, line 854:
```suggestion
			
			if (pokemon.terastallized) {
				text += `<small>(Terastallized)</small><br />`;
			} else if (pokemon.hasEverTerastallized) {
				text += `<small>(Previously Terastallized)</small><br />`;
```

andrebastosdias: Maybe it is a bit verbose, but it looks strange to have just `(Terastallized)` if it is not terastallized anymore. Feel free to change my suggested tag.

andrebastosdias:
It looks good. I think you are just missing here one thing. Add `pokemon.hasEverTerastallized = oldpokemon.hasEverTerastallized;` and `oldpokemon.hasEverTerastallized = false;` here https://github.com/smogon/pokemon-showdown-client/blob/87a8997278d70b77ed8baf2ebef0088eb7d57902/play.pokemonshowdown.com/src/battle.ts#L880-L883

hammy358:
> It looks good. I think you are just missing one thing. Add `pokemon.hasEverTerastallized = oldpokemon.hasEverTerastallized;` and `oldpokemon.hasEverTerastallized = false;` here
> 
> https://github.com/smogon/pokemon-showdown-client/blob/87a8997278d70b77ed8baf2ebef0088eb7d57902/play.pokemonshowdown.com/src/battle.ts#L880-L883

Should be good now. Appreciate all the feedback man.

PR #2465, opened 2025-06-21
Track abilities and items from request data

andrebastosdias:
Fixes:
- https://www.smogon.com/forums/threads/skill-swap-and-known-abilities.3765878/
- https://www.smogon.com/forums/threads/tooltip-and-known-items.3763936/

We could and maybe should update more things with the request.


andrebastosdias:
Superseded by #2732

PR #2466, opened 2025-06-24
Preact: Allow applying smogon analysis sets

DieterReinert:
Allows users to apply a Smogon analysis set from the abilities pane.


Zarel:
(Also, I think this code should be out of dexsearch)

DieterReinert:
@Zarel implemented feedback:
image

Zarel:
image

Okay, here's my redesign.

Zarel:
I think maybe above the ability list makes more sense. But this being your master branch, I would rather not do more development inside your PR...

In the future, _please_ don't PR your master branch.

PR #2470, opened 2025-07-12
Tooltip: Prevent PP from displaying as negative due to Pressure

hammer798:
Bug report: https://www.smogon.com/forums/threads/pressure-makes-pp-go-to-1.3767358/


Karthik99999:
The display shows negative pp on purpose. This allows for custom moves that would not show the correct max pp to at least show how many times the move was used through the negative pp.

Karthik99999:
One solution to keep the scenario I described in tact is to check if the move actually exists in the dex before doing the 0 pp min here.

hammer798:
That all makes sense (although I'll admit I have no idea how custom moves work). It sounds like this is broadly more trouble than it's worth, I can just close this.

PR #2474, opened 2025-07-17
Preact: Support autojoining minimized rooms

ISenseAura:
Fixes:
-  Any rooms opened on the left panel are removed from the autojoin list
-  Rooms minimised into a DM box are also not added to autojoin

### DIFF
--- a/play.pokemonshowdown.com/src/client-main.ts
+++ b/play.pokemonshowdown.com/src/client-main.ts
@@ -2656,7 +2656,7 @@ export const PS = new class extends PSModel {
 		if (!PS.server.registered) return;
 		let autojoins: string[] = [];
 		let autojoinCount = 0;
-		let rooms = this.rightRoomList;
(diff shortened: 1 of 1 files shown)


ISenseAura:
> Hm. They should probably also retain position. Which would require different syntax for how it's stored.

minirooms first, then left then right?

ISenseAura:
ok classic mistake, forgot to change branch

Zarel:
Okay, the thing is that this is part of the results table, and I actually wanted it to be separate from the results table. I want it to look very different from abilities. More like buttons than table rows, so it's obvious it's doing something else.

I took so long to comment because I was undecided between fixing this myself and telling you to fix it. But I think I have other higher priority fixes to Preact client right now, so I'm gonna push this to you for now.

Zarel:
So I think that sort of thing can't really be stored in a way that's backwards compatible with the old client. I've been going through possibilities in my head and I think probably the best way to do it is to wait until Preact client is fully deployed. 

The alternative would be to have a prefix on room names, for the other locations. That'd be backwards compatible.

PR #2477, opened 2025-07-22
Add a resource page indexing all public commands

mia-pi-git:
Went with a build script that was part of full build rather than a CRQ after some discussion. Figured I'd fold in some basic resources (from /intro and the like) since it's going to be on the homepage, might as well make it useful. Not wedded to the 'resources' title, just couldn't think of a better name offhand. 

Yes, the script has to be exactly as hacky as it is in order to easily programmatically load all of the chat commands from the server. No, it would not be cleaner to reimplement the entire thing ourselves. Yes, I tried.


mia-pi-git:
> At this point I'm just not accepting PRs unless they're for Preact client.

Aren't we still some months away from launching?

In any case, I ported it to preact in addition to the old version.

Zarel:
Can you PLEASE not merge stuff like this without a review???

Zarel:
Like, fair if it's been weeks and no one's gotten around to it. But how long did you wait? A few seconds?

image

Zarel:
> Aren't we still some months away from launching?

Should be less than a month if all goes well.

PR #2480, opened 2025-07-23
Replays: Support real-time playback

mia-pi-git:
This was... painful. If only we used promises here. @Zarel, would appreciate feedback - the actual battle engine isn't usually my cup of tea.

### DIFF
--- a/play.pokemonshowdown.com/src/battle.ts
+++ b/play.pokemonshowdown.com/src/battle.ts
@@ -3397,7 +3397,7 @@ export class Battle {
 	runMajor(args: Args, kwArgs: KWArgs, preempt?: boolean) {
 		switch (args[0]) {
 		case 't:': {
-			if (this.realtime && this.timerCount) {
+			if (this.realtime && this.timerCount && !this.seeking) {
 				if (this.timeWait) {
 					setTimeout(() => {
 						this.timeWait = null;
--- a/play.pokemonshowdown.com/src/battle.ts
+++ b/play.pokemonshowdown.com/src/battle.ts
@@ -1135,6 +1135,9 @@ export class Battle {
 	 * paused, not just waiting for the opponent to make a move.
 	 */
 	paused: boolean;
+	realtime = false;
+	timeWait: number | null = null;
+	timerCount = 0;
 
 	constructor(options: {
 		$frame?: JQuery,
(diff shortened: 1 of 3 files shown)


Zarel:
We don't use Promises? I thought I refactored it so the main loop worked on Promises, just not async/await.

play.pokemonshowdown.com/src/battle.ts
```diff
@@ -3393,6 +3396,32 @@ export class Battle {
...
+						this.nextStep();
+					}, this.timeWait);
+					return true;
+				}
+				const curStep = this.stepQueue.indexOf(`|t:|${args[1]}`);
+				const curTime = Number(args[1]);
+
+				const extractTime = (line: string) => Number(line.split('|')[2]);
+				const nextStep = this.stepQueue.findIndex((line, index) => {
+					if (index < curStep) return false;
```
Zarel, line 3413: You should be diffing this with the previous timestamp, not the next timestamp. You should be including chat, too.

In theory you should be able to save the previous timestamp. And if there's no saved previous timestamp you can just skip ahead.

play.pokemonshowdown.com/src/battle.ts
```diff
@@ -3393,6 +3396,32 @@ export class Battle {
 	}
 	runMajor(args: Args, kwArgs: KWArgs, preempt?: boolean) {
 		switch (args[0]) {
+		case 't:': {
+			if (this.realtime && this.timerCount) {
+				if (this.timeWait) {
+					setTimeout(() => {
```
Zarel, line 3402: I don't see a check for fast forward. You should be skipping all timeouts when fast forwarding.

play.pokemonshowdown.com/src/battle.ts
mia-pi-git: I made it diff with the next one so that the turn plays out while the timer goes along.

play.pokemonshowdown.com/src/battle.ts
mia-pi-git: Right, forgot about that. Accounted for it.

play.pokemonshowdown.com/src/battle.ts
```diff
@@ -3393,6 +3396,32 @@ export class Battle {
 	}
 	runMajor(args: Args, kwArgs: KWArgs, preempt?: boolean) {
 		switch (args[0]) {
+		case 't:': {
```
Zarel, line 3399: In theory all this code should be in scene...

play.pokemonshowdown.com/src/battle.ts
Zarel: But that's wrong...

Zarel: Wait. Is it? Actually let me see.

Zarel: Okay. I'm right. Timestamps get added at the beginning of turns.

https://github.com/smogon/pokemon-showdown/blob/master/sim/battle.ts#L2945

So everything after a timestamp happens after the timestamp occurs. So it needs to be diffed with the preceding time, not the next time.

mia-pi-git: I don't follow? If you want the turn to play out as the timer waits for the next interval, you diff against the timestamp for the next turn, no?

Zarel: Here, maybe it would be an easier to work through an example. Imagine most turns take 5 seconds, but between turn 2 and 3, someone spends an entire minute thinking:

```
|t:|0
[turn 1 happens]
|t:|5
[turn 2 happens]
|t:|65
[turn 3 happens]
```

After turn 1, when you hit `|t:|5`, you would want to wait 5 seconds. But your code makes it wait 60 seconds.

play.pokemonshowdown.com/src/battle.ts
Zarel: As in, in `BattleScene` in `battle-animations.ts`, rather than in `Battle` in `battle.ts`

play.pokemonshowdown.com/src/battle.ts
Zarel: The actual thing you should do is cache when `|t:|0` was, and when we hit it. That would let us know more precisely how long to wait.

PR #2485, opened 2025-07-29
Preact: Disallow challenges within 5 seconds of the last attempt

ISenseAura:


play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -1149,6 +1156,10 @@ class ChatPanel extends PSRoomPanel<ChatRoom> {
 			formatName: format,
 			teamFormat: format,
 		};
+		PS.mainmenu.canChallenge = false;
+		setTimeout(() => {
+			PS.mainmenu.canChallenge = true;
+		}, 10 * 1000);
```
Zarel: I would have a `lastChallenge` timestamp rather than a timeout. Timeouts have race conditions.

Zarel:
Thanks!

PR #2488, opened 2025-07-31
Preact: notify and tempnotify highlights

dot-Comfey:
Mainly did this so my room that uses /notifyrank a lot doesn't drive me crazy with irrelevant highlights.


Zarel:
Thanks!

PR #2492, opened 2025-08-03
Preact: Fix bug with accept & reject challenge

ISenseAura:


play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -314,6 +314,7 @@ export class ChatRoom extends PSRoom {
 		'reject'(target) {
 			this.challenged = null;
 			this.update(null);
+			this.sendDirect(`/reject ${this.pmTarget!}`);
```
```suggestion
			this.sendDirect(`/reject ${this.pmTarget || ''}`);
```

`pmTarget` is not guaranteed to exist here.

play.pokemonshowdown.com/src/panel-mainmenu.tsx
```diff
@@ -774,7 +774,7 @@ export class TeamForm extends preact.Component<{
 	onSubmit: ((e: Event, format: string, team?: Team) => void) | null,
 	onValidate?: ((e: Event, format: string, team?: Team) => void) | null,
 }> {
-	format = '';
+	format = this.props.format || '';
```
Zarel: Uh. I think there's a reason these are different. What is this for?

ISenseAura: When theres an Incoming challenge, we pass formatName as a prop to TeamForm

Zarel: Yeah, and `format` should be entirely unused in this case.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -314,6 +314,7 @@ export class ChatRoom extends PSRoom {
 		'reject'(target) {
 			this.challenged = null;
 			this.update(null);
+			this.sendDirect(`/reject ${this.pmTarget || ''}`);
```
Zarel: Actually. Is this even necessary? `/reject` is already treated correctly by the server.

ISenseAura: Old client sends /reject user tho

Zarel:
There we go. Thanks for the catch! This was the challenge bug I was talking about on the kanban board.

PR #2493, opened 2025-08-03
Preact: Update team order in localStorage after drag-and-drop

ISenseAura:


play.pokemonshowdown.com/src/panels.tsx
```diff
@@ -616,6 +616,7 @@ export class PSView extends preact.Component {
 					return;
 				}
 			}
+			if (PS.dragging?.type === 'team') return;
```
Zarel: This should be handled in `handleDrop`, not here. See above (lines 614-617).

play.pokemonshowdown.com/src/panel-teambuilder.tsx
```diff
@@ -168,6 +168,7 @@ class TeambuilderPanel extends PSRoomPanel<TeambuilderRoom> {
 		// before: Dragged goes before Over, splice at i
 		// after: Dragged goes after Over, splice at i - 1 + 1
 		PS.teams.list.splice(iOver, 0, draggedTeam);
+		PS.teams.save();
```
Zarel: Hm. This is saving too often, I think.

Zarel: Like this is happening every team you drag past. Feels excessive to save/load based on that...

play.pokemonshowdown.com/src/panels.tsx
ISenseAura: done

play.pokemonshowdown.com/src/panel-teambuilder.tsx
ISenseAura: moved inside `handleDrop`

play.pokemonshowdown.com/src/panels.tsx
Zarel: Not done, this still exists.

play.pokemonshowdown.com/src/panel-teambuilder-team.tsx
```diff
@@ -107,6 +107,11 @@ class TeamPanel extends PSRoomPanel<TeamRoom> {
 
 	static formatResources = {} as Record<string, FormatResource>;
 
+	static handleDrop = (e: Event) => {
+		PS.teams.save();
+		return true;
+	};
```
Zarel: This should only happen if the thing being dragged is a team. The `return true` part stops the error message, which is bad if you drag an Excel spreadsheet into the teambuilder or something.

Zarel: So the way to think about this is, there are five different things that can be dragged and dropped:

PS custom:
- teams
- roomtabs

Browser built-in:
- chunks of HTML
- plaintext
- files

And all of these call `handleDrop` and need to be handled differently.

Zarel: Also: This should go in `panel-teambuilder`, not `panel-teambuilder-team`.

PR #2495, opened 2025-08-04
Use gen 5 animated sprites if available

andrebastosdias:
https://www.smogon.com/forums/threads/use-available-2d-animations-with-pok%C3%A9mon-that-do-not-have-3d-models-yet.3766462/

### DIFF
--- a/play.pokemonshowdown.com/src/battle-dex.ts
+++ b/play.pokemonshowdown.com/src/battle-dex.ts
@@ -634,19 +634,11 @@ export const Dex = new class implements ModdedDex {
 		spriteData.gen = Math.max(graphicsGen, Math.min(species.gen, 5));
 		const baseDir = ['', 'gen1', 'gen2', 'gen3', 'gen4', 'gen5', '', '', '', ''][spriteData.gen];
 
-		let animationData = null;
 		let miscData = null;
 		let speciesid = species.id;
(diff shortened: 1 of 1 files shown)


andrebastosdias:
Some Gen 5 sprites are duplicated in `ani` and `gen5ani`, with the former containing sprites it shouldn't. This PR allows them to be removed.
Example:
https://play.pokemonshowdown.com/sprites/ani/colossoil.gif
https://play.pokemonshowdown.com/sprites/gen5ani/colossoil.gif

Marty-D:
> Gen 5 back static sprites (for those without designated back sprites) are just mirrored versions of the front sprites. If we want to support Gen 5 animated back sprites, we'd need @Marty-D to mirror the missing animated sprites (both regular and shiny).

When I asked about this last year, the spriters explicitly requested that this not happen. When they send in an actual static back sprite they want it to be seen instead of a flipped animated front sprite.

andrebastosdias:
> When I asked about this last year, the spriters explicitly requested that this not happen. When they send in an actual static back sprite they want it to be seen instead of a flipped animated front sprite.

How cumbersome would it be to mirror the images in real time? That way, we could perform all the checks and establish the priorities programmatically. Of course, this would require deleting the inverted Gen 5 back sprites.

PR #2499, opened 2025-08-08
Preact: Add automatic open team sheet requests

Daniel1464:
Fixes a visual glitch where the timer in battles would visually continue to count down even after the user has made their choice.
Also adds the option to automatically request for open team sheets in vgc battles(does not do anything for singles).

### DIFF
--- a/play.pokemonshowdown.com/src/panel-battle.tsx
+++ b/play.pokemonshowdown.com/src/panel-battle.tsx
@@ -414,7 +414,8 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
 			this.send('/timer on');
 			room.autoTimerActivated = true;
 		}
-		if (PS.prefs.autoopenteamsheets) {
+    const formatName = room.battle.id.split("-")[1]?.toLowerCase();
+		if (PS.prefs.autoopenteamsheets && formatName?.includes("vgc")) {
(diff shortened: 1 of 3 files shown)


Zarel:
Uh. Is there an auto-OTS thread in Suggestions? I don't play VGC so I can't evaluate the feature.

Zarel:
This should really be two different PRs.

Daniel1464:
> Uh. Is there an auto-OTS thread in Suggestions? I don't play VGC so I can't evaluate the feature.
Yeah there is - i think its here https://www.smogon.com/forums/threads/changes-to-battle-options.3765932/

> This should really be two different PRs.
Yeah it should be; mb, i was a little lazy since the first change is literally one line of code

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -201,6 +201,7 @@ class TimerButton extends preact.Component<{ room: BattleRoom }> {
 		const room = this.props.room;
 		if (!this.timerInterval && room.battle.kickingInactive) {
 			this.timerInterval = setInterval(() => {
+				if (room.choices?.isDone()) return;
```
Zarel:
```suggestion
```

Split off into d728ac2

PR #2500, opened 2025-08-08
Mirror sprites automatically

andrebastosdias:
Read this first: https://github.com/smogon/pokemon-showdown-client/pull/2495#issuecomment-3168687478

---

Everything seems to be working, but `battle-animations.ts` needs a second check, as I’m not sure if all the locations are correct.
Of course, this would require @Marty-D to delete the mirrored static Gen 5 back sprites.

Here, you can see the result, modded to force front sprites.

https://github.com/user-attachments/assets/46cc80b1-db08-4bd0-ae78-0d20c102ee2e

### DIFF
--- a/play.pokemonshowdown.com/src/battle-dex.ts
+++ b/play.pokemonshowdown.com/src/battle-dex.ts
@@ -198,6 +198,7 @@ export interface SpriteData {
 	isFrontSprite?: boolean;
 	cryurl?: string;
 	shiny?: boolean;
+	mirror?: boolean;
 }
 
 export interface TeambuilderSpriteData {
@@ -604,6 +605,7 @@ export const Dex = new class implements ModdedDex {
 			isFrontSprite: false,
 			cryurl: '',
 			shiny: options.shiny,
+			mirror: false,
 		};
 		let name = species.spriteid;
 		let dir;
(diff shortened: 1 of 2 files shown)


play.pokemonshowdown.com/src/battle-dex.ts
```diff
@@ -707,41 +709,57 @@ export const Dex = new class implements ModdedDex {
...
-				if (facing.endsWith('f')) name += '-f';
-				if (spriteData.gen >= 6) spriteData.pixelated = false;
-				dir = animDir + 'ani' + dir;
-				spriteData.w = animationData[facing].w;
-				spriteData.h = animationData[facing].h;
-				spriteData.url += dir + '/' + name + '.gif';
-				animatedSprite = true;
-				break;
+		const possibleSprites: AnyObject[] = [];
+		for (const possibleFacing of ['back', 'front']) {
```
andrebastosdias, line 713: Just to be clear, it tries all the back sprites first.

play.pokemonshowdown.com/src/battle-dex.ts
```diff
@@ -707,41 +709,56 @@ export const Dex = new class implements ModdedDex {
...
-				if (facing.endsWith('f')) name += '-f';
-				if (spriteData.gen >= 6) spriteData.pixelated = false;
-				dir = animDir + 'ani' + dir;
-				spriteData.w = animationData[facing].w;
-				spriteData.h = animationData[facing].h;
-				spriteData.url += dir + '/' + name + '.gif';
-				animatedSprite = true;
-				break;
+		const possibleSprites: AnyObject[] = [];
+		for (const possibleFacing of facing === 'front' ? ['front'] : ['back', 'front']) {
```
andrebastosdias, line 713: Just to be clear, it tries all the back sprites first.

andrebastosdias:
In this PR https://github.com/smogon/pokemon-showdown-client/pull/2495, we added Gen 5 animated sprites for the opponent’s Pokémon. Our own Pokémon still use static sprites.

As a QoL improvement, we no longer need to import sprites and mirrored sprites (for the back) separately.

andrebastosdias:
Front Pokemon:
1. ani
2. gen5ani
3. gen5

Back Pokemon:
1. ani-back
2. gen5ani-back
3. gen5-back
4. ani (mirrored)
5. gen5ani (mirrored)
6. gen5 (mirrored)

This is what this PR implements. Currently, steps 2 (mostly irrelevant), 4 (very irrelevant), and 5 (**very relevant**) are skipped.

Zarel:
Are there pokemon without a `gen5-back`? I thought they all had one...

Zarel:
What have we been using, then?

Zarel:
The thing I'm saying is, my impression is that we should have `gen5-back` entries for every single pokemon, even if they are placeholder "the front-sprite but flipped" entries:

https://play.pokemonshowdown.com/sprites/gen5-back/ogerpon-cornerstone.png

So the code should never have a reason to fall back to possibilities 4-6.

andrebastosdias:
They have. The point of this PR was to use mirrored animated sprites when no _**real**_ back sprites are available.

andrebastosdias:
Because you would delete the mirrored sprites from `*-back`. You don't need them, you can mirror sprites directly in the client.

Let’s take Cinderace, Skeledirge and Pecharunt as examples (and imagine none of them have 3D sprites).

Cinderace has a "real" back sprite. Skeledirge does not have a "real" back sprite but does have an animated front sprite. Pecharunt only has a static front sprite.

Cinderace would go:

1. ani-back ❌
2. gen5ani-back ❌
3. gen5-back ✅

Skeledirge would go:

1. ani-back ❌
2. gen5ani-back ❌
3. gen5-back ❌
4. ani (mirrored) ❌
5. gen5ani (mirrored) ✅

Pecharunt would go:

1. ani-back ❌
2. gen5ani-back ❌
3. gen5-back ❌
4. ani (mirrored) ❌
5. gen5ani (mirrored) ❌
6. gen5 (mirrored) ✅

andrebastosdias:
> I don't think this would improve things. Partially because I think flipped sprites would look worse animated, and partially because I think we do want gen5-back entries for every sprite even if they're placeholders.

You can see how it looks right now in turn 1 (static mirrored Chien-Pao vs. animated front Chien-Pao) https://replay.pokemonshowdown.com/gen9bssregj-2419344111

> Also, this PR doesn't even change the thing where we currently do not track whether or not a given gen5-back sprite exists, so are you telling me this PR does nothing?

What do you mean? You would track it because the non-real back sprites inside gen5-back would be deleted. And it does something: in the replay above, Player 1’s Chien-Pao would also be animated (but flipped).

PR #2502, opened 2025-08-09
Preact: Do not subtle notify on seen messages

ISenseAura:


Zarel:
There's gotta be a better way to do this... oh well, I suppose we'll figure that one out later.

PR #2504, opened 2025-08-09
Battle: Ignore nicknames on tooltip

ISenseAura:


Zarel:
While I like the simplicity of your PR, cooshaug is less frequently a contributor so I figure I'll give them the credit to be nice.

Zarel:
Hm. As far as I can tell there's no way to mark as duplicate? Anyway this is a duplicate of #2508.

play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -834,6 +834,7 @@ export class BattleTooltips {
 		if (pokemon.speciesForme !== pokemon.name) {
 			name += ` <small>(${BattleLog.escapeHTML(pokemon.speciesForme)})</small>`;
 		}
+		if (this.battle.ignoreNicks) name = `${BattleLog.escapeHTML(pokemon.speciesForme)}`;
```
Zarel, line 837: For the record, default nickname is `baseSpecies`, not `speciesForme`. Not available on `pokemon` so you'd need to `getSpecies` to grab it.

Zarel, line 837: Also what's with the cast?

ISenseAura: > Also what's with the cast?

Since i was making this change on phone, i just copy pasted the above line, removed the small tag but forgot to remove the `

PR #2505, opened 2025-08-09
Preact: Fix bug with accept challenge

ISenseAura:
i missed this last time
so basically when you have starred a format that requires a team and when someone challenges you to a format that requires no team, the `TeamForm` considers the starred format instead of the incoming challenge format and hence shows the error 'This format requires team'

hence suggesting use `props.format` in case of incoming challenge, i probably used it at wrong place last time

### DIFF
--- a/play.pokemonshowdown.com/src/panel-mainmenu.tsx
+++ b/play.pokemonshowdown.com/src/panel-mainmenu.tsx
@@ -782,7 +782,7 @@ export class TeamForm extends preact.Component<{
 	};
 	submit = (ev: Event, validate?: 'validate') => {
 		ev.preventDefault();
-		const format = this.format;
+		const format = this.props.format || this.format;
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/src/panel-mainmenu.tsx
```diff
@@ -782,7 +782,7 @@ export class TeamForm extends preact.Component<{
 	};
 	submit = (ev: Event, validate?: 'validate') => {
 		ev.preventDefault();
-		const format = this.format;
+		const format = this.props.format || this.format;
```
Zarel, line 785: Ah yes, I considered fixing it that way, but then I decided to just make it so `this.format` would always match `this.props.format`, so we wouldn't have to write `this.props.format || this.format` everywhere it showed up.

Zarel:
Fixed in #2538

PR #2508, opened 2025-08-09
Hide nickname in battle tooltip

cooshaug:
Bug reported [here](https://www.smogon.com/forums/threads/ignore-nicknames-doesnt-ignore-in-tooltips.3747709/).

### DIFF
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -830,9 +830,10 @@ export class BattleTooltips {
 			genderBuf = ` <img src="${Dex.fxPrefix}gender-${gender.toLowerCase()}.png" alt="${gender}" width="7" height="10" class="pixelated" /> `;
 		}
 
-		let ignoreNicks = this.battle.ignoreNicks || this.battle.ignoreOpponent;
-		let name = BattleLog.escapeHTML(ignoreNicks ? pokemon.speciesForme : pokemon.name);
-		if (!ignoreNicks && pokemon.speciesForme !== pokemon.name) {
+		const ignoreNicks = this.battle.ignoreNicks || this.battle.ignoreOpponent;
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -830,8 +830,9 @@ export class BattleTooltips {
 			genderBuf = ` <img src="${Dex.fxPrefix}gender-${gender.toLowerCase()}.png" alt="${gender}" width="7" height="10" class="pixelated" /> `;
 		}
 
-		let name = BattleLog.escapeHTML(pokemon.name);
-		if (pokemon.speciesForme !== pokemon.name) {
+		let ignoreNicks = this.battle.ignoreNicks || this.battle.ignoreOpponent;
```
Zarel: Does `ignoreOpponent` usually also apply `ignoreNicks`? I can see arguments for both approaches. I think because `ignoreNicks` is a separate setting, I would default to no.

Zarel: Looks like yes. Unless someone objects, I'm going to change that, but that's out of scope for this PR.

play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -830,8 +830,9 @@ export class BattleTooltips {
...
+		let name = BattleLog.escapeHTML(ignoreNicks ? pokemon.speciesForme : pokemon.name);
+		if (!ignoreNicks && pokemon.speciesForme !== pokemon.name) {
```
Zarel:
```suggestion
		const ignoreNicks = this.battle.ignoreNicks || this.battle.ignoreOpponent;
		const nickname = ignoreNicks ? Dex.species.get(pokemon.speciesForme).baseSpecies : pokemon.name;
		let name = BattleLog.escapeHTML(nickname);
		if (pokemon.speciesForme !== nickname) {
```

Zarel:
Thanks!

PR #2510, opened 2025-08-15
[Client] Refactor Xerneas formes

andrebastosdias:
Server PR: https://github.com/smogon/pokemon-showdown/pull/11369
The sprites' names and the icons' indices have to be updated.

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -383,10 +383,10 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 				}
 				if (isGen9BH) {
 					if ((species.natDexTier === 'Illegal' || species.forme.includes('Totem')) &&
-						!['Floette-Eternal', 'Greninja-Ash', 'Xerneas-Neutral'].includes(species.name)) {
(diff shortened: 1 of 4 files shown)


Marty-D:
It seems like a bad idea to not only break 12 years of replays but also break every other project that interacts with the server and these files. What problem is this solving that the current implementation can't handle?

andrebastosdias:
> It seems like a bad idea to not only break 12 years of replays but also break every other project that interacts with the server and these files. What problem is this solving that the current implementation can't handle?

This (the server PR) is the accurate implementation of the Xerneas forms. You don't need to add checks for Xerneas in combination with its ability in Hackmons or other mods, because it automatically transforms from `Xerneas-Neutral` to `Xerneas-Active` at the start of a battle.

I see two ways to handle the replays:

* Use the date to route `Xerneas` to the correct sprite.
* From now on, make Xerneas explicitly use the full form name (`Xerneas-Neutral` or `Xerneas-Active`), so we can continue (incorrectly) redirecting `Xerneas` to `Xerneas-Active` for replay compatibility.

Neither solution is ideal, but personally I prefer having the accurate implementation on the server side.

PR #2515, opened 2025-09-03
Preact: Update cookie expiry

ISenseAura:


play.pokemonshowdown.com/src/panel-mainmenu.tsx
```diff
@@ -439,7 +439,7 @@ class NewsPanel extends PSRoomPanel {
 	change = (ev: Event) => {
 		const target = ev.currentTarget as HTMLInputElement;
 		if (target.value === '1') {
-			document.cookie = "preactalpha=1; expires=Thu, 1 Sep 2025 12:00:00 UTC; path=/";
+			document.cookie = "preactalpha=1; expires=Thu, 1 Oct 2025 12:00:00 UTC; path=/";
```
mia-pi-git:
```suggestion
			document.cookie = "preactalpha=1; expires=Thu, 1 Oct 2035 12:00:00 UTC; path=/";
```
Having to do this once a month is annoying. Let's just bump it out far enough it hopefully won't be a problem.

PR #2518, opened 2025-09-06
[Preact] Add backup button to top of teambuilder panel

justin-xing:
Small change: Refactor backup button into subcomponent and put it at both top and bottom of teambuilder panel, as per request https://www.smogon.com/forums/threads/add-another-backup-restore-teams-button-section-to-the-top-of-the-page.3763108/


Zarel:
Hm. I was thinking of hiding it in a `...` menu somewhere. I'd be afraid of making it look too cluttered just right there.

PR #2520, opened 2025-09-09
Preact: Do not autojoin deleted rooms

ISenseAura:
Bug:
Auto joins deleted room, does not remove it from autojoin list by closing room.

Steps to reproduce:

1. Create groupchat/room 
2. Join it
3. Delete it
4. Close it
5. Refresh

FIx:
Initially mark all rooms disconnected (``room.connected = false``)
do not add disconnected rooms to autojoin list when updating the list

### DIFF
--- a/play.pokemonshowdown.com/src/client-main.ts
+++ b/play.pokemonshowdown.com/src/client-main.ts
@@ -1922,7 +1922,7 @@ export const PS = new class extends PSModel {
 			}
 			let rooms = autojoin[this.server.id] || '';
 			for (let title of rooms.split(",")) {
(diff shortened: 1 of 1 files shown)


Zarel:
I guess my worry with this implementation is that sync errors or bad connections could delete your autojoins.

I also don't want to delete my autojoins if I have hidden rooms in my autojoins, and I open PS without logging in. I think it shouldn't remove a room from autojoins unless I explicitly close it.

Zarel:
I think my preferred solution is just to remove deleted rooms from autojoin after they're explicitly closed.

PR #2521, opened 2025-09-09
Preact: Add Go to turn button in battle room

ISenseAura:
Forum approved:
https://www.smogon.com/forums/threads/go-to-turn-button-when-spectating-battles.3666813/

### DIFF
--- a/play.pokemonshowdown.com/src/panel-battle.tsx
+++ b/play.pokemonshowdown.com/src/panel-battle.tsx
@@ -485,6 +485,9 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
 			<p>
 				<button class="button" data-cmd="/switchsides">
 					<i class="fa fa-random" aria-hidden></i> Switch viewpoint
+				</button> {}
+				<button class="button" data-href="gototurnpopup">
+					<i class="fa fa-random" aria-hidden></i> Go to turn
 				</button>
 			</p>
 		</div>;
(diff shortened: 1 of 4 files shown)


DieterReinert (dismissed):
Nice work implementing a community-requested feature! The popup pattern implementation is solid and follows the existing codebase conventions well.

A few issues to address:

**Icon confusion**: Both "Switch viewpoint" and "Go to turn" buttons use `fa-random`, making them visually identical. Consider using `fa-step-forward` for the go-to-turn button.

**Input validation**: The turn input accepts any string and passes it directly to `/ffto` without validation. Should check if it's a valid turn number/format before sending.

**User feedback**: Silent failures when no battle room exists or invalid input is provided. Users won't know why nothing happened.

**Accessibility**: Input field could use `type="number"` and `placeholder` to guide users on expected format.

The core functionality looks good - just needs some UX polish before merge.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -467,8 +467,7 @@ export class ChatRoom extends PSRoom {
 			this.update(null);
 		},
 		'switchsides'() {
-			if (!this.battle) return this.add('|error|You are not in a battle');
```
Zarel, line 470: ?

I think having an error message on failure is a good thing.

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -1012,7 +1015,10 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
 				</p>
 			) : (
 				<p>
-					<button class="button" data-cmd="/switchsides"><i class="fa fa-random" aria-hidden></i> Switch viewpoint</button>
+					<button class="button" data-cmd="/switchsides"><i class="fa fa-random" aria-hidden></i> Switch viewpoint</button> {}
+					<button class="button" data-href="gototurnpopup">
```
Zarel: Hm. I think this would work better with a `PS.prompt` than with a custom popup. Your custom popup looks like it doesn't autofocus, either, and it's missing some styling.

Zarel: Plausibly this should go to `/ffto`, and then `/ffto` without a turn would open the prompt.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -467,8 +467,7 @@ export class ChatRoom extends PSRoom {
...
-			this.battle.switchViewpoint();
+			this.battle?.switchViewpoint();
```
```suggestion
			if (!this.battle) return this.add('|error|You are not in a battle');
			this.battle.switchViewpoint();
```

PR #2533, opened 2025-09-26
Bring back PS' true mascot Meloetta

ISenseAura:


ISenseAura:
> What is this slander against My Friend Malamar

How many megas dooes malamar have? One
How many does meloetta have? Zero

Hence Meloetta needs the push!
case closed 🔨 🕶️

PR #2536, opened 2025-10-01
Preact: Shift+arrows to move rooms

ISenseAura:


play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -913,10 +913,21 @@ export class ChatTextEntry extends preact.Component<{
...
-		// 	const newValue = `/pm ${PS.user.lastPM}, `;
-		// 	this.setValue(newValue, newValue.length);
-		// 	return true;
+			// } else if (e.keyCode === 32 && PS.user.lastPM && ['/reply', '/r', '/R'].includes(this.getValue())) { // '/reply ' is being written
+			// 	const newValue = `/pm ${PS.user.lastPM}, `;
+			// 	this.setValue(newValue, newValue.length);
+			// 	return true;
+		} else if (ev.shiftKey && ev.keyCode === 37) {
+			const toIndex = PS.rightRoomList.indexOf(PS.room.id) - 1;
+			PS.moveRoom(PS.room, 'right', false, toIndex);
```
Zarel: Would this work if the room wasn't a right room? Plausibly the buttons should be different in vertical tabs mode.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -913,10 +913,21 @@ export class ChatTextEntry extends preact.Component<{
...
+			PS.update();
+			return true;
+		} else if (ev.shiftKey && ev.keyCode === 39) {
+			let toIndex = PS.rightRoomList.indexOf(PS.room.id) + 1;
+			if (toIndex >= PS.rightRoomList.length - 1) toIndex = 0;
```
Zarel: It looks like this is incapable of moving rooms to the left room list, which the old client could do.

Ideally, these commands could be used to move rooms between the left list, right list, and mini-windows.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -913,10 +913,83 @@ export class ChatTextEntry extends preact.Component<{
...
 				return true;
 			}
-		// } else if (e.keyCode === 32 && PS.user.lastPM && ['/reply', '/r', '/R'].includes(this.getValue())) { // '/reply ' is being written
-		// 	const newValue = `/pm ${PS.user.lastPM}, `;
-		// 	this.setValue(newValue, newValue.length);
-		// 	return true;
+			// } else if (e.keyCode === 32 && PS.user.lastPM && ['/reply', '/r', '/R'].includes(this.getValue())) { // '/reply ' is being written
+			// 	const newValue = `/pm ${PS.user.lastPM}, `;
+			// 	this.setValue(newValue, newValue.length);
+			// 	return true;
```
```suggestion
		// } else if (e.keyCode === 32 && PS.user.lastPM && ['/reply', '/r', '/R'].includes(this.getValue())) { // '/reply ' is being written
		// 	const newValue = `/pm ${PS.user.lastPM}, `;
		// 	this.setValue(newValue, newValue.length);
		// 	return true;
```

Zarel:
Thanks!

Zarel:
This checks `PS.prefs.onepanel` for layout. But our actual layout depends on how wide the screen is. It should instead check `PS.leftPanelWidth`.

PR #2537, opened 2025-10-01
Preact: Bump cookie expiration

ISenseAura:
let me know if you wanna directly push the change yourself and if i should stop PRing this


ISenseAura:
Closing this coz it seems like the change has been made live already

PR #2538, opened 2025-10-17
Preact: Support `/challenge [user], [format]`

ISenseAura:
ok this was _surprisingly_ difficult to implement without making _major_ changes to existing structures and functions.

if this alternate approach of ``app.on('response:*')`` is approved, it may help fix the "failing validator not cancelling the search" bug also

### DIFF
--- a/play.pokemonshowdown.com/src/panel-chat.tsx
+++ b/play.pokemonshowdown.com/src/panel-chat.tsx
@@ -303,8 +303,8 @@ export class ChatRoom extends PSRoom {
 		'chall,challenge'(target) {
 			if (target) {
 				let [targetUser, format] = target.split(',');
-				if (this.pmTarget && format === undefined) {
-					format = targetUser;
+				if (this.pmTarget && (format === undefined || targetUser.includes('@@@'))) {
+					format = target;
 					targetUser = this.pmTarget;
 				}
 				format = (format || '').trim();


play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -303,7 +303,14 @@ export class ChatRoom extends PSRoom {
 		'chall,challenge'(target) {
 			if (target) {
 				const [targetUser, format] = target.split(',');
-				PS.join(`challenge-${toID(targetUser)}` as RoomID);
+				const callback = (data: any) => {
+					PS.mainmenu.listeners.userdetails = null;
+					if (data.rooms === false) return this.errorReply('This player does not exist or is not online.');
```
Zarel, line 313: Hm. I don't think this is necessary. I'd just send the challenge (or open the challenge window, if incomplete) and let the rest sort itself out.

ISenseAura: Ah, i tried to make it work like how old client does and its better UX i feel

Zarel: Does old client check if user exists first? Hm.

ISenseAura: Yes

ISenseAura: The error message is just a c/p from old client’s

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -2418,7 +2418,7 @@ export const PS = new class extends PSModel {
 		if (options.id.startsWith('challenge-')) {
 			this.requestNotifications();
 			options.id = `dm-${options.id.slice(10)}` as RoomID;
-			options.args = { challengeMenuOpen: true };
+			options.args = { challengeMenuOpen: true, ...options.args };
```
Zarel, line 2421: I do really like this code, though. Good idea.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -1184,7 +1188,7 @@ class ChatPanel extends PSRoomPanel<ChatRoom> {
 				<button data-cmd="/cancelchallenge" class="button">Cancel</button>
 			</TeamForm>
 		</div> : room.challengeMenuOpen ? <div class="challenge">
-			<TeamForm onSubmit={this.makeChallenge}>
+			<TeamForm onSubmit={this.makeChallenge} format={room.args?.format as string || undefined}>
```
Zarel: Wait, this will gray out the format selector, right? Ideally we should have a different way to do this.

play.pokemonshowdown.com/src/panel-mainmenu.tsx
```diff
@@ -775,6 +797,7 @@ export class TeamForm extends preact.Component<{
 	class?: string, format?: string, teamFormat?: string, hideFormat?: boolean, selectType?: SelectType,
 	onSubmit: ((e: Event, format: string, team?: Team) => void) | null,
 	onValidate?: ((e: Event, format: string, team?: Team) => void) | null,
+	disableFormatDropdown?: boolean,
```
Zarel: Oh, this is how you do it. I would have had `format` and `initialFormat` be different options, rather than changing every other use of `<TeamForm>` and `<FormatDropdown>`.

play.pokemonshowdown.com/src/panel-chat.tsx
ISenseAura: No, it doesnt, i created an optional argument to mark the format selector disabled

play.pokemonshowdown.com/src/panel-mainmenu.tsx
Zarel: In general, React attributes are not a great abstraction for the thing we want. This still probably causes problems if you try to `/challenge` while the challenge menu is already open, for instance. There are a few ways to deal with this but they're all annoyingly complex. But my point is, I don't like `disableFormatDropdown` as the way.

PR #2541, opened 2025-10-26
Update sugar's name to Luna

sugar700:
sugar (previously xfix) has ultimately decided on the name Luna, rather than previously listed name.

This is connected to https://github.com/smogon/pokemon-showdown-client/pull/2176, except, well, decided on a different name (Kamila was essentially me experimenting).


Zarel:
Thanks!

PR #2544, opened 2025-11-03
Preact: Update PS! Favicon when new notification is received

ISenseAura:


Zarel:
I ended up reimplementing the favicon thing. I'm willing to review the easter egg separately though.

PR #2547, opened 2025-11-04
Preact: Show room rank when user popup is opened from chat

ISenseAura:


play.pokemonshowdown.com/src/panel-popups.tsx
```diff
@@ -54,7 +54,10 @@ class UserPanel extends PSRoomPanel<UserRoom> {
 		}
 		const hideInteraction = room.id.startsWith('viewuser-');
 
-		const group = PS.server.getGroup(room.name);
+		const parentRoomId = this.props.room.parentRoomid;
+		const parentRoom = PS.rooms[parentRoomId!] as ChatRoom;
+		const name = parentRoom?.users?.[user.userid] || room.name;
+		const group = PS.server.getGroup(name);
```
Zarel: Wait. How did this ever work? `getGroup` requires a symbol, not a name...

Zarel: Oh, I see, it takes either symbol or name.

Zarel:
Either way, this approach is fewer lines. It won't follow through promotions but I think that's probably a good thing?

PR #2548, opened 2025-11-04
Preact: Disable play button properly in battles

ISenseAura:


play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -455,7 +455,7 @@ class BattlePanel extends PSRoomPanel<BattleRoom> {
 		return <div class="controls">
 			<p>
 				{atEnd ? (
-					<button class="button disabled" aria-disabled data-cmd="/play" style="min-width:4.5em">
+					<button class="button disabled" aria-disabled style="min-width:4.5em">
```
Zarel, line 458: Maybe this should be Replay? Isn't that what it is in battles currently?

ISenseAura: its "Pause" actually, in old client 
image

ISenseAura: Pause does make sense tho

Zarel: In which case it probably shouldn't look disabled?

ISenseAura: > In which case it probably shouldn't look disabled?

the play button? actually a disabled play button doesnt make sense there
to mimic old client behavior, if battle is paused then show play button if not then show pause button.
ill commit these changes shortly

PR #2552, opened 2025-11-09
Preact: Fix bug with /challenge

ISenseAura:


urkerab:
(I'd better hope nobody still plays any of the ~80 old formats on my side server which were written before the gen prefix was a thing...)

ISenseAura:
> (I'd better hope nobody still plays any of the ~80 old formats on my side server which were written before the gen prefix was a thing...)

you can always rename them to support the prefix on server side, no?

PR #2554, opened 2025-11-11
Preact: Treat popups as server response to challenges and search

swordfishtr:
This fixes an issue where challenging or searching with an invalid team causes the UI to incorrectly assume success and go out of sync until the user cancels out.

- Added `challengedSent` and `challengingSent` states much like `searchSent` in mainmenu.
- Rewrote protocol logic and linked documentation.
- Treated popups as server response to challenges and search.

### DIFF
--- a/play.pokemonshowdown.com/src/panel-chat.tsx
+++ b/play.pokemonshowdown.com/src/panel-chat.tsx
@@ -1210,20 +1210,21 @@ class ChatPanel extends PSRoomPanel<ChatRoom> {
 			room.args!.format = undefined;
 		}
 
-		const challengeTo = room.challenging ? <div class="challenge">
+		const challengeSent = room.teamSent && !room.challenged;
+		const challengeTo = room.challenging ? <div class="challenge outgoing">
 			<p>Waiting for {room.pmTarget}...</p>
 			<TeamForm format={room.challenging.formatName} teamFormat={room.challenging.teamFormat} onSubmit={null}>
 				<button data-cmd="/cancelchallenge" class="button">Cancel</button>
 			</TeamForm>
-		</div> : room.challengeMenuOpen ? <div class="challenge">
+		</div> : room.challengeMenuOpen ? <div class="challenge outgoing">
 			<TeamForm onSubmit={this.makeChallenge} defaultFormat={defaultFormat}>
-				{room.teamSent && <button class="button" disabled>
+				{challengeSent && <button class="button" disabled>
 					Challenging...
 				</button>}
-				{!room.teamSent && <button type="submit" class="button button-first">
+				{!challengeSent && <button type="submit" class="button button-first" disabled={!!room.challenged}>
 					<strong>Challenge</strong>
 				</button>}
-				{!room.teamSent && <button data-href="battleoptions" class="button button-last" aria-label="Battle options">
+				{!challengeSent && <button data-href="battleoptions" class="button button-last" aria-label="Battle options">
 					<i class="fa fa-caret-down" aria-hidden></i>
 				</button>} {}
 				<button data-cmd="/cancelchallenge" class="button">Cancel</button>
--- a/play.pokemonshowdown.com/style/client2.css
+++ b/play.pokemonshowdown.com/style/client2.css
@@ -896,11 +896,24 @@ p.or:after {
 
 .challenge {
 	margin: 0 5px 5px;
-	background: #fcd2b3;
 	border: 1px solid #f57b21;
-	border-radius: 5px;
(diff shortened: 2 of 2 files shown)


play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -315,17 +317,14 @@ export class ChatRoom extends PSRoom {
 				});
 				return;
 			}
-			if (this.challengeMenuOpen) {
-				this.cancelChallenge();
-				return;
-			}
```
Zarel, line 321: Still not sure what this change is about. The idea here is supposed to be that the challenge menu is already open, then `/chall` should close it instead.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -544,13 +543,13 @@ export class ChatRoom extends PSRoom {
 			this.add(`|error|Can only be used in a PM.`);
 			return;
 		}
-		if (this.challenging) {
+		if (this.teamSent || this.challenging) {
 			this.sendDirect('/cancelchallenge');
-			this.challenging = null;
-			this.challengeMenuOpen = true;
```
Zarel, line 550: I'm also confused why these were removed.

Zarel: I just tested this, and it makes it so that when you cancel a challenge, it no longer reopens the challenge menu. I think this is bad.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -575,26 +574,28 @@ export class ChatRoom extends PSRoom {
...
-			// we are sending the challenge
-			this.challenging = challenge;
+		// Protocol documentation: https://github.com/smogon/pokemon-showdown-client/pull/1799
+
+		this.teamSent = null;
+		if (!challenge) {
+			// rejected or canceled. maybe also accepted?
+			// (when we reject, we are sender; when we cancel, we are not)
+			this.challenged = null;
+			this.challenging = null;
```
Zarel: Ok. I think this change is bad. It makes it so that when you reject a challenge, you also cancel a challenge, and vice versa, but these are different things. I think the old code didn't have this problem.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -315,17 +317,14 @@ export class ChatRoom extends PSRoom {
 				});
 				return;
 			}
```
Zarel, line 319:
```suggestion
			}
			if (this.challengeMenuOpen) {
				this.cancelChallenge();
				return;
			}
```

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -574,27 +579,25 @@ export class ChatRoom extends PSRoom {
 		const challenge = this.parseChallenge(challengeString);
 		const userid = toID(name);
 		if (this.args?.format) this.args.format = null;
+		this.teamSent = null;
 
+		// Protocol documentation: https://github.com/smogon/pokemon-showdown-client/pull/1799
 		if (userid === PS.user.userid) {
-			if (!challenge && !this.challenging) {
-				// this is also used for canceling challenges
```
Zarel, line 580: Wait. Was I wrong? Ugh why would it work that way?

play.pokemonshowdown.com/src/panel-chat.tsx
swordfishtr: IIRC my conclusion there was that, those 2 actions are the same, since you can't be both challenging and challenged by the same person at the same time. The protocol tells you who was the challenger, but it never makes a difference.

play.pokemonshowdown.com/src/panel-chat.tsx
swordfishtr: There are subjective changes included from my playing around with the UI. I should've included that point in the original post, sorry. Feel free to revert anything you don't agree with.

play.pokemonshowdown.com/src/panel-chat.tsx
swordfishtr: The protocol logic was fine, just had the comments the wrong way around for rejecting vs canceling. I think my rewrite is more clear in that regard

play.pokemonshowdown.com/src/panel-chat.tsx
Zarel: Oh, okay, good to know, I'll update the comment.

Zarel:
image image

PR #2556, opened 2025-11-21
Preact: Fix idle indicator persisting in userlist after users become active

TurboRx:
users' idle indicators (`@!`) sometimes remained visible in the userlist even after they became active, particularly in rooms with custom rules or tournaments. the namecolor would update correctly but the idle marker would not clear.

### Root Cause

when `addUser()` updates an existing user's entry, it modifies `onlineUsers[index]` but doesn't re-sort the list. Since the sort key includes idle state (`!name.endsWith('@!')`), stale entries remain in the wrong position with incorrect idle indicators.

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -308,10 +308,7 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 		pokemon.sort();
 		const tierTable = {};
 		const overrideTier = {};
-		const ubersUUBans = {};
-		const ndDoublesBans = {};
-		const thirtyfivePokes = {};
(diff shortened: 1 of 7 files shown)


play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -669,11 +671,6 @@ export class ChatRoom extends PSRoom {
 	}
 
 	handleJoinLeave(action: 'join' | 'leave', name: string, silent: boolean) {
-		if (action === 'join') {
```
ISenseAura, line 690: good catch

PR #2558, opened 2025-11-23
Preact: Skip PM notifications for challenge acceptance messages

TurboRx:


play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -303,6 +308,8 @@ export class ChatRoom extends PSRoom {
 		'chall,challenge'(target) {
 			if (target) {
 				let [targetUser, format] = target.split(',');
+				// In PM rooms, autofill the username if only format is provided
+				// or if the target includes custom rules (@@@)
```
TurboRx, line 312: in here, added a comment to clarify the logic

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -139,10 +139,15 @@ export class ChatRoom extends PSRoom {
...
 				if (toID(fromUser) === PS.user.userid) break;
 				const message = args[args[0] === 'c:' ? 3 : 2];
-				this.notify({
-					title: `${this.title}`,
-					body: message,
-				});
+				const isChallengeAcceptance =
+					message.includes(' accepted the challenge') &&
+					message.includes('starting battle-');
+				if (!isChallengeAcceptance) {
```
ISenseAura, line 145: I’m not sure this is the right approach. We should first understand why this needs to be done this way.

TurboRx: hmm.. so the current approach filters messages client-side, but there may be a better way.

TurboRx: can you provide more details so I can try any other best way to add it?

TurboRx: like at which way it can be implemented?

TurboRx: i have some suggestions like server-side approach, special message flag, and current filtering approach??

ISenseAura: its either we do not notify for all /log messages or we do, im not sure how old client does it but hard coding to not notify for challenge messages should be our last option

TurboRx: After investigating both old and new client codes. `/nonotify` command sets `noNotify: true` this is the intended way for servers to send messages without notifications`/log` command does NOT set `noNotify` in either old or new client`/text` command: Sets `noNotify` in old client but not in new client`/raw` command: Sets `noNotify: true` in new client sooo. I have options...

1. **Make `/log` messages not notify**: Add `noNotify: true` to the `/log` case in `battle-log.ts` for consistent behavior
2. **Server-side fix**: Server should send challenge acceptance messages using `/nonotify` instead of regular chat

TurboRx: what do you think about it?

TurboRx: If we want `/log` messages to not notify, I can add `noNotify: true` to the `/log` case in `battle-log.ts`. Otherwise, the server should send challenge acceptance messages using `/nonotify` instead of regular chat.

Zarel:
Superseded by #2565. Thanks for the effort, though!

PR #2562, opened 2025-12-01
Replays: Fix controls when skipping to first turn

boughtabigk1:
We need to re-render the battle when skipping back to the first turn as the skip turn / last turn button won't render back to their enabled state until the first turn starts. This means if you skip to the end then decide to go back, you can't skip turns forward again until the first turn begins. That doesn't sound like it matters, but try watching [this replay](https://replay.pokemonshowdown.com/gen9doublesou-2491309465-p0hdpwk1mkaioitx24k54tr4h45iwhmpw) by skipping to the end and then back to the first turn. it can take a long time to get to the first turn from preview.


Zarel:
Oh, I see the problem. Good catch!

PR #2565, opened 2025-12-06
Preact: Fix /nonotify and raw message parsing in PMs

ISenseAura:


play.pokemonshowdown.com/src/battle-log.ts
```diff
@@ -1295,7 +1284,7 @@ export class BattleLog {
 		case 'data-move':
 			return ['chat message-error', '[outdated code no longer supported]'];
 		case 'text':
-			return ['chat', BattleLog.parseMessage(target)];
+			return ['chat', BattleLog.parseMessage(target), true];
```
Zarel, line 1287: Why was this changed?

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -139,15 +133,25 @@ export class ChatRoom extends PSRoom {
...
+						.split(' ')
+						.slice(1)
+						.join(' ')
+						.replace(/<[^>]*?>/g, '');
+				}
+				if (!nonotify) this.notify({
 					title: `${this.title}`,
-					body: message,
+					body: textContent,
 				});
```
Zarel: This should all go inside the `if (!nonotify)` block, right? Also I think it should be `noNotify`.

play.pokemonshowdown.com/src/battle-log.ts
ISenseAura: To match with old client implementation plus it was sending notification for empty “/text” messages (prob server side issue?)

Zarel: I suppose that works for me.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -139,16 +133,28 @@ export class ChatRoom extends PSRoom {
...
-				});
+				const noNotify = this.log?.parseChatMessage(message, name, args[1])?.[2];
+				if (!noNotify) {
+					let textContent = message;
+					if (/^\/(log|raw|html)\b/.test(message)) {
+						textContent = message
+							.split(' ')
+							.slice(1)
+							.join(' ')
+							.replace(/<[^>]*?>/g, '');
```
```suggestion
						textContent = message.split(' ').slice(1).join(' ')
							.replace(/<[^>]*?>/g, '');
```

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -139,16 +133,25 @@ export class ChatRoom extends PSRoom {
...
 				if (toID(fromUser) === PS.user.userid) break;
 				const message = args[args[0] === 'c:' ? 3 : 2];
-				this.notify({
-					title: `${this.title}`,
-					body: message,
-				});
+				const noNotify = this.log?.parseChatMessage(message, name, args[1])?.[2];
+				if (!noNotify) {
+					let textContent = message;
+					if (/^\/(log|raw|html)\b/.test(message)) {
```
```suggestion
					if (/^\/(log|raw|html) /.test(message)) {
```

Zarel:
(The code splits by space, so it's not like matching by arbitrary word boundary is useful)

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -139,16 +133,25 @@ export class ChatRoom extends PSRoom {
...
 				if (toID(fromUser) === PS.user.userid) break;
 				const message = args[args[0] === 'c:' ? 3 : 2];
-				this.notify({
-					title: `${this.title}`,
-					body: message,
-				});
+				const noNotify = this.log?.parseChatMessage(message, name, args[1])?.[2];
+				if (!noNotify) {
+					let textContent = message;
+					if (/^\/(log|raw|html) /.test(message)) {
```
Zarel: Should we also detect `uhtml` and `uhtmlchange` here?

ISenseAura: those are supported in pms? sure then

ISenseAura: `✅ |pm| my3amthoughts|+p9|/uhtml test, <b> test </b>` ok yeah

PR #2567, opened 2025-12-09
Preact: handle modal and html popup messages

ISenseAura:


play.pokemonshowdown.com/src/panel-mainmenu.tsx
```diff
@@ -177,7 +177,7 @@ export class MainMenuRoom extends PSRoom {
 					room.update(null);
 				}
 			}
-			PS.alert(message.replace(/\|\|/g, '\n'));
+			PS.alert(message.replace(/^(\|modal\||\|html\||\|wide\|)/, "").replace(/\|\|/g, '\n'));
```
Zarel, line 180: Should these not be handled inside `PS.alert`?

ISenseAura: we can if we want
but i think its better that we keep this outside ``PS.alert``
those prefixes are protocol stuff so handling them outside of the function keeps ``PS.alert`` simple, as it is just a helper function to show the given message in a popup.

Zarel: They aren't handled at all, though? This PR just ignores them rather than handling them.

ISenseAura: oh wait right, ugh im so bad
i forgot modal, html, wide actually mean something here
but i still think they should be handled outside alert, let me know and ill push changes accordingly

Zarel: You can handle it wherever you'd like, it just needs to be handled somewhere.

Zarel: Also, it looks like `|html|` is already supported.

PR #2568, opened 2025-12-09
Preact: properly extract targetRoomId from roomId in UserOptionsPanel

ISenseAura:
this makes it work for DM rooms

### DIFF
--- a/play.pokemonshowdown.com/src/panel-popups.tsx
+++ b/play.pokemonshowdown.com/src/panel-popups.tsx
@@ -1,5 +1,5 @@
 import preact from "../js/lib/preact";
-import { toID, toRoomid, toUserid, Dex } from "./battle-dex";
+import { toID, toRoomid, toUserid, Dex, PSUtils } from "./battle-dex";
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/src/panel-popups.tsx
```diff
@@ -230,7 +230,8 @@ class UserOptionsPanel extends PSRoomPanel {
 		data?: Record<string, string>,
 	};
 	getTargets() {
-		const [, targetUser, targetRoomid] = this.props.room.id.split('-');
+		const [, targetUser, ...rest] = this.props.room.id.split('-');
+		const targetRoomid = rest.join('-');
```
Zarel: Wait, isn't there `Utils.splitFirst` for this?

Image

Zarel:
```suggestion
		const [, targetUser, targetRoomid] = PSUtils.splitFirst(this.props.room.id, '-', 2);
```

PR #2570, opened 2025-12-10
'Copy of...' Compression: Duplicate Naming Conventions

snehalpaliwal:
**Link To Issue**: https://www.smogon.com/forums/threads/copy-of-compression.3743603/

**What**: Changing the way copy of names are created with three different settings (v. (n), (n), Copy of)

**Why Important**: Request from users to not just see a line of 'Copy of Copy of Copy of Copy of Copy of Copy of Untitled ___'
when making numerous versions. Mention that it would be "more useful to have the copy system become Untitled ____ (1)/" as an example. Gives users more autonomy and easily changes convention. 

**Proof of Testing**: 
[WITH REQUESTED CHANGES]
Simplified_UI_ALL
Simplified_UI

[OLD Version]
Updates_To_All_Teams
Dropdown
CopyOf_Dropdown
(v  n)_Dropdown
(n)_Dropdown


mia-pi-git:
I think allowing users to configure this overcomplicates the UI and will likely confuse users for minimal gain. How about we just go with `(1)`, `(2)`, etc?

mia-pi-git (changes requested):
see prior comment: change convention from configurable to just `name (1)`, `name (2)`, etc.

iforgetwhyimhere:
@Zarel currently Preact uses "team name + (copy)" to title its copies, is this something to be replaced or kept as-is?

PR #2571, opened 2025-12-11
Teambuilder: Show everything legal for custom game format

ISenseAura:
on preact, it is now possible to select Custom game as format in teambuilder but it still shows legalities of OU format.

this fixes it.

(also when no format is selected the dex search assumes its custom game so by default all mons are legal)

### DIFF
--- a/play.pokemonshowdown.com/src/battle-dex-search.ts
+++ b/play.pokemonshowdown.com/src/battle-dex-search.ts
@@ -742,9 +742,8 @@ abstract class BattleTypedSearch<T extends SearchType> {
 			}
 			this.baseIllegalResults = [];
 			this.illegalReasons = {};
-
 			for (const id in this.getTable()) {
-				if (!(id in legalityFilter)) {
+				if (!(id in legalityFilter) && this.format !== 'customgame') {
 					this.baseIllegalResults.push([this.searchType, id as ID]);
 					this.illegalReasons[id] = 'Illegal';
 				}
--- a/play.pokemonshowdown.com/src/battle-dex-search.ts
+++ b/play.pokemonshowdown.com/src/battle-dex-search.ts
@@ -742,6 +742,7 @@ abstract class BattleTypedSearch<T extends SearchType> {
 			}
 			this.baseIllegalResults = [];
 			this.illegalReasons = {};
+
 			for (const id in this.getTable()) {
 				if (!(id in legalityFilter) && this.format !== 'customgame') {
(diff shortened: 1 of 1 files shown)


iforgetwhyimhere:
> people don't accidentally use them.

If you're _picking_ custom game from the selector, I think 99% of people are intending to use wacky things like G-max formes, CAPs, Missingno, etc. with normally illegal moves and abilities. I think they should all display as being legal (since they _are_.) It feels to me like marking them as Illegal doesn't help the average user, since they may think that it's not allowed in Custom Game, after all its marked "Illegal" on "Custom Game".

> no format is selected the dex search assumes its custom game so by default 

I still think the teambuilder default format should be current gen OU, since that _is_ played a lot more than CG, and CG having its own validation could trip people up since CG allows everything, while other formats are pretty notorious for having rules.

Zarel:
> If you're picking custom game from the selector, I think 99% of people are intending to use wacky things like G-max formes, CAPs, Missingno, etc. with normally illegal moves and abilities.

I guess I don't know what people are actually using it for in practice. But I intended it to be used mostly for stuff like "I want to play OU but with one unban" without needing to learn the custom rule syntax.

> I still think the teambuilder default format should be current gen OU

Teambuilder default format is current gen, format unspecified, i.e. `gen9` not `gen9customgame`. 

Format unspecified is intended to be relatively usable for building teams in any format. I want it to be broadly similar to AG in terms of what it shows as legal/illegal.

iforgetwhyimhere:
> intended it to be used mostly for stuff like "I want to play OU but with one unban" without needing to learn the custom rule syntax.

I do think this is a fair idea, but there is a certain division of people that want a completely free teambuilder - the people who are building OU + one or two things are using OU's builder.

> Teambuilder default format is current gen, format unspecified

Just something I don't think about a lot. My main point is that with the CG builder being completely free, it shouldn't be the default. I don't have a problem with "current gen Dex" and I don't think most other users do either.

Zarel:
But I do think Custom Game is used for trying out new OM ideas. I think the average use case wants to use one category of illegal thing, not everything, and being able to see legality helps you decide to use one illegal thing while avoiding others.

Zarel:
So does my proposal in https://github.com/smogon/pokemon-showdown-client/pull/2571#issuecomment-4063902216 solve your problem?

iforgetwhyimhere:
@Zarel can we revisit this? It feels like we reached an agreement that the PR satisfies, but it was drafted instead of merged.

PR #2574, opened 2025-12-12
Doom Desire & Future Sight Timers UI

mollup:
# Future Sight and Doom Desire Timer UI
Added Future Sight and Doom Desire UI timers so that both users can easily see when damage will actually be dealt. Currently spawns in with 3 turns remaining when used, then ticks down to 2, then 1, then attacks.

[Link to issue](https://www.smogon.com/forums/threads/future-sight-indicator.3684086/)

### `battle.ts`
Added Future Sight and Doom Desire as sideConditions on client side. Not affected by Court Change because courtchange in moves file on server side manually selects the sideConditions it affects. Also added them as -sidestart cases.
### `battle.css`
Added an optional animation implementation for Future Sight and Doom Desire. The animation is currently unused, as it did not feel fully faithful to in-game behavior, but the underlying implementation is preserved for potential future use.

## Example:
image

### DIFF
--- a/play.pokemonshowdown.com/src/battle.ts
+++ b/play.pokemonshowdown.com/src/battle.ts
@@ -2745,8 +2745,8 @@ export class Battle {
 				this.scene.updateWeather();
 				break;
 			}
-			if (!(effect.id === 'typechange' && poke.terastallized) && 
-			    effect.id !== 'futuresight' && effect.id !== 'doomdesire') {
+			if (!(effect.id === 'typechange' && poke.terastallized) &&
+				effect.id !== 'futuresight' && effect.id !== 'doomdesire') {
 				poke.addVolatile(effect.id);
 			}
 			this.scene.updateStatbar(poke);


play.pokemonshowdown.com/style/battle.css
```diff
@@ -560,6 +560,18 @@ License: GPLv2
...
+	border: 1px solid #8844AA;
+	display: none;
+	position: absolute;
+}
+.sidecondition-doomdesire {
+	background: #CCCCCC;
+	border: 1px solid #999999;
+	display: none;
+	position: absolute;
+}
```
Zarel:
```suggestion
```

I appreciate the eagerness, but unless you actually want to implement them visually, I think this just clutters the CSS file.

play.pokemonshowdown.com/src/battle.ts
```diff
@@ -3048,18 +3067,19 @@ export class Battle {
...
-			let side = this.getSide(args[1]);
-			let effect = Dex.getEffect(args[2]);
-			side.addSideCondition(effect, !!kwArgs.persistent);
-
-			switch (effect.id) {
+	case '-sidestart': {
+		let side = this.getSide(args[1]);
+		let effect = Dex.getEffect(args[2]);
+		side.addSideCondition(effect, !!kwArgs.persistent);			
+		switch (effect.id) {
```
Zarel:
```suggestion
		case '-sidestart': {
			let side = this.getSide(args[1]);
			let effect = Dex.getEffect(args[2]);
			side.addSideCondition(effect, !!kwArgs.persistent);			

			switch (effect.id) {
```

This indentation change is wrong, and also makes the diff harder to read.

play.pokemonshowdown.com/src/battle.ts
```diff
@@ -3051,7 +3070,7 @@ export class Battle {
 		case '-sidestart': {
 			let side = this.getSide(args[1]);
 			let effect = Dex.getEffect(args[2]);
-			side.addSideCondition(effect, !!kwArgs.persistent);
+			side.addSideCondition(effect, !!kwArgs.persistent);
```
Zarel:
```suggestion
			side.addSideCondition(effect, !!kwArgs.persistent);
```

play.pokemonshowdown.com/src/battle.ts
```diff
@@ -2730,8 +2736,17 @@ export class Battle {
...
+				this.scene.updateWeather();
+				break;
+			case 'doomdesire':
+				poke.side.addSideCondition(effect, false);
+				this.scene.updateWeather();
+				break;
 			}
-			if (!(effect.id === 'typechange' && poke.terastallized)) {
+			if (!(effect.id === 'typechange' && poke.terastallized) && 
+			    effect.id !== 'futuresight' && effect.id !== 'doomdesire') {
```
Zarel:
```suggestion
			if (!(effect.id === 'typechange' && poke.terastallized) &&
				effect.id !== 'futuresight' && effect.id !== 'doomdesire') {
```

PR #2587, opened 2026-01-03
Preact: Low height screen battle handling

PokerPuppy2514:
When the client window is in a smaller screen, the battle screen will match the original handling with attack and switch buttons
image

Client rewrite issue: https://github.com/orgs/smogon/projects/2?pane=issue&itemId=111984798

### DIFF
--- a/play.pokemonshowdown.com/style/client2.css
+++ b/play.pokemonshowdown.com/style/client2.css
@@ -1886,6 +1886,91 @@ pre.textbox.textbox-empty[placeholder]:before {
 	background: #009AA4;
 }
 
+@media (max-height:570px) and (min-width: 440px) {
+	/*
+	 * Black move/switch menu for low-res screens (parity with client1)
+	 */
+	.controls {
+		position: absolute;
+		bottom: 10px;
+		left: 0;
+		right: 0;
+		width: auto;
+		background: #444444;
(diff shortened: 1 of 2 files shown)


play.pokemonshowdown.com/style/client2.css
```diff
@@ -1886,6 +1886,91 @@ pre.textbox.textbox-empty[placeholder]:before {
...
+		position: absolute;
+		right: 20px;
+		bottom: 20px;
+	}
+	.shiftselect {
+		position: absolute;
+		right: 150px;
+		bottom: 20px;
+	}
+	.moveselect button, .switchselect button, .shiftselect button {
```
ISenseAura: the `button` selector isn’t really needed here, `moveselect`, `switchselect` and `shiftselect` are applied directly to the button elements, so targeting the classes alone should be enough. 

(rn its effectively looking for a button inside a button which is prob why the styles arent being applied to the attack and switch buttons in your screenshot)

Zarel:
I appreciate the effort, but I've been wanting to redesign this for a while...

Zarel:
I ended up fixing this in 24b86547c01bcd4424f

Thanks again for the contribution! I appreciate the effort, even if I ended up not using it because I wanted to redesign it.

PR #2589, opened 2026-01-03
Fix join/leave messages appearing at wrong turn in battle log

surfnWOB:
In the Preact client, spectators and players joining a battle would appear grouped together at the current turn instead of at their correct historical positions in the battle log.

Changes:

1.  `BattleRoom.receiveLine` (new override in `panel-battle.tsx`):
  - Handles user list updates for join/leave/name
  - Routes messages to battle (if exists) or backlog (for later processing by BattlePanel)
  - Keeps ChatRoom clean with no battle-specific conditionals

2. `Battle.runMajor` (simplified in `battle.ts`):
  - Join/leave/name handlers now just log the message
  - User list management delegated to BattleRoom

Testing:
  - Spectate a live battle or watch a replay
  - Join messages should appear at their correct historical positions

### DIFF
--- a/play.pokemonshowdown.com/src/battle.ts
+++ b/play.pokemonshowdown.com/src/battle.ts
@@ -1150,6 +1150,7 @@ export class Battle {
 		autoresize?: boolean,
 	} = {}) {
 		this.id = options.id || '';
+		this.roomid = options.id || '';
 
 		if (options.$frame && options.$logFrame) {
 			this.scene = new BattleScene(this, options.$frame, options.$logFrame);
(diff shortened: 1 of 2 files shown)


Zarel:
idk how the Git log got that messed up, but I fixed it

play.pokemonshowdown.com/src/battle.ts
```diff
@@ -3617,55 +3540,19 @@ export class Battle {
...
-				if (!room.users[userid]) room.userCount.users++;
-				room.users[userid] = user;
-				room.userList.add(userid);
-				room.userList.updateUserCount();
-				room.userList.updateNoUsersOnline();
-			}
+		case 'join': case 'j': case 'J':
+		case 'leave': case 'l': case 'L':
+			// User list management is handled by ChatRoom.receiveLine
+			// Here we just log the message to the battle log
```
Zarel, line 3546: This change would presumably break how it works in the old client...

play.pokemonshowdown.com/src/panel-battle.tsx
```diff
@@ -147,10 +147,34 @@ export class BattleRoom extends ChatRoom {
...
+
+	override receiveLine(args: Args) {
+		switch (args[0]) {
+		case 'users':
+			const usernames = args[1].split(',');
+			const count = parseInt(usernames.shift()!, 10);
+			this.setUsers(count, usernames);
+			return;
+		case 'join': case 'j': case 'J':
+			this.addUser(args[1]);
```
Zarel, line 159: You've added this but not removed the old handling. What's going on?

Zarel:
In theory, the thing you want to do is set `room.joinLeave = null` whenever something happens that makes you not want to add join/leave messages in a specific location.

Zarel:
Fixed in 8336ed5673405baba7f8ff2ab8

PR #2590, opened 2026-01-05
[Client] Implement Mega Stones as pairs {key: value}

andrebastosdias:
Server PR: https://github.com/smogon/pokemon-showdown/pull/11684
I don't think the client was even updated when `megaStone` and `megaEvolves` became `string | string[]`.

### DIFF
--- a/play.pokemonshowdown.com/js/client-teambuilder.js
+++ b/play.pokemonshowdown.com/js/client-teambuilder.js
@@ -1521,7 +1521,7 @@
 					}
 					if (this.curTeam.gen >= 6) {
 						var item = dex.items.get(set.item);
-						if (item.megaStone && item.megaEvolves[species.baseSpecies]) {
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/js/client-teambuilder.js
```diff
@@ -1521,8 +1521,8 @@
 					}
 					if (this.curTeam.gen >= 6) {
 						var item = dex.items.get(set.item);
-						if (item.megaStone && species.baseSpecies === item.megaEvolves) {
-							species = dex.species.get(item.megaStone);
+						if (item.megaStone && item.megaEvolves[species.baseSpecies]) {
```
KrisXV:
```suggestion
						if (item.megaEvolves && item.megaEvolves[species.baseSpecies]) {
```

PR #2594, opened 2026-01-09
Add a check to get rid of unneeded Ubers header

duckycrater:
Currently the Ubers header displays even if no Ubers mons are legal, leading to a useless header(pictured below in biomechmons builder). This gets rid of that.
image

### DIFF
--- a/play.pokemonshowdown.com/src/battle-dex-search.ts
+++ b/play.pokemonshowdown.com/src/battle-dex-search.ts
@@ -1176,14 +1176,63 @@ class BattlePokemonSearch extends BattleTypedSearch<'pokemon'> {
 						'calyrexshadow' in table.metagameBans[format] &&
 						type === 'header' && id === 'AG'
 					) return false;
+					const gen9ubers = [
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/src/battle-dex-search.ts
```diff
@@ -1171,23 +1171,18 @@ class BattlePokemonSearch extends BattleTypedSearch<'pokemon'> {
...
 					if ((dex.gen === 7 || dex.gen === 6) &&
 						(id === 'AG' || id === 'rayquazamega') &&
 						'megarayquazaclause' in table.metagameBans[format]
 					) return false;
 					return true;
 				});
 			}
+			tierSet = tierSet.filter(([type], i) => {
+				if (type === 'header' && i < tierSet.length - 1 && tierSet[i + 1][0] === 'header')
+					return false;
```
KrisXV:
```suggestion
				if (type === 'header' && i < tierSet.length - 1 && tierSet[i + 1][0] === 'header') return false;
```

PR #2598, opened 2026-01-19
Fix teambuilder support for OMs

Zrp200:
Under the current implementation, Natdex formats are actually using the metagame bans for SV formats, if they exist.

The current implementation actively enables this by **specifically checking both formats**.

However, the metagames are actually 100% independent, resulting in extra pokemon being listed as banned.

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -447,11 +447,11 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 					let banned = ruleTable.isBannedSpecies(species);
 					if (!banned) {
 						const requiredItems = species.requiredItems;
-						if (requiredItems?.length) {
+						if (requiredItems && requiredItems.length) {
 							for (const itemName of requiredItems) {
 								const item = Dex.items.get(itemName);
-								if ((item.itemUser && item.itemUser.includes(species.name) ||
-									item.megaStone && Object.values(item.megaStone).includes(species.name)
+								if (((item.itemUser && item.itemUser.includes(species.name)) ||
+									(item.megaStone && Object.values(item.megaStone).includes(species.name))
 								) && ruleTable.isBanned('item:' + item.id)) {
 									banned = true;
 									break;


play.pokemonshowdown.com/src/battle-dex-search.ts
```diff
@@ -1164,13 +1164,12 @@ class BattlePokemonSearch extends BattleTypedSearch<'pokemon'> {
 			'nationaldexaaa', 'nationaldexbh', 'nationaldexgodlygift', 'nationaldexstabmons', 'tiershift',
 		];
 		if (dex.gen >= 6) {
-			if (
-				(customBanlists.includes(format) && table.metagameBans?.[format]) ||
-				(this.formatType === 'natdex' && customBanlists.includes('nationaldex' + format) &&
-					table.metagameBans?.['nationaldex' + format])) {
+			// fixme this should be unneeded, as there is no risk of natdex tiers overwriting regular tiers in build-indexes
+			// eslint-disable-next-line @typescript-eslint/no-shadow
+			const format = this.formatType === 'natdex' ? 'nationaldex' + this.format : this.format;
```
KrisXV: its giving an error because format is declared in the upper scope, you need to use a different variable name

Zrp200: fixed, and I knew that, hence me silencing it instead of fixing it. I *wanted* to shadow it in order to avoid changing the code underneath it

play.pokemonshowdown.com/src/battle-dex-search.ts
```diff
@@ -1164,13 +1164,9 @@ class BattlePokemonSearch extends BattleTypedSearch<'pokemon'> {
 			'nationaldexaaa', 'nationaldexbh', 'nationaldexgodlygift', 'nationaldexstabmons', 'tiershift',
 		];
 		if (dex.gen >= 6) {
-			if (
-				(customBanlists.includes(format) && table.metagameBans?.[format]) ||
-				(this.formatType === 'natdex' && customBanlists.includes('nationaldex' + format) &&
-					table.metagameBans?.['nationaldex' + format])) {
+			if (customBanlists.includes(format) && table.metagameBans?.[format]) {
```
KrisXV, line 1168: this wont pick up on national dex formats because `format` is sanitized by removing `nationaldex` from the formatid string and changing the formatType to `'natdex'`. you'd need to adjust the array above

Zrp200: - 

> this wont pick up on national dex formats because `format` is sanitized by removing `nationaldex` from the formatid string and changing the formatType to `'natdex'`. you'd need to adjust the array above

shouldn't the arrays match the ones in build-indices

Zrp200: > this wont pick up on national dex formats because `format` is sanitized by removing `nationaldex` from the formatid string and changing the formatType to `'natdex'`. you'd need to adjust the array above

cp'd from discord for completeness:
It looks to me that battle-dex-search automatically trims out the natdex, so the format being checked already lacks the 'nationaldex'
which is a contributor to the bugs that are currently live
In your implementation, you manually added 'nationaldex' to format, but you also checked the base format. And because both 'ubersuu' and 'nationaldexubersuu' were in metagameBans, both were checked
gen9natdex shouldn't need to know the metagame bans of sv formats, so it can use 'ubersuu' and such without risk of conflict 
this removes the need to manually add 'nationaldex' in battle-dex-search.ts

build-tools/build-indexes
```diff
@@ -433,36 +433,41 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
-						}
-						if (species.requiredItems && species.requiredItems.length) {
-							for (const itemName of species.requiredItems) {
+					const format = Dex.formats.get(`${gen}${isNatDex ? 'nationaldex' : ''}${meta}`);
+					if (!format.exists) continue;
+					const ruleTable = Dex.formats.getRuleTable(format);
+					let banned = ruleTable.isBannedSpecies(species);
+					if (!banned) {
+						const requiredItems = species.requiredItems;
+						if (requiredItems?.length) {
```
KrisXV:
```suggestion
						if (requiredItems && requiredItems.length) {
```

build-tools/build-indexes
```diff
@@ -433,36 +433,41 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
+							for (const itemName of requiredItems) {
 								const item = Dex.items.get(itemName);
-								if (item.itemUser && item.itemUser.includes(species.name) &&
-									Dex.formats.getRuleTable(format).isBanned('item:' + item.id)) {
-									metagameBans[meta][species.id] = 1;
-								}
-								if (item.megaStone && Object.values(item.megaStone).includes(species.name) &&
-									Dex.formats.getRuleTable(format).isBanned('item:' + item.id)) {
-									metagameBans[meta][species.id] = 1;
+								if ((item.itemUser && item.itemUser.includes(species.name) ||
```
KrisXV:
```suggestion
								if (((item.itemUser && item.itemUser.includes(species.name)) ||
```

build-tools/build-indexes
```diff
@@ -433,36 +433,41 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
...
+									item.megaStone && Object.values(item.megaStone).includes(species.name)
```
KrisXV:
```suggestion
									(item.megaStone && Object.values(item.megaStone).includes(species.name))
```

PR #2601, opened 2026-01-29
Preact: Keyboard shortcut to jump to next unread room

EthansTrader:


play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -2401,6 +2401,26 @@ export const PS = new class extends PSModel {
 		}
 		return this.focusRoom(rooms[index + 1]);
 	}
+	focusLeftUnreadRoom() {
```
ISenseAura: i think these two functions can be merged into a single function that takes a direction (left/right) as argument since the logic is almost identical and only differs by navigation direction.

like this maybe?: 

```
focusUnreadRoom(direction: 'left' | 'right') {
	const { rooms, index } = this.horizontalNav();
	if (index === -1) return;

	const unreadRooms = rooms.filter((room, i) =>
		PS.rooms[room]?.isSubtleNotifying &&
		(direction === 'left' ? i < index : i > index)
	);

	if (!unreadRooms.length) return;

	const target =
		direction === 'left'
			? unreadRooms[unreadRooms.length - 1]
			: unreadRooms[0];

	return this.focusRoom(target);
}
```

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -2401,6 +2401,26 @@ export const PS = new class extends PSModel {
...
+		);
+
+		if (!unreadRooms.length) return;
+
+		const target =
+			direction === 'left'
+				? unreadRooms[unreadRooms.length - 1]
+				: unreadRooms[0];
+			console.log("Target = " + target);
+			console.log("Unread Rooms = " + unreadRooms);
```
Zarel:
```suggestion
```

I'm sure these were useful while debugging but they're just clutter now.

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -2401,6 +2401,26 @@ export const PS = new class extends PSModel {
...
+			PS.rooms[room]?.isSubtleNotifying &&
+			(direction === 'left' ? i < index : i > index)
+		);
+
+		if (!unreadRooms.length) return;
+
+		const target =
+			direction === 'left'
+				? unreadRooms[unreadRooms.length - 1]
+				: unreadRooms[0];
```
Zarel:
```suggestion
		const target = direction === 'left' ? unreadRooms[unreadRooms.length - 1] : unreadRooms[0];
```

play.pokemonshowdown.com/src/client-main.ts
```diff
@@ -2401,6 +2401,21 @@ export const PS = new class extends PSModel {
 		}
 		return this.focusRoom(rooms[index + 1]);
 	}
+	focusUnreadRoom(direction: 'left' | 'right') {
+		const { rooms, index } = this.horizontalNav();
+		if (index === -1) return;
+
+		const unreadRooms = rooms.filter((room, i) =>
+			PS.rooms[room]?.isSubtleNotifying &&
+			(direction === 'left' ? i < index : i > index)
```
Zarel, line 2410: Looks like this wouldn't wrap.

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -929,6 +929,7 @@ export class ChatTextEntry extends preact.Component<{
 	}
 	handleKey(ev: KeyboardEvent) {
 		const cmdKey = ((ev.metaKey ? 1 : 0) + (ev.ctrlKey ? 1 : 0) === 1) && !ev.altKey && !ev.shiftKey;
+		const altKey = ev.altKey
```
```suggestion
		const altKey = ev.altKey;
```

PR #2604, opened 2026-02-02
Replace (fainted) text with 0/X hp

EthansTrader:
Based on https://www.smogon.com/forums/threads/show-0-x-hp-instead-of-fainted-in-battle.3765617/

This only works for the Pokeicons below the trainer. Kind of a band-aid implementation, but not sure how interested we are in refactoring the HP system to avoid setting maxhp to 100.

### DIFF
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -873,7 +873,7 @@ export class BattleTooltips {
 		if (pokemon.fainted && pokemon.maxhp === 100) {
 			text += `<p><small>HP:</small> (fainted)</p>`;
 		} else if (pokemon.fainted) {
-			text += `<p><small>HP:</small> <s class="gray">0/${pokemon.maxhp}</s></p>`;
+			text += `<p><small>HP:</small> <span class="gray">0/${pokemon.maxhp} (fainted)</span></p>`;
 		} else if (this.battle.hardcoreMode) {
 			if (serverPokemon) {
 				const status = pokemon.status ? ` <span class="status ${pokemon.status}">${pokemon.status.toUpperCase()}</span>` : '';


play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -870,8 +870,10 @@ export class BattleTooltips {
 			text += `<p class="tooltip-section"><strong>Possible Illusion #${illusionIndex}</strong>${levelBuf}</p>`;
 		}
 
-		if (pokemon.fainted) {
-			text += '<p><small>HP:</small> (fainted)</p>';
+		if (pokemon.fainted && pokemon.maxhp === 100) {
+			text += `<p><small>HP:</small> (fainted)</p>`;
+		} else if (pokemon.fainted) {
+			text += `<p><small>HP:</small> 0/${pokemon.maxhp}</p>`;
```
Zarel: I think there should be _some_ easily visible indication that it's fainted.

Thoughts on this?

```suggestion
			text += `<p><small>HP:</small> <s class="gray">0/${pokemon.maxhp}</s></p>`;
```

play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -870,8 +870,10 @@ export class BattleTooltips {
 			text += `<p class="tooltip-section"><strong>Possible Illusion #${illusionIndex}</strong>${levelBuf}</p>`;
 		}
 
-		if (pokemon.fainted) {
-			text += '<p><small>HP:</small> (fainted)</p>';
+		if (pokemon.fainted && pokemon.maxhp === 100) {
+			text += `<p><small>HP:</small> (fainted)</p>`;
+		} else if (pokemon.fainted) {
+			text += `<p><small>HP:</small> <s class="gray">0/${pokemon.maxhp}</s></p>`;
```
Zarel: Hm. Strikethroughs are not super readable. Maybe something more like


```suggestion
			text += `<p><small>HP:</small> <span class="gray">0/${pokemon.maxhp} (fainted)</span></p>`;
```

Zarel:
Thanks!

PR #2610, opened 2026-03-01
Preact: Show additional rule options in challenge forms

ISenseAura:
fixes a bug where:
Additional rule options (i.e. Tera Type Preview in NatDex Draft) don't appear when initiating a challenge

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -430,17 +430,17 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 			if (!tierTable[speciesTier]) tierTable[speciesTier] = [];
 			tierTable[speciesTier].push(id);
 
-			if (genNum >= 6) {
-				for (const meta of [
+			if (genNum >= 5) {
+				for (const meta of (isVGC && genNum === 9 ? ['4v4doublesuu'] : [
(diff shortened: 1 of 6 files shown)


ISenseAura:
> Looks like that build error was just because there was a build error in master two weeks ago. In theory I think that could also have been fixed if you clicked the "re-run checks" button, although I'm not sure you have access to it.

yup i knew the build issue was from another commit, dhelmise fixed it right away but I couldnt find the rerun button so i left it

play.pokemonshowdown.com/src/panel-mainmenu.tsx
```diff
@@ -827,10 +838,25 @@ export class TeamForm extends preact.Component<{
...
 		PS.teams.loadTeam(team).then(() => {
 			(validate === 'validate' ? this.props.onValidate : this.props.onSubmit)?.(ev, format, team);
 		});
 	};
+	toggleCustomRule = (ev: Event) => {
+		const checked = (ev.target as HTMLInputElement)?.checked;
+		const rule = (ev.target as HTMLInputElement)?.name;
+		if (rule === 'terapreview') this.teraPreview = checked;
+		if (rule === 'bestof') this.bestOf = checked;
+	};
```
Zarel, line 859: Hm. Should this be state? If we just read it from the DOM every time, there would be no way to desync.

Zarel: You know what they say. There's state, and there's cache. And there's no need for cache if it's not actually there to optimize anything.

PR #2611, opened 2026-03-02
Fix stale Leech Seed visual after Shed Tail switch

frxyoz:
Fixes a client-side visual desync where Leech Seed particles could remain on the incoming Pokémon after a seeded user switched out with Shed Tail.
Root cause: volatile state in copyVolatileForm reused shared object references, then removed entries without fully clearing source-side sprite effects.
Fix: clone boosts/volatiles instead of aliasing, and clear source volatile visuals after transfer so old effects (including Leech Seed) are removed reliably.
Changed file: play.pokemonshowdown.com/src/battle.ts
Validation: local build succeeds (npm run build), TS compile passes.

Closes [https://github.com/smogon/pokemon-showdown-client/issues/2447](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)

### DIFF
--- a/play.pokemonshowdown.com/src/battle.ts
+++ b/play.pokemonshowdown.com/src/battle.ts
@@ -441,8 +441,8 @@ export class Pokemon implements PokemonDetails, PokemonHealth {
 	 * copyAll = 'shedtail' means Shed Tail
 	 */
 	copyVolatileFrom(pokemon: Pokemon, copySource?: | 'shedtail' | boolean) {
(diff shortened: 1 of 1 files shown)


Zarel:
Wait, why are we copying volatiles with Shed Tail in the first place?

frxyoz:
Could it be that the code might've been forked from the baton pass code?

PR #2621, opened 2026-03-22
Highlight Help text removal

iforgetwhyimhere:
When using !help highlight, the client overrides the command and doesn't properly send a message. Since help for the highlight commands are on the server, these cases can safely be removed.

Depends on server #11843


Zarel:
I guess this is so you can `!help highlight`...

Zarel:
This is not how we've done it in the past. Client commands being documented in the client have a variety of advantages (offline support, ease of updating when we update the command).

But I can believe those are minor compared to supporting `!help`. Ugh...

Zarel:
What if we just always forwarded `!` to the server? Rather than removing the client help?

iforgetwhyimhere:
> variety of advantages (offline support,

Is there even any way to trigger the helptext while offline? Disconnecting removes all your text input boxes where you could enter a command afaik

> What if we just always forwarded ! to the server?

This could work. Also should probably be extended to all client-side commands.

PR #2622, opened 2026-03-25
Update tooltips of Galvanize/Dragonize/Mega Sol

ry4242:
Updated Galvanize/Dragonize to be more accurate, and added Mega Sol tooltip to Weather Ball damage boost


play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -2021,6 +2021,7 @@ export class BattleTooltips {
 			}
 		}
 		if (move.id === 'weatherball') {
+			value.abilityModify(1.5, "Mega Sol");
```
KrisXV: this is wrong

PR #2625, opened 2026-03-29
Hide formats with no games or Elo=1000 in untargeted /rank

Slayer95:
- QoL improvement for the lowest ladder players.
- Declutters and hides away formats that were only tried long ago.

### DIFF
--- a/play.pokemonshowdown.com/js/client-chat.js
+++ b/play.pokemonshowdown.com/js/client-chat.js
@@ -1004,7 +1004,7 @@
 						var row = data[i];
 						if (!row) return self.add('|raw|Error: corrupted ranking data');
 						var formatId = toID(row.formatid);
-						if (!formatTargeting || formats[formatId] || gens[formatId.slice(0, 4)] || (gens['gen6'] && formatId.substr(0, 3) !== 'gen')) {
+						if (!formatTargeting && row.elo >= 1000.5 || formats[formatId] || gens[formatId.slice(0, 4)] || (gens['gen6'] && formatId.substr(0, 3) !== 'gen')) {
 							buffer += '<tr>';
 						} else {
 							buffer += '<tr class="hidden">';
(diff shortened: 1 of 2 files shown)


Zarel:
I don't think this is a particularly meaningful place to draw the line. This is for formats you tried once and lost? I guess certain kinds of users will have a lot of those, but I don't imagine it'd be that common. And you might as well retain a memory of them.

Slayer95:
> This is for formats you tried once and lost?

Yeah, that would be the second bullet point.

> I guess certain kinds of users will have a lot of those, but I don't imagine it'd be that common. 

Well, that's 11/22 formats for me, and 3/3 for you. Furthermore, if a user tried an OM or old-gen format once, and lost, and the ladder is gone, that record is gonna stick out forever like a sore thumb per status quo, so veteran users are bound to have some of them.

> And you might as well retain a memory of them.

I mean, I don't intend to fully remove them. I just move them behind the ``"format1, format2, format3, and X other formats not shown"`` message. While keeping memory is nice, there isn't much of a point in keeping them in front, and making them *that* memorable, specially if they turn out to be unplayable.

The first point is about people stuck in the bottom of the ladder, who are likely to be bullied to some degree after their ``/rank`` is quickly checked by other chat users (basically mandatory check when a lesser known user claims stuff about a tier). While this doesn't fully hide their dark records (and is ineffective if they happened to win their last battle), it does hide their rating behind a "See more" interaction. Unfortunately, this sort of bullying isn't likely to prompt reports nor, therefore, moderation action; so I'd hope some gating/obscurity would help, even if only a bit.

> I don't think this is a particularly meaningful place to draw the line.

I have just noticed that https://pokemonshowdown.com/users/userid hides ratings for formats with 0 games played (even if they have Elo>1000). Would you find that to be more meaningful? Actually, I now recall that I have seen people complain that formats remain in their ``/rank`` even after reset. So, using 0 games as the line would let users hide formats through W/L resets.

Still, using 0 games as exclusive criterion would forfeit the anti-bullying goal. Why not use both criteria together?

Zarel:
I guess. If there's an expand button, sure.

Zarel:
Looks like this is an oldclient-specific change. How does Preact client handle this?

Slayer95:
> Looks like this is an oldclient-specific change. How does Preact client handle this?

Oh, it's pretty much the same. Updated.

Also:

> updates the cutoff from 1000.5 to 1001, to account for rounding changes

andrebastosdias:
What if we could simply reset our ELO by removing all entries? We can reset the W/L record, but we can't completely remove the entries from our account.

Edit: I just read the "Elo=1000" part. Then this is good!

Slayer95:
> What if we could simply reset our ELO by removing all entries? We can reset the W/L record, but we can't completely remove the entries from our account.

For the health of ladders, I don't support resetting rating values other than uncertainty (when appliable, and even then, with restrictions). Also, it seems to be out of scope for this PR Q_Q

It would be a different story if the ladder MMR were Glicko rather than Elo, but that proposal didn't fly on PR.

Zarel:
Resetting W/L should in theory hide the rating from your account, at least until you play another game.

Slayer95:
Right. So, this is good to go...?

Zarel:
Yeah, it's just waiting on me doing some local testing. Which I'll do after I work on some other Preact bugs.

Zarel:
Thanks!

PR #2627, opened 2026-03-29
Preact: Add room privacy control buttons

iforgetwhyimhere:
https://www.smogon.com/forums/threads/publicroom-button.3710317/

image


ISenseAura:
i think we can simply just add another checkbox "Hide current battle" or "Hide this battle" instead of those buttons. it would look cleaner

PR #2636, opened 2026-04-11
Always round down when displaying elo

ISenseAura:
https://www.smogon.com/forums/threads/always-round-down-when-displaying-elo.3780515/


Zarel:
There are at least three other places Elo is displayed. `pokemonshowdown.com/ladder.php`, `pokemonshowdown.com/users.php`, and `play.pokemonshowdown.com/ladder.php`.

There might be others, too, I'm not sure. Take a more thorough look.

Zarel:
I took a more thorough look and didn't find any others.

PR #2641, opened 2026-04-13
Preact: Fix "friend came online" console messages now leave dm box as unread

TurboRx:
when a friend comes online the server sends a `|pm|~|USER|/nonotify Your friend ... has just connected!` PM. the `/nonotify` is intentional, it suppresses the full desktop notification popup. however, in panel-chat.tsx the `dm-` room's handler only skipped `notify()` for `nonotify()` messages but never called `subtleNotify()` either, so the room tab showed no indication of new content.

### DIFF
--- a/play.pokemonshowdown.com/src/panel-chat.tsx
+++ b/play.pokemonshowdown.com/src/panel-chat.tsx
@@ -158,6 +158,9 @@ export class ChatRoom extends PSRoom {
 						title: `${this.title}`,
 						body: textContent,
 					});
+				} else if (!isIgnored) {
+					// /nonotify messages (e.g. "friend came online") should still mark the room as unread
+					this.subtleNotify();
 				}
 			} else {
 				this.subtleNotify();


TurboRx:
bug fix: `"Friend came online" Console messages do not audibly notify users (or even leave the box as unread)` https://github.com/orgs/smogon/projects/2?pane=issue&itemId=111176277

Zarel:
We may be misusing `nonotify` here...

Zarel:
Looking at the other uses of `/nonotify`, I think probably it should not notify at all, and we should add `/subtlenotify` for this use-case.

Zarel:
Closing. I already implemented `/subtlenotify`; the rest should be done server-side if we choose to do it.

~~In the future, I'd also prefer if you didn't phrase feature requests like bug reports. Feature requests should go through https://psim.us/suggestions.~~ Futher conversation suggests ifwih thought it was a bug; that's not your fault.

PR #2642, opened 2026-04-13
Preact: Skip choices for tatsugiri when in commanding mode

ISenseAura:


Zarel:
Thanks! I really wish this protocol message had been named something more generic than "commanding", but too late now I guess...

PR #2644, opened 2026-04-13
Champions: Fix Mega Sol tooltips

andrebastosdias:
https://www.smogon.com/forums/threads/display-fire-type-weather-ball-with-mega-sol.3780775/

### DIFF
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -2276,7 +2276,7 @@ export class BattleTooltips {
 			}
 		}
 		if (move.id === 'weatherball') {
-			if (value.abilityModify(2, "Mega Sol") || this.battle.weather !== 'deltastream') {
+			if (!value.abilityModify(2, "Mega Sol") && this.battle.weather !== 'deltastream') {
 				value.weatherModify(2);
 			}
 		}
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -2276,8 +2276,7 @@ export class BattleTooltips {
 			}
 		}
 		if (move.id === 'weatherball') {
-			value.abilityModify(2, "Mega Sol");
-			if (this.battle.weather !== 'deltastream') {
+			if (value.abilityModify(2, "Mega Sol") || this.battle.weather !== 'deltastream') {
 				value.weatherModify(2);
```
KrisXV: uh what?

KrisXV: wouldn't this make weather ball display as having 200 base power?

andrebastosdias: You shouldn't duplicate the move power twice when sun is on the field.

KrisXV: I know, but how does this fix that? these both call `this.modify` so it'd be directly mutating weather ball's base power twice regardless of this change as far as i can tell

andrebastosdias: Fixed, I though the first part was a check

PR #2646, opened 2026-04-13
Fix learnsets for formes with empty objects

andrebastosdias:
The correct way to do this would be to run the methods from the cached server in order to obtain the complete learnset.

### DIFF
--- a/build-tools/build-indexes
+++ b/build-tools/build-indexes
@@ -990,6 +990,7 @@ process.stdout.write("Building `data/teambuilder-tables.js`... ");
 	const G2Learnsets = Dex.mod('gen2').data.Learnsets;
 	for (const id in G2Learnsets) {
 		const learnset = G2Learnsets[id].learnset;
+		if (!learnset) continue;
 		for (const moveid in learnset) {
 			const gens = learnset[moveid].map(x => Number(x[0]));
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/src/battle-dex-search.ts
```diff
@@ -815,6 +815,14 @@ abstract class BattleTypedSearch<T extends SearchType> {
 		}
 		return results;
 	}
+	protected getLearnsetEntry(table: any, speciesid: ID) {
+		let learnset = table.learnsets?.[speciesid];
+		if (!learnset || typeof learnset !== 'object') return null;
+		if ('learnset' in learnset) learnset = learnset.learnset;
```
KrisXV: This isn’t how learnsets look client side

andrebastosdias: How is it? It fixes the problem

KrisXV: Yeah this if statement is just superfluous because it’s never true, client-side learnset objects are just

```
speciesid: {
    moveid: '123456789'
}
```

andrebastosdias: You can re-review it

PR #2648, opened 2026-04-14
Preact: subscribe ChatUserList to room to prevent blank user list on join/refresh

TurboRx:
bug fix: `User list is blank when you join a room or refresh, happens rarely` https://github.com/orgs/smogon/projects/2?pane=issue&itemId=111176277

### DIFF
--- a/play.pokemonshowdown.com/src/panel-chat.tsx
+++ b/play.pokemonshowdown.com/src/panel-chat.tsx
@@ -1428,12 +1428,18 @@ export class ChatUserList extends preact.Component<{
 	room: ChatRoom, left?: number, top?: number, minimized?: boolean, static?: boolean,
 }> {
 	subscription: PSSubscription | null = null;
+	subscribeTimeout: number | null = null;
 	override componentDidMount() {
-		this.subscription = this.props.room.subscribe(args => {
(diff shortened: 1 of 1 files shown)


play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -1427,6 +1427,21 @@ class ChatPanel extends PSRoomPanel<ChatRoom> {
 export class ChatUserList extends preact.Component<{
 	room: ChatRoom, left?: number, top?: number, minimized?: boolean, static?: boolean,
 }> {
+	subscription: PSSubscription | null = null;
+	subscribeTimeout: number | null = null;
+	override componentDidMount() {
+		this.subscribeTimeout = window.setTimeout(() => {
```
Zarel, line 1433: Is the timeout necessary?

Zarel:
My thoughts:

- There's no particularly explained mechanism of action that this fixes the bug.
- It also seems likely to be overkill. We don't have reason to believe every room change is missing a userlist update, for a bug that apparently only happens on first load.
- You've PRed things that did not do what they say they're intended to do in the past. We mostly end up rejecting your PRs.

For these reasons, I'm just going to close this now.

PR #2649, opened 2026-04-14
Preact: Fix scrolling news tab UI freeze

TurboRx:
`NewsPanel` passed `fullSize` to `PSPanelWrapper`, which adds the `mini-window-flex` CSS class `height: auto`. in wide mode the `.mainmenu-mini-windows` container is `position: absolute`, so its content is excluded from the parent scroll container's `scrollHeight` scrolling the page couldn't reach the news content, making the UI appear frozen. bug fix: `Scrolling News tab freezes UI` https://github.com/orgs/smogon/projects/2?pane=issue&itemId=111176277

### DIFF
--- a/play.pokemonshowdown.com/src/panel-mainmenu.tsx
+++ b/play.pokemonshowdown.com/src/panel-mainmenu.tsx
@@ -495,7 +495,8 @@ class NewsPanel extends PSRoomPanel {
 	};
 	override render() {
 		const cookieSet = document.cookie.includes('preactalpha=1');
-		return <PSPanelWrapper room={this.props.room} fullSize scrollable>
+		const tinyLayout = PS.mainmenu.width < 620;
+		return <PSPanelWrapper room={this.props.room} fullSize={tinyLayout} scrollable={!tinyLayout}>
 			<div class="construction">
 				This is the client rewrite beta test.
 				<form>
--- a/play.pokemonshowdown.com/src/panel-mainmenu.tsx
+++ b/play.pokemonshowdown.com/src/panel-mainmenu.tsx
@@ -495,7 +495,7 @@ class NewsPanel extends PSRoomPanel {
 	};
 	override render() {
 		const cookieSet = document.cookie.includes('preactalpha=1');
-		return <PSPanelWrapper room={this.props.room} fullSize scrollable>
+		return <PSPanelWrapper room={this.props.room} scrollable>
 			<div class="construction">
 				This is the client rewrite beta test.
 				<form>
--- a/play.pokemonshowdown.com/src/panel-mainmenu.tsx
+++ b/play.pokemonshowdown.com/src/panel-mainmenu.tsx
@@ -495,7 +495,7 @@ class NewsPanel extends PSRoomPanel {
 	};
 	override render() {
 		const cookieSet = document.cookie.includes('preactalpha=1');
-		return <PSPanelWrapper room={this.props.room} scrollable>
+		return <PSPanelWrapper room={this.props.room} fullSize scrollable>
 			<div class="construction">
 				This is the client rewrite beta test.
 				<form>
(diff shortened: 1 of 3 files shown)


ISenseAura:
did you test this? this seems like it wouldnt work.

also,(i know this is how it works in old client but still) im not sure whether we want this behavior on all mini rooms, dm windows have overflow y hidden and have their own scrollable content and scrolling inside it should only scroll the content of the dm window not the outside page.

the news window doesnt have overflow hidden so theres no scrollable content inside it hence its fine if scrolling on/inside it scrolls the outside page as well.

TurboRx:
yeah, ur right. and that commit is on my local i forgot to push. now it will work properly.

ISenseAura:
still not working on my end

TurboRx:
i tested it too idk why it's not working rn, i need to further check. maybe i clicked in the wrong place and scrolled.

TurboRx:
now it's working on my end. tested on pc and mobile both.

play.pokemonshowdown.com/src/panel-mainmenu.tsx
```diff
@@ -495,7 +495,8 @@ class NewsPanel extends PSRoomPanel {
 	};
 	override render() {
 		const cookieSet = document.cookie.includes('preactalpha=1');
-		return <PSPanelWrapper room={this.props.room} fullSize scrollable>
+		const tinyLayout = PS.mainmenu.width < 620;
+		return <PSPanelWrapper room={this.props.room} fullSize={tinyLayout} scrollable={!tinyLayout}>
```
Zarel: What's this part for?

TurboRx: The only difference made is in scrolling according to screen size. Mobile devices < 620px use `height: auto`, allowing natural scrolling of the page. Desktop devices ≥ 620px use an internal scroll box height of 500px. DMs are not affected by this change.

Zarel: Shouldn't this be implicit?

TurboRx: alr, done. moved the width check into `PSPanelWrapper's` mini-window branch, so NewsPanel no longer needs to pass `fullSize` or `scrollable` at all. The wrapper now derives both implicitly from `PS.mainmenu.width`.

Zarel: I'm still confused about what this is for.

My original code has `fullSize` and `scrollable` for a reason.

As a tab, it should be scrollable. As a mini-window, it should have `height: auto` (and not be scrollable).

Zarel: Can you be specific about what this changes? The before and after.

Zarel: Okay. I just tested the behavior and I finally understand. I decided to fix it my own way.

play.pokemonshowdown.com/src/panel-topbar.tsx
```diff
@@ -297,14 +297,20 @@ export class PSHeader extends preact.Component {
 }
 
 export class PSMiniHeader extends preact.Component {
+	scrollRaf: number | null = null;
```
Zarel, line 300: This is unnecessary. `forceUpdate` actually enqueues an update, instead of immediately updating, so there's no particular need to manually requestAnimationFrame.

TurboRx: thanks for looking into this and explaining the `forceUpdate` behavior. i overcomplicated the fix trying to make the props implicit. glad the core issue is fixed now. learned a lot from this review, will keep the explicit prop design in mind for the future.

PR #2651, opened 2026-04-17
Preact: Update footer

ry4242:
Add Privacy policy and fix bgcredit in Preact, and add wrapping.


ISenseAura:
probably better to be added along with a fix for the following change
https://www.smogon.com/forums/threads/fix-the-layering-on-the-privacy-policy-link.3777213/

play.pokemonshowdown.com/style/client.css
```diff
@@ -606,6 +606,7 @@ p.or:after {
 	position: absolute;
 	left: 325px;
 	top: 40px;
+	z-index: 1;
```
Zarel: ?

This doesn't look like the Preact client.

ry4242: It's fixing the above mentioned issue with clipping, I'll update the title to reflect that

Zarel:
I can't imagine why we'd need to deal with Z-index clipping at all. Shouldn't we be wrapping the link or something?

ry4242:
I implemented it like this because this is how it was already implemented in the modern client. Should both legacy and Preact use wrapping instead of z-indexing?

Zarel:
I think so. The main one I care about is Preact client, because I am still aiming for a very-soon deployment of it.

play.pokemonshowdown.com/style/client2.css
```diff
@@ -723,7 +723,6 @@ p.or:after {
 	width: 312px;
 	min-height: 2px;
 	padding: 25px 0 51px 0;
-	z-index: 1;
```
Zarel, line 726: Okay. This change makes it impossible to use DMs.

Zarel: ...This makes me very concerned whether you've tested this change at all...

ry4242: I am still able to use DMs on the testclient. How can i recreate the issue?

Zarel: Are you using testclient-new?

Zarel:
Okay. After trying a large variety of hacks, I have successfully come up with code that will reliably put the footer in the right place.

play.pokemonshowdown.com/style/client2.css
ry4242: Yes, but I'm not sure how to use it "online".

Zarel: Use `testclient-key.js`; it's documented in the README.

ry4242: Yeah that works for testclient-old, but testclient-new (and previously testclient-beta) have never worked for me other than purely visual changes.

ry4242: Oh, actually, I figured it out.

Zarel:
Thanks!

PR #2653, opened 2026-04-18
Preact: Cancel challenge when closing a DM

ISenseAura:


Zarel:
I wonder if it should cancel the challenge instead of closing...

ISenseAura:
> I wonder if it should cancel the challenge instead of closing...

thats exactly the old client behavior btw

Zarel:
Yeah, if we do change that it should probably go in a different PR. Thanks!

PR #2654, opened 2026-04-19
Display more speeds in tooltip

iforgetwhyimhere:
https://www.smogon.com/forums/threads/displaying-uninvested-speed-in-battle-tooltips.3653714/


Zarel:
tyty!

PR #2655, opened 2026-04-19
Champions: Disable IVs selector

andrebastosdias:
image


andrebastosdias:
I didn't remove it because it looks a bit weird

image

PR #2656, opened 2026-04-19
Tooltip: Fix Normalize type for status moves

andrebastosdias:
https://www.smogon.com/forums/threads/type-effectiveness-normalize-status-moves.3780449/


Marty-D:
Should this check exist at all? Status moves have always had their type changed. Galvanize Roar into Volt Absorb activates the Ability, for example.

andrebastosdias:
Oh, the check in the server is "non-status Z-move".

Zarel:
Thanks!

PR #2659, opened 2026-04-20
Preact: update styles for background panel buttons

TurboRx:
Updates the button styles to use ps default style.

| Screenshot 1 | Screenshot 2 |
| :---: | :---: |
| Screenshot | IMG_20260420_140532 |


Zarel:
The random button is intended to look like one of the existing background options. I wouldn't want it to look like the other buttons.

The upload button... I don't mind, although you don't need to make it imitate the system dialogue if you're going to do that. Just make it say "Choose file or drag image".

PR #2660, opened 2026-04-20
Tooltip: Fix status moves effectiveness during extreme weather

andrebastosdias:
https://www.smogon.com/forums/threads/status-moves-are-not-affected-by-extreme-weather.3780974/


Zarel:
Thanks!

PR #2661, opened 2026-04-22
UUBL teambuilder: include (OU) by technicality mons in the picker

surfnWOB:
## Summary

In Gen 3 UUBL, three Pokémon — **Porygon2**, **Raikou**, and **Regice** — are `(OU)` by technicality but are explicitly unbanned from the format (server-side change: smogon/pokemon-showdown#11980). The teambuilder picker slices the tier set at `slices.UUBL`, which is set *after* the `(OU)` section — so these mons fall into `baseIllegalResults` and display as **Illegal**.

The `"OU by technicality"` header and its entries are already emitted into the tier list by `build-indexes`; the fix is just to start the slice a little earlier for UUBL.

## Changes

- **`build-tools/build-indexes`**: add `"(OU)"` to the tiers that get a `formatSlices` entry, so the generated `teambuilder-tables.js` exposes `slices['(OU)']` pointing at the start of the "OU by technicality" section.
- **`play.pokemonshowdown.com/src/battle-dex-search.ts`**: for `format === 'uubl'`, use `slices['(OU)'] || slices.UUBL`, matching the fallback pattern already used by RU/NU/PU/ZU.

No hardcoded pokemon IDs, no `gen === N` checks — the data layer does the work, and the "OU by technicality" header appears naturally above the UUBL section.

## Test plan

- [ ] Run `node build-tools/build-indexes && node build`
- [ ] Open teambuilder, select **[Gen 3] UUBL**
- [ ] Verify an **OU by technicality** header appears at the top, containing **(OU) Porygon2**, **(OU) Raikou**, and **(OU) Regice**
- [ ] Verify none of the three are marked Illegal
- [ ] Verify the **UUBL** section still appears below with its normal contents (Alakazam, Armaldo, ...)


KrisXV:
Why are you using Claude for this

surfnWOB:
Apologies, I never received guidance on this. I will be mindful of this in the future. I'm a professional software developer and I did test these changes and feel that they address the issue

Zarel:
I don't think this is the first time you've produced a PR like this. Why is this based on 43bb7a1c6340df176 and not 200757ef6808383? are you rebasing on the wrong commit or something?

Zarel:
My main concern is whether _every_ UUBL format is going to include OU by technicality. Or if this is something we need to hardcode for a specific generation.

surfnWOB:
I did flag this when speaking to goldmason about it. In practice, to my understanding, `gen3uubl` is the only format that is played with any regularity and that has a specific format buildable on client. 

If I'm understanding your concern, if there was a theoretical `gen4uubl` that wanted `(OU)`s removed then I think that would require changes here for the builder.

PR #2662, opened 2026-04-22
Preact: Add user mod actions panel

ISenseAura:
Screenshot from 2026-04-22 13-32-01


Zarel:
Shouldn't the `...` menu cover it? Why a different menu?

ISenseAura:
> Shouldn't the `...` menu cover it? Why a different menu?

Too much on a single menu, plus mod actions now have enough buttons to have its own menu

Zarel:
...honestly I'm not a fan of that.

Zarel:
Yeah, I've thought about this and I still prefer having them all in one menu.

PR #2663, opened 2026-04-22
Preact: Fix old status not clearing in userdetails cache

ISenseAura:


play.pokemonshowdown.com/src/panel-mainmenu.tsx
```diff
@@ -402,6 +402,7 @@ export class MainMenuRoom extends PSRoom {
 			if (!userdetails) {
 				this.userdetailsCache[userid] = response;
 			} else {
+				if (!response.status) response.status = '';
```
Zarel:
```suggestion
				response.status ||= '';
```

play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -734,11 +734,6 @@ export class ChatRoom extends PSRoom {
 	}
 
 	handleJoinLeave(action: 'join' | 'leave', name: string, silent: boolean) {
-		if (action === 'join') {
-			this.addUser(name);
-		} else if (action === 'leave') {
-			this.removeUser(name);
-		}
```
Zarel, line 741: What's this for?

ISenseAura: that part there was unnecessary coz we were already adding/removing users in ``receive`` and this function only handles showjoins

Zarel: Hm. Okay, sounds good.

PR #2667, opened 2026-04-26
Preact: Add tab autocomplete

Daniel1464:


play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -1163,6 +1178,28 @@ export class ChatTextEntry extends preact.Component<{
...
+						currentCandidates.unshift({ type: "command", command: cmd });
+					}
+					this.tabComplete = {
+						candidates: currentCandidates,
+						candidateIndex: 0,
+						prefix: nextCmd,
+						cursor: nextCmd,
+					};
+				});
+				return;
```
Zarel: Hm. This isn't a valid return value, and I think it's wrong here. It's making it so that if a tabcomplete isn't found, it's tabbing out of the box. It also currently does a weird thing where it tabs out, and then refocuses if it a finds a tabcomplete.

Zarel:
Yeah, fixing the return value makes it work great! Thanks!

PR #2670, opened 2026-05-02
Preact: Fix DM challenge edge cases

abhishekpradhan:
## Summary
- Prevent opening the outgoing challenge form in a PM while an incoming challenge from that user is active.
- Send PM challenge reject actions as global challenge commands targeting the PM user, matching old-client behavior for offline users.
- Send the incoming-challenge Accept button as a global targeted `/accept` command instead of routing it through the PM room.

## Validation
- `npm test`
- Targeted protocol smoke against rebuilt `panel-chat.js`:
  - Incoming challenge blocks the outgoing challenge menu.
  - PM `/reject` emits global `/reject alice`.
  - Typed PM `/accept` remains routed through the PM room and is intentionally not special-cased.

Note: the build step still logs `php: command not found` while trying to retrieve news, but `npm test` completed successfully.

### DIFF
--- a/play.pokemonshowdown.com/src/panel-chat.tsx
+++ b/play.pokemonshowdown.com/src/panel-chat.tsx
@@ -351,7 +351,12 @@ export class ChatRoom extends PSRoom {
 			this.challenged = null;
 			this.teamSent = null;
 			this.update(null);
-			this.sendDirect(`/reject ${target}`);
+			const userid = toID(target || this.pmTarget || '');
+			PS.send(userid ? `/reject ${userid}` : `/reject`);
+		},
+		'accept'(target) {
+			const userid = toID(target || this.pmTarget || '');
+			PS.send(userid ? `/accept ${userid}` : `/accept`);
 		},
 		'clear'() {
 			this.log?.reset();
@@ -598,6 +603,7 @@ export class ChatRoom extends PSRoom {
 			this.add(`|error|Can only be used in a PM.`);
 			return;
 		}
+		if (this.challenged) return;
 		this.challengeMenuOpen = true;
 		this.update(null);
 	}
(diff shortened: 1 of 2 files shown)


play.pokemonshowdown.com/src/panel-chat.tsx
```diff
@@ -348,10 +348,15 @@ export class ChatRoom extends PSRoom {
...
 		'reject'(target) {
+			target ||= this.pmTarget || '';
 			this.challenged = null;
 			this.teamSent = null;
 			this.update(null);
-			this.sendDirect(`/reject ${target}`);
+			PS.send(target ? `/reject ${toID(target)}` : `/reject`);
+		},
+		'accept'(target) {
+			target ||= this.pmTarget || '';
```
```suggestion
```

Pretty sure this is meaningless?

abhishekpradhan: Good catch, the original version made this look redundant. It’s meant to support typed `/accept` from a DM by defaulting to the PM target, similar to `/reject`; otherwise it can send a bare `/accept`. I rewrote it to compute `userid` directly so the fallback is clearer.

abhishekpradhan:
Sorry for the force-push churn here. For reviewability, the incremental change from the previous pushed head to the current one is only the accept/reject fallback cleanup in `panel-chat.tsx`:

https://github.com/smogon/pokemon-showdown-client/compare/9eba65f4bbb6858038804072f0f70098a637f5ed...922172e2fc7ff39c582ba8a7241b8a378b1545ab

That change rewrites the fallback from mutating `target` to computing `userid` directly; behavior is intended to remain the same.

abhishekpradhan:
Closing this draft for now since follow-up work is being handed off.

PR #2678, opened 2026-05-14
Champions: Prefer max IVs

shrianshChari:
You cannot adjust IVs in Champions, so all IVs should be set to the maximum IV possible. ~~It seems like it already doesn't allow Speed to be set to 0, so all that needs to be adjusted is Attack.~~
Screenshot from 2026-05-13 20-04-41
Screenshot from 2026-05-13 20-04-29


Zarel:
Feels like the wrong place to do this fix.

shrianshChari:
The only places I see `minAtk` in are `play.pokemonshowdown.com/js/client-teambuilder.js` and `play.pokemonshowdown.com/src/battle-team-editor.tsx`. Maybe I should modify the former as well.

Zarel:
Don't worry, I'll push something that covers this.

PR #2681, opened 2026-05-21
Make speed tooltip values a configurable option

iforgetwhyimhere:
https://www.smogon.com/forums/threads/new-speed-tier-change.3782753/

Some people are not a fan of #2654 . This changes the forced full 4-value option to allow users to select between partial (Simple, 2 value) and full (Full, 4 value) in most battle formats. No changes affect Let's Go or Random Battle tiers.

The default is still full, but this allows users who are not a fan of full to use partial instead.

image

### DIFF
--- a/play.pokemonshowdown.com/js/client-topbar.js
+++ b/play.pokemonshowdown.com/js/client-topbar.js
@@ -518,7 +518,7 @@
 			}
 			buf += '<p><label class="optlabel">Background: <button class="button" name="background">Change background</button></label></p>';
 			var speedTipType = Dex.prefs('speedtiptype') || 'full';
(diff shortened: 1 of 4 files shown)


Zarel:
I'm not really a fan of having configuration for things like this. I think it's worth thinking of other options. Maybe having the intermediate values in gray?

iforgetwhyimhere:
> Maybe having the intermediate values in gray?

I honestly think that would cause even _more_ confusion than this solution. Maybe it belongs in Battle Options more? But I didn't want to put a drop-down in Battle Options when there aren't any before. I suppose it _could_ be a checkbox "Show more potential speeds"?

mynameisthunder:
> > Maybe having the intermediate values in gray?
> 
> I honestly think that would cause even _more_ confusion than this solution. Maybe it belongs in Battle Options more? But I didn't want to put a drop-down in Battle Options when there aren't any before. I suppose it _could_ be a checkbox "Show more potential speeds"?

a checkbox in settings sounds perfect for things like this. with the default of the normal way of doing things. Otherwise we can just revert the speed text back to how it was before.

PR #2684, opened 2026-05-22
Remove empty move lines from set export

Karthik99999:
Currently, export set always adds at least 4 dashes to the moves section, which causes empty dashes to show up when using sets like Fake Out + Last Resort. This simply removes that behavior.

image


Zarel:
This is done for the teambuilder; the teambuilder needs those lines. You can set them as optional, though.

PR #2686, opened 2026-05-23
Preact: Fix EV slider behavior for Champions

shrianshChari:
The EV slider doesn't give you the 2 remaining SPs if you drag it beyond the limit in Champions formats.

New behavior:

[Screencast 2026-05-23 10:52:36.webm](https://github.com/user-attachments/assets/2eca2b85-8acd-48fe-9214-10a22509af56)


Zarel:
Already fixed in ebd4d7f419f6ef18ead

PR #2687, opened 2026-05-23
Meteor Beam animation should use the space background instead of sand

shrianshChari:
https://www.smogon.com/forums/threads/have-meteor-beam-use-the-space-background-instead-of-the-sand-background.3690923/

[Screencast 2026-05-23 11:32:49.webm](https://github.com/user-attachments/assets/f57be9d0-8f36-4ffa-981f-5d21b78ffbae)


Zarel:
Looks good; thanks!

PR #2688, opened 2026-05-23
Prevent client-command help text response from triggering on broadcasted commands

iforgetwhyimhere:
dependency for upcoming server PR


Zarel:
Thanks!

PR #2689, opened 2026-05-25
Preact: Fix bug with command tabcomplete

ISenseAura:
the current implementation treats the entire line as one argument for tabcomplete which allowed it to perform tabcomplete on the whole line itself instead of individual words, so it would tabcomplete "/re ply" to "/reply" and also not allowing username tabcomplete on command arguments. for ex. "/forcerename 67skib" wouldnt tabcomplete the username argument.

this fixes it.


Zarel:
tyty!

PR #2691, opened 2026-05-25
Force Struggle's type effectiveness to 1x

iforgetwhyimhere:
https://www.smogon.com/forums/threads/bug-report-mechanics.3783002/


Zarel:
Not in Gen 1, right? Didn't it take a few gens before Ghost wasn't immune to Struggle?

Zarel:
Looks like it's Normal-type only in Gen 1.

Zarel:
You also shouldn't display Struggle effectiveness in gen 1-6 Hardcore Mode... As is, in this PR, you're showing effectiveness for Struggle and no other move in Hardcore Mode.

Zarel:
Thanks!

PR #2692, opened 2026-05-28
Make Tera Blast show up as Usually Useless for formats it's banned in

23PLevinJ:
fixes https://www.smogon.com/forums/threads/add-tera-blast-to-usually-useless-if-tera-is-not-available-in-the-meta.3783114/

### DIFF
--- a/play.pokemonshowdown.com/src/battle-dex-search.ts
+++ b/play.pokemonshowdown.com/src/battle-dex-search.ts
@@ -1725,6 +1725,8 @@ class BattleMoveSearch extends BattleTypedSearch<'move'> {
 		case 'temperflare':
 			return (!moves.includes('flareblitz') && !moves.includes('pyroball') && !moves.includes('sacredfire') &&
 				!moves.includes('bitterblade') && !moves.includes('firepunch')) || this.isDoubles;
+		case 'terablast':
+         return !(dex.gen === 9 && !this.formatType && ['ou', 'uu', 'ru', 'nu', 'pu', 'zu'].includes(format));
 		case 'terrainpulse': case 'waterpulse':
 			return ['megalauncher', 'technician'].includes(abilityid) && !moves.includes('originpulse');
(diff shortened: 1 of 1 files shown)


iforgetwhyimhere:
This should probably check for: is Gen 9, does not have `-Tera Blast`, and does not have `Terastal Clause` (in the banlist/ruleset) instead (dynamic checking).

23PLevinJ:
how would i implement those edits?

iforgetwhyimhere:
Hm, I suppose there's no easy way to call the ruleTable to check it. Probably would he something for another PR?

23PLevinJ:
yeah maybe

play.pokemonshowdown.com/src/battle-dex-search.ts
```diff
@@ -1725,6 +1725,8 @@ class BattleMoveSearch extends BattleTypedSearch<'move'> {
 		case 'temperflare':
 			return (!moves.includes('flareblitz') && !moves.includes('pyroball') && !moves.includes('sacredfire') &&
 				!moves.includes('bitterblade') && !moves.includes('firepunch')) || this.isDoubles;
+		case 'terablast':
+         return !(dex.gen === 9 && !this.formatType && ['ou', 'uu', 'ru', 'nu', 'pu', 'zu'].includes(format));
```
iforgetwhyimhere, line 1729: This logic is reversed

Daniel1464: Also, it would be better if this actually pulled from the formats where tera blast is banned through some kind of code logic instead of hardcoding them.

iforgetwhyimhere: that is what the rest of the conversation was. its on my to-do list but shouldnt stop this from merge

iforgetwhyimhere:
Superceded by #2712

PR #2693, opened 2026-05-30
Prioritize HP when assigning the remaining EVs in Stat Guesser

andrebastosdias:
Try to assign remaining EVs to HP before Def or SpD. In Champions formats, this will suggest allocating the remaining 2 stat points to HP. In formats such as Gen 9 OU, some Pokémon like Excadrill will have the remaining 4 EVs allocated to HP, since doing so does not make its HP divisible by SR damage.


Zarel:
Thanks!

PR #2695, opened 2026-06-02
Recognize Champions VGC as doubles in teambuilder

Abkwreu:


play.pokemonshowdown.com/src/battle-dex-search.ts
```diff
@@ -654,6 +654,7 @@ abstract class BattleTypedSearch<T extends SearchType> {
 			this.formatType = 'champions';
 			this.dex = Dex.mod('champions' as ID);
 			format = format.slice(9) as ID;
+			if (format.startsWith('vgc')) this.isDoubles = true;
 			if (format.startsWith('vgc') || format.startsWith('bss')) format = 'ubers' as ID;
```
DaWoblefet, line 658: This line should just not have the VGC check. The next if statement already is supposed to set the appropriate format and the isDoubles flag.

KrisXV: If this check doesnt exist then the format will show CAP pokemon, so idk

PR #2700, opened 2026-06-10
Add offensive coverage chart to teambuilder

Samarinara:
This is a small quality of life change that enhances the teambuilding experience. I was inspired to make this after testing out the client rewrite and really enjoying the defensive coverage feature. I tried to make it as simple as possible.

### DIFF
--- a/play.pokemonshowdown.com/src/battle-team-editor.tsx
+++ b/play.pokemonshowdown.com/src/battle-team-editor.tsx
@@ -764,6 +764,31 @@ export class TeamEditorState extends PSModel {
 		}
 		return counters;
 	}
+	teamOffensiveCoverage() {
+		type OCounter = {
+			type: Dex.TypeName,
+			superEffective: number,
+			neutral: number,
+			resisted: number,
+		};
+		const counters = {} as Record<Dex.TypeName, OCounter>;
+		for (const type of this.dex.types.names()) {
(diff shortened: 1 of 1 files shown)


iforgetwhyimhere:
Offensive coverage is a lot more complicated than defensive - Pokemon's "coverage" moves can be considered separate from each other, STAB vs non-STAB, Tera coverage, etc. It gets quite complex, and not easy at all to reduce to a set of numbers.

IMO, _if_ anything, this should consider raw typing of the team, and maybe Tera type offensive coverage, if it doesn't match a base type.

Samarinara:
> Offensive coverage is a lot more complicated than defensive - Pokemon's "coverage" moves can be considered separate from each other, STAB vs non-STAB, Tera coverage, etc. It gets quite complex, and not easy at all to reduce to a set of numbers.
> 
> IMO, _if_ anything, this should consider raw typing of the team, and maybe Tera type offensive coverage, if it doesn't match a base type.

Reducing it to just STAB and Tera coverage is a good idea, but I'm afraid it still won't get a good picture of the team. Is there room for a scoring system or similar that would weigh more based on how strong the moves would be?

This could make it so a strong mon like Basculegion doesn't have its high base power moves and high attack stat counted the same as smt like snarl. It would be more work but I would be willing to do it

PR #2701, opened 2026-06-11
Support `-brokeprotect`

iforgetwhyimhere:
Client dependency for recognizing changes in https://github.com/smogon/pokemon-showdown/pull/12107

### DIFF
--- a/play.pokemonshowdown.com/src/battle-text-parser.ts
+++ b/play.pokemonshowdown.com/src/battle-text-parser.ts
@@ -1199,6 +1199,7 @@ export class BattleTextParser {
 			return template.replace('[POKEMON]', this.pokemon(pokemon));
 		}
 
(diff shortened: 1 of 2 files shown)


urkerab:
You need to retain support for the old protocol because of replays (and other reasons).

iforgetwhyimhere:
> You need to retain support for the old protocol because of replays (and other reasons).

Yeah, I realized that overnight. Fixed now.

> What’s the message for moves that fully remove Protect for the rest of the turn like Feint?

"POKEMON fell for the feint!" Is what Feint uses, but other moves like Hyperspace Hole/Fury and Shadow Force use "It broke through [TARGETS]'s protection!" as an `|-activate|`. Same case with other moves that bypass protect. This code handles specifically Protect not _fully_ blocking, but reducing, the damage taken.

PR #2702, opened 2026-06-12
Champions: Track freeze turns

andrebastosdias:


Daniel1464:
Why would this only be for champions formats?

andrebastosdias:
> Why would this only be for champions formats?

In Champions, freeze expires after 3 turns. In previous games, freeze didn't have a duration.

PR #2703, opened 2026-06-15
Clean up `getSpeedRange` variables

iforgetwhyimhere:
When I changed the speed tooltips in https://github.com/smogon/pokemon-showdown-client/commit/6d6e8748a66e9e5c9e0246296a52e73a7868b581, some variables in `getSpeedRange` became redundant as their different values are already handled in their individual cases. This finishes that change by replacing them with their effective values in the location where they are referenced.


Zarel:
tyty

PR #2705, opened 2026-06-20
Champions: Add Natures from OTS to tooltips

ShivaD173:
Add Natures from OTS to tooltips. Displays known speed range based on nature.
Screenshot 2026-06-20 at 6 14 08 PM

### DIFF
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -1472,7 +1472,7 @@ export class BattleTooltips {
 				return `<p><small>Spe</small> ${min}<small class="gray">&ndash;${ev0}&ndash;</small>${max} <small>(before external modifiers)</small></p>`;
 			} else if (clientPokemon.nature) {
 				// Nature already taken into account in min/max
-				return `<p><small>Spe</small> ${min} to ${max}<br><small>(before external modifiers)</small></p>`;
+				return `<p><small>Spe</small> ${min} to ${max} <small>(before external modifiers)</small></p>`;
 			} else {
 				return `<p><small>Spe</small> ${min}<small class="gray">&ndash;${ev0}&ndash;${ev252}&ndash;</small>${max}<br><small>(before external modifiers)</small></p>`;
 			}
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -976,6 +976,12 @@ export class BattleTooltips {
 			text += '</p>';
 		}
 
+		// Only display when you don't also have stats
+		if (clientPokemon?.nature && !serverPokemon) {
+			let natureText = '<small>Nature:</small> ' + clientPokemon.nature;
+			text += `<p>${natureText}</p>`;
+		}
+
 		text += this.renderStats(clientPokemon, serverPokemon, !isActive);
 
 		if (serverPokemon && !isActive) {
@@ -1453,7 +1459,7 @@ export class BattleTooltips {
 		const isTransformed = clientPokemon?.volatiles.transform;
 		if (!serverPokemon || isTransformed) {
 			if (!clientPokemon) throw new Error('Must pass either clientPokemon or serverPokemon');
-			let { min, ev0, ev84, ev252, max } = this.getSpeedRange(clientPokemon);
+			let { min, ev0, ev0plus, ev84, ev252minus, ev252, max } = this.getSpeedRange(clientPokemon);
 			if (this.battle.gen < 3) {
 				if (this.battle.tier.includes('Random')) {
 					return `<p><small>Spe</small> ${max} <small>(before stat stage changes)</small></p>`;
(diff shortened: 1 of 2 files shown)


play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -1453,7 +1459,7 @@ export class BattleTooltips {
 		const isTransformed = clientPokemon?.volatiles.transform;
 		if (!serverPokemon || isTransformed) {
 			if (!clientPokemon) throw new Error('Must pass either clientPokemon or serverPokemon');
-			let { min, ev0, ev84, ev252, max } = this.getSpeedRange(clientPokemon);
+			let { min, ev0, ev0plus, ev84, ev252minus, ev252, max } = this.getSpeedRange(clientPokemon);
```
iforgetwhyimhere: Can this not add new variables? Just use a natureIsKnown or something, and multiply the nature value in getSpeedRange

play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -1587,37 +1609,47 @@ export class BattleTooltips {
...
 		let ev84;
+		let ev252minus;
 		let ev252;
 		let max;
 		if (tier.includes("Let's Go")) {
 			min = tr(tr(tr(2 * baseSpe * level / 100 + 5) * 0.9) * tr((70 / 255 / 10 + 1) * 100) / 100);
 			ev0 = tr(tr(tr((2 * baseSpe + 31) * level / 100 + 5)) * tr((70 / 255 / 10 + 1) * 100) / 100);
+			ev0plus = tr(tr(tr(2 * baseSpe * level / 100 + 5) * 1.1) * tr((70 / 255 / 10 + 1) * 100) / 100);
 			ev84 = tr(tr(tr((2 * baseSpe + 31) * level / 100 + 5)) * tr((70 / 255 / 10 + 1) * 100) / 100);
+			ev252minus = tr(tr(tr((2 * baseSpe + 31) * level / 100 + 5) * 0.9) * tr((70 / 255 / 10 + 1) * 100) / 100);
```
iforgetwhyimhere: Even if you don't opt for the former, just calculate ev252 first, then set ev252minus to 0.9 * ev252

iforgetwhyimhere (changes requested):

play.pokemonshowdown.com/src/battle-tooltips.ts
ShivaD173: Good call. Now it populates min/max directly, which by side effect will automatically make it work with gyro ball calculation.

play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -1585,37 +1594,50 @@ export class BattleTooltips {
...
+		let maxNatureMult = 1.1;
+		if (pokemon.nature) {
+			let natureVals = BattleNatures[pokemon.nature];
+			if (natureVals.minus === 'spe') {
+				minNatureMult = 0.9;
+				maxNatureMult = 0.9;
+			} else if (natureVals.plus === 'spe') {
+				minNatureMult = 1.1;
+				maxNatureMult = 1.1;
+			}
```
iforgetwhyimhere, line 1606:
```suggestion
			let natureVals = BattleNatures[pokemon.nature];
			if (natureVals.minus === 'spe') maxNatureMult = 0.9;
			if (natureVals.plus === 'spe') minNatureMult = 1.1;
```
No need to redefine vars as their default value

iforgetwhyimhere (commented):
LGTM + small nitpick

iforgetwhyimhere (approved):

play.pokemonshowdown.com/src/battle-tooltips.ts
```diff
@@ -1464,9 +1470,12 @@ export class BattleTooltips {
 				return `<p><small>Spe</small> ${min} or ${ev84} <small>(before external modifiers)</small></p>`;
 			} else if (this.battle.tier.includes("Let's Go")) {
 				return `<p><small>Spe</small> ${min}<small class="gray">&ndash;${ev0}&ndash;</small>${max} <small>(before external modifiers)</small></p>`;
+			} else if (clientPokemon.nature) {
+				// Nature already taken into account in min/max
+				return `<p><small>Spe</small> ${min} to ${max}<br><small>(before external modifiers)</small></p>`;
```
iforgetwhyimhere: on second thought, the `<br>` also isn't needed since there's only 2 numbers.

ShivaD173: Fixed

DaWoblefet:
This is an excellent change, but if we're going to account for Speed on the tooltips for +/- Speed natures, we should do it for neutral Natures too.

For example, this Ninetales is Timid, and it correctly displays the 141-177 range:
image

This Kingambit is Adamant, so it should show a range of 70-102, not 63-112.
image

ShivaD173:
Ah, I think I accidentally removed it when trying to address one of the PR comments. Support for Neutral natures has been readded.
Screenshot 2026-07-08 at 4 08 20 PM

DaWoblefet:
Very nice work. Thank you!

PR #2709, opened 2026-06-23
Fix Weather Ball type/damage display

ry4242:
Fixes Weather Ball's type and damage display when disabling/ignoring effects are taken into account (Utility Umbrella, Klutz, Magic Room)


Zarel:
Thanks!

PR #2710, opened 2026-06-23
Update the teambuilder big buttons

ry4242:
Removes the "Shiny" tab and change the Level tab into "Info", where the Shiny icon is now located. Also adds the gender icon and potential G-max icon into Info as well.

Also lets Hidden Power type tab appear if the editor is in Natdex.

G-max icon would of course require a hosted gmax.png.

image
c35cef887c3ff3e207d1fb0eb789e28b


Zarel:
Sorry. We ended up completely removing the Wizard editor. If you'd like to add better Gmax support, though, that could be nice.

PR #2712, opened 2026-06-24
Teambuilder: Removes Tera button in formats with Terastal clause

ry4242:
Also moves Tera Blast to Usually Useless in formats with Tera Clause, and adds HP type button to NatDex.

Requires https://github.com/smogon/pokemon-showdown/pull/12132


play.pokemonshowdown.com/src/battle-dex-search.ts
```diff
@@ -1720,6 +1722,8 @@ class BattleMoveSearch extends BattleTypedSearch<'move'> {
 		case 'temperflare':
 			return (!moves.includes('flareblitz') && !moves.includes('pyroball') && !moves.includes('sacredfire') &&
 				!moves.includes('bitterblade') && !moves.includes('firepunch')) || this.isDoubles;
+		case 'terablast':
+			return !this.terastalClause;
```
iforgetwhyimhere:
```suggestion
			return !this.terastalClause && dex.gen === 9;
```

PR #2716, opened 2026-07-03
Battles: Disable autocomplete on username-prefix search

a96lex:
When i go to the battles tab to expectate, my username is prefilled on the input field.

This is counterintuitive to me since I never want to look for my battles to expectate, as I am already in them. So I always need to manually delete my username manually.

This change disables autocomplete for that input field so it starts empty.


Zarel:
It's really unfortunate that our text boxes mostly don't benefit from autocomplete. I keep on thinking maybe we should keep it on in case there's a certain type of person who likes it, but that's probably a lost cause.

a96lex:
in this case, i find it counterintuitive. I never want to filter battles to spectate by my username since i am already in them as player.

Zarel:
Yeah, fair enough.

PR #2718, opened 2026-07-08
Assume neutral Natures in Random Battles

iforgetwhyimhere:
Fixes https://www.smogon.com/forums/threads/bug-report-client.3784995/

Natures are always neutral in Random Battle, so using `min`, which uses a -Spe nature, was returning a value lower than possible in rands.
I put the mult setter above the Nature set, so that in the case where a Pokemon DOES have a nature, through whatever means, it won't break the tooltip with incorrect values (assuming neutral when it has pos/neg).

### DIFF
--- a/play.pokemonshowdown.com/src/battle-tooltips.ts
+++ b/play.pokemonshowdown.com/src/battle-tooltips.ts
@@ -1596,6 +1596,10 @@ export class BattleTooltips {
 
 		let minNatureMult = 0.9;
 		let maxNatureMult = 1.1;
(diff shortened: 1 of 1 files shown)


DaWoblefet:
Is this always true? I thought that some of the Trick Room or Gyro Ball sets ran Speed lowering Natures.

iforgetwhyimhere:
I did so too! Turns out it's just 0 EV / 0 IV, and the Nature doesn't actually change.

DaWoblefet:
Thanks for the fix!

PR #2719, opened 2026-07-16
Mega Meowstic F base abilities showing incorrectly in team builder

BumaTehEwok:
This is my first time contributing, so I hope I am following the process ok.

I debugged and fixed this issue.
https://www.smogon.com/forums/threads/female-mega-meowstic-shows-base-forme-with-incorrect-ability.3785036/

Cause:
When base form abilities are loaded in the team builder they default to base species. Since Meowstic is the base species, but diverges from Meowstic-F it created a new edge case.

Fix:
Use changesFrom first when getting the species for a mega evolution when doing an ability search. This requires no changes to the pokedex data since battleOnly was already set to Meowstic-F for Mega Meowstic-F. 

Falls back to baseSpecies when changeForm is undefined so it does not effect other Megas. 

image

image

image

### DIFF
--- a/play.pokemonshowdown.com/src/battle-dex-search.ts
+++ b/play.pokemonshowdown.com/src/battle-dex-search.ts
@@ -1390,7 +1390,7 @@ class BattleAbilitySearch extends BattleTypedSearch<'ability'> {
 
 		if (species.isMega) {
 			abilitySet.unshift(['html', `Will be <strong>${species.abilities['0']}</strong> after Mega Evolving.`]);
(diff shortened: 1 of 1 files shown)


iforgetwhyimhere (approved):

iforgetwhyimhere:
This would also fix any issue with Floette-Eternal should there be disparity between it and base Floette. I assume there wouldn't be, but it also futureproofs other pre-mega forme-based Megas.

iforgetwhyimhere:
https://www.smogon.com/forums/threads/bug-report-teambuilder.3787290/

PR #2730, opened 2026-08-07
Preact: Respect ignored nicknames in move target controls

wbstx:
This is a fix for a reported bug
https://www.smogon.com/forums/threads/nicknames-in-target-buttons-dont-get-ignored.3785984/

When "Ignore nicknames" or "Ignore opponent" is enabled, pokemon nicknames should not appear in the move target buttons or old choice summary.


wbstx:
Resolved the conflicts. Could someone take another look when they have time? Thanks!

PR #2731, opened 2026-08-08
Allow Sorting Pokémon in the Teambuilder by Usage

andrebastosdias:
https://www.smogon.com/forums/threads/allow-sorting-pok%C3%A9mon-in-the-teambuilder-by-usage.3783376/
image


build-tools/build-usage-stats
andrebastosdias, line 1: I'd like to get @Marty-D's feedback on whether we can create a single file containing only the usage values to speed up the build process to a single request.

PR #2750, opened 2026-09-05
Teambuilder: Fix header's Ability column to match whether abilities are shown in result rows

andrebastosdias:


play.pokemonshowdown.com/src/battle-searchresults.tsx
```diff
@@ -65,13 +65,16 @@ export class PSSearchResults extends preact.Component<{
 	renderPokemonSortRowHTML(index: number) {
 		const search = this.props.search;
 		const sortCol = search.sortCol;
+		const abilityWidthClass = search.dex.gen < 3 || search.dex.modid === 'gen7letsgo' ? null :
+			(search.dex.gen < 5 ? 'singleabilitysortcol' : 'abilitysortcol');
```
Zarel: Hm. Plausible we should just calculate `numAbilityCols` as `0`, `1`, or `2` here, instead of needing to recalculate it every time.

play.pokemonshowdown.com/src/battle-dex-search.ts
```diff
@@ -90,6 +90,7 @@ export class DexSearch {
 	}
 	static unselectableResultTypes = ['header', 'html', 'sortpokemon', 'sortmove'];
 	firstPokemonColumn: 'Tier' | 'Number' = 'Number';
+	numAbilityCols: 0 | 1 | 2 = 0;
```
Zarel, line 93: I think this is the wrong place for this. The other one, `firstPokemonColumn` is configuration (you decide how to label it). `numAbilityCols` is automatic. It should probably just be a const somewhere.

Zarel: Okay. On further reflection, I don't actually think there's a better place to put it if we want to cache it. It's just weird to have a cache next to configuration in a way that makes them look like the same kind of thing.

andrebastosdias: What's the verdict?

Zarel: I guess we'll just do this for now.

